PR #49, opened 2016-01-25
Changed more print statements to log statements.

leonardr:
The same as before. In this case I sometimes need to use the generic 'logging' library instead of self.log because the CirculationParser class, unlike Monitor, doesn't define a log object.

### DIFF
--- a/threem.py
+++ b/threem.py
@@ -4,6 +4,7 @@ import itertools
 import datetime
 import os
 import re
+import logging
 
 from nose.tools import set_trace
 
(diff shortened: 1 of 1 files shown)


threem.py
```diff
@@ -70,7 +71,10 @@ def get_events_between(self, start, end, cache_result=False):
         try:
             events = EventParser().process_all(response.content)
         except Exception, e:
-            print response.content
+            self.log.error(
```
courte: Is `self.log` automatically set when you import logging?

leonardr: No, self.log is a property of Monitor (core/monitor.py) which looks up a logger based on the monitor's .service_name. If you're not inside a Monitor (or something similar like CoverageProvider) there is no self.log.

courte: Okay, so this self.log needs to be changed to `logging.error`. All the others are good to go.

courte: Wait, jk, ThreeMAPI's has the log.

courte:
Closing manually because I accidentally rebased against master locally. (Sorry! :skull:) This has been merged into master under the sha: 39cd8486ef49cb92b193d176f95acb9378a17d00.

Issue #65, opened 2016-02-05
Overdrive books on reserve not handled correctly


leonardr:

leonardr:
When you put an Overdrive book on hold, and then the book becomes available, Overdrive presents the book as still being on hold  but with the option to check it out. This wouldn't be so bad except it preserves your hold position _within the list of people who have the option to check the book out_. So it still looks like you're #3 in the hold queue instead of being (one of the people) at the front of the line.

leonardr:
It looks like I already have code that handles this, but it only works if you're #1 in the hold queue. I didn't anticipate that the book might be available _regardless_ of your purported queue position.

PR #66, opened 2016-02-05
It doesn't matter what Overdrive says the patron's position is--if th…

leonardr:
…ere is a checkout action on a hold, we consider their position to be zero.

This branch relaxes an assumption that if you have an Overdrive book on hold but ready to check out, your hold queue position will be 1. Actually, your hold queue position can be any number. The important thing is not the hold queue position but the presence of a 'checkout' option.

This fixes bug #65.


aslagle:
:+1:

PR #79, opened 2016-02-23
Open-access books are not subject to a site policy that prohibits holdds, since you never need to put one on hold.

leonardr:
The test for this one is a little unusual. The case being tested previously shouldn't have worked, since works created in tests are open-access by default. I changed the case previously tested to give the right behavior, and added another test that verifies what I _thought_ I was testing before--you can't put a hold on a non-open-access book.


aslagle:
:+1:

PR #80, opened 2016-02-24
Setup Google Oauth

### LINKED ISSUE #74: Set up authentication for NYPL staff

courte:
Allows folks with nypl.org emails to become admins. Sets up Google Oauth for authentication.

Fixes #74.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -80,6 +80,7 @@ from config import (
 from lanes import make_lanes
 
 from adobe_vendor_id import AdobeVendorIDController
+from oauth import GoogleAuthService
 from axis import (
     Axis360API,
 )
@@ -227,6 +228,8 @@ class CirculationManager(object):
             self.log.warn("Adobe Vendor ID controller is disabled due to missing or incomplete configuration.")
             self.adobe_vendor_id = None
 
+    def google(self):
+        return GoogleAuthService(self._db, self.url_for('google_auth_signin'))
 
     def annotator(self, lane, *args, **kwargs):
         """Create an appropriate OPDS annotator for the given lane."""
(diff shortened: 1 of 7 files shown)


controller.py
```diff
@@ -394,6 +394,53 @@ def apply_borrowing_policy(self, patron, license_pool):
...
+            test_mode=self.manager.testing
+        )
+
+    def authenticated_admin_from_request(self):
+        authorization = flask.request.environ.get('HTTP_AUTHORIZATION')
+        if authorization:
+            admin = get_one(self._db, Admin, access_token=authorization[7:])
+            if admin and self.google.active_credentials(admin):
+                return admin
+        return self.google.auth_uri
```
aslagle: Is it possible to do the redirect here?

controller.py
```diff
@@ -394,6 +394,53 @@ def apply_borrowing_policy(self, patron, license_pool):
...
+
+    def authenticated_admin(self, credentials):
+        """Creates or updates an admin with the given credentials"""
+        admin = Admin(email=credentials['email'])
+        admin.update_credentials(
+            self._db, credentials['access_token'], credentials['credentials']
```
aslagle: what's in credentials["credentials"]?

aslagle: oh I see, you've already extracted things from the google response.

controller.py
courte: There were some type issues, but I've found a less bad-regex-y way to return the redirect from here.

controller.py
courte: I also changed the naming here, since "credentials" wasn't accurate.

courte:
Replaced by #86.

Issue #82, opened 2016-02-26
Best-sellers query has become very slow


leonardr:
```
circulation=> explain select * FROM mv_works_editions_datasources_identifiers JOIN licensepools ON licensepools.id = mv_works_editions_datasources_identifiers.license_pool_id JOIN customlistentries ON licensepools.id = customlistentries.license_pool_id JOIN customlists ON customlists.id = customlistentries.list_id 
WHERE mv_works_editions_datasources_identifiers.language IN ('eng') AND mv_works_editions_datasources_identifiers.medium IN ('Book') AND (EXISTS (SELECT 1 
FROM licensepooldeliveries, deliverymechanisms 
WHERE licensepools.id = licensepooldeliveries.license_pool_id AND deliverymechanisms.default_client_can_fulfill = true)) AND 16 = customlists.data_source_id AND customlistentries.most_recent_appearance >= '2014-02-26T16:02:20.474957'::timestamp order by sort_author limit 10;
```

```
 Limit  (cost=119402.26..119402.26 rows=1 width=1047)
   ->  Sort  (cost=119402.26..119402.26 rows=1 width=1047)
         Sort Key: mv_works_editions_datasources_identifiers.sort_author
         ->  Nested Loop  (cost=0.84..119402.25 rows=1 width=1047)
               Join Filter: (customlistentries.list_id = customlists.id)
               ->  Nested Loop  (cost=0.84..119399.91 rows=1 width=888)
                     Join Filter: (licensepools.id = customlistentries.license_pool_id)
                     ->  Seq Scan on customlistentries  (cost=0.00..676.14 rows=7647 width=134)
                           Filter: (most_recent_appearance >= '2014-02-26 16:02:20.474957'::timestamp without time zone)
                     ->  Materialize  (cost=0.84..118494.37 rows=2 width=758)
                           ->  Nested Loop Semi Join  (cost=0.84..118494.36 rows=2 width=758)
                                 ->  Nested Loop  (cost=0.42..118488.54 rows=5 width=754)
                                       ->  Seq Scan on mv_works_editions_datasources_identifiers  (cost=0.00..118446.29 rows=5 width=701)
                                             Filter: (((language)::text = 'eng'::text) AND (medium = 'Book'::medium))
                                       ->  Index Scan using licensepools_pkey on licensepools  (cost=0.42..8.44 rows=1 width=53)
                                             Index Cond: (id = mv_works_editions_datasources_identifiers.license_pool_id)
[...]

PR #83, opened 2016-02-26
Don't use materialized works when finding books on a certain list or set of lists.

leonardr:
This is for bug #82.

The best-sellers query was incredibly slow, and although I'm interested in understanding why it's so slow, there's a pretty simple short-term solution solution: don't use the materialized work for the query.


aslagle:
:+1:

Issue #81, opened 2016-02-24
Use Configuration for Admin authentication details


courte:
Not all libraries will use Google Oauth to authenticate their administrators, so make sure the Configuration file is being used to set relevant variables: the location of the client json file & the domain name to match against.

PR #86, opened 2016-02-29
Google config

### LINKED ISSUE #74: Set up authentication for NYPL staff

courte:
Fixes #74 & #81. Creates admins. Uses Google OAuth to authenticate them. Handles callbacks. Holds onto access codes. Allows for library-specific configuration. Tests it. Creates peace & love on earth. Remains *very* commit-ted.

Replaces/includes #80.

### DIFF
--- a/controller.py
+++ b/controller.py
@@ -456,6 +456,9 @@ class AdminController(CirculationManagerController):
             return redirect(admin_url, Response=Response)
 
     def staff_email(self, email):
+        """Checks the domain of an email address against the admin-authorized
+        domain"""
+
         staff_domain = Configuration.policy(
             Configuration.ADMIN_AUTH_DOMAIN, required=True
         )
--- a/controller.py
+++ b/controller.py
@@ -80,6 +80,7 @@ from config import (
 from lanes import make_lanes
 
 from adobe_vendor_id import AdobeVendorIDController
+from oauth import GoogleAuthService
 from axis import (
     Axis360API,
 )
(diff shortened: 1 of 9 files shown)


controller.py
```diff
@@ -394,6 +394,70 @@ def apply_borrowing_policy(self, patron, license_pool):
...
+    def redirect_after_signin(self):
+        """Uses the Google OAuth client to determine admin details upon
+        callback. Barring error, redirects to the provided redirect url.."""
+
+        admin_details, redirect_url = self.google.callback(flask.request.args)
+        if isinstance(admin_details, ProblemDetail):
+            return ProblemDetail
+
+        if not self.staff_email(admin_details['email']):
+            return INVALID_ADMIN_CREDENTIALS
```
aslagle: I don't think this should return a problem detail. It gets returned where you'd normally be redirected, so the response will be displayed to the user.

aslagle: maybe this route needs a decorator to convert the problem detail to html.

oauth.py
```diff
@@ -0,0 +1,101 @@
...
+            return dict(
+                email=credentials.id_token.get('email'),
+                access_token=credentials.get_access_token()[0],
+                credentials=credentials.to_json(),
+            ), redirect_url
+
+    def google_error_problem_detail(self, error):
+        detail = "There was an error connecting with Google OAuth: %s" % error
+        return pd(
+            "http://librarysimplified.org/terms/problem/google-oauth-error",
```
leonardr: This should be in the module with the rest of the problem detail constants, since we occasionally need to provide a list of all the possible problem detail URLs.

PR #89, opened 2016-03-02
API and UI for suppressing/unsuppressing

### LINKED ISSUE #87: Allow staff to suppress a license pool

aslagle:
This fixes #87.

Depends on: https://github.com/NYPL-Simplified/circulation-web/pull/1


api/admin/controller.py
```diff
@@ -0,0 +1,189 @@
...
+    
+    def staff_email(self, email):
+        """Checks the domain of an email address against the admin-authorized
+        domain"""
+
+        staff_domain = Configuration.policy(
+            Configuration.ADMIN_AUTH_DOMAIN, required=True
+        )
+        domain = email[email.index('@')+1:]
+        return domain == staff_domain
```
leonardr: Is the case of the email address a potential issue here? (e.g. FOO@NYPL.ORG)

aslagle: It comes directly from google, but it could be an issue if the config file had upper case.

api/admin/routes.py
```diff
@@ -0,0 +1,96 @@
...
+def google_auth_callback():
+    return app.manager.admin_signin_controller.redirect_after_signin()
+
+@app.route('/admin/signin')
+@returns_problem_detail
+def admin_signin():
+    return app.manager.admin_signin_controller.signin()
+
+
+@app.route('/admin/works/<data_source>/<identifier>/details', methods=['GET'])
```
leonardr: I think you can lose the `/details` bit.

PR #92, opened 2016-03-07
Sync before claiming no active loan

leonardr:
This branch changes the CirculationAPI so that if a patron tries to fulfill a book for which they have no active loan, we sync the patron's bookshelf to make _sure_ they have no active loan before giving them an error. There are a number of edge cases where the patron may have a legitimate loan that we just don't know about.

### DIFF
--- a/app.py
+++ b/app.py
@@ -8,6 +8,7 @@ import flask
 from flask import (
     Flask, 
     Response,
+    redirect,
 )
 
 from config import Configuration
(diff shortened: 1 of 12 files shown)


aslagle:
The fulfill change looks good :+1:

bibliographic_refresh is for something else?

leonardr:
Yeah, I'm just adding a circulation version of a script i already wrote in core.

PR #94, opened 2016-03-07
Sync only finds current patrons loans

leonardr:
This branch was originally designed to diagnose a problem on an active site, and I've still got the debugging code in place, but while writing it I think I found the problem.

sync_bookshelf() starts with a set of `local_loans` and `local_holds` for the given patron. It asks the external data sources about the list of loans _they_ have for the given patron, and any loans not on the remote list get removed locally.

...except... it doesn't limit the `local_loans` and `local_holds` to the given patron. It gets _all_ the active loans and holds. So every time someone syncs their bookshelf they're destroying the local records of every other patron's loans, but not their own.

This branch fixes that stupid bug.


aslagle:
:+1:

PR #99, opened 2016-03-10
Moved scripts back to top-level and fixed paths. (Fixes #98)

### LINKED ISSUE #98: Update scripts for new api module

aslagle:
Scripts aren't part of the api so I thought it made sense to move them back.


leonardr:
This makes sense.

PR #100, opened 2016-03-11
Open access cdn

leonardr:
This branch uses a CDN to link to open-access content, if such a CDN is defined in the configuration.


tests/test_opds.py
```diff
@@ -40,6 +44,41 @@
...
+        # becomes the `href` of an open-access `link` tag.
+        [lpdm] = self.work.license_pools[0].delivery_mechanisms
+        link_tag = self.annotator.open_access_link(lpdm)
+        eq_(lpdm.resource.url, link_tag.get('href'))
+
+        # If we have a CDN set up for open-access links, the CDN hostname
+        # replaces the original hostname.
+        with temp_config() as config:
+            cdn_host = "https://cdn.com/"
+            set_trace()
```
aslagle: you should take this out

tests/test_opds.py
```diff
@@ -9,6 +9,10 @@
 import feedparser
 from . import DatabaseTest
 
+from config import (
```
aslagle: this needs to be api.config now

aslagle:
:+1:

PR #104, opened 2016-03-15
Added edit route and linked to it in admin details.

aslagle:
This connects to #103

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -157,11 +157,29 @@ class WorkController(CirculationManagerController):
         if isinstance(pool, ProblemDetail):
             return pool
         work = pool.work
+
         annotator = AdminAnnotator(self.circulation)
         return entry_response(
             AcquisitionFeed.single_entry(self._db, work, annotator)
         )
 
+    def edit(self, data_source, identifier):
+        """Edit a work's metadata."""
+
+        pool = self.load_licensepool(data_source, identifier)
+        if isinstance(pool, ProblemDetail):
+            return pool
+        work = pool.work
+        changed = False
+
+        new_title = flask.request.form.get("title")
+        if new_title and work.title != new_title:
+            work.primary_edition.title = new_title
+            changed = True
+        
+        if changed:
+            work.calculate_opds_entries()
+        return Response("", 200)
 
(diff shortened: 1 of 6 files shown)


api/admin/controller.py
```diff
@@ -157,11 +157,29 @@ def details(self, data_source, identifier):
...
+            return pool
+        work = pool.work
+        changed = False
+
+        new_title = flask.request.form.get("title")
+        if new_title and work.title != new_title:
+            work.primary_edition.title = new_title
+            changed = True
+        
+        if changed:
```
leonardr: An edition has a `sort_title` as well as a `title`. Usually the `sort_title` and `title` are the same, so if the `title` changes, so should the `sort_title`.

But if `sort_title` and `title` are not the same to begin with, then changing the `title` almost always implies a change to the `sort_title`. Most of the time this change can be done automatically because it's a matter of moving "A", "An", or "The" from the beginning of `title` to the end. But this isn't always the case, and sometimes the problem is that the `title` is correct but the `sort_title` is wrong.

You need to handle all these cases:

a) sort_title and title are the same, and a change in one is reflected in the other
b) sort_title and title are the same but should not be the same
c) sort_title needs to be edited and title should stay the same, or vice versa

It's probably best to do this in a separate branch, and maybe this should even be a backlog item for now.

api/admin/controller.py
```diff
@@ -157,11 +157,29 @@ def details(self, data_source, identifier):
...
+            work.calculate_opds_entries()
```
leonardr: The old data will also be kept in the materialized views (at least until the views are refreshed) and the search index. I don't think it's worth doing anything about the materialized views, but you should update the work's entry in the search index.

This won't happen just from updating the title, but you may also need to call Work.calculate_presentation() after an edit operation.

api/admin/controller.py
aslagle: https://github.com/NYPL-Simplified/circulation/issues/106

api/admin/controller.py
aslagle: Fixed

skomputer:
:+1:

PR #105, opened 2016-03-15
Admin metadata refresh

courte:
- A new route for refreshing a book's metadata, included in its Admin-annotated OPDS entry
- Metadata Wrangler-related problem details
- A test refactoring that made more sense before, but is now just *here*.
- Annoying whitespace changes to make reviewing more difficult!!

### DIFF
--- a/api/problem_details.py
+++ b/api/problem_details.py
@@ -198,7 +198,7 @@ METADATA_REFRESH_PENDING = pd(
 )
 
 METADATA_REFRESH_FAILURE = pd(
-      "http://librarysimplified.org/terms/problem/metadata-refresh-pending",
+      "http://librarysimplified.org/terms/problem/metadata-refresh-failure",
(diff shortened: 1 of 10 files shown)


api/problem_details.py
```diff
@@ -189,3 +189,17 @@
...
+
+METADATA_REFRESH_PENDING = pd(
+      "http://librarysimplified.org/terms/problem/metadata-refresh-pending",
+      201,
+      "Metadata refresh pending.",
+      "The Metadata Wrangler is looking for new data. Check back later.",
+)
+
+METADATA_REFRESH_FAILURE = pd(
+      "http://librarysimplified.org/terms/problem/metadata-refresh-pending",
```
aslagle: should this be metadata-refresh-failure?

aslagle:
:+1:

PR #108, opened 2016-03-17
Allow 3M Monitor to accept a default start date

courte:
Because the ThreeMEventMonitor is also how we keep track of ThreeM circulation overall, it needs to be able to be initialized to gather all data possible given a particular start date. This update allows an administrator to pass in a date and begin gathering circulation events (and create `licensepools` & `identifiers`) from that date until now.

### DIFF
--- a/api/threem.py
+++ b/api/threem.py
@@ -5,7 +5,6 @@ import datetime
 import os
 import re
 import logging
-import sys
 
 from nose.tools import set_trace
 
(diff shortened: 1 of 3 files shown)


api/threem.py
```diff
@@ -603,15 +604,38 @@ class ThreeMEventMonitor(Monitor):
     associated with it until the ThreeMCirculationMonitor runs.
     """
 
+    TWO_YEARS_AGO = datetime.timedelta(365*2)
+    name = "3M Event Monitor"
+
     def __init__(self, _db, default_start_time=None,
                  account_id=None, library_id=None, account_key=None):
+        if not default_start_time:
+            default_start_time = self.create_default_start_time(sys.argv[1:])
```
leonardr: For testability I like to keep the sys.argv stuff in the python script that's kept in bin/, not in these classes (the main exception is IdentifierInputScript, where the sys.argv management is in its own method). This isn't a big deal but see if you can move this to 3m_monitor.

api/threem.py
```diff
@@ -603,15 +604,38 @@ class ThreeMEventMonitor(Monitor):
...
+    @classmethod
+    def create_default_start_time(cls, date_args=None):
+        """Sets the default start time if it's passed as an argument.
+
+        The command line date argument should have the format YYYY-MM-DD.
+        """
+        if date_args:
+            date = date_args[0]
+            try:
+                year, month, day = [int(digit) for digit in date.split('-')]
```
leonardr: Use datetime.strptime("%Y-%m-%d", date) here

api/threem.py
```diff
@@ -603,15 +604,38 @@ class ThreeMEventMonitor(Monitor):
...
+                return datetime.datetime(year, month, day)
+            except ValueError:
+                # Date argument wasn't in the proper format.
+                two_years_ago = datetime.datetime.utcnow() - cls.TWO_YEARS_AGO
+                logging.getLogger(cls.name).warn(
```
leonardr: You should be able to use self.log anywhere inside a Script method.

leonardr: Oh, but it's a class method. Yesterday I ran into a similar problem. It's tough to create Script objects during tests; we need to fix that. This is fine for now.

courte: I set it up like this because the log requires a `service_name` property, and it hasn't yet been created when this method is called. I'll investigate a better arrangement, though.

leonardr:
+1

PR #111, opened 2016-03-21
Added route for admins to sign back in from a new tab.

aslagle:
This connects to https://github.com/NYPL-Simplified/circulation-web/issues/7

I also changed signin to two words everywhere.


skomputer:
:+1:

PR #116, opened 2016-03-22
Complaint details

skomputer:


api/admin/controller.py
```diff
@@ -168,6 +170,24 @@ def details(self, data_source, identifier):
         return entry_response(
             AcquisitionFeed.single_entry(self._db, work, annotator)
         )
+        
+    def complaints(self, data_source, identifier):
+        """Return detailed complaint information for admins."""
+        
+        pool = self.load_licensepool(data_source, identifier)
```
leonardr: If you're starting with a LicensePool I don't know if it makes sense to consolidate complaints for all the LicensePools associated with that work. 

Complaints are registered with the LicensePool. Some can only be fixed by fixing the Work ("wrong audience"), some by fixing the Edition associated with the work ("wrong title") and some by fixing the LicensePool itself ("book won't render"). Right now there's generally one Edition and one LicensePool associated with each book, so treating all those errors the same way is okay for now... I guess I'm just worried about the future here.

api/admin/controller.py
```diff
@@ -168,6 +170,24 @@ def details(self, data_source, identifier):
...
+        if isinstance(pool, ProblemDetail):
+            return pool
+        work = pool.work
+        counter = self._count_complaints_for_work(work)
+        response = dict({
+            "book": {
```
leonardr: I use "book" in code comments as a kind of shorthand, but in code and data I try not to use that term, since the 'book' here could just as easily be an audiobook or borrowable video. Can you use a more generic term, or even an internal term like 'license_pool'?

leonardr:
Looks good, I just brought up a couple things I'd like you to think about.

api/admin/controller.py
skomputer: how about "work"? "book" felt awkward when i wrote it, but that's what we use in circulation-web, which might be a problem too.

leonardr: "work" is no good here because a Work is a specific thing in this system which is not identified by data source + identifier. Either Edition or LicensePool can be identified by data source + identifier. I would maybe call it 'edition'.

api/admin/controller.py
skomputer: aha, i'll just return complaints for that one license pool if that's better.

leonardr: I think that's better, but it really depends on how you handle a future case where there are 3 different editions of Frankenstein with different overlapping sets of complaints. But since it's the future you can mostly worry about it later.

leonardr:
+1

PR #117, opened 2016-03-23
Axis 360 bibliographic coverage provider

leonardr:
This branch is a companion to https://github.com/NYPL-Simplified/server_core/pull/124: it removes the code that was moved into core.


aslagle:
:+1:

PR #120, opened 2016-03-25
Change calculate_presentation() call to use a PresentationCalculationPolicy.

leonardr:


api/admin/controller.py
```diff
@@ -183,7 +184,14 @@ def edit(self, data_source, identifier):
...
         if changed:
-            work.calculate_presentation(calculate_opds_entry=True)
+            # Even if the presentation doesn't visibly change, we want
+            # to regenerate the OPDS entries and update the search
+            # index for the work, because that might be the 'real'
+            # problem the user is trying to fix.
+            policy = PresentationCalculationPolicy(
+                regenerate_opds_entries=True, update_search_index=True,
+            )
+            work.calculate_presentation()
```
courte: You need to pass the policy here. :sunglasses:

PR #122, opened 2016-03-25
Availability repair script

leonardr:
This branch creates a new 'repair' script for checking and updating the availability of a small number of specific books. It's useful when debugging failure or slowness in the automates systems used to keep availability information current.


scripts.py
```diff
@@ -64,9 +68,8 @@
 )
 from api.axis import (
     Axis360API,
-    Axis360BibliographicCoverageProvider,
 )
-
+from core.axis import Axis360BibliographicCoverageProvider
```
skomputer: this is a duplicate import

skomputer:
+1

PR #124, opened 2016-03-28
Fix a test that broke when the core API changed.

leonardr:


skomputer:
This is a duplicate of #121.

PR #128, opened 2016-03-31
Service status log script

courte:
All the timing of methods you could ask for! Or, much of it! Or, well... Two specific cases.

There's a monitor to check the request times of our various third-party behemoths when it comes to loans and other patron activity, as well as a script for timing the duration of checkout out, fulfilling, and checking in a book.

Tests? What are thooooose?

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -36,6 +36,8 @@ from core.app_server import (
 )
 from core.opds import AcquisitionFeed
 from opds import AdminAnnotator, AdminFeed
+from collections import Counter
+
 
 def setup_admin_controllers(manager):
     """Set up all the controllers that will be used by the admin parts of the web app."""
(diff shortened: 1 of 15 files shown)


leonardr:
My change is not related to this branch; it's an error I discovered while testing. I told `service_checkout_status` to check out an Overdrive audio book by mistake, and it ran into an error while formatting the 'this is an audio book' message.

leonardr:
This looks good and I've verified that the scripts work. Two things before you merge.

1. Please put the scripts in a new `timing/` subdirectory. Call them `timing/borrow` and `timing/patron_activity`. I would like them grouped in a subdirectory because I anticipate creating more of them in the future.
2. It's not a big deal but I don't understand why one of your scripts is a Monitor and the other is a regular Script. What's your rationale?

leonardr:
:smile_cat:

PR #129, opened 2016-04-01
Changed migration to subclass script.

aslagle:
Made search index alias migration a subclass of script so it can access config.


leonardr:
+1

PR #131, opened 2016-04-05
Commit changes made during startup

leonardr:
This branch fixes https://github.com/NYPL-Simplified/server_core/issues/154 by making sure that a)  `CirculationManager` is not initialized more than once when running with the reloader turned on, b) the database session is committed after `CirculationManager` is initialized, ensuring that documents like the Overdrive list of library collections only ever have to be obtained once.

As a bonus, this cuts startup time in half (from 6 seconds to 3 seconds on my computer).

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -43,7 +43,7 @@ def run():
         import socket
         socket.setdefaulttimeout(None)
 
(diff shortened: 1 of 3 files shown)


aslagle:
:+1:

PR #135, opened 2016-04-07
More book edit fields

skomputer:


api/admin/controller.py
```diff
@@ -201,7 +203,21 @@ def edit(self, data_source, identifier):
...
+            work.primary_edition.publisher = unicode(new_publisher)
+            changed = True
+
+        new_summary = flask.request.form.get("summary")
+        if new_summary != work.summary_text:
+            staff_data_source = DataSource.lookup(self._db, DataSource.LIBRARY_STAFF)
+            (link, is_new) =  work.primary_edition.primary_identifier.add_link(
+                Hyperlink.DESCRIPTION, None, 
+                staff_data_source, content=new_summary)
+            work.set_summary(link.resource)
```
leonardr: We're going to need something to make sure that this summary continues to be _the_ summary even if Work.calculate_presentation() is called again. Doesn't have to be in this branch though.

aslagle:
:+1:

Issue #132, opened 2016-04-06
Allow ThreeM Monitor to initialize itself


courte:
If the ThreeM Event Monitor doesn't have any timestamps in the database, it should start looking for events from two years ago and work its way forward. Having the monitor do this for itself (as opposed to depending on the script we use currently) will simplify the initialization process by a lot.

PR #137, opened 2016-04-08
Initialize 3M event monitor #132

courte:
The 3M Events Monitor needs to run from a much earlier date if it is being initialized for the first time. Before, this was being managed by passing in a date on the command line. Now, the Monitor checks to see if it's been initialized. If it hasn't ever created a Timestamp before, it initializes from two years before the date it's being run.


leonardr:
:+1:

PR #140, opened 2016-04-13
Stop unconfigured services from logging errors

courte:
Unconfigured APIs were turning up as errors in the logs, which distracts from the work the script is actually doing. This branch ignores unconfigured APIs completely. Who needs em??!

It also uses an API's Data Source name, instead of whatever higgledy-piggledy was hard-coded there before (i.e. I called it "Axis" instead of "Axis 360" 😑 ).

### DIFF
--- a/api/services.py
+++ b/api/services.py
@@ -47,19 +47,17 @@ class ServiceStatus(object):
             return status
         patron = patrons[0]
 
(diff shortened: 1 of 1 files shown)


aslagle:
:+1:

PR #144, opened 2016-04-15
Add cutoff_time to CoverageProvider constructors.

leonardr:
Note that any other CoverageProviders that override `__init__` will need to be changed as well; there may be some in metadata.


courte:
:+1:

PR #145, opened 2016-04-15
Stop using configuration in test versions of the Google Oauth service

courte:
This will fix the two test errors that pop up with configuration files that don't have Google Oauth set up; it bypasses environmental configuration entirely, instead using knowledge of the test environment to build a test-friendly Google Oauth "Client".


aslagle:
:+1:

PR #149, opened 2016-04-20
Admin edit summary

### LINKED ISSUE #146: Admin edit: setting summary to empty string causes 500 error

aslagle:
This uses https://github.com/NYPL-Simplified/server_core/pull/189

This fixes #146, and also a bug where staff couldn't edit the summary multiple times.

I also fixed the problem where a bunch of the admin tests would fail any time an earlier test had failed.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -230,10 +230,24 @@ class WorkController(CirculationManagerController):
 
         new_summary = flask.request.form.get("summary")
         if new_summary != work.summary_text:
-            (link, is_new) =  work.primary_edition.primary_identifier.add_link(
-                Hyperlink.DESCRIPTION, None, 
-                staff_data_source, content=new_summary)
-            work.set_summary(link.resource)
+            old_summary = None
+            if work.summary and work.summary.data_source == staff_data_source:
+                old_summary = work.summary
+
+            if new_summary:
+                (link, is_new) =  work.primary_edition.primary_identifier.add_link(
+                    Hyperlink.DESCRIPTION, None, 
+                    staff_data_source, content=new_summary)
+                work.set_summary(link.resource)
+            else:
+                work.set_summary(None)
+
+            # Delete previous staff summary
+            if old_summary:
+                for link in old_summary.links:
+                    self._db.delete(link)
+                self._db.delete(old_summary)
+
             changed = True
 
         if changed:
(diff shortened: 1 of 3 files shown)


leonardr:
Got it.

aslagle:
I changed this to create a description resource with an empty string instead of setting the resource to None, but this currently won't let you delete an existing summary from another source. We'll need to make sure it uses the staff data source summary even if it's empty.

aslagle:
I changed Identifier.evaluate_summary_quality in core so it could privilege staff above the license source.

api/admin/controller.py
```diff
@@ -228,12 +228,23 @@ def edit(self, data_source, identifier):
...
         if new_summary != work.summary_text:
+            old_summary = None
+            if work.summary and work.summary.data_source == staff_data_source:
+                old_summary = work.summary
+
             (link, is_new) =  work.primary_edition.primary_identifier.add_link(
                 Hyperlink.DESCRIPTION, None, 
                 staff_data_source, content=new_summary)
             work.set_summary(link.resource)
+
```
leonardr: If you change the description multiple times, what happens? My mental image is that the old descriptions are going to stick around and potentially confuse things, because nothing stops an identifier from having multiple DESCRIPTION links from the same data source.

tests/admin/test_controller.py
```diff
@@ -99,10 +98,34 @@ def test_edit(self):
...
+            flask.request.form = ImmutableMultiDict([
+                ("title", "New title"),
+                ("audience", "Young Adult"),
+                ("summary", "abcd")
+            ])
+            response = self.manager.admin_work_controller.edit(lp.data_source.name, lp.identifier.identifier)
+            eq_(200, response.status_code)
+            eq_("abcd", self.english_1.summary_text)
+            assert 'New summary' not in self.english_1.simple_opds_entry
+
```
leonardr: That is, in my previous comment I'm saying that this might work by accident. Unless you do something with the old description I think it might still be hanging around, and at some point in the future it might come back.

api/admin/controller.py
aslagle: That's what the code below this is for. If the work's previous summary has the staff data source, it and its link get deleted.

leonardr: Duh! OK.

aslagle: I guess there might be a race condition where you could end up with two?

leonardr: That's true but I don't think there's anything you can do within this code to solve it 100%. The two links would be created in separate database transactions.

PR #151, opened 2016-04-21
added viewport meta tag to admin page

skomputer:
This branch adds a viewport meta tag necessary for responsive design. It also capitalizes "Books" in the title of the suppressed books feed.


aslagle:
:+1:

PR #154, opened 2016-04-21
Multiple admin sessions

skomputer:
This branch authenticates a request using the email instead of the access token in the session. This is the smallest possible change that enables multiple clients to simultaneously authenticate using the same Google account, which is necessary for circulation-web's selenium-based Firefox and Chrome tests to run in parallel.


aslagle:
I think we should add an expiration to the flask sessions before merging this: http://flask.pocoo.org/docs/0.10/api/#flask.session.permanent

aslagle:
I finally added the flask session expiration. I also ended up having to move CSRF tokens out of the session, because I don't want them to expire.

api/admin/routes.py
```diff
@@ -23,10 +23,15 @@
...
 from StringIO import StringIO
 import urllib
+from datetime import timedelta
 
 # The secret key is used for signing cookies for admin login
 app.secret_key = Configuration.get(Configuration.SECRET_KEY)
 
+# An admin's session will expire after this amount of time and
+# the admin will have to log in again.
+app.permanent_session_lifetime = timedelta(hours=4)
```
aslagle: I'm not sure how long this should be.

api/admin/routes.py
```diff
@@ -43,7 +48,14 @@ def requires_admin(f):
     @wraps(f)
     def decorated(*args, **kwargs):
         if 'setting_up' in kwargs:
-            setting_up = kwargs.pop('setting_up')
+            # If the function also requires a CSRF token,
+            # setting_up needs to stay in the arguments for
+            # the next decorator. Otherwise, it should be
+            # removed before the route function.
```
aslagle: When I tested this I realized the decorators were in the wrong order before. The admin check needs to happen before the CSRF token, so it won't check the CSRF token if you're not logged in. That means the requires_admin decorator can't always remove the 'setting_up' argument anymore.

api/admin/routes.py
leonardr: I would set it at at 9 hours so a librarian doesn't have to log in again during their workday.

api/admin/routes.py
```diff
@@ -295,8 +308,7 @@ def admin_view(collection=None, book=None, **kwargs):
     setting_up = (app.manager.admin_sign_in_controller.auth == None)
     if not setting_up:
         admin = app.manager.admin_sign_in_controller.authenticated_admin_from_request()
-        csrf_token = app.manager.admin_sign_in_controller.get_csrf_token()
-        if isinstance(admin, ProblemDetail) or csrf_token is None or isinstance(csrf_token, ProblemDetail):
```
leonardr: It looks like these extra checks were removed (and not added back below) because there's no way `generate_csrf_token()` or `get_csrf_token()` can return None or a ProblemDetail. Is that right?

api/admin/controller.py
```diff
@@ -112,9 +112,9 @@ def authenticated_admin_from_request(self):
         if not self.auth:
             return ADMIN_AUTH_NOT_CONFIGURED
 
-        access_token = flask.session.get("admin_access_token")
-        if access_token:
-            admin = get_one(self._db, Admin, access_token=access_token)
```
leonardr: With the removal of this stuff, are we using access_token for anything anymore?

leonardr:
I have some basic-level questions but the parts I can understand seem good.

api/admin/controller.py
aslagle: I don't think we are. We could probably stop storing it in the db.

api/admin/routes.py
aslagle: Right.

api/admin/controller.py
leonardr: I filed https://github.com/NYPL-Simplified/server_core/issues/503 for this.

aslagle: I'm modifying that table for my next branch so I can do that too.

Issue #159, opened 2016-04-27
Move preload list from content server to circulation


leonardr:
We've got a feature in the content server that creates a feed from a custom list of preload books. Aferdita requested that this feed be available from the circulation manager instead, so that it can be fetched at runtime rather than build time.

This will also make it easy for different libraries to put different books in their preload list.

If there's not already a link relation for linking to the preload feed, now is a good time to come up with one. This feed should be linked to, either from the main feed or the patron's bookshelf feed. I'm not sure which one makes more sense.

PR #161, opened 2016-04-28
Preload feed

aslagle:
This connects to #159

### DIFF
--- a/api/opds.py
+++ b/api/opds.py
@@ -4,6 +4,8 @@ from flask import url_for
 from lxml import etree
 from collections import defaultdict
 
+from sqlalchemy.orm import lazyload
+
 from config import Configuration
 from core.opds import (
     Annotator,
@@ -15,9 +17,12 @@ from core.opds import (
 )
 from core.model import (
     Identifier,
+    LicensePool,
     LicensePoolDeliveryMechanism,
     Session,
     BaseMaterializedWork,
+    Work,
+    Edition,
 )
 from core.lane import Lane
 from circulation import BaseCirculationAPI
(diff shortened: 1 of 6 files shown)


aslagle:
I'm going to move the link so it can be used before someone logs in.

aslagle:
Actually I already had the link in all the feeds - I wasn't seeing it because I was using the cached feed.

api/opds.py
```diff
@@ -556,3 +569,40 @@ def single_hold_feed(cls, circulation, hold, test_mode=False):
...
+    def page(cls, _db, title, url, annotator=None,
+             use_materialized_works=True):
+
+        """Create a feed of content to preload on devices."""
+        configured_content = Configuration.policy(Configuration.PRELOADED_CONTENT)
+
+        identifiers = [Identifier.parse_urn(_db, urn)[0] for urn in configured_content]
+        identifier_ids = [identifier.id for identifier in identifiers]
+
+        if use_materialized_works:
```
leonardr: This seems like copy-and-pasted code from Lane.materialized_works. Is it possible to refactor Lane.materialized_works to avoid the duplication?

Not a big deal and not necessary now, especially if this is just code you copied from content to circulation.

aslagle: Yeah, some of it is. I'm not using a lane here, so I would have to move that code somewhere else to use it.

aslagle: Maybe a good way to do this would be to make the preloads a custom list and use that to make a lane.

leonardr: Oh, I like that idea. Then librarians will be able to edit it with the editor.

leonardr: But this is fine for now. Let's file a list to turn preloads into a custom list.

aslagle:
I wonder if I should do something to leave out the preload link if you haven't configured any preloads?

leonardr:
I think we can do that at the same time we make the preloads a custom list.

PR #164, opened 2016-04-29
fixed group_uri() and added test

skomputer:
This branch fixes a bug in my previous refactor of `CirculationMangerAnnotator.group_uri()` and tests the particular scenario that failed.


aslagle:
:+1:

PR #165, opened 2016-04-29
added "up" links to complaints and suppressed feeds

skomputer:
This branch adds "up" links (with the same url and title as "start" links) to admin complaints and suppressed feeds.


aslagle:
:+1:

PR #166, opened 2016-04-29
Scope the database session

leonardr:
This branch makes `_db` a mandatory argument to the `CirculationManager` constructor. In the tests, we use the same database session we were using before. But in the actual app, we use a `flask_scoped_session` that creates a new database session for every request.

I made changes to bring the test setup and teardown up-to-date with my recent core branch.

I also make some changes designed to make tests pass more reliably. I had a couple cases where lists were in a semi-random order and occasionally I'd get the wrong order and the test would fail.

This is not ready to merge yet because I need to add at least some kind of test of the `flask_scoped_session`. My plan to use it for all the controller tests has fallen through but I should be able to do a basic test that two different requests use different sessions.

### DIFF
--- a/tests/test_controller.py
+++ b/tests/test_controller.py
@@ -122,7 +122,6 @@ class CirculationControllerTest(ControllerTest):
             "Quite British", "John Bull", language="eng", fiction=True,
             with_open_access_download=True
         )
-        print self.english_1.license_pools
         self.english_2 = self._work(
             "Totally American", "Uncle Sam", language="eng", fiction=False,
             with_open_access_download=True
--- a/tests/test_controller.py
+++ b/tests/test_controller.py
@@ -856,10 +856,10 @@ class TestScopedSession(ControllerTest):
 
             # But if we were to use flask_scoped_session to create a
             # brand new session, it would not see the Identifier,
-            # because it's running in a new request-specific.
+            # because it's running in a new request-specific
+            # transaction.
             new_session = self.app.manager._db.session_factory()
             eq_([], new_session.query(Identifier).all())
-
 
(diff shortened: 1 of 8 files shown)


aslagle:
The failing search test is using the fallback database search [here](https://github.com/NYPL-Simplified/server_core/blob/master/lane.py#L1154) which doesn't order the results.

tests/test_controller.py
```diff
@@ -81,14 +92,37 @@ def setup(self):
...
+    def setup(self):
+        super(CirculationControllerTest, self).setup()
 
         # Create two English books and a French book.
         self.english_1 = self._work(
             "Quite British", "John Bull", language="eng", fiction=True,
             with_open_access_download=True
         )
-
+        print self.english_1.license_pools
```
aslagle: you can take out this print

tests/test_controller.py
```diff
@@ -765,3 +798,59 @@ def test_preload(self):
...
+            # The Identifier immediately shows up in the session that
+            # created it.
+            [identifier] = session1.query(Identifier).all()
+            eq_("1024", identifier.identifier)
+
+            # It doesn't show up in self._db, the database session
+            # used by most other unit tests, because it was created
+            # within the (still-active) context of a Flask request,
+            # which happens within a nested database session.
+            eq_([], self._db.query(Identifier).all())
```
aslagle: and it also doesn't show up in app.manager._db?

leonardr: It does show up in app.manager._db, and I'm pretty sure that's because app.manager._db is managing the parent transaction.

aslagle:
:+1:

PR #167, opened 2016-05-02
Use client details for wrangler

### LINKED ISSUE #156: Use LS client details to request lookups from the Metadata Wrangler

courte:
Some changes included in this branch:

- Update circ to use new test setup from core
- Fix a couple tests broken due to changing database ordering
- Includes exception messages in CoverageFailures - not just the status code integer
- Adds a Coverage Provider-type reaper to remove newly-unlicensed works from an authenticated collection on the Metadata Wrangler (MW)
- Adjusts the existing MW Coverage Provider to re-run relicensed works.

The CoverageRecords created by these two providers work as a sort of in-collection/out-of-collection toggle. In collection, the Metadata Wrangler creates a record when works are added to the circ manager or when they're relicensed (even if the reaper has already removed them). When items exit the collection and no licenses are owned, the reaper deletes that record and adds a its own record.

One particular area of uncertainty is the name of the Metadata Wrangler Collection Reaper `data_source`. It needs to have a DataSource in order to create CoverageRecords, but it seems odd to have a whole separate source for this one particular action. Ideas in this (or any other) regard are welcome.

Fixes #156.

### DIFF
--- a/api/coverage.py
+++ b/api/coverage.py
@@ -76,10 +76,7 @@ class MetadataWranglerCoverageProvider(OPDSImportCoverageProvider):
         )
 
         if not metadata_lookup:
-            url = Configuration.integration_url(
-                Configuration.METADATA_WRANGLER_INTEGRATION
-            )
-            metadata_lookup = SimplifiedOPDSLookup(url)
+            metadata_lookup = SimplifiedOPDSLookup.from_config()
         self.lookup = metadata_lookup
 
         super(MetadataWranglerCoverageProvider, self).__init__(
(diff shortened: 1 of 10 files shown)


api/coverage.py
```diff
@@ -68,25 +70,43 @@ def __init__(self, _db, identifier_types=None, metadata_lookup=None,
...
-            self.coverage_source,
+            self.output_source,
             workset_size=20,
             cutoff_time=cutoff_time,
         )
 
+    @property
+    def items_that_need_coverage(self):
+        reaper_source = DataSource.lookup(
+            self._db, DataSource.METADATA_WRANGLER_COLLECTION
```
leonardr: Rather than create special data sources for the metadata wrangler collection and the reaper, how about using the 'operation' field of CoverageRecord? Then you can use METADATA_WRANGLER as the data source instead of creating a new data source.

leonardr:
:+1:

PR #168, opened 2016-05-03
Dont store data model objects in CirculationAPI

leonardr:
This branch fixes a problem where data model objects (DataSource objects) were being used after the database session that created them went out of scope. This caused bookshelf sync to fail silently.

The app server creates a single CirculationAPI object and uses it to manage all circulation-type interaction with external APIs. Previous to this branch, the CirculationAPI constructor looked up some DataSource objects and stored them in instance variables. But when an actual request came in, it was assigned a new database connection, and those DataSource objects were no longer usable.

This branch changes the CirculationAPI constructor so that it stores database names instead of DataSource objects. The DataSource objects are looked up as necessary using DataSource.lookup(). This hurts performance, but circulation transactions are relatively rare, so I'm not too worried about a number of redundant DataSource lookups, and this is something I think we can optimize in a later branch.

The tests in test_overdrive.py verify that the code still works as before, but I can't think of a way to test the specific case where the database connection that creates the CirculationAPI object goes out of scope.

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -107,24 +107,25 @@ class CirculationAPI(object):
         # to include loans from all licensed data sources.  We do not
         # need to include loans from open-access sources because we
         # are the authorities on those.
-        self.data_sources_for_sync = []
+        data_sources_for_sync = []
         if self.overdrive:
-            self.data_sources_for_sync.append(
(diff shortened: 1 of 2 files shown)


aslagle:
Looks good to me. I can't think of a way to test that case either.

PR #169, opened 2016-05-04
Don't join threads until all threads have been created.

leonardr:
I originally wrote the loan sync code to check each loan source (Overdrive, 3M, Axis, etc.) in a separate thread. The idea was to run the sync in approximately the time taken by the slowest data source, rather than the sum of the times taken by each data source.

Unfortunately a slight bug meant we haven't been seeing that performance improvement. Calling `Thread.join()` blocks the process until the thread completes, and I was calling `Thread.join()` on each thread right after calling `Thread.start()`. This meant that thread #1 had to complete before thread #2 even started.

This branch makes sure to call `Thread.start()` on all the threads before calling `Thread.join()`. I also converted the print statements in circulation.py to use the logging system, and added logging for exceptions that happen during sync (which are otherwise ignored as they happen inside a child thread).

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -5,6 +5,7 @@ from collections import defaultdict
 from threading import Thread
 import logging
 import re
+import time
 
 from core.model import (
     get_one,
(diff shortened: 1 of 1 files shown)


api/circulation.py
```diff
@@ -415,29 +416,48 @@ def patron_activity(self, patron, pin):
...
 
         threads = []
         import time
+        before = time.time()
         for api in self.apis:
             thread = PatronActivityThread(api, patron, pin)
             threads.append(thread)
         for thread in threads:
             thread.start()
+        for thread in threads:
```
leonardr: This is the essential line.

courte:
:+1:

PR #170, opened 2016-05-04
Fix circulation tests for composite

leonardr:
This is Darya's branch that gets the circulation manager to work with the licensepool-superceded core branch. Apparently it's mostly test fixes. We'll find out!

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -523,14 +523,13 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI):
             license_pool.identifier.identifier)
         edition.title = edition.title or book.get('title')
 
-        # if the pool does not already have a presetation edition, 
+        # If the pool does not already have a presentation edition, 
         # and if this edition is newly made, then associate pool and edition
         # as presentation_edition
         if ((not license_pool.presentation_edition) and is_new_edition): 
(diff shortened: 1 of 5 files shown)


api/overdrive.py
```diff
@@ -505,26 +505,38 @@ def update_licensepool(self, book):
...
-        if is_new:
+
+        # if the pool does not already have a presetation edition, 
+        # and if this edition is newly made, then associate pool and edition
+        # as presentation_edition
+        if ((not license_pool.presentation_edition) and is_new_edition): 
+            # NOTE:  I think it might not be guaranteed that edition will be set as 
+            # pool's presentation_edition here.  Also, let's keep an eye on this code 
+            # bit's performance speed.
+            edition_changed = license_pool.set_presentation_edition(None)
```
leonardr: There's no guarantee that _this_ edition will be made the license pool's presentation edition (although it should happen in every real instance), but by the time we `set_presentation_edition` there should be _a_ presentation edition of some kind.

So I would take out this NOTE.

leonardr:
Looks good pending a very small change and the approval of the core branch it uses.

aslagle:
This looks fine to me. It needs master merged again and it still needs the very small change Leonard proposed.

PR #171, opened 2016-05-04
Edit genres

skomputer:
This branch adds endpoints for updating a work's genres, retrieving a work's classifications, and retrieving a complete genre tree.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -41,6 +41,7 @@ from core.app_server import (
 from core.opds import AcquisitionFeed
 from opds import AdminAnnotator, AdminFeed
 from collections import Counter
+from core.classifier import genres
 
 
 def setup_admin_controllers(manager):
@@ -191,6 +192,33 @@ class WorkController(CirculationManagerController):
         })
         
         return response
+
+    def subjects(self, data_source, identifier):
+        """Return list of subjects for this work that are linked to genres."""
+
+        pool = self.load_licensepool(data_source, identifier)
+        if isinstance(pool, ProblemDetail):
+            return pool
+
+        results = Classification.for_work_with_genre(
+            self._db, pool.work)
+
+        data = []
+        for result in results:
+            data.append(dict({
+                "type": result.subject.type,
+                "name": result.subject.identifier,
+                "source": result.data_source.name,
+                "weight": result.weight
+            }))
+
+        return dict({
+            "book": {
+                "data_source": data_source,
+                "identifier": identifier
+            },
+            "subjects": data
+        })
 
     def edit(self, data_source, identifier):
         """Edit a work's metadata."""
@@ -420,3 +448,17 @@ class FeedController(CirculationManagerController):
             pagination=pagination
         )
         return feed_response(opds_feed)
+
+    def genres(self):
+        data = dict({
+            "Fiction": dict({}),
+            "Nonfiction": dict({})
+        })
+        for name in genres:
+            top = "Fiction" if genres[name].is_fiction else "Nonfiction"
+            data[top][name] = dict({
+                "name": name,
+                "parents": [parent.name for parent in genres[name].parents],
+                "subgenres": [subgenre.name for subgenre in genres[name].subgenres]
+            })
+        return data
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -418,12 +418,13 @@ class WorkController(CirculationManagerController):
             return COMPLAINT_ALREADY_RESOLVED
         return Response("", 200)
 
-    def update_genres(self, data_source, identifier, new_genres):
+    def update_genres(self, data_source, identifier):
         pool = self.load_licensepool(data_source, identifier)
         if isinstance(pool, ProblemDetail):
             return pool
         work = pool.work
         staff_data_source = DataSource.lookup(self._db, DataSource.LIBRARY_STAFF)
+        new_genres = flask.request.form.get("genres")
 
         current_classifications = Classification \
             .for_work_with_genre(work) \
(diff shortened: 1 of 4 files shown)


api/problem_details.py
```diff
@@ -224,3 +224,17 @@
...
+
+GENRE_NOT_FOUND = pd(
+    "http://librarysimplified.org/terms/problem/complaint-already-resolved",
+    status_code=404,
+    title="Genre not found.",
+    detail="The specified work does not have the specified genre."
+)
+
+INCOMPATIBLE_GENRE = pd(
+    "http://librarysimplified.org/terms/problem/complaint-already-resolved",
```
aslagle: these need their own uris

api/admin/controller.py
```diff
@@ -233,7 +264,7 @@ def edit(self, data_source, identifier):
                 data_source=staff_data_source,
                 subject_type=Subject.SIMPLIFIED_FICTION_STATUS,
                 subject_identifier=fiction_term,
-                weight=STAFF_WEIGHT * 100,
+                weight=WorkController.STAFF_WEIGHT * 100,
```
aslagle: We can come back to this later, but fiction status and genres should be edited in the same request. That way changing fiction status won't need to invalidate staff-selected genres, and we won't need to give it a higher weight.

api/admin/controller.py
```diff
@@ -388,6 +419,69 @@ def resolve_complaints(self, data_source, identifier):
...
+            genre, is_new = Genre.lookup(self._db, name)
+            if not isinstance(genre, Genre):
+                return GENRE_NOT_FOUND
+            if work.fiction != genres[name].is_fiction:
+                return INCOMPATIBLE_GENRE
+
+        # delete existing staff classifications for genres that aren't being kept
+        for c in current_staff_classifications:
+            name = c.subject.genre.name
+            if work.fiction != genres[name].is_fiction or name not in new_genres:
```
aslagle: you can remove the fiction check here

api/admin/controller.py
```diff
@@ -388,6 +419,69 @@ def resolve_complaints(self, data_source, identifier):
...
+                self._db.delete(c)
+
+        # add new staff classifications for new genres
+        for genre in new_genres:
+            if genre not in current_staff_genres:
+                classification = work.primary_edition.primary_identifier.classify(
+                    data_source=staff_data_source,
+                    subject_type=Subject.SIMPLIFIED_GENRE,
+                    subject_identifier=genre,
+                    weight=WorkController.STAFF_WEIGHT * 100
```
aslagle: you don't need to multiply by 100

api/admin/routes.py
```diff
@@ -63,6 +63,15 @@ def admin_sign_in():
 def work_details(data_source, identifier):
     return app.manager.admin_work_controller.details(data_source, identifier)
 
+@app.route('/admin/works/<data_source>/<identifier>/classifications', methods=['GET'])
```
aslagle: This only returns genre classifications so maybe it should be called genres? Or do you want to expand it later?

api/problem_details.py
```diff
@@ -224,3 +224,17 @@
     title="Complaint already resolved.",
     detail="You can't resolve a complaint that is already resolved."
 )
+
+GENRE_NOT_FOUND = pd(
+    "http://librarysimplified.org/terms/problem/complaint-already-resolved",
+    status_code=404,
+    title="Genre not found.",
+    detail="The specified work does not have the specified genre."
```
aslagle: maybe something like "one of the submitted genres does not exist"

tests/admin/test_controller.py
```diff
@@ -179,6 +183,68 @@ def test_edit_invalid_input(self):
             eq_(400, response.status_code)
             eq_(INVALID_EDIT.uri, response.uri)
 
+    def test_update_genres(self):
+        # start with no genres
```
aslagle: it would be good to start with something and make sure the staff changes override it

tests/admin/test_controller.py
```diff
@@ -179,6 +183,68 @@ def test_edit_invalid_input(self):
...
+        previous_genres = requested_genres
+
+        # try to add a nonfiction genre
+        with self.app.test_request_context("/"):
+            requested_genres = ["Drama", "Women's Fiction", "Cooking"]
+            form = MultiDict()
+            for genre in requested_genres:
+                form.add("genres", genre)
+            flask.request.form = form
+            response = self.manager.admin_work_controller.update_genres(lp.data_source.name, lp.identifier.identifier)
```
aslagle: check that the response was the problem detail

api/admin/controller.py
```diff
@@ -388,6 +419,68 @@ def resolve_complaints(self, data_source, identifier):
...
+                    data_source=staff_data_source,
+                    subject_type=Subject.SIMPLIFIED_GENRE,
+                    subject_identifier=genre,
+                    weight=WorkController.STAFF_WEIGHT
+                )
+
+        # add NONE classification if we aren't keeping any genres
+        staff_classifications = work.classifications_with_genre() \
+            .filter(Classification.data_source_id == staff_data_source.id) \
+            .all()
```
aslagle: This should be the same as checking in len(new_genres) is 0, which would save a database query. Maybe it's safer to do this to make sure?

api/problem_details.py
```diff
@@ -224,3 +224,17 @@
     title="Complaint already resolved.",
     detail="You can't resolve a complaint that is already resolved."
 )
+
+GENRE_NOT_FOUND = pd(
+    "http://librarysimplified.org/terms/problem/complaint-already-resolved",
```
aslagle: this also needs a new uri

aslagle:
:+1:

PR #172, opened 2016-05-06
 Don't roll back database session if database session hasn't been created yet 

leonardr:
This branch fixes the display of errors that happen early on in the startup process, by removing code that can mask that error with another, more stupid error--the attempt to roll back a database session that hasn't been initialized yet.

I also added a monkeypatch taken from an unreleased version of Flask, which stops Flask from handling requests while the @app.before_first_request functions are running.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -271,7 +271,9 @@ class CirculationManagerController(object):
         try:
             patron = self.authenticated_patron(header.username, header.password)
         except RemoteInitiatedServerError,e:
-            return problem(REMOTE_INTEGRATION_FAILED, e.message, 503)
+            return REMOTE_INTEGRATION_FAILED.detailed(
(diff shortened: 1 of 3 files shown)


api/controller.py
```diff
@@ -271,7 +271,9 @@ def authenticated_patron_from_request(self):
         try:
             patron = self.authenticated_patron(header.username, header.password)
         except RemoteInitiatedServerError,e:
-            return problem(REMOTE_INTEGRATION_FAILED, e.message, 503)
```
leonardr: REMOTE_INTEGRATION_FAILED is already a ProblemDetail. This was creating a ProblemDetail that contained another ProblemDetail, which makes no sense.

aslagle:
:+1:

PR #173, opened 2016-05-10
Upgrade command line arguments

leonardr:
This branch makes minor changes and improvements that go along with https://github.com/NYPL-Simplified/server_core/pull/223.

* The `MetadataWranglerCollectionReaper` constructor needs to take the same arguments as other `CoverageProvider` constructors, because the `RunCoverageProviderScript` passes in those arguments when it instantiates the `CoverageProvider`.

* The `AvailabilityRefreshScript` no longer needs to call `parse_identifiers()`; it just needs to call `parse_command_line(self._db)` and the identifiers will be properly loaded from the database.


aslagle:
Looks good

PR #174, opened 2016-05-10
Dont prestore data source

leonardr:
This fixes a serious issue that prevents new Adobe vendor IDs from being registered. There was one more piece of code that was keeping a database object: the Adobe controller. In a multithreaded environment you can't hold on to this information outside of a request context. I changed it to look up the DataSource every time.


aslagle:
:+1:

PR #175, opened 2016-05-11
Added repair script for work consolidations.

aslagle:
See https://github.com/NYPL-Simplified/server_core/pull/227


leonardr:
I'm a little concerned that this makes it really easy to wipe out all the works in the system, at which point they have to be recreated. :+1: but I've added a backlog issue for this: https://github.com/NYPL-Simplified/server_core/issues/228

PR #176, opened 2016-05-12
Improve date parsing robustness

leonardr:
This branch stops a crash of the 3M EventParser when the patron ID and/or end date is `None`. These fields were treated as required, but they're not actually required, and they're no longer present on `license_add` events.

The branch also improves test coverage a bit--there was previously no test of the 3M EventParser at all.


courte:
:+1:

PR #179, opened 2016-05-13
3M books are never open access. Record this fact.

leonardr:
This branch makes sure that newly created 3M LicensePools are marked as non-open-access. (Their .open_access was being left blank.)

Also, don't record a work as being 'removed' when its `.licenses_owned` was already set to zero on the local side--it was already removed and you're filling the log with redundant junk.


aslagle:
:+1: to the changes but I think this branch has a bad submodule reference

PR #180, opened 2016-05-13
Edit classifications

skomputer:
This branch combines audience and fiction status editing with genre editing into a single controller method.


api/admin/controller.py
```diff
@@ -419,55 +320,178 @@ def resolve_complaints(self, data_source, identifier):
...
+                if c.subject.type == Subject.SIMPLIFIED_FICTION_STATUS:
+                    self._db.delete(c)
+
+            # Create a new classification with a high weight (higher than genre)
+            fiction_term = "Fiction" if new_fiction else "Nonfiction"
+            classification = primary_identifier.classify(
+                data_source=staff_data_source,
+                subject_type=Subject.SIMPLIFIED_FICTION_STATUS,
+                subject_identifier=fiction_term,
+                weight=WorkController.STAFF_WEIGHT * 100,
```
aslagle: now that staff can't create incompatible fiction/genre combinations, fiction classifications shouldn't need the extra high weight.

api/problem_details.py
```diff
@@ -237,4 +237,11 @@
     status_code=409,
     title="Incompatible genre.",
     detail="The genre is incompatible with the fiction status of the work."
+)
+
+EROTICA_FOR_ADULTS_ONLY = pd(
+    "http://librarysimplified.org/terms/problem/erotica-for-adults-only",
+    status_code=409,
```
aslagle: 400 might be better for these since the things that conflict are all in the same request

aslagle:
:+1:

PR #181, opened 2016-05-13
Bring back bibliographic coverage scripts

leonardr:
This branch brings back bibliographic coverage scripts for each licensed book source. Previously I thought these were not necessary in normal operation and could be treated as repair scripts, but they totally are necessary. The bibliographic coverage scripts are how we find out about format availability--information that the licensed book sources make available but that is not captured by the metadata wrangler.

### DIFF
--- a/bin/3m_bibliographic_coverage
+++ b/bin/3m_bibliographic_coverage
@@ -1,0 +1,10 @@
+#!/usr/bin/env python
+"""Make sure all 3M books have bibliographic coverage."""
+import os
+import sys
+bin_dir = os.path.split(__file__)[0]
+package_dir = os.path.join(bin_dir, "..")
+sys.path.append(os.path.abspath(package_dir))
+from core.scripts import RunCoverageProviderScript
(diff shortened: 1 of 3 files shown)


courte:
:+1:

courte:
BPL had 3M books just fine without the 3M bibliographic script; could we only need the Overdrive one?

leonardr:
You're right, it's not necessary for 3M books. When we discover a new book we explicitly do a bibliographic lookup and set the delivery mechanisms. (api/threem.py handle_event()) I think it would be better if that code called ensure_coverage on the appropriate BibliographicCoverageProvider. That would make more efficient use of the 3M API since we would get all the bibliographic informaiton. On the flip side, we might want to ensure bibliographic coverage of new Overdrive books as we discover them. I think I kept those two operations separate to reduce the chance of an error in metadata lookup causing us to lose real-time circulation data.

In any event I think we should have this script in place as a safety measure, to cover anything that falls through the cracks.

PR #183, opened 2016-05-13
Move bibliographic_coverage back out of `repair/` now that it can work as a general-purpose bibliographic coverage script.

leonardr:


aslagle:
:+1:

Issue #186, opened 2016-05-17
Work lookup gives 500 error


leonardr:
http://qa.circulation.librarysimplified.org/works/Gutenberg/128 gives a 500 error. It's a rarely used function but it shouldn't be giving 500.

PR #188, opened 2016-05-17
Fix work lookup 500

leonardr:
This branch fixes https://github.com/NYPL-Simplified/circulation/issues/186 by making `CirculationManagerAnnotator.annotate_work_entry` a little more robust. It's not the worst error but the fact that it causes 500 errors creates noise that can mask other, more serious errors in our reports.

### DIFF
--- a/bin/repair/work_consolidation
+++ b/bin/repair/work_consolidation
@@ -6,4 +6,4 @@ bin_dir = os.path.split(__file__)[0]
 package_dir = os.path.join(bin_dir, "..", "..")
 sys.path.append(os.path.abspath(package_dir))
 from core.scripts import WorkConsolidationScript
-WorkConsolidationScript(force=True).run()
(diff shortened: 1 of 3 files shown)


bin/repair/work_consolidation
```diff
@@ -6,4 +6,4 @@ bin_dir = os.path.split(__file__)[0]
 package_dir = os.path.join(bin_dir, "..", "..")
 sys.path.append(os.path.abspath(package_dir))
 from core.scripts import WorkConsolidationScript
-WorkConsolidationScript(force=True).run()
+WorkConsolidationScript(force=False).run()
```
leonardr: This isn't relevant to the main bug. It's the circulation part of the change I made as part of https://github.com/NYPL-Simplified/server_core/pull/245.

courte:
:+1:

PR #192, opened 2016-05-18
Immediate bibliographic coverage upon discovery of book

leonardr:
This branch makes sure that 3M and Overdrive books get presentation-ready Works as soon as we learn about them. It does this by running the identifier through the appropriate BibliographicCoverageProvider. This means an extra request to the API interrupts the processing of circulation data, but if that request fails for whatever reason, it gets turned into a `CoverageFailure` for the normal `bibliographic_coverage` script to retry later, and the processing of circulation data isn't interrupted.

As part of this, this branch adds the first end-to-end test of OverdriveAPI.update_licensepool(), verifying that the Work is created and that bibliographic coverage is provided.

This branch gets rid of the circulation-specific subclasses of BibliographicCoverageProvider. All they did was mark the work presentation-ready after providing coverage, and that's something that can go (and has gone) into core.

I got rid of `ThreeMCirculationMonitor` after discovering that the class was no longer used.

I moved the provider-specific bibliographic coverage provider scripts from bin/ to bin/repair. I think the fact that `bibliographic_coverage` supports the `--identifier-type` argument means we don't need those scripts anymore, but there's no harm keeping them in `bin/repair` for a bit, we can decide that later.

What about Axis 360? Well, the Axis 360 API combines circulation and bibliographic data, so in a sense we don't need this code--we already create the Work and the LicensePool simultaneously, without even needing to make another request. However I think there are two smaller problems: first, we are in fact making that other request later, because we don't provide a bibliographic coverage record for the Axis 360 book, leading the Axis360BibliographicMonitor to get the same data and do the same work again. Second, I don't think we actually set the Work as presentation-ready after creating it. This means that Axis 360 books don't show up to patrons until the official Axis360BibliographicCoverageProvider runs. I'll do another branch to investigate these issues.


aslagle:
:+1:

PR #195, opened 2016-05-18
Move novelist from metadata

courte:
Based on our conversation with NoveList On Monday, each circulation manager is going to have to pull its own bibliographic weight on this one. This branch just pulls the changes from the metadata wrangler over to circulation wholesale, along with a small test utility method.


leonardr:
I'm basically trusting that this is the same code that I already reviewed in metadata. It all looks familiar except for some of the sample_data code, which looks good. :+1:

PR #196, opened 2016-05-18
Axis 360 circulation monitor effectively does the work of the bibliographic monitor as well

leonardr:
When we find out about a new book in the Axis 360 circulation monitor, automatically create a presentation-ready work and mark the book as covered by the Axis 360 bibliographic monitor. This will save the bibliographic monitor the trouble of making another request for the same data.


courte:
:+1:

PR #197, opened 2016-05-18
Fix circulation edge cases

leonardr:
This branch fixes and tests three edge cases I discovered while investigating a problem with the 3M circulation server:

* If you go to check out a book, and the remote side says you already have the book on hold, we raised an exception which resulted in a 500 error. Now, we treat it the same as if you go to check out a book, the remote side says there are no copies available, and you put the book on hold instead.

* If you specify an invalid license pool when revoking a hold, or an invalid distribution mechanism when borrowing, you triggered old code that expected failure conditions to be conveyed with a custom `Response`. We now use a `ProblemDetail` to convey those error conditions, so I changed the code to look for a `ProblemDetail`.

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -202,6 +202,7 @@ class CirculationAPI(object):
 
         # First, try to check out the book.
         loan_info = None
+        hold_info = None
         try:
             loan_info = api.checkout(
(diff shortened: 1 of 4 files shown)


courte:
:+1: when the tests roll in :)

PR #199, opened 2016-05-18
Edit classifications fix

skomputer:
This branch makes sure that edit_classifications() changes staff genre classifications only if the provided genres doesn't match the existing **computed** genres. Previously it would compare the provided genres with existing **staff** genres. Therefore if genres were left untouched by staff, new staff genre classifications would still be added, and if all genres had never been edited and were then removed, a NONE genre would not be created in their place.


aslagle:
:+1:

PR #200, opened 2016-05-19
Fixed some code that was still referencing pool.edition.

aslagle:
This is throwing exceptions in the 3M monitor on production.


leonardr:
:+1:

PR #202, opened 2016-05-19
3M's internal problems handled as 502 errors

leonardr:
This branch builds on https://github.com/NYPL-Simplified/server_core/pull/259. It changes RemoteInitiatedServerError and InternalServerError so that the core ErrorHandler turns them into problem detail documents.

The rationale is that these errors are really, really serious errors. We don't know when they might happen, and they stop a client request in its tracks. The request can't be completed. So normal "catch the exception and turn it into a nice problem detail" code isn't going to work. 

At the same time, we want to handle RemoteInitiatedServerError (a problem with some external service we called) differently from InternalServerError (a problem with the web app itself). We want the clients to be able to render a RemoteInitiatedServerError to the patron so when they call tech support they'll be able to say which service caused the problem, and that measn problem detail documents.

As a proof of concept, this branch also changes the 3M ErrorParser to return RemoteInitiatedServerError whenever the server returns a document that can't be parsed as a normal error document. (Generally because the app server is down.) I've simulated this error in a running web app and verified that it's handled correctly, when combined with the code in the core branch.

### DIFF
--- a/api/axis.py
+++ b/api/axis.py
@@ -57,6 +57,8 @@ class Axis360API(BaseAxis360API, Authenticator, BaseCirculationAPI):
 
     SET_DELIVERY_MECHANISM_AT = BaseCirculationAPI.BORROW_STEP
 
+    SERVICE_NAME = "Axis 360"
+
     # Create a lookup table between common DeliveryMechanism identifiers
     # and Overdrive format types.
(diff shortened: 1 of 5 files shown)


aslagle:
Looks good once the tests are fixed.

PR #203, opened 2016-05-20
Changed URLs that identify a LicensePool to include identifier type.

### LINKED ISSUE #187: URLs that identify LicensePools need to use data source name/identifier type/identifier, not data source name/identifier
Currently our URLs look like https://circulation.openebooks.us/works/Axis%20360/0014213917/borrow/2. This URL conflates a number of separate issues which can't remain conflated much longer:

1. It says "works" but we're not actually identifying a work.

2. In fact, it's not clear what we're identifying. An `Identifier` is identified by type and identifier. A `LicensePool` is identified by an `Identifier` plus a`DataSource`. Here we've got the name of a data source ("Axis 360") plus an identifier ("0014213917").

If we're identifying a `LicensePool`, we need to specify data source, identifier type, and identifier. If we're identifying an `Identifier` we need to specify identifier type and identifier. The time where we could reliably identify a `LicensePool` based solely on data source and identifier is coming to a close.

Fortunately, we should be able to change these URLs without affecting client behavior.

aslagle:
Fixes #187 - URLs now have data source, identifier type, and identifier.

Depends on https://github.com/NYPL-Simplified/server_core/pull/260

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -167,13 +167,13 @@ class WorkController(CirculationManagerController):
 
     STAFF_WEIGHT = 100000
 
-    def details(self, data_source, identifier):
+    def details(self, data_source, identifier_type, identifier):
         """Return an OPDS entry with detailed information for admins.
         
         This includes relevant links for editing the book.
         """
 
-        pool = self.load_licensepool(data_source, identifier)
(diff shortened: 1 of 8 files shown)


leonardr:
:+1:

PR #204, opened 2016-05-20
Test circulation API

leonardr:
This branch fixes an edge case in renewals. Overdrive (and possibly other distributors) prohibits renewing a loan if there are people waiting in line for the book. If you try, the `OverdriveAPI` raises `NoAvailableCopies`. In `CirculationAPI` weren't prepared for this--we only prepared for `CannotRenew`, which is raised if there _are_ available copies but your loan isn't near enough to its expiry date for renewal to be an option.

This real improvement in this branch is `test_circulationapi.py`, which contains the first tests of the troublesome `CirculationAPI` class. I had put this off because properly testing the code requires two levels of mock objects--a mock `CirculationAPI` that calls mock remote API objects--but this is so important to the circ manager and there are so many weird edge cases that I finally had to do it. I've added tests for all the `borrow()` cases that end with an active loan for the book. There are a lot more tests to be written and I'll get to them as I can.

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -221,8 +221,6 @@ class CirculationAPI(object):
                 on_multiple='interchangeable'
             )
 
-        loan_info = None
-        hold_info = None
         try:
             loan_info = api.checkout(
                 patron, pin, licensepool, internal_format
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -200,7 +200,28 @@ class CirculationAPI(object):
                 if parse_fines(patron.fines) >= parse_fines(max_fines):
                     raise OutstandingFines()
 
-        # First, try to check out the book.
+        # Do we (think we) already have this book out on loan?
+        existing_loan = get_one(
+             self._db, Loan, patron=patron, license_pool=licensepool,
+             on_multiple='interchangeable'
+        )
+        
+        loan_info = None
+        hold_info = None
+        if existing_loan:
+            # Sync with the API to see if the loan still exists.  If
+            # it does, we still want to perform a 'checkout' operation
+            # on the API, because that's how loans are renewed, but
+            # certain error conditions (like NoAvailableCopies) mean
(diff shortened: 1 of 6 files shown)


api/circulation.py
```diff
@@ -200,7 +200,28 @@ def parse_fines(fines):
...
 
-        # First, try to check out the book.
+        # Do we (think we) already have this book out on loan?
+        existing_loan = get_one(
+             self._db, Loan, patron=patron, license_pool=licensepool,
+             on_multiple='interchangeable'
+        )
+        
+        loan_info = None
+        hold_info = None
```
aslagle: these two lines are duplicated now

leonardr:
I found a WHOLE OTHER SET of mock objects for this code lying unused and (slightly) broken in `circulation.py` I moved it to the test class, got it to work again, and rewrote the tests to use it. I also added two more tests to show the difference in how `NoAvailableCopies` is handled when you already have a loan (it's a renewal error) versus when you have no loan (it gets turned into a hold).

leonardr:
Turns out those mock objects weren't unused; they're actually used in test_controller.py, which contains some tests that really ought to be in test_circulationapi.py. They're not really testing the controller, it's just that testing the controller was the only convenient way to test features of CirculationAPI.

api/testing.py
```diff
@@ -0,0 +1,144 @@
...
+            DataSource.lookup(self._db, name).id for name in 
+            self.identifier_type_to_data_source_name.values()
+        ]
+        self.remotes = {}
+
+    def local_loans(self, patron):
+        return self._db.query(Loan).filter(Loan.patron==patron)
+
+    def local_holds(self, patron):
+        return self._db.query(Hold).filter(Loan.patron==patron)
```
aslagle: The filter should be Hold.patron

aslagle:
Looks good except for that one copy/paste error

PR #209, opened 2016-05-23
Fix everything lane url

skomputer:
This branch modifies CirculationManagerAnnotator.lane_url() so that it returns the default_lane_url() when the given lane has no parent.


leonardr:
:+1:

PR #210, opened 2016-05-24
Handle internal server errors for Axis 360 and 3M

leonardr:
This branch makes a number of changes to ensure that if something goes wrong with the Axis 360 or 3M integration, the exception that gets raised is a RemoteInitiatedServerError or a RemoteIntegrationException. This also happens when we're acting on a patron's behalf using Overdrive.

We could potentially replace RemoteInitiatedServerError with RemoteIntegrationException, but I like
RemoteInitiatedServerError for cases like Axis 360 where the request returns a 200 status code and a document that says "just kidding, everything's terrible". But maybe that's not a good reason to have two very similar exceptions.


api/overdrive.py
```diff
@@ -78,13 +79,15 @@ def patron_request(self, patron, pin, url, extra_headers={}, data=None,
...
+            method = method.lower()
         else:
             if data:
-                method = requests.post
+                method = 'post'
             else:
-                method = requests.get
-        response = method(url, headers=headers, data=data)
+                method = 'get'
+        response = HTTP.request_with_timeout(
```
leonardr: This is OverdriveAPI.patron_request. This code ensures that if we make any request using a patron's access token, and we get an error, the patron sees a 502 error rather than a 500.

aslagle:
:+1:

Issue #207, opened 2016-05-23
Bad error message when you try to borrow a work with no licenses_owned


leonardr:
I tried to borrow 3M book dxhoor9, which has no licenses_owned. I got this error:

```
{"status": 502, "debug_message": "the patron document status was CAN_WISH and not one of CAN_LOAN,RESERVATION", "type": "http://librarysimplified.org/terms/problem/cannot-issue-loan", "detail": "Could not issue loan (reason unknown).", "title": "Could not issue loan."}
```

First, this should be a better error--it should be a 404. Second, I saw this happen in production. We need something that forces an immedate refresh of availability information from 3M if this sort of problem happens. We have similar code for Overdrive.

leonardr:
It is good to force a refresh of availability information if this happens, but it won't solve this problem. According to https://github.com/NYPL-Simplified/circulation/issues/213, we know we have no licenses for the book but we're putting it in feeds anyway.

PR #212, opened 2016-05-25
Raise NoLicenses when patron tries to borrow a 3M book with no licenses

leonardr:
3M treats a book with no licenses as being in the state "CAN_WISH". We didn't handle that case so if you tried to borrow a book with no licenses, you got a 500 error.

This branch creates a NoLicenses exception which returns the NO_LICENSES problem detail document. I also changed NO_LICENSES to make it a 404 error instead of a 403. It might be a 403 error if the library was asking the distributor on its own behalf, since the library can theoretically buy more licenses, but since the patron is asking for the book I think the correct response code is 404.

This branch partially fixes https://github.com/NYPL-Simplified/circulation/issues/207


api/threem.py
```diff
@@ -386,6 +386,9 @@ def process_one(self, error_tag, namespaces):
         actual, expected = m.groups()
         expected = expected.split(",")
 
+        if actual == 'CAN_WISH' and actual == 'CAN_HOLD':
```
aslagle: this can't be right

aslagle:
:+1:

Issue #213, opened 2016-05-25
We're showing books for which we have no licenses


leonardr:
The first result here: https://circulation.librarysimplified.org/search/?q=the+professor%27s+daughter

```
  <entry schema:additionalType="http://schema.org/Book">
    <id>urn:librarysimplified.org/terms/id/3M%20ID/dxhoor9</id>
    <title>The Professor's Daughter</title>
    <alternativeHeadline>A Novel</alternativeHeadline>
    <bibframe:distribution bibframe:ProviderName="3M"/>
    <author>
      <name>Emily Raboteau</name>
    </author>
    <summary type="html">&lt;b&gt;A daughter's future and her father's past converge in Emily Raboteau's explosive first novel exploring identity, assimilation, and the legacy of race&lt;/b&gt;&lt;br /&gt;&lt;br /&gt;"My father is black and my mother is white and my brother is a vegetable." When Emma Boudreaux's older brother, Bernie, winds up in a coma after a freak accident, it's as if she loses a part of herself. All their lives, he has served as her compass, her stronger, better half: Bernie was brilliant when Emma was smart, charismatic when she was awkward, and confident when she was shy. Only Bernie was able to navigate-if not always diplomatically-the terrain of their biracial identity. Now, as the chronic rash that's flared up throughout her life returns with a vengeance, Emma is sleepwalking through her first year at college, left alone to grow into herself.&lt;br /&gt;&lt;br /&gt;The key to Emma's self-discovery lies in her father's past. Esteemed Princeton professor Bernard Boudreaux is emotionally absent and secretive about his family history. Little does Emma know just how haunted that history is, how tortured the path from the Deep South town to his present Ivy League success has been. Though her father and brother are bound by the past, Emma might just escape. &lt;br /&gt;&lt;br /&gt;In exhilarating, magical prose, &lt;i&gt;The Professor's Daughter&lt;/i&gt; traces the borderlands of race and family, the contested territory that gives birth to rage, confusion, madness, and invisibility. This striking debut marks the arrival of an astonishingly original voice that surges with energy and purpose.&lt;br /&gt;&lt;br /&gt;&lt;br /&gt;</summary>
    <updated>2016-05-13T13:44:04Z</updated>
    <simplified:pwid>949da39e-2dd5-7e18-42cc-f687b573a6ef</simplified:pwid>
[...]

PR #214, opened 2016-05-25
Immediately update a book's availability if exception indicates our data is out-of-date

leonardr:
If you try to borrow a book and get a NoAvailableCopies or NoLicenses exception, this branch immediately asks the license provider for up-to-date licensing information, to reduce the amount of time our licensing information is out of date.

This branch adds a new method, `update_availability(self, licensepool)` to the contract implemented by all the APIs used by `CirculationAPI`. Its job is to update the licensing information for a single `LicensePool`. I implemented and tested all three cases (Overdrive, 3M, Axis 360), and it's always a wrapper around a little bit of preexisting code.

This is useful code to have around, so I finished the branch even though this sort of race condition isn't currently the major problem driving `NoAvailableCopies` and `NoLicenses` exceptions: see https://github.com/NYPL-Simplified/circulation/issues/213

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -221,6 +221,7 @@ class CirculationAPI(object):
                 on_multiple='interchangeable'
             )
 
+        needs_update = False
         try:
             loan_info = api.checkout(
                 patron, pin, licensepool, internal_format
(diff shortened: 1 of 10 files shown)


api/circulation.py
```diff
@@ -221,6 +221,7 @@ def parse_fines(fines):
                 on_multiple='interchangeable'
             )
 
+        needs_update = False
```
aslagle: this variable isn't used

aslagle:
:+1:

PR #216, opened 2016-05-25
Institute a single code path for updating the circulation information for 3M books

leonardr:
This branch changes the implementation of ThreeMAPI.update_availability so that it checks the status of a single LicensePool by instantiating a ThreeMCirculationSweep and calling its `process_batch()` method on a single identifier. This may seem like overkill, and the code should certainly be refactored, but it solves two pretty bad problems:

1. My previous branch introduced a typo into ThreeMCirculationSweep.process_batch() which I didn't catch because that method isn't tested. It is tested now, albeit only accidentally.

2. The whole point of my previous branch is that if a book leaves the collection and then someone tries to check it out, we find out that the book is gone. In my previous branch, that didn't happen, because the essential code doesn't run. The essential code compares the list of identifiers we sent 3M against the list of identifiers we got back, and deduces that any identifier not mentioned in the list must have left our collection. That code is currently in ThreeMCirculationSweep.

### DIFF
--- a/api/threem.py
+++ b/api/threem.py
@@ -101,12 +101,9 @@ class ThreeMAPI(BaseThreeMAPI, BaseCirculationAPI):
 
     def update_availability(self, licensepool):
         """Update the availability information for a single LicensePool."""
-        for circ in self.get_circulation_for(
-                [licensepool.identifier.identifier]
-        ):
-            self.apply_circulation_information_to_licensepool(
-                circ, licensepool
-            )
+        return ThreeMCirculationSweep(self._db, api=self).process_batch(
+            [licensepool.identifier]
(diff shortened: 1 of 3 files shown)


api/threem.py
```diff
@@ -623,7 +623,7 @@ def process_batch(self, identifiers):
                     self._db, pool, CirculationEvent.TITLE_ADD,
                     None, None, start=now)
 
-            self.api.apply_circulation_data_to_licensepool(pool)
+            self.api.apply_circulation_information_to_licensepool(circ, pool)
```
leonardr: Here's the typo.

aslagle:
:+1:

PR #218, opened 2016-05-26
Patron oauth

aslagle:
This branch adds support for OAuth authentication providers. A circulation manager can support multiple OAuth providers and one basic auth provider.

The first OAuth provider is Clever - this branch has an initial implementation but it needs more work once we know the requirements.

Providers are now configured by module name in the config file, eg `['api.firstbook', 'api.clever']` - and this requires changing the existing config files. In the future we could move the auth provider modules to a different package or repo.

### DIFF
--- a/api/admin/config.py
+++ b/api/admin/config.py
@@ -10,7 +10,6 @@ import contextlib
 class Configuration(CirculationManagerConfiguration):
 
     ADMIN_AUTH_DOMAIN = "admin_authentication_domain"
-    SECRET_KEY = "secret_key"
     GOOGLE_OAUTH_INTEGRATION = "Google OAuth"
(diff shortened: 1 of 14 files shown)


api/admin/config.py
```diff
@@ -10,7 +10,6 @@
 class Configuration(CirculationManagerConfiguration):
 
     ADMIN_AUTH_DOMAIN = "admin_authentication_domain"
-    SECRET_KEY = "secret_key"
```
aslagle: This is used for patron authentication as well as admin authentication now.

leonardr:
I like this branch and everything seems right, but I'm not comfortable enough with the finer details of OAuth to give it a +1, so I'd like Courteney to review it as well.

courte:
:+1: yesssssssss 🎈

PR #221, opened 2016-05-27
Fix language lane urls

aslagle:
This branch fixes the urls for top-level language lanes. Since they have no parent, they were incorrectly assumed to be the overall top-level lane and given the default lane url.

This fix is kind of hacky, but it seemed better than checking the name of the lane or the languages. Eventually, it would be nice to make an actual Lane for the top-level instead of using the Circ Manager.


skomputer:
👍

leonardr:
:+1: 

I filed an issue to resolve the underlying problem.

PR #223, opened 2016-05-27
'Other Language' sublanes

aslagle:
This branch changes the "Other Languages" lane to have sublanes for languages specified in the config file, instead of audiences and fiction status.

It also has a script that prints out all the languages with at least one non-open access work and the number of works they have.

### DIFF
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -42,14 +42,14 @@ def make_lanes_default(_db):
     # Fiction", "Adult Nonfiction", and "Children/YA" sublanes.
     #
     # Finally the top-level LaneList includes an "Other Languages" sublane
-    # which covers all other languages. This lane contains "Adult Fiction",
-    # "Adult Nonfiction", and "Children/YA" sublanes.
+    # which covers all other languages. This lane contains sublanes for each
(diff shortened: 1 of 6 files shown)


api/lanes.py
```diff
@@ -339,44 +339,27 @@ def lane_for_small_collection(_db, languages):
...
-    )
-
-    ya_children = Lane(
-        _db, 
-        full_name="Children & Young Adult", 
-        fiction=Lane.BOTH_FICTION_AND_NONFICTION,
-        audiences=[YA, CHILDREN],
-        **common_args
-    )
+    for language_set in other_languages:
```
leonardr: What is going to happen if other_languages is empty? Is the "Other Languages" name going to have nothing in it, or will it just be a big scrolling list of all books not in exclude_languages?

aslagle: It should be the big scrolling list of all books not in exclude_languages. But I just checked, and it turns out the lane url is wrong again when this happens, so I'll look into that.

aslagle: Okay, I changed it so there is no "Other Languages" lane in this case.

leonardr:
:+1:

PR #227, opened 2016-05-27
Translate NoveList recommendations into a feed

### LINKED ISSUE #215: Include NoveList recommendations

courte:
This branch grabs the recommendations suggested by NoveList--changing its lookup return values a bit--and finds the relevant works in the Circulation Manager. All this work is wrapped up into a handy-dandy link on each work's entry. There are also a few changes to make testing possible.

Fixes #215. Depends on NYPL-Simplified/server_core#281.


api/novelist.py
```diff
@@ -39,17 +39,28 @@ class NoveListAPI(object):
     # from the database.
     QUERY_ENDPOINT = "http://novselect.ebscohost.com/Data/ContentByQuery?\
             ISBN=%(ISBN)s&ClientIdentifier=%(ClientIdentifier)s&version=%(version)s"
-    MAX_REPRESENTATION_AGE = 6*30*24*60*60      # six months
+    MAX_REPRESENTATION_AGE = 14*24*60*60      # two weeks
```
courte: I brought this down a lot because we're now using it for recommendations, instead of less volatile background data like equivalent ISBNs.

leonardr:
:+1: once the issues with the core branch are resolved.

PR #228, opened 2016-05-31
More accurate language count

leonardr:
This branch changes the LanguageScript to exclude things that aren't books (i.e. foreign-language movies) and LicensePools where no licenses are currently owned. I also improved the script output to a) give human-readable language names to the person running the script, who probably doesn't know all the language codes, and b) provide a sample value for `tiny_language_collections` which can be edited to taste.

I also changed the code so that if two languages have the same number of books, they're ordered by the English name of the language.


aslagle:
:+1:

PR #230, opened 2016-05-31
get google oauth config from circulation.json instead of separate file

### LINKED ISSUE #229: Don't keep Google OAuth client configuration in a separate file
Curently the Google OAuth client configuration is kept in a file whose location is configured as client_json_file in the circulation manager's configuration document. Since this document is JSON and contains sensitive information, it should be included wholesale in the master JSON configuration file, so that it can be controlled by the same methods used to control other sensitive information like the database password.

For the time being, it's okay to continue to use client_json_file as a fallback if the actual configuration is not present.

skomputer:
Closes #229.

### DIFF
--- a/api/admin/oauth.py
+++ b/api/admin/oauth.py
@@ -7,15 +7,13 @@ from oauth2client import client as GoogleClient
 
 class GoogleAuthService(object):
 
-    def __init__(self, client_json_file, redirect_uri, test_mode=False):
+    def __init__(self, config, redirect_uri, test_mode=False):
         if test_mode:
             self.client = DummyGoogleClient()
         else:
-            self.client = GoogleClient.flow_from_clientsecrets(
-                client_json_file,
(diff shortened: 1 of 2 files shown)


api/admin/oauth.py
```diff
@@ -7,15 +7,13 @@
...
+    def __init__(self, config, redirect_uri, test_mode=False):
         if test_mode:
             self.client = DummyGoogleClient()
         else:
-            self.client = GoogleClient.flow_from_clientsecrets(
-                client_json_file,
-                scope='https://www.googleapis.com/auth/userinfo.email',
-                redirect_uri=redirect_uri
-            )
+            config['redirect_uri'] = redirect_uri
```
leonardr: Before modifying `config` you should create a new object, i.e. `config = dict(config)`. Otherwise you'll be modifying the global site configuration.

leonardr:
:+1: after one minor change.

PR #233, opened 2016-06-02
Analytics

skomputer:
This branch introduces a Google Analytics provider, uses the "local" analytics provider to log license pool availability changes as circulation events in the database, and adds an endpoint for logging analytics events from clients.


tests/test_axis.py
```diff
@@ -140,60 +149,65 @@ class TestCirculationMonitor(DatabaseTest):
...
-
-        # A CoverageRecord has been provided for this book in the Axis
-        # 360 bibliographic coverage provider, so that in the future
-        # it doesn't have to make a separate API request to ask about
-        # this book.
-        records = [x for x in license_pool.identifier.coverage_records
-                   if x.data_source.name == DataSource.AXIS_360
-                   and x.operation is None]
-        eq_(1, len(records))
+        with temp_config() as config:
```
leonardr: This is fine for now but you don't literally need to set up a LocalAnalyticsProvider here, just to maintain the previous functionality. It'd be a bit nicer if you used a MockAnalyticsProvider instead, and verified that the appropriate events were recorded in the MockAnalyticsProvider. (This would require changing MockAnalyticsProvider so that it kept a log of all the events it saw, not just the most recent one.)

tests/test_google_analytics_provider.py
```diff
@@ -0,0 +1,59 @@
...
+    temp_config,
+)
+from core.analytics import Analytics
+from api.google_analytics_provider import GoogleAnalyticsProvider
+from . import DatabaseTest
+from core.model import CirculationEvent
+import urlparse
+import datetime
+
+class MockGoogleAnalyticsProvider(GoogleAnalyticsProvider):
```
leonardr: Again, in this sort of mock class it's frequently useful to see a complete log of things that passed through it, not just the most recent thing.

leonardr:
I have some nitpicks but they can be taken care of in the future as necessary. :+1:

PR #238, opened 2016-06-06
Localization

aslagle:
This branch wraps error messages in a function as in https://github.com/NYPL-Simplified/server_core/pull/289, and adds babel localization to the flask app. 

`bin/util/generate_translation_template` is a script to generate a messages.pot file, which a translator can fill in for each language. I'm not sure if messages.pot should be checked in. 

The translator will produce a messages.po file for each language, which should go in translations/<2_letter_language_code>/LC_MESSAGES/messages.po. 

`bin/util/compile_translations` will convert the messages.po files to messages.mo files which are used by the application.

This connects to https://github.com/NYPL-Simplified/circulation/issues/244


aslagle:
It might be worth breaking up the translations into separate domains for circulation, core, and the admin interface. Then the core translations could be reused in other applications and the admin ones could be optional. This would be annoying if we wanted to use some of the same strings in core and circulation or circulation and admin - you would have to make sure you set up the right domain for individual strings, or have them in both translation files.

api/admin/oauth.py
```diff
@@ -46,7 +47,7 @@ def callback(self, request={}):
             ), redirect_url
 
     def google_error_problem_detail(self, error):
-        error_detail = GOOGLE_OAUTH_FAILURE.detail + " Error: " + error
+        error_detail = GOOGLE_OAUTH_FAILURE.detail + _("Error: ") + error
```
leonardr: I think the rule is that this should be "Error: %(error)s" so translators can vary the placement of the variable.

api/app.py
```diff
@@ -13,17 +13,21 @@
...
 
-
 testing = 'TESTING' in os.environ
 db_url = Configuration.database_url(testing)
 SessionManager.initialize(db_url)
 session_factory = SessionManager.sessionmaker(db_url)
 _db = flask_scoped_session(session_factory, app)
 SessionManager.initialize_data(_db)
 
+app.config['BABEL_DEFAULT_LOCALE'] = 'en'
```
leonardr: You should be able to get this from the config object.

api/authenticator.py
```diff
@@ -166,8 +167,8 @@ def create_authentication_document(self):
                 links[rel] = dict(href=value, type="text/html")
 
         doc = OPDSAuthenticationDocument.fill_in(
-            base_opds_document, auth_type, "Library", opds_id, None, "Barcode",
-            "PIN", links=links
+            base_opds_document, auth_type, _("Library"), opds_id, None, _("Barcode"),
+            _("PIN"), links=links
```
leonardr: I'm a little worried about this because individual libraries will want to modify this sort of vocabulary, but we _also_ want it to be localized. I don't see a better solution for now, though.

leonardr:
Breaking up the translations is a good idea because we need the circulation and core translated into a _lot_ more languages than we need to admin interface, and we want to make it as easy a job as possible.

api/admin/oauth.py
```diff
@@ -47,7 +47,15 @@ def callback(self, request={}):
...
 
     def google_error_problem_detail(self, error):
-        error_detail = GOOGLE_OAUTH_FAILURE.detail + _("Error: ") + error
+        error_detail = _("Error: %(error)s", error=error)
+
+        # ProblemDetail.detailed requires the detail to be an internationalized
+        # string, so pass the combined string through _ as well even though the
+        # components were translated already. Space is a variable so it doesn't
+        # end up in the translation template.
+        space = " "
```
leonardr: Will this cause an error when the already-localized string can't be further localized?

scripts.py
```diff
@@ -522,3 +523,26 @@ def do_run(self):
...
+
+class CompileTranslationsScript(Script):
+    """A script to combine translation files for circulation, core
+    and the admin interface, and compile the result to be used by the
+    app. The combination step is necessary because Flask-Babel does not
+    support multiple domains yet.
+    """
+
+    def run(self):
+        languages = Configuration.localization_languages()
```
leonardr: Since you do the three_to_two lookup at least, is it worth putting that code in localization_languages? The only reason I suggested using the 3-codes in the config file was for consistency. If every code point that uses then uses the 2-codes, it's fine to do the lookup immediately after reading from the config file.

api/admin/oauth.py
aslagle: No. It will try to look up a translation for the new string, and there won't be one, so it will just return the string.

scripts.py
aslagle: done

leonardr:
:+1:

scripts.py
```diff
@@ -525,6 +526,25 @@ def do_run(self):
...
+        for language in languages:
+            base_path = "translations/%s/LC_MESSAGES" % language
+            if not os.path.exists(base_path):
+                logging.warn("No translations for configured language %s" % language)
+                continue
+
+            os.system("rm %(path)s/messages.po" % dict(path=base_path))
+            os.system("cat %(path)s/*.po > %(path)s/messages.po" % dict(path=base_path))
```
aslagle: pybabel's update apparently doesn't do what I thought it did, but simply concatenating the files seems to work fine.

leonardr:
:+1:

PR #240, opened 2016-06-07
Added temp config with secret key to authenticator tests.

### LINKED ISSUE #239: TestAuthenticator.test_authenticated_patron assumes a secret key is configured
A number of test_authenticator.py tests fail when run against my minimal configuration. They assume there's a real value in configuration['secret_key']. If I have these problems, other users will as well.

Please change these tests to use temp_config() to create a temporary Configuration object with a dummy value for 'secret_key'.

aslagle:
Fixes #239


leonardr:
:+1:

PR #241, opened 2016-06-07
Got the content server bibliographic coverage provider to work.

leonardr:
This branch makes ContentServerBibliographicCoverageProvider actually do what it's supposed to--update the metadata of open-access books by running them through the content server's lookup protocol. It looks like this stopped working once I refactored coverage.py on May 27. This time there are tests.

### DIFF
--- a/api/coverage.py
+++ b/api/coverage.py
@@ -139,7 +139,10 @@ class OPDSImportCoverageProvider(CoverageProvider):
         # id_mapping maps our local identifiers to identifiers the
         # foreign data source will reocgnize.
         id_mapping = self.create_identifier_mapping(batch)
-        foreign_identifiers = id_mapping.keys()
+        if id_mapping:
+            foreign_identifiers = id_mapping.keys()
+        else:
+            foreign_identifiers = batch
 
         response = self.lookup.lookup(foreign_identifiers)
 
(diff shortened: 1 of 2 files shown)


api/coverage.py
```diff
@@ -139,7 +139,10 @@ def lookup_and_import_batch(self, batch):
         # id_mapping maps our local identifiers to identifiers the
         # foreign data source will reocgnize.
         id_mapping = self.create_identifier_mapping(batch)
-        foreign_identifiers = id_mapping.keys()
+        if id_mapping:
```
leonardr: I don't have a good way of testing this because it's in a method overridden by MockOPDSImportCoverageProvider. I have tested it in a real situation. This means you don't _have_ to have a mapping; you can return None (which happens by default) and the original identifiers will be used as themselves.

api/coverage.py
```diff
@@ -344,7 +347,20 @@ def __init__(self, _db, service_name=None, lookup=None, **kwargs):
             _db, DataSource.OA_CONTENT_SERVER
         )
         super(ContentServerBibliographicCoverageProvider, self).__init__(
-            service_name, lookup=lookup, output_source=output_source,
-            expect_license_pool=True, presentation_ready_on_success=True
+            service_name, input_identifier_types=None,
```
leonardr: This was just a missing argument.

aslagle:
:+1:

PR #242, opened 2016-06-07
Remove license data source

daryanypl:


leonardr:
This looks good. 3M needs to be changed to use this system, but it's so out-of-date that it doesn't use CirculationData in the first place, so it doesn't need to happen here.

PR #245, opened 2016-06-08
Add series lanes and a groups feed for all books related to a particular work.

### LINKED ISSUE #234: Include OPDS feeds of works in a series to entries

courte:
This branch:
- Makes sure that NoveList series information is applied to a licensed edition if series data isn't currently captured.
- Creates a lane for series.
- Creates a groups-feed-intended parent lane RelatedBooksLane, and trains the CirculationManagerAnnotator on how to deal with it and its chil'rens.
- Allows LicensePool-based sublanes to have proper URLs.
- Puts a link to related books in a Work's OPDS entry if appropriate.

I mostly hate the `DISPLAY_NAME` for these lanes and wish they were simpler while staying clear. 😕 

Fixes #234.

### DIFF
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -453,9 +453,11 @@ class RelatedBooksLane(LicensePoolBasedLane):
             lane_name = "Recommendations for %s by %s" % (
                 license_pool.work.title, license_pool.work.author
             )
-            sublanes.append(RecommendationLane(
+            recommendation_lane = RecommendationLane(
                 _db, license_pool, lane_name, novelist_api=novelist_api
-            ))
+            )
+            if recommendation_lane.recommendations:
+                sublanes.append(recommendation_lane)
         except ValueError, e:
             # NoveList isn't configured.
             pass
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -432,8 +432,8 @@ class RelatedBooksLane(LicensePoolBasedLane):
     DISPLAY_NAME = "Related Books"
 
     def __init__(self, _db, license_pool, full_name, display_name=None,
-                 mock_api=None):
-        sublanes = self._get_sublanes(_db, license_pool, mock_api=mock_api)
+                 novelist_api=None):
+        sublanes = self._get_sublanes(_db, license_pool, novelist_api=novelist_api)
         if not sublanes:
             edition = license_pool.presentation_edition
             raise ValueError(
@@ -445,7 +445,7 @@ class RelatedBooksLane(LicensePoolBasedLane):
             invisible=True
         )
 
-    def _get_sublanes(self, _db, license_pool, mock_api=None):
+    def _get_sublanes(self, _db, license_pool, novelist_api=None):
         sublanes = []
 
         # Create a recommendations sublane.
@@ -454,7 +454,7 @@ class RelatedBooksLane(LicensePoolBasedLane):
                 license_pool.work.title, license_pool.work.author
             )
             sublanes.append(RecommendationLane(
-                _db, license_pool, lane_name, mock_api=mock_api
+                _db, license_pool, lane_name, novelist_api=novelist_api
             ))
         except ValueError, e:
             # NoveList isn't configured.
@@ -513,8 +513,8 @@ class RecommendationLane(LicensePoolBasedLane):
     MAX_CACHE_AGE = 7*24*60*60      # one week
 
     def __init__(self, _db, license_pool, full_name, display_name=None,
-            mock_api=None):
-        self.api = mock_api or NoveListAPI.from_config(_db)
+            novelist_api=None):
+        self.api = novelist_api or NoveListAPI.from_config(_db)
         super(RecommendationLane, self).__init__(
             _db, license_pool, full_name, display_name=display_name
         )
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -465,8 +465,7 @@ class RelatedBooksLane(LicensePoolBasedLane):
         # Create a series sublane.
         series = license_pool.presentation_edition.series
         if series:
-            lane_name = SeriesLane.lane_name_from_series_title(series)
-            sublanes.append(SeriesLane(_db, license_pool, lane_name))
+            sublanes.append(SeriesLane(_db, license_pool))
 
         return sublanes
 
@@ -479,7 +478,12 @@ class RelatedBooksLane(LicensePoolBasedLane):
 class SeriesLane(LicensePoolBasedLane):
     """A lane of Works in a series based on a particular LicensePool"""
 
-    DISPLAY_NAME = "Other Books in this Series"
+    def __init__(self, _db, license_pool):
+        series_name = license_pool.presentation_edition.series
+        full_name = display_name = series_name
+        super(SeriesLane, self).__init__(
+            _db, license_pool, full_name, display_name=display_name
+        )
 
     def apply_filters(self, qu, work_model=Work, *args, **kwargs):
         edition = self.license_pool.presentation_edition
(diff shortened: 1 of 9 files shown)


api/lanes.py
```diff
@@ -374,32 +377,167 @@ def lane_for_other_languages(_db, exclude_languages):
...
+
+class RelatedBooksLane(LicensePoolBasedLane):
+    """A lane of Works all related to the Work of a particular LicensePool
+
+    Sublanes currently include a SeriesLane and a RecommendationLane"""
+
+    DISPLAY_NAME = "Related Books"
+
+    def __init__(self, _db, license_pool, full_name, display_name=None,
+                 mock_api=None):
```
leonardr: I guess I didn't notice this in your last branch, but can you call this `novelist_api` instead of `mock_api`? There's potentially going to be more than one API going in to the `RelatedBooksLane`, and it also should work fine if you pass in the real API. For an example look at CirculationAPI in api/circulation.py--its constructor has `overdrive`, `threem`, and `axis` arguments and they can be real APIs or mocks.

api/lanes.py
```diff
@@ -374,32 +377,167 @@ def lane_for_other_languages(_db, exclude_languages):
...
+
+    @classmethod
+    def lane_name_from_series_title(cls, series_title):
+        feed_title = "Other Books in "
+        if series_title[:3].lower() != 'the':
+            feed_title += "the "
+        feed_title += series_title
+        if series_title.lower().endswith(' series'):
+            return feed_title
+        return feed_title + ' series'
```
leonardr: Internationalizing this is going to be a pain but we can tackle it when we do a pass of internationalizing the lane names in general.

courte: I could change it to "Other Books in This Series" for simplicity.

api/novelist.py
```diff
@@ -392,6 +391,23 @@ def process_item(self, identifier):
...
+            # identifier's edition.
+            self._apply_series_data(
+                identifier, edition.series, edition.series_position
+            )
+
         return identifier
+
+    def _apply_series_data(self, identifier, series, series_position):
+        """Provides series data on the given identifier"""
+        licensed_edition = self.edition(identifier)
```
leonardr: I see what you're doing here but CoverageProvider.edition() is designed for CoverageProviders that are talking directly to the LicensePool source (e.g. the OverdriveBibliographicCoverageProvider). As such, it finds the Edition where the DataSource is the DataSource associated with the LicensePool. I think you want to create a new Edition where the DataSource is Novelist. 

You can do this by creating a Metadata object that includes primary_identifier, data_source, series_title, and series_position. Then call `edition()` on the Metadata object, and then `Metadata.apply(edition)`.

I'm not sure what to do about CoverageProvider.edition(), but it's going to need some work once it's possible to have multiple LicensePools for the same Identifier from different data sources.

api/opds.py
```diff
@@ -245,20 +267,28 @@ def annotate_work_entry(self, work, active_license_pool, edition, identifier, fe
...
                 )
             )
 
+    @classmethod
+    def related_books_available(cls, license_pool):
+        """:return: bool asserting whether related books are available for a
+        particular work
+        """
+        series = license_pool.presentation_edition.series
+        return NoveListAPI.is_configured() or series
```
leonardr: It looks like is_configured() runs pretty quick but if it's going to be called for every single book in a feed it would be good to cache the result.

api/opds.py
```diff
@@ -245,20 +267,28 @@ def annotate_work_entry(self, work, active_license_pool, edition, identifier, fe
...
+        # Add a link for related books if available.
+        if self.related_books_available(active_license_pool):
             feed.add_link_to_entry(
                 entry,
                 rel='related',
                 type=OPDSFeed.ACQUISITION_FEED_TYPE,
                 title='Recommended Works',
                 href=self.url_for(
-                    'recommendations',
+                    'related_books',
```
leonardr: I think the IANA 'related' link relation would work here.

_The value "related" signifies that the IRI in the value of the href attribute identifies a resource related to the resource described by the containing element._

api/lanes.py
```diff
@@ -374,32 +377,167 @@ def lane_for_other_languages(_db, exclude_languages):
...
+
+
+class RecommendationLane(LicensePoolBasedLane):
+    """A lane of recommended Works based on a particular LicensePool"""
+
+    DISPLAY_NAME = "Recommended Books"
```
leonardr: Why switch from 'related' to 'recommended' when you switched the link relation from 'recommended' to 'related'?

tests/test_lanes.py
```diff
@@ -152,3 +152,14 @@ def test_make_lanes_default(self):
...
+class TestRelatedBooksLane(DatabaseTest):
+
+    def test_initialization(self):
+        work = self._work(with_license_pool=True)
+        [lp] = work.license_pools
+        with temp_config() as config:
+            config['integrations'][Configuration.NOVELIST_INTEGRATION] = {}
+            assert_raises(
+                ValueError, RelatedBooksLane, self._db, lp, ""
+            )
```
leonardr: You should also test a successful initialization.

api/lanes.py
leonardr: This discussion has actually made me think that the series feed should not be hanging on a LicensePool at all. I think we should have one feed per series, and it should be named after the series. 

Every LicensePool that belongs to a given series would link to the same series feed.

This would reduce the naming problem (the feed would just be named after the series, no "Other Books in...") and also reduce the number of distinct feeds that need to be calculated.

What do you think of this? This is a chance you could make in a separate branch.

api/lanes.py
```diff
@@ -374,32 +377,167 @@ def lane_for_other_languages(_db, exclude_languages):
...
+
+        # Aliasing Edition here allows this query to function
+        # regardless of work_model and existing joins.
+        work_edition = aliased(Edition)
+        qu = qu.join(work_edition).filter(
+            work_edition.series==series,
+            work_edition.id!=edition.id
+        )
+        return qu
+
```
leonardr: How are the series feeds sorted? I would expect them to be sorted by series_position and then by title. This should handle both the 'Harry Potter' case and the 'For Dummies' case.

api/opds.py
courte: This isn't the link relation, it's the method from `api/routes.py` for which the URL is being created. The "related" rel is set 4 lines up.

api/lanes.py
courte: I think it's a good idea. Does it matter that the book would be included in its own "also in this series" related feed?

courte: This work is: #253.

api/novelist.py
```diff
@@ -391,23 +394,19 @@ def process_item(self, identifier):
...
         edition, ignore = metadata.edition(self._db)
         metadata.apply(edition)
+
         if edition.series or edition.series_position:
-            # Place any missing series data that NoveList gathered into this
-            # identifier's edition.
-            self._apply_series_data(
-                identifier, edition.series, edition.series_position
-            )
+            metadata.primary_identifier = identifier
```
leonardr: Extend your comment to make it explicit that you're modifying the Metadata you got from Novelist so that it can be applied directly to the original identifier.

tests/test_controller.py
```diff
@@ -856,14 +864,16 @@ def test_series(self):
         eq_(200, response.status_code)
         feed = feedparser.parse(response.data)
         eq_('Other Books in the Like As If Whatever Mysteries series', feed['feed']['title'])
```
leonardr: Since you now include a book in its own feed, 'Other Books in...' isn't apropriate anymore.

leonardr:
:+1: after addressing my minor comments.

PR #246, opened 2016-06-09
Make metadata wrangler coverage provider respect cutoff_time

leonardr:
This makes it possible to force a refresh of coverage. Other coverage providers support this but the essential **kwargs weren't being propagated to the superconstructor.


aslagle:
:+1:

PR #248, opened 2016-06-10
Search coverage provider

leonardr:
This branch removes the search index update monitor, replacing it with the much more flexible SearchIndexCoverageProvider.

Many of the changes have to do with other coverage providers, dealing with the change in the signature of `items_that_need_coverage`.


aslagle:
:+1:

PR #256, opened 2016-06-14
Remove licensepools from series lanes

### LINKED ISSUE #253: Don't use license pools to create series lanes
As of #245, we will hold SeriesLanes with a LicensePool, meaning each edition in a series makes its own series Lane. It makes more sense to cache the entire series--without a LicensePool reference--minimizing lookups and showing a work among its peers.

To wit:
> This discussion has actually made me think that the series feed should not be hanging on a LicensePool at all. I think we should have one feed per series, and it should be named after the series.
>
> Every LicensePool that belongs to a given series would link to the same series feed.
>
> This would reduce the naming problem (the feed would just be named after the series, no "Other Books in...") and also reduce the number of distinct feeds that need to be calculated.

(from #245 line notes)

courte:
Fixes #253, removing licensepools from SeriesLanes. It also:
- changes the way that URLs are created for some of these QueryGeneratedLanes (which is not a great name, since all Lanes are QueryGenerated - very open to ideas on this).
- moves some tests out of controller and instead tests the `works()` and `materialized_works()` methods directly for QueryGeneratedLanes.

### DIFF
--- a/api/opds.py
+++ b/api/opds.py
@@ -118,6 +118,12 @@ class CirculationManagerAnnotator(Annotator):
     def default_lane_url(self):
         return self.groups_url(None)
 
+    def custom_lane_url(self, lane):
+        if not hasattr(lane, 'url_arguments'):
+            raise ValueError('This lane has no arguments for a custom URL')
+        route, kwargs = lane.url_arguments
+        return self.cdn_url_for(route, **kwargs)
+
     def feed_url(self, lane, facets=None, pagination=None):
         lane_name, languages = self._lane_name_and_languages(lane)
         kwargs = dict({})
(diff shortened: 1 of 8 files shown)


api/opds.py
```diff
@@ -118,6 +118,12 @@ def groups_url(self, lane):
     def default_lane_url(self):
         return self.groups_url(None)
 
+    def custom_lane_url(self, lane):
```
leonardr: is there ever going to be a case where you'll want both custom_lane_url and feed_url()? If not, you could make this the QueryGeneratedLane implementation of feed_url().

migration/20160614-1-remove-licensepools-from-series-lanes.sql
```diff
@@ -0,0 +1 @@
+UPDATE cachedfeeds set license_pool_id=NULL where type='series';
```
leonardr: Rename this file to 20160617-... to avoid (my) confusion). I just run a bunch of these migration scripts, including a 20160615, and the 17th will be the earliest we'll be able to run this script for real.

Eventually these will be numbered with version numbers instead of dates. That won't eliminate the need to rename files occasionally but it should make it less common.

tests/test_controller.py
```diff
@@ -846,66 +844,23 @@ def test_report_problem_post_success(self):
...
-                self.datasource, self.identifier.type, self.identifier.identifier
+                "Like As If Whatever Mysteries"
             )
         eq_(200, response.status_code)
         feed = feedparser.parse(response.data)
         eq_("Like As If Whatever Mysteries", feed['feed']['title'])
         [entry] = feed['entries']
         eq_(self.english_1.title, entry['title'])
 
-        # Remove cache.
```
leonardr: Confirm this for me: you don't need this test anymore because you test the same code on a lower level in TestSeriesLane?

courte: Confirmed! In `TestSeriesLane.test_works_query` I'm making sure that the number and order of queries is what I expect, so here I'm just making sure that the query result is returned.

leonardr:
What happened to the SQL migration script? It looks like you removed it rather than renaming it.

Everything looks good otherwise.

courte:
Urgh, something with the git rearranging. It's back, now!

leonardr:
cool :+1:

PR #257, opened 2016-06-15
Scrape sample epubs

daryanypl:


daryanypl:
Cleaned up calls to the moved OPDS feed.

leonardr:
This looks good now!

PR #262, opened 2016-06-16
Temporary fix for search urls.

aslagle:


tests/test_opds.py
```diff
@@ -149,49 +149,50 @@ def test_single_entry_no_active_license_pool(self):
 
 class TestOPDS(DatabaseTest):
 
+    def setup(self):
+        super(TestOPDS, self).setup()
+        parent = Lane(self._db, "Fiction", languages=["eng"], fiction=True)
+        fantasy_lane = Lane(self._db, "Fantasy", languages=["eng"], genres=[Fantasy], parent=parent);
```
courte: You've got an extra semi-colon at the end here.

courte:
:+1: post-tests

tests/test_opds.py
aslagle: oops

PR #264, opened 2016-06-17
Propagate counts_as_covered when calling items_that_need_coverage

leonardr:
This branch changes the circulation manager coverage providers so that they pass along `counts_as_covered` when calling `items_that_need_coverage`. The `CoverageProviderScript` passes in that argument, and a CoverageProvider that doesn't accept it will crash.

I added tests for two cases: first, the MetadataWranglerCoverageProvider, which behaves like normal coverage providers, and second, the MetadataWranglerCollectionReaper, which ignores `counts_as_covered` because it is always looking for identifiers that were successfully covered by a _different_ CoverageProvider (MetadataWranglerCoverageProvider).

### DIFF
--- a/api/coverage.py
+++ b/api/coverage.py
@@ -208,9 +208,9 @@ class MetadataWranglerCoverageProvider(OPDSImportCoverageProvider):
                 "it will not know which collection you're asking about."
             )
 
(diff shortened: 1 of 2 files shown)


aslagle:
:+1:

PR #265, opened 2016-06-17
Fixed overdrive sync when no format is locked in.

aslagle:
This restores books that were checked out in Overdrive but not locked in to a format to the loans feed.


leonardr:
:+1: I guess I should have reviewed this before I deployed, but it looks good.

PR #267, opened 2016-06-17
added unit test to make sure authentication headers have proper quotes

daryanypl:


tests/test_controller.py
```diff
@@ -170,6 +173,28 @@ def test_authenticated_patron_correct_credentials(self):
...
+
+    def test_authentication_sends_proper_headers(self):
+        '''
+        Make sure the reals header has quotes around the realm name.  
+        Without quotes, some iOS versions don't recognize the header value.
+        '''
+        response = self.controller.authenticate()
+        
+        found_realm = False
+        for header_key, header_value in response.headers:
```
leonardr: You should also be able to look up the header directly in response.headers without iterating over all of them.

daryanypl: done

leonardr:
:+1: pending tests passing

PR #268, opened 2016-06-17
lane sweeper script uses real top-level lane

skomputer:


aslagle:
:+1: but we should try to write a test for this at some point

PR #269, opened 2016-06-17
Any format that can be fulfilled through simplified should show up in the loans feed.

aslagle:
This change will fix the StaleDataErrors we're occasionally seeing on fulfill.

The problem is that you can fulfill a loan as a format that's not supported by the default client (eg PDF), and if you also get your loans feed at the same time, the loan can get removed from the database after the fulfill request has fetched it. 

Any format that can be fulfilled by the circulation manager shouldn't be removed when you sync.

DEFAULT_READABLE_FORMATS was put in for Kindle format, and that's not in FORMATS either, so this should still work. It means PDFs borrowed via Overdrive will show up in the loans feed again, and you'll get an error when you try to download them.

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -406,16 +406,12 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI):
         start = cls._pd(checkout.get('checkoutDate'))
         end = cls._pd(checkout.get('expires'))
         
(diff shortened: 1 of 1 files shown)


leonardr:
:+1: I filed https://github.com/NYPL-Simplified/circulation/issues/270 to deal with the Kindle case.

PR #272, opened 2016-06-20
Removed code that only updated Axis 360 with circulation data once.

leonardr:
Previously, the Axis 360 monitor only called CirculationData.apply() when a LicensePool was new. This bug meant updates never showed up. 

I have left in place the code that only calls Metadata.apply() when an Edition is new, because that update takes a long time and bibliographic metadata generally doesn't change.


skomputer:
👍

PR #273, opened 2016-06-21
Add a configurable blacklist of strings found in invalid authorization identifiers.

leonardr:
Some NYPL patrons have multiple barcodes, all but one indicating previous cards that have been lost. When a patron has multiple barcodes our interface to the Millenium Patron API picks the last one in the list, but we want to make sure we pick a card that hasn't been lost. This branch offers a simple way to blacklist invalid barcodes based on regular expressions.


aslagle:
:+1:

PR #276, opened 2016-06-21
download analytics data

skomputer:
Added admin endpoint for downloading CSV file with circulation events between optional start and end dates.


api/admin/controller.py
```diff
@@ -606,16 +607,16 @@ def genres(self):
             })
         return data
 
-    def circulation_events(self):
+    def circulation_events(self, csv=False):
```
leonardr: It doesn't look like you use `csv` anywhere.

api/admin/controller.py
```diff
@@ -631,3 +632,58 @@ def circulation_events(self):
...
+
+        header = ["time", "type", "book_id", "title", "author", "fiction",
+                "audience", "publisher", "language", "target_age", "genre"]
+
+        def result_to_row(result):
+            (event, identifier, work, edition, genre) = result
+            return [
+                str(event.start) or "",
+                event.type,
+                identifier.identifier,
```
leonardr: You might also want to provide identifier.type.

api/admin/controller.py
```diff
@@ -631,3 +631,61 @@ def circulation_events(self):
         }, results)
 
         return dict({ "circulation_events": events })
+
+    def bulk_circulation_events(self):
+        annotator = AdminAnnotator(self.circulation)
```
leonardr: It looks like you don't use this Annotator (which makes sense since it's for OPDS feeds).

api/admin/controller.py
```diff
@@ -631,3 +631,61 @@ def circulation_events(self):
...
+            .filter(WorkGenre.affinity == t.c.max_affinity) \
+            .order_by(CirculationEvent.start.asc())
+
+        if start:
+            query = query.filter(CirculationEvent.start > start)
+
+        if end:
+            query = query.filter(CirculationEvent.end < end)
+
+        results = query.all()
```
leonardr: Now that I'm looking at the whole system, I'm uncomfortable with running this kind of open-ended query on the same machine that is trying to serve books to patrons. Can you send a problem detail instead if the gap between start and end is too wide (the default value of end being right now)? Let's set it at one day to start with.

This way someone who really does want all that data can get it, but they won't be tying up a connection for ten minutes, and they can space out the requests.

api/admin/controller.py
```diff
@@ -633,10 +634,9 @@ def circulation_events(self):
         return dict({ "circulation_events": events })
 
     def bulk_circulation_events(self):
-        annotator = AdminAnnotator(self.circulation)
-
-        start = flask.request.args.get("start")
-        end = flask.request.args.get("end")
+        default = str(datetime.today()).split(" ")[0]
+        date = flask.request.args.get("date", default)
```
leonardr: Is this a string or a date? If it's a string, how does line 659 work? If it's a date, why do you need to parse it on line 639?

skomputer: it's a string. line 659 works timestamp comparisons hold when timestamps are cast to strings, e.g. `"2016-06-23" < "2016-06-23 09:00:00" < "2016-06-24"`.

leonardr:
:+1:

PR #277, opened 2016-06-22
Get service_status working again

leonardr:
This branch fixes the service_status controller, which broke when we switched the way authentication mechanisms are configured.

This code has no test coverage because it involves the interaction of a lot of pieces that are difficult to test. This is why we didn't notice service_status was broken in the first place. I'm not happy about this but I'm not unhappy enough to put in the work necessary to cover this code in a way that's useful.

I've manually verified that this works for Millenium Patron and FirstBook authenticators. We also get an appropriate error message if the configuration is invalid or not accepted by the authenticator.

Currently we only verify a site's basic auth provider. In theory we should verify that we can authenticate test patrons with all auth providers, but I don't know how this would even work with an OAuth provider, so I didn't try to write the code.

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -178,3 +178,18 @@ class Authenticator(object):
             doc[type] = provider_info
 
         return json.dumps(doc)
+
+
+class BasicAuthAuthenticator(Authenticator):
+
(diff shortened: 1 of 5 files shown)


aslagle:
:+1:

PR #278, opened 2016-06-24
Fix import error

leonardr:
This branch fixes an import error and changes the CSV dialect used by the staff picks updater to use tab-separated files (which I generally think are less risky than comma-separated files).

I thought I was also removing an NYPL-specific script, but it turns out I never committed that script to the repository in the first place.


courte:
👍

PR #279, opened 2016-06-24
better analytics download

skomputer:
This branch improves admin/bulk_circulation_events by returning a list of genres in order of affinity for each book, including the relevant date in the name of the downloaded file, and returning an error instead of partial results when CSV encoding fails.


leonardr:
This all looks good. Is having multiple genres in a single field going to make it difficult to analyze the data?

leonardr:
Confirmed that multiple genres in a single field is better. :+1:

PR #280, opened 2016-06-24
Changed to the new cdnify API.

leonardr:
This branch changes the circ manager to use the new `cdnify` API created by https://github.com/NYPL-Simplified/server_core/pull/322.


courte:
👍

PR #281, opened 2016-06-29
OPDS import creates a Work even if LicensePool and Edition don't come from the same DataSource

leonardr:
This branch corrects OPDSImporter to work even when LicensePool and Edition come from different DataSources (which can happen--in fact almost always happens--when importing books from the open-access content server).

### DIFF
--- a/api/coverage.py
+++ b/api/coverage.py
@@ -70,9 +70,32 @@ class OPDSImportCoverageProvider(CoverageProvider):
         )
 
         results = []
+        leftover_identifiers = set()
+        # We grant coverage if an Edition was created from the operation.
         for edition in imported_editions:
-            self.finalize_edition(edition)
-            results.append(edition.primary_identifier)
+            identifier = edition.primary_identifier
+            results.append(identifier)
+            leftover_identifiers.add(identifier)
+
+        # We may also have created a LicensePool from the operation.
+        for pool in pools:
+            self.finalize_license_pool(pool)
+            identifier = pool.identifier
+            if identifier in leftover_identifiers:
+                leftover_identifiers.remove(identifier)
+            else:
+                msg = "OPDS import operation imported LicensePool, but no Edition."
+                results.append(
+                    CoverageFailure(
+                        identifier, msg, data_source=self.output_source,
+                        transient=True
+                    )
+                )
+        for identifier in leftover_identifiers:
+            self.log.warn(
+                "OPDS import operation imported Edition for %r, but no LicensePool.", 
+                identifier
+            )
 
         for failure in error_messages_by_id.values():
             results.append(failure)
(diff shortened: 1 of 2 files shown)


api/coverage.py
```diff
@@ -70,9 +70,32 @@ def process_batch(self, batch):
         )
 
         results = []
+        leftover_identifiers = set()
+        # We grant coverage if an Edition was created from the operation.
         for edition in imported_editions:
-            self.finalize_edition(edition)
-            results.append(edition.primary_identifier)
+            identifier = edition.primary_identifier
```
leonardr: Previously we called finalize_edition, but that code required edition.license_pool to be set, and it's not set if the DataSource don't match. Now we call finalize_license_pool and operate directly on the LicensePool (which may involve creating or setting a presentation_edition).

However, since it's reasonable to create an Edition but no LicensePool when doing an OPDS import, we count the operation as a success if an Edition was created.

api/coverage.py
```diff
@@ -70,9 +70,32 @@ def process_batch(self, batch):
...
+                leftover_identifiers.remove(identifier)
+            else:
+                msg = "OPDS import operation imported LicensePool, but no Edition."
+                results.append(
+                    CoverageFailure(
+                        identifier, msg, data_source=self.output_source,
+                        transient=True
+                    )
+                )
+        for identifier in leftover_identifiers:
```
leonardr: This is a little weird but not weird enough to be an error. It just means the data source has metadata about the book but no copies of the book.

api/coverage.py
```diff
@@ -86,22 +109,11 @@ def process_item(self, identifier):
...
         return result
 
-    def finalize_edition(self, edition):
-        """An OPDS entry has become an Edition. This method may (depending on
-        configuration) create a Work for that book and mark it as
+    def finalize_license_pool(self, pool):
+        """An OPDS entry has become a LicensePool. This method may (depending
+        on configuration) create a Work for that book and mark it as
         presentation-ready.
-        """
```
leonardr: All this method did was look for a LicensePool and run code on the LicensePool, so it was an easy change to make it take the LicensePool in the first place.

api/coverage.py
```diff
@@ -343,8 +358,9 @@ def __init__(self, _db, service_name=None, lookup=None, **kwargs):
         output_source = DataSource.lookup(
             _db, DataSource.OA_CONTENT_SERVER
         )
+        kwargs['input_identifier_types'] = None
         super(ContentServerBibliographicCoverageProvider, self).__init__(
-            service_name, input_identifier_types=None,
+            service_name,
```
leonardr: This is the fix tested by test_script_instantiation.

aslagle:
:+1: once the tests are passing

PR #282, opened 2016-06-30
Removed assumptions from tests

leonardr:
This branch makes the circulation tests pass with my rock-bottom minimal configuration which specifies nothing except the database connection information. Previously there were two tests that assumed that the configuration had a (potentially empty) ''policies' section.


skomputer:
👍

Issue #283, opened 2016-06-30
unglue.it books can't be borrowed


leonardr:
`https://circulation.librarysimplified.org/works/unglue.it/URI/https://unglue.it/api/id/work/4983//borrow" rel="http://opds-spec.org/acquisition/borrow` is a valid borrow link but gives a 404, probably because it contains embedded slashes.

leonardr:
The problem is in nginx. We need to add the following line to the `server` section:

`merge_slashes off;`

leonardr:
This has been fixed on Open Ebooks and SimplyE.

PR #284, opened 2016-06-30
Clarified the error message that occurs when the patron asks about a book we can't find.

leonardr:
This branch will help me diagnose https://github.com/NYPL-Simplified/circulation/issues/283


aslagle:
:+1:

Issue #274, opened 2016-06-21
Add code to determine Open Ebooks eligibility


leonardr:
For Open Ebooks, verifying a patron's Clever credentials are only half the story. We also need to verify that the patron is eligible for free/reduced school lunch or attends a Title I school.

Since this is an additional requirement specifically for Open Ebooks it should be possible to turn this verification on or off.

PR #285, opened 2016-06-30
Clever update

### LINKED ISSUE #231: Scripts that cache opds feeds don't work when an oauth provider is enabled
```
  File "bin/cache_opds_lane_facets", line 11, in <module>
    CacheFacetListsPerLane().run()
  File "/Users/amyslagle/dev/circulation/scripts.py", line 304, in __init__
    super(CacheRepresentationPerLane, self).__init__()
  File "/Users/amyslagle/dev/circulation/scripts.py", line 266, in __init__
    app.manager = CirculationManager(self._db)
  File "/Users/amyslagle/dev/circulation/api/controller.py", line 143, in __init__
    self.opds_authentication_document = self.auth.create_authentication_document()
  File "/Users/amyslagle/dev/circulation/api/authenticator.py", line 175, in create_authentication_document
    authenticate=provider.authenticate_url(),
  File "/Users/amyslagle/dev/circulation/api/clever.py", line 40, in authenticate_url
    redirect_uri = url_for('oauth_callback', _external=True)
  File "/Users/amyslagle/dev/circulation/env/lib/python2.7/site-packages/flask/helpers.py", line 258, in url_for
    raise RuntimeError('Attempted to generate a URL without the '
RuntimeError: Attempted to generate a URL without the application context being pushed. This has to be executed when application context is available.
```

aslagle:
This connects to #274 

This branch includes:
- New OPDS authentication document format that supports multiple providers better
- First Book has a different login label, instead of requiring the config to override it
- Clever is now a module that hopefully we can pull out later
- A script to import lists of districts with Title I funding
- The oauth callback for clever checks the user type and the district
- More tests for clever and authenticator

Also, this fixes #231


leonardr:
As I mentioned in person I have only two critiques:

1. You're using the long 'id' from Clever as both the authorization_identifier and the external_identifier. It's more in keeping with other usage (though not 100%) to use the 'id' as the external_identifier and the student number (the thing the student may have memorized) as the authorization_identifier. This means changing `authenticated_patron` to do a lookup based on external_identifier instead of external_identifier.
2. If it were my branch I'd put the Excel scraping script in the operations repo. It's not going to be around for the long term, and in the short term, all that's really necessary for the circulation manager is the _output_ of the script, which you've also included.

aslagle:
Changed the authorization identifier and the import script is here now: https://github.com/NYPL-Simplified/operations/pull/31

leonardr:
For the record, I was totally wrong about the authorization identifier. "Student number" isn't globally unique, and for teachers that number isn't even a "student number", it's a "teacher number", so the best thing to do is what you had before.

:+1:

api/authenticator.py
```diff
@@ -101,6 +101,15 @@ def create_token(self, provider_name, provider_token):
         )
         return jwt.encode(payload, self.secret_key, algorithm='HS256')
 
+    def get_credential_from_header(self, header):
+        if self.basic_auth_provider and 'password' in header:
+            return header['password']
+        elif self.oauth_providers and 'bearer' in header.lower():
+            simplified_token = header.split(' ')[1]
+            privder, token = self.decode_token(simplified_token)
```
leonardr: This is misspelled but it doesn't matter.

aslagle: I got rid of this function in the next commit.

api/authenticator.py
```diff
@@ -101,15 +101,6 @@ def create_token(self, provider_name, provider_token):
         )
         return jwt.encode(payload, self.secret_key, algorithm='HS256')
 
-    def get_credential_from_header(self, header):
```
leonardr: This method would still be useful for avoiding the copy-and-paste lookup of 'password'.

leonardr:
:+1: as is, but I think you could improve things a bit by bringing back a smaller version of get_credential_from_header.

aslagle:
Should get_credential_from_header return None for an oauth provider then?

leonardr:
Yeah, that's what I was thinking.

aslagle:
I've changed the way the OAuth flow works, so the circ manager is more like an OAuth proxy.


1. The circ manager creates an OPDS authentication document authenticate urls for each OAuth provider pointing to its own oauth_authenticate endpoint, e.g. /oauth_authenticate?provider=Clever
2. The client allows the user to choose an OAuth provider, then gets the authenticate url for that provider and optionally appends a redirect_uri to it. For mobile clients, the redirect_uri will be something like "open-ebooks-clever-1234://oauth". For web clients, it will be the page the user is currently on. If no redirect_uri is provided, the default is the root url.
3. The client sends the user to the authenticate url, eg. /oauth_authenticate?provider=Clever&redirect_uri=open-ebooks-clever-1234://oauth
4. The circ manager encodes the provider name and redirect_uri in a state parameter and redirects to the third-party provider's authentication page, with it's own oauth_callback endpoint as the redirect_uri.
5. The user logs in, and the third party provider redirects the user to the circ manager's /oauth_callback with a code and the state.
6. The circ manager extracts the provider name and redirect uri from the state, and uses the code to obtain an access token and optionally patron info from the provider. It creates a JSON web token containing the access token and the provider name.
7. The circ manager redirects the user to the redirect uri provided by the client, with the JSON web token and patron info in the hash fragment, e.g. open-ebooks-clever-1234://oauth#access_token=abcd.efgh.ijkl&patron_info=abcd
8. The client obtains the JSON web token and includes it in an "Authentication: Bearer" header in subsequent requests.
9. The circ manager parses the JSON web token to obtain the provider name and access token, and hits the provider's API to check that the access token is valid.

api/problem_details.py
```diff
@@ -184,6 +184,13 @@
     detail=_("The OAuth callback must contain a code and a state parameter with the OAuth provider name."),
 )
 
+UNKNOWN_OAUTH_PROVIDER = pd(
+    "http://librarysimplified.org/terms/problem/unknown-oauth-provider",
+    status_code=400,
+    title=_("Unknown OAuth provider."),
+    detail=_("The specified OAuth provider name isn't one of the known providers."),
```
leonardr: When you use this would it be useful to .detail() it with a list of the providers that are known?

leonardr:
This makes sense. :+1:

api/authenticator.py
```diff
@@ -123,32 +123,44 @@ def authenticated_patron(self, _db, header):
                     return provider.authenticated_patron(_db, provider_token)
         return None
 
+    def _redirect_with_error(self, redirect_uri, problem_detail):
+        problem_detail_json = json.dumps(dict(uri=problem_detail.uri, title=str(problem_detail.title), detail=str(problem_detail.detail)))
```
leonardr: Why do this instead of calling the `json` function in core/util/problem_detail.py?

aslagle: Good point, I changed it

leonardr:
:+1: once and for all!

PR #286, opened 2016-06-30
Handle two more server-side 3M errors.

leonardr:
This branch turns two more 3M errors into RemoteInitiatedServerErrors, meaning that they get shown as 502 errors instead of 500 errors, helping us distinguish between problems we need to fix and problems 3M needs to fix.

This time it's "Unknown error" and "Authentication failed". "Authentication failed" sure _sounds_ like it's our problem, but 3M doesn't check patron credentials, and if you give 3M the wrong client credentials you don't get this "Authentication failed" message at all; you get a 401 error with no entity-body. So we can reliably say that when 3M sends us "Authentication failed" it's always because something went wrong on the 3M side.

### DIFF
--- a/api/threem.py
+++ b/api/threem.py
@@ -1,4 +1,5 @@
 from lxml import etree
+
 from cStringIO import StringIO
 import itertools
 import datetime
@@ -146,10 +147,6 @@ class ThreeMAPI(BaseThreeMAPI, BaseCirculationAPI):
         else:
             # Error condition.
             error = ErrorParser().process_all(response.content)
-            if error.message == 'Unknown error':
-                raise RemoteInitiatedServerError(
-                    response.content, ThreeMAPI.SERVICE_NAME
-                )
             if isinstance(error, AlreadyCheckedOut):
                 # It's already checked out. No problem.
                 pass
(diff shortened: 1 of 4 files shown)


api/threem.py
```diff
@@ -404,6 +401,19 @@ def process_one(self, error_tag, namespaces):
...
+            #
+            # 'Authentication failed' could _in theory_ be an error on
+            # our side, but if authentication is set up improperly we
+            # actually get a 401 and no body. When we get a real error
+            # document with 'Authentication failed', it's always a
+            # transient error on the 3M side. Possibly some
+            # authentication internal to 3M has failed? Anyway, it
+            # happens relatively frequently.
+            return RemoteInitiatedServerError(
+                response.content.ThreeMAPI.SERVICE_NAME
```
aslagle: I think you lost a comma

leonardr:
I guess 'unknown error' was already being handled this way, but it wasn't tested.

api/threem.py
leonardr: Hm... that means my tests don't actually exercise this code.

leonardr: Aha! That code was raising an exception, but then the exception was being treated as an indicator of bad data from 3M, and a RemoteInitiatedServerError was created to represent the bad data. So my test was exercising this code and it was passing, but for totally the wrong reason.

aslagle:
:+1:

PR #288, opened 2016-07-06
admin authentication properly escapes redirect urls

skomputer:
Previously if an unauthenticated user tried to visit an admin page showing a collection or book containing an escaped "/" in its URL, then after authenticating they would be redirected to an unescaped and broken version of the original URL. The root of this bug is that WSGI unescapes all escaped "/" characters in any request URL, so the `admin_web` route couldn't access the original escaped URL to redirect to. This branch solves this problem by adding `admin_web` route patterns that explicitly identify the `collection` and `book` parts of the route so that they can be re-escaped in the redirect URL passed to the authentication flow.

### DIFF
--- a/api/admin/routes.py
+++ b/api/admin/routes.py
@@ -22,6 +22,7 @@ from templates import (
 
 import csv, codecs, cStringIO
 from StringIO import StringIO
+import urllib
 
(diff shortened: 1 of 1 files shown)


leonardr:
Sucks that you gotta do this but it looks good.

PR #292, opened 2016-07-08
Use os.path.abspath to get static js file path.

aslagle:
This was necessary to make the admin interface work on qa.


skomputer:
👍

PR #295, opened 2016-07-11
Preserve overdrive email address

leonardr:
Before putting an Overdrive book on hold, this branch checks with Overdrive to see what email address (if any) the patron used the last time they put a book on hold.

If such an email address exists, we tell Overdrive to send the notification for the book we're about to put on hold, to the same address. The only time we use the site address is when the patron has never before put an Overdrive book on hold.


aslagle:
Looks good once the tests are fixed

aslagle:
:+1:

PR #296, opened 2016-07-12
Handle the bizarre but real case of an Axis book which is checked out but not downloadable.

leonardr:
I don't know how this happened, but it did happen in production, and it revealed a case where a variable can be used before it was defined.

Hopefully stopping the crash will reveal the actual problem in the form of a patron complaint.

I think this only happens for Axis 360 books that are on hold, a scenario that currently almost never happens.


aslagle:
:+1:

PR #298, opened 2016-07-13
When a patron has multiple authorization identifiers, privilege the o…

### LINKED ISSUE #297: When a patron has multiple authorization identifiers, privilege the one they actually use to log in
Let's say a patron has two authorization identifiers in the ILS, A and B. They use A to log in, but for whatever reason we choose B to represent them to third parties such as 3M. When they check out a 3M book, it will be registered under authorization identifier B.

Now suppose this person _also_ uses A to log in to the 3M Cloud Reader website, and checks out a book on that website. That book is registered under authorization identifier A. 

Now the patron effectively has two 3M accounts, one for the website and one for the mobile app, even though they consistently used authorization identifier A to log in both places.

The best solution going forward is, when deciding which of a patron's identifiers to set as `Patron.authorization_identifier`, to privilege a) the one they already have set in `Patron.authorization_identifier`, or failing that b) the one they actually used to log in.

This does not solve the problem for patrons who are already in this mess. We can identify which patrons have the problem -- we use the ILS to determine which patrons have multiple authorization identifiers, and we use the 3M API to determine which of those patrons have 3M books for both sets of identifiers. But we can't automatically fix the problem because one way or another patrons will lose books.

So I think we'll need an operational script that we run to correct a patron's authorization identifier on an as-needed basis, as part of a support task.

aslagle:
…ne they actually use to log in. (Fixes #297)

### DIFF
--- a/api/millenium_patron.py
+++ b/api/millenium_patron.py
@@ -29,7 +29,7 @@ class MilleniumPatronAPI(BasicAuthAuthenticator, XMLParser):
     FINES_FIELD = 'MONEY OWED[p96]'
     EXPIRATION_DATE_FORMAT = '%m-%d-%y'
 
-    MULTIVALUE_FIELDS = set(['NOTE[px]'])
+    MULTIVALUE_FIELDS = set(['NOTE[px]', BARCODE_FIELD])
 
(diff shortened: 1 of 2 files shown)


leonardr:
:+1: looks great!

PR #301, opened 2016-07-14
Identical delivery mechanisms are interchangeable.

leonardr:
This branch stops a crash when a book has multiple delivery mechanisms with the same media type and DRM scheme. (This happens currently for EPUBs published by Standard Ebooks - they have an "EPUB" version, presumably EPUB 2, and an EPUB 3 version.)

The 'solution' is to just pick one at random. This is good enough to stop the crash but it would be better to always prefer the EPUB 3 version, based on the extension associated with the Resource associated with the LicensePoolDeliveryMechanism object.


leonardr:
Filed https://github.com/NYPL-Simplified/circulation/issues/302 for the long-term fix.

aslagle:
:+1:

PR #307, opened 2016-07-15
Update search index when license pool availability changes.

aslagle:
In https://github.com/NYPL-Simplified/server_core/pull/345, I added license pool availability info to the search index, so the search index will need to be updated whenever availability changes. 

This connects to https://github.com/NYPL-Simplified/server_core/issues/339

### DIFF
--- a/api/axis.py
+++ b/api/axis.py
@@ -166,14 +166,15 @@ class Axis360API(BaseAxis360API, Authenticator, BaseCirculationAPI):
         return list(AvailabilityResponseParser().process_all(
             availability.content))
 
-    def update_availability(self, licensepool):
+    def update_availability(self, licensepool, search_index_client=None):
         """Update the availability information for a single LicensePool.
 
         Part of the CirculationAPI interface.
         """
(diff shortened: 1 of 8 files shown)


api/axis.py
```diff
@@ -219,7 +223,10 @@ def update_licensepools_for_identifiers(self, identifiers):
                 licenses_reserved=0,
                 patrons_in_hold_queue=0,
             )
-            availability.apply(pool, False)
+            availability.apply(pool)
```
aslagle: This wasn't working before, but we didn't have test coverage for it.

api/threem.py
```diff
@@ -245,7 +245,7 @@ def apply_circulation_information_to_licensepool(self, circ, pool):
             self.log.info("Updating %s (%s)", e.title, e.author)
         else:
             self.log.info(
-                "Updating unknown work %s", identifier.identifier
+                "Updating unknown work %s", pool.identifier.identifier
```
aslagle: This also wasn't working before.

leonardr:
This was made moot by our recent changes to store collection information in the search index.

PR #308, opened 2016-07-15
Look up 3M books on the metadata wrangler by ISBN

### LINKED ISSUE #247: Send ISBNs for 3M books to the metadata wrangler, not the 3M IDs
I've discovered that 3M won't serve bibliographic data for a book that's not in your collection. This means it's pretty much useless for anyone to send a 3M ID to the metadata wrangler. As a circ manager, you can get the 3M metadata yourself and put it into a 3M edition. Then you can ask the metadata wrangler about the _ISBN_, and create a metadata wrangler edition based on that. 

This is what Axis 360 does already, so it's not that big a change. The real challenge is making the metadata wrangler do more useful work when given an ISBN.

https://github.com/NYPL-Simplified/metadata_wrangler/issues/66 tracks the metadata wrangler side of this work.

aslagle:
This will allow other libraries to look up 3M books that aren't in NYPL's 3M collection.

This fixes #247

### DIFF
--- a/api/coverage.py
+++ b/api/coverage.py
@@ -250,7 +250,7 @@ class MetadataWranglerCoverageProvider(OPDSImportCoverageProvider):
         """
         mapping = dict()
         for identifier in batch:
-            if identifier.type == Identifier.AXIS_360_ID:
+            if identifier.type in [Identifier.AXIS_360_ID, Identifier.THREEM_ID]:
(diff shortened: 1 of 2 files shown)


leonardr:
I tested 1200 randomly chosen 3M ISBNs and only two would have gotten no cover image from this lookup. That's good enough that this will work as an initial approach. :+1:

PR #309, opened 2016-07-18
Scripts to display raw data from Overdrive

leonardr:
Sometimes it's useful to see exactly what Overdrive says when asked for bibliographic or circulation data. This branch creates two new informational scripts for dumping this information to standard output. Previously I was writing one-off scripts for this and I'm tired of it.

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -548,7 +548,7 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI):
         else:
             book_id = book['id']
             circulation_link = book['availability_link']
-        return book_id, self.get(circulation_link, {})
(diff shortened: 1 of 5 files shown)


api/overdrive.py
```diff
@@ -549,35 +562,27 @@ def update_licensepool(self, book):
...
             self.log.error(
                 "HTTP exception communicating with Overdrive",
                 exc_info=e
             )
+
         if status_code != 200:
             self.log.error(
                 "Could not get availability for %s: status code %s",
-                book['id'], status_code
+                book_id, status_code
```
aslagle: I think this line needs to be changed back

aslagle:
Looks good except that one line

PR #311, opened 2016-07-19
Instead of parsing the Millenium patron dump as HTML, parse it as a text file delimited by <BR>\n

leonardr:
This seems like a clunkier solution than what we had before, but it's more robust on the data we actually have. The problem with parsing the dump as HTML is that the values may themselves contain HTML, including (presumably, though I haven't seen it) BR tags. Whereas I've never seen a value in the patron dump that contained embedded newlines--it looks like they get changed to dollar signs.


leonardr:
This fixes issue 310.

aslagle:
:+1:

PR #312, opened 2016-07-19
Properly handle an error if it occurs when getting the list of 3M events

leonardr:
This branch stops a couple minor problems from cluttering up our error log.

First: When the 3M monitor gets an error handling a 3M event it's logged and reraised, killing the script. When the monitor get an error _getting the list of events_ it's not logged because of an undefined variable. The script is still killed but the reason why is lost. (Although the reason why is almost certainly a timeout.) This simple branch makes sure the error is properly logged before being reraised.

Second: When a patron's card is blocked or expired, we don't allow them to carry out transactions but we do let them sync their list of loans. But Overdrive won't even let such a patron get their list of loans. We were logging this at the 'error' level because it seemed impossible that the ILS would authorize the patron but Overdrive would refuse to do so; but actually it happens all the time and is almost never a serious problem.


aslagle:
:+1:

PR #314, opened 2016-07-21
Update content server monitor

leonardr:
A couple small changes:

* Creates scripts to dump raw circulation and bibliographic data from 3M.
* Brings the content server OPDS import script up-to-date with the new OPDS import API.


bin/content_server_monitor
```diff
@@ -12,19 +12,13 @@ from core.model import DataSource
...
 print "Importing from %s" % content_server_url
-if import_presentation_ready:
-    print "Works will be made presentation-ready immediately."
 OPDSImportScript(
-    content_server_url, DataSource.OA_CONTENT_SERVER, OPDSImporter, 
-    register_timestamp, 
-    immediately_presentation_ready=import_presentation_ready
+    feed_url=content_server_url, opds_data_source=DataSource.OA_CONTENT_SERVER,
+    importer_class=OPDSImporter, 
+    immediately_presentation_ready=True
```
aslagle: I thought we didn't want certain data sources to be immediately presentation ready.

leonardr: That was true once but I don't think it's true anymore. My old design was that you'd get data from numerous sources and the book would be marked presentation ready once everything was perfect. But that led to huge delays, so I switched to a system where we marked a work as presentation ready once the basic info was in place. 

I definitely think once we hear from the content server we've got enough information about a book to start serving it to patrons. In theory there could still be an OPDS feed where importing wouldn't make it presentation ready immediately (maybe if the data was _wrong_ and needed to be corrected), but the content server isn't that feed.

aslagle:
:+1:

PR #315, opened 2016-07-21
Update overdrive formats when a delivery mechanism is no longer available

aslagle:
When we are unable to lock in a format that we thought was available, we will now update the license pool delivery mechanisms and give the user a specific 502. 

There's also a sweep script to update formats for all the overdrive books in our collection.


api/overdrive.py
```diff
@@ -262,7 +278,12 @@ def get_fulfillment_link(self, patron, pin, overdrive_id, format_type):
             response = self.lock_in_format(
                 patron, pin, overdrive_id, format_type)
             if response.status_code not in (201, 200):
-                raise CannotFulfill("Could not lock in format %s" % format_type)
+                if response.status_code == 400:
+                    message = json.loads(response.content).get("message")
+                    if message == "The selected format may not be available for this title.":
+                        raise FormatNotAvailable("This book is not Available the requested format")
```
leonardr: Copyedit: "This book is not available in the format you requested."

I don't normally copyedit but this string will go out to patrons.

tests/files/overdrive/lock_in_format_not_available.json
```diff
@@ -0,0 +1 @@
+{"errorCode":"PatronTitleProcessingFailed","message":"The selected format may not be available for this title.","token":"bf3b1876-20fa-4755-a923-acc809740002"}
```
leonardr: This is what Overdrive says? That's so wishy-washy.

leonardr:
:+1: after a copyediting change.

api/overdrive.py
aslagle: :+1:

tests/files/overdrive/lock_in_format_not_available.json
aslagle: It is :(

PR #316, opened 2016-07-22
Migration to restore missing open access hyperlinks.

aslagle:


leonardr:
:+1:

aslagle:
This version has been run:

Found 43019 open access pools without open access links
Fixed 40083 pools
1852 pools with no resource were not fixed
1084 pools with no identifier were not fixed

leonardr:
Looks good. Go ahead and merge it in case this is a recurring problem and we need to fix it again.

PR #321, opened 2016-07-25
WWW-Authenticate header omitted for XMLHttpRequests

skomputer:
In order for NYPL-Simplified/opds-web-client to display a custom login form based on OPDS authentication data from the server (title, login and password labels, etc), the `WWW-Authenticate` response header must be omitted because it triggers the browser's default basic authentication popup. This branch changes basic authentication so that this response header is omitted when the `X-Requested-With` request header is present and set to the commonly used `XMLHttpRequest`.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -299,8 +299,11 @@ class CirculationManagerController(object):
     def authenticate(self):
         """Sends a 401 response that demands authentication."""
         data = self.manager.opds_authentication_document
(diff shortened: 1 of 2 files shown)


aslagle:
I've moved this code in https://github.com/NYPL-Simplified/circulation/pull/285 :(

aslagle:
:+1: to commit https://github.com/NYPL-Simplified/circulation/pull/285/commits/1a1b25713d57ec854d1afd4271b30b7f2570ce35 in #285 

I'd like to close this, if we can wait until that branch is merged.

PR #322, opened 2016-07-26
Allow a client to do authdata token-based authentication through the interface normally used for username/password authentication.

leonardr:
This branch changes our implementation of the Adobe Vendor ID service to support token-based authentication even for clients that don't normally support it. This is all clients, because Adobe's client-side implementation can only use username/password authentication in some key places.

Here's how it works: you can get a single-use token by triggering the `create_authdata` endpoint. 
At this point you're supposed to sign in with an authData request, including the token as your authData. Instead, we have you sign in with a standard request (_not_ an authData request) and pass the single-use token in as the username, leaving the password blank.

If the password is blank, we try doing a token lookup based on the "username". Only if that fails do we try doing a username/password lookup.

### DIFF
--- a/api/adobe_vendor_id.py
+++ b/api/adobe_vendor_id.py
@@ -249,20 +249,39 @@ class AdobeVendorIDModel(object):
         UUID and their human-readable label, creating a Credential
         object to hold the UUID if necessary.
         """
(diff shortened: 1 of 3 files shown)


aslagle:
:+1:

PR #323, opened 2016-07-26
If the ILS gives an invalid expiration date for the patron's account, treat it as though the account never expires.

leonardr:
This code treats a patron whose EXP DATE field is set in the ILS to an invalid date (I saw "  -  -  ") as though their expiration date was None (meaning that their account never expires). Without this code, a patron whose EXP DATE is set to an invalid date can't log in.


aslagle:
:+1:

PR #326, opened 2016-07-26
Create contributor lane

### LINKED ISSUE #235: Link a feed of other works by an author to work entries
This first comes with the big step of merging or otherwise identifying and eradicating duplicate contributors from the db.

courte:
This branch creates an endpoint for all of the works of a particular contributor and includes results of that search to the RelatedBooksLane included in the OPDS feed.

Closes #235.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -860,7 +860,6 @@ class WorkController(CirculationManagerController):
             # No related books were found.
             return NO_SUCH_LANE.detailed(e.message)
 
-
         url = self.cdn_url_for(
             'related_books', data_source=data_source,
             identifier_type=identifier_type, identifier=identifier
--- a/api/controller.py
+++ b/api/controller.py
@@ -795,7 +795,7 @@ class WorkController(CirculationManagerController):
         """Serve a feed of books written by a particular author"""
 
         if not contributor_name:
-            return NO_SUCH_LANE.detailed("No series provided")
+            return NO_SUCH_LANE.detailed("No contributor provided")
 
         lane = ContributorLane(self._db, contributor_name)
 
--- a/api/controller.py
+++ b/api/controller.py
@@ -791,15 +791,13 @@ class LoanController(CirculationManagerController):
 
 class WorkController(CirculationManagerController):
 
-    def contributor(self, contributor_name, contributor_id=None):
+    def contributor(self, contributor_name):
         """Serve a feed of books written by a particular author"""
 
         if not contributor_name:
             return NO_SUCH_LANE.detailed("No series provided")
 
-        lane = ContributorLane(
-            self._db, contributor_name, contributor_id=contributor_id
-        )
+        lane = ContributorLane(self._db, contributor_name)
 
         annotator = self.manager.annotator(lane)
         url = annotator.feed_url(
--- a/api/controller.py
+++ b/api/controller.py
@@ -23,8 +23,6 @@ from core.app_server import (
     load_lending_policy,
     load_facets_from_request,
     load_pagination_from_request,
-    load_facets,
-    load_pagination,
     ComplaintController,
     HeartbeatController,
     URNLookupController,
@@ -84,6 +82,7 @@ from config import (
 
 from lanes import (
     make_lanes,
+    ContributorLane,
     RecommendationLane,
     RelatedBooksLane,
     SeriesLane,
@@ -792,6 +791,29 @@ class LoanController(CirculationManagerController):
 
 class WorkController(CirculationManagerController):
 
+    def contributor(self, contributor_name, contributor_id=None):
+        """Serve a feed of books written by a particular author"""
+
+        if not contributor_name:
+            return NO_SUCH_LANE.detailed("No series provided")
+
+        lane = ContributorLane(
+            self._db, contributor_name, contributor_id=contributor_id
+        )
+
+        annotator = self.manager.annotator(lane)
+        url = annotator.feed_url(
+            lane,
+            facets=load_facets_from_request(),
+            pagination=load_pagination_from_request()
+        )
+
+        feed = AcquisitionFeed.page(
+            self._db, lane.display_name, url, lane,
+            annotator=annotator, cache_type=CachedFeed.CONTRIBUTOR_TYPE
+        )
+        return feed_response(unicode(feed.content))
+
     def permalink(self, data_source, identifier_type, identifier):
         """Serve an entry for a single book.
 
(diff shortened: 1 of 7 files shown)


api/controller.py
```diff
@@ -792,6 +791,29 @@ def detail(self, data_source, identifier_type, identifier):
 
 class WorkController(CirculationManagerController):
 
+    def contributor(self, contributor_name, contributor_id=None):
+        """Serve a feed of books written by a particular author"""
+
+        if not contributor_name:
+            return NO_SUCH_LANE.detailed("No series provided")
```
leonardr: Should be "no contributor"

api/controller.py
```diff
@@ -829,16 +851,17 @@ def recommendations(self, data_source, identifier_type, identifier,
             # NoveList isn't configured.
             return NO_SUCH_LANE.detailed(_("Recommendations not available"))
 
-        url = self.cdn_url_for(
-            'recommendations', data_source=data_source,
-            identifier_type=identifier_type, identifier=identifier
-        )
         annotator = self.manager.annotator(lane)
+        url = annotator.feed_url(
```
leonardr: This is a good change.

api/lanes.py
```diff
@@ -562,3 +590,59 @@ def apply_filters(self, qu, work_model=Work, *args, **kwargs):
...
         return qu
+
+
+class ContributorLane(QueryGeneratedLane):
+    """A lane of Works written by a particular contributor"""
+
+    ROUTE = 'contributor'
+    MAX_CACHE_AGE = 48*60*60    # 48 hours
+
+    def __init__(self, _db, contributor_name, contributor_id=None):
```
leonardr: I'm a little concerned about the use of database IDs in URLs because contributors can get merged into each other and deleted. A person's database ID at the time the URL is generated might not be their ID at the time the link is followed. 

I think this is okay for now but I would like to move towards using VIAF ID in the URLs, once we have more reliable sources of VIAF IDs, and use the database ID as a fallback for when we don't have VIAF ID.

api/lanes.py
```diff
@@ -562,3 +590,59 @@ def apply_filters(self, qu, work_model=Work, *args, **kwargs):
...
+        if contributor_id:
+            self.contributor = get_one(_db, Contributor, id=contributor_id)
+            if (self.contributor_name!=self.contributor.display_name and
+                self.contributor_name!=self.contributor.name):
+                raise ValueError(
+                    "ContributorLane can't be created with inaccurate"
+                    " Contributor data."
+                )
+
+        full_name = display_name = "Books by %s" % self.contributor_name
```
leonardr: This string needs to be localized but that's a big job that can't go into this branch. I filed https://github.com/NYPL-Simplified/server_core/issues/358 to deal with it.

api/lanes.py
```diff
@@ -562,3 +590,59 @@ def apply_filters(self, qu, work_model=Work, *args, **kwargs):
...
+        return self.ROUTE, kwargs
+
+    def apply_filters(self, qu, work_model=Work, *args, **kwargs):
+        if not self.contributor_name:
+            return None
+        qu = self.only_show_ready_deliverable_works(qu, work_model)
+
+        work_edition = aliased(Edition)
+        qu = qu.join(work_edition).join(work_edition.contributions)
+        qu = qu.join(Contribution.contributor).order_by(False)
```
leonardr: I don't know what `order_by(False)` does.

tests/test_controller.py
```diff
@@ -868,35 +901,54 @@ def test_related_books(self):
...
+        eq_(True, href.endswith(expected_series_link))
+
+        # The other book by this contributor is in the contributor feed.
+        [e3] = [e for e in feed['entries'] if e['title'] == same_author.title]
+        title, href = collection_link(e3)
+        eq_("Books by John Bull", title)
+        expected_contributor_link = urllib.quote('contributor/John Bull')
+        eq_(True, href.endswith(expected_contributor_link))
+
+        # The original book is listed in both the series and contributor feeds.
```
leonardr: Nice test.

api/lanes.py
courte: At this moment, there's no point when the URL has the contributor_id; it's only passed in when the RelatedBooksLane is initialized. I think I may have been thinking about the connection between route and lane too strictly. I can remove contributor_id from the route, but leave it here. Either way, the cached feed will have used the ID, but it won't be present or shared anywhere in the feed; the URL that the lane generates uses only the contributor display_name.

api/lanes.py
courte: This was necessary when I was experimenting with unions, but no longer. *&#42;deletes aggressively&#42;*

api/lanes.py
leonardr: Yes, that sounds great. I'd still like to move towards VIAF to avoid the problem of combining books by different people with the same name. But long term, like I said.

aslagle:
If you add any new strings that should be localized, you can wrap them in `_` - e.g. `_("No contributor provided")` and they will be added to the translation files we export. Lane names can wait, but the problem details will already work, once we get the translations for them.

courte:
Thanks, @aslagle!

leonardr:
:+1: but I have a final question that can be addressed later if it's a problem.

When we create the sublane for 'books by this author' that's going to go into the "related books" lane, what are the facets? I remember saying that it should be ordered alphabetically, but I was thinking of the main lane you get when you click through.

Does the sublane only show books that are currently available? If all of an author's books are checked out, does the lane not show up?

Does the sublane apply the default quality filter? If all of an author's books are rated very poorly, does the lane not show up?

I guess these questions also apply to the other "related books" sublanes like series and recommendations, but this is the first time I really thought in terms of a patron expecting to see specific books and not finding them.

My dream scenario is that the "by this author" sublane would contain a selection of the highest-rated books by that author, with preference given to books that are available now. But showing books in alphabetical order is fine for now, assuming that the lane shows books even if they're checked out or lousy.

This should go into an issue but I want to see whether there's even a problem that needs to be addressed, or if we just have an opportunity to make a good thing better.

courte:
The books won't be ordered according to availability or quality (although I think we have some code for that somewhere for the Staff Picks & Bestseller feeds, right?). Right now they don't take quality or age range or a number of other items into account, but that wouldn't be a difficult change to make. Can we make a follow-up issue for it, and I'll work on that next?

As for what's happening now:
- The sublanes use `Lane#only_show_ready_deliverable_works`, which means it only shows deliverable, unsuppressed works with owned licenses. If holds are allowed, it will show books that have been checked out.
- Empty QueryBasedLanes are possible. An empty RelatedBooksLane is the most likely annoyance the client would face, since its generation is only based on whether there's a potential for related books, not whether there are any present in the database.
- If we only access series/recommendation/contributor lanes via the RelatedBooksLane, we shouldn't have an issue with empty lanes since the collection links wouldn't have entries to be added to. If we use those routes independently, though, those could also turn up empty.

PR #327, opened 2016-07-27
convert strings to utf8 before posting data to google analytics

skomputer:
`GoogleAnalyticsProvider` uses `urllib.urlencode` to post data to an analytics API endpoint, but when data includes `Unicode` strings it produces an error like this:

```
UnicodeEncodeError: 'ascii' codec can't encode character u'\xf3' in position 9: ordinal not in range(128)
```

This branch solves this problem by converting all strings to `UTF-8` before passing them to `urllib.urlencode`.

### DIFF
--- a/api/google_analytics_provider.py
+++ b/api/google_analytics_provider.py
@@ -45,6 +45,8 @@ class GoogleAnalyticsProvider(object):
                 'cd10': edition.language,
                 'cd11': work.top_genre()
             })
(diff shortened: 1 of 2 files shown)


leonardr:
This looks good. Can you take this opportunity to remove the test logging you added when debugging the configuration problem?

leonardr:
:+1:

At some point take the logging out of core as well.

Issue #329, opened 2016-07-28
A book with no fulfillable delivery mechanisms should not show up in a feed


leonardr:
Currently a book with no delivery mechanisms shows up in a feed with a `borrow` link that does not list any indirect acquisition mechanisms (since there are none). The problem is that this looks just like a generic OPDS feed where the acquisition mechanisms are present but not specified.

A book with no delivery mechanisms should not show up in the feed, period.

leonardr:
This will not be terribly important soon because this mostly happens with Overdrive books, and Overdrive books will in the near future have Overdrive Read as a fulfillable delivery mechanism.

leonardr:
Amy resolved this problem in a recent commit.

PR #332, opened 2016-07-29
Added fulfill links for Overdrive's streaming reader.

aslagle:


api/circulation.py
```diff
@@ -342,7 +342,7 @@ def fulfill(self, patron, pin, licensepool, delivery_mechanism, sync_on_failure=
                 )
             else:
                 raise NoActiveLoan(_("Cannot find your active loan for this work."))
-        if loan.fulfillment is not None and loan.fulfillment != delivery_mechanism:
+        if loan.fulfillment is not None and loan.fulfillment != delivery_mechanism and not delivery_mechanism.delivery_mechanism.is_streaming:
```
leonardr: In theory a streaming delivery mechanism could be exclusive of other delivery mechanisms, so we'd have to store that information per data source per mechanism and check if there are any conflicts. Fortunately for now that doesn't seem to be the case.

api/opds.py
```diff
@@ -430,16 +430,20 @@ def acquisition_links(self, active_license_pool, active_loan, active_hold,
...
+                    if lpdm is active_loan.fulfillment or lpdm.delivery_mechanism.is_streaming:
+                        fulfill_links.append(
+                            self.fulfill_link(
+                                data_source_name,
+                                identifier,
+                                active_license_pool,
+                                active_loan,
+                                lpdm.delivery_mechanism
+                            )
+                        )
```
leonardr: Is it trivial to add on to this code that fixes https://github.com/NYPL-Simplified/circulation/issues/329 by cancelling the creation of the entry if there are no delivery mechanisms whatsoever?

I think the answer is no but I also think you know better than I do, having just looked at this code.

api/overdrive.py
```diff
@@ -311,6 +319,11 @@ def get_fulfillment_link(self, patron, pin, overdrive_id, format_type):
             return response
 
     def get_fulfillment_link_from_download_link(self, patron, pin, download_link):
+        # If this for Overdrive's streaming reader, and the link expires,
+        # the patron can go back to this url again to get a new one.
+        auth_url = flask.request.url
```
leonardr: I'd like to keep Flask references out of the circulation API. Could this URL be passed in as an argument?

tests/test_controller.py
```diff
@@ -465,6 +466,127 @@ def test_borrow_success(self):
...
+            response = self.manager.loans.borrow(
+                data_source.name, identifier.type, identifier.identifier)
+
+            # A loan has been created for this license pool.
+            loan = get_one(self._db, Loan, license_pool=pool)
+            assert loan != None
+            # The loan has yet to be fulfilled.
+            eq_(None, loan.fulfillment)
+
+            # We've been given an OPDS feed with two entries, which tell us how
```
leonardr: Two delivery mechanisms

api/opds.py
aslagle: I can try that, the core code handles an UnfulfillableWork exception so I can probably raise that somewhere in the code that calls this.

aslagle: Although, I think that would put an error in the feed which we don't want.

leonardr: I think it's probably okay to include that error in the feed, at least at the start. It won't get shown to patrons and it should stop us from tearing our hear out wondering why a book doesn't show up.

aslagle: It's a little more complicated so I will make that a separate branch.

leonardr:
Are you sure that the sample data you included doesn't contain any reusable credentials or identifying information? If so then :+1:, otherwise garble it.

aslagle:
The download link's expired already, and I think the reserve id only identifies a particular loan.

api/opds.py
aslagle: It wasn't that bad so I added it on here after all. We'll have to make sure the clients don't choke on the error messages.

leonardr:
:+1:

PR #333, opened 2016-08-02
The authdata token used in Adobe authentication is now persistent.

leonardr:
This branch changes the name of the `Credential` used to authenticate a patron through Adobe authdata. This `Credential` is now persistent--it does not expire and is not consumed on use. Having this credential allows a client to deactivate a device through Adobe without patron input, which is useful when a patron's credentials have changed.


tests/test_adobe_vendor_id.py
```diff
@@ -74,12 +74,19 @@ def test_standard_lookup_success(self):
...
     def test_authdata_lookup_success(self):
+        
+        # Create an authdata token for Bob.
         now = datetime.datetime.utcnow()
-        temp_token, ignore = Credential.temporary_token_create(
-            self._db, self.data_source, self.model.TEMPORARY_TOKEN_TYPE,
-            self.bob_patron, datetime.timedelta(seconds=60))
-        old_expires = temp_token.expires
-        assert temp_token.expires > now
+        temp_token, ignore = Credential.persistent_token_create(
```
aslagle: you might want to change this variable name

aslagle:
Looks good except for those persistent temp_tokens :)

PR #334, opened 2016-08-03
Handle firstbook error

leonardr:
This branch ensures that errors from the First Book API are properly turned into `RemoteInitiatedServerError` exceptions. Previously the constructor was initiated without the required service name.


aslagle:
:+1:

PR #335, opened 2016-08-04
Made fulfilling a streaming delivery mechanism return an OPDS entry with the streaming fulfill link.

aslagle:


leonardr:
This is more complicated than I thought it would be but it makes sense. :+1:

PR #337, opened 2016-08-04
Added CORS headers for patron web client.

aslagle:
The patron web client will be on a different domain than the circ manager, so the circ manager needs Cross-Origin Resource Sharing headers for browsers to let it make requests.

This branch adds a decorator that will allow CORS requests from a specific patron web client URL in the config file, and allow authentication headers and cookies in those requests.

The decorator is only used on routes that will be used by the patron web client.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -44,6 +44,7 @@ class Configuration(CoreConfiguration):
     MILLENIUM_INTEGRATION = "Millenium"
     AUTHORIZATION_IDENTIFIER_BLACKLIST = "authorization_identifier_blacklist"
     STAFF_PICKS_INTEGRATION = "Staff Picks"
+    PATRON_WEB_CLIENT_INTEGRATION = "Patron Web Client"
 
     LIST_FIELDS = "fields"
    
--- a/api/routes.py
+++ b/api/routes.py
@@ -117,36 +117,46 @@ else:
             return f
         return decorated
 
+def dir_route(path_without_slash, *args, **kwargs):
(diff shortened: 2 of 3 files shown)


aslagle:
By default, if there's a route with a trailing slash, like '/groups/', Werkzeug will automatically redirect a request without the slash to the route with the slash. This happens even if you defined a route without the slash too - we had done that on some routes, but it wasn't doing anything. Werkzeug handles all this outside of the app and doesn't call the route code or any of the decorators.

When the browser makes a CORS request, it first sends an OPTIONS "preflight" request and only makes the real request if it gets a response with the appropriate headers. The decorator will check if the request type is "OPTIONS" and return a 200 with CORS headers, but the redirects from Werkzeug aren't valid preflight responses, so the browser wouldn't follow them. 

I'm not sure how best to handle this, but in commit f376100 I disabled Werkzeug's redirects, so a request to a URL like '/groups' will return the same response as '/groups/'. I added explicit routes without slashes in a couple cases where we didn't have them, since they'll no longer be automatically redirected.

leonardr:
How about defining a decorator that registers both the path as is and with a slash?

`@app.routedir('/works')`

would register `/works` and `/works/`.

You'd have to remember to use that decorator when appropriate, but it seems easier to remember than having to remember to call `route` twice.

I tried to think of a way to combine this with your `allows_patron_web` but I think the two decorators are orthogonal.

leonardr:
:+1:

PR #341, opened 2016-08-16
Finalize batch

leonardr:
This branch updates the circ manager to reflect the fact that `CoverageProvider.finalize_batch()` now commits the database session. We have a coverage provider that implements `finalize_batch` and we want to make sure it calls the superclass.


aslagle:
:+1:

PR #347, opened 2016-08-22
Fix and test the client for the metadata wrangler collection reaper

leonardr:
This branch changes the metadata wrangler collection reaper to accommodate the changes made by
https://github.com/NYPL-Simplified/metadata_wrangler/pull/85, and adds a test for previously untested code. Instead of attempting a full OPDS feed import, we just extract OPDSMessage objects from the feed. Those messages will tel us which items were successfully reaped and which ones failed for some reason.

### DIFF
--- a/api/coverage.py
+++ b/api/coverage.py
@@ -1,6 +1,8 @@
 import logging
+from lxml import etree
 from nose.tools import set_trace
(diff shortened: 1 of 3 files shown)


courte:
👍

PR #348, opened 2016-08-23
Changed search index update script to use a Monitor instead of a CoverageProvider.

aslagle:
With https://github.com/NYPL-Simplified/server_core/pull/377, creating search documents is very fast, so it's no longer expensive to reindex all the works. Without the CoverageRecords, I can make a complete search index in 10 minutes!


leonardr:
:+1:

Issue #351, opened 2016-08-30
Someone isn't checking to see if `load_pagination_from_request` returns a ProblemDetail


leonardr:
I believe this is the request that causes the problem:

GET /works/series/Tin%20Star?available=now&after=50&order=author&collection=main&size=50

```
Traceback (most recent call last):
  File "/var/lib/app/env/local/lib/python2.7/site-packages/flask/app.py", line 1639, in full_dispatch_request
    rv = self.dispatch_request()
  File "/var/lib/app/env/local/lib/python2.7/site-packages/flask/app.py", line 1625, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File "/var/lib/app/env/local/lib64/python2.7/site-packages/newrelic-2.66.0.49/newrelic/hooks/framework_flask.py", line 40, in _nr_wrapper_handler_
    return wrapped(*args, **kwargs)
  File "./core/app_server.py", line 167, in decorated
    v = f(*args, **kwargs)
  File "./api/routes.py", line 239, in series
    return app.manager.work_controller.series(series_name)
  File "./api/controller.py", line 945, in series
    pagination=load_pagination_from_request()
  File "./api/opds.py", line 130, in feed_url
    kwargs.update(dict(facets.items()))
AttributeError: 'ProblemDetail' object has no attribute 'items'
```

leonardr:
I can't duplicate the 500 error on production or locally, but I do see what the problem is.

Issue #350, opened 2016-08-30
Overdrive representation edge cases crash scripts


leonardr:
I see a couple cases where an unexpected bit of missing data or weird representation causes an Overdrive script to crash. We should be able to handle these representations assuming they're not totally broken. And if there's one representation we can't handle it shouldn't crash the whole script, we should just move on to the next representation.

leonardr:
```

Traceback (most recent call last):
  File "/var/lib/app/circulation/core/scripts.py", line 111, in run
    self.do_run()
  File "/var/lib/app/circulation/core/scripts.py", line 132, in do_run
    self.monitor.run()
  File "/var/lib/app/circulation/api/overdrive.py", line 810, in run
    super(OverdriveCirculationMonitor, self).run()
  File "/var/lib/app/circulation/core/monitor.py", line 74, in run
    new_timestamp = self.run_once(start, cutoff) or cutoff
  File "/var/lib/app/circulation/api/overdrive.py", line 820, in run_once
    for i, book in enumerate(self.recently_changed_ids(start, cutoff)):
  File "/var/lib/app/circulation/core/overdrive.py", line 236, in all_ids
    next_link, 'next'
  File "/var/lib/app/circulation/core/overdrive.py", line 258, in _get_book_list_page
    OverdriveRepresentationExtractor.availability_link_list(data))
  File "/var/lib/app/circulation/core/overdrive.py", line 405, in availability_link_list
    title=product['title'],
KeyError: 'title'
```

leonardr:
This crashed the Overdrive Collection Reaper:

```
Traceback (most recent call last):
  File "/var/lib/app/circulation/core/monitor.py", line 231, in run
    new_offset = self.run_once(offset)
  File "/var/lib/app/circulation/core/monitor.py", line 260, in run_once
    self.process_batch(identifiers)
  File "/var/lib/app/circulation/api/overdrive.py", line 888, in process_batch
    self.api.update_licensepool(i.identifier)
  File "/var/lib/app/circulation/api/overdrive.py", line 637, in update_licensepool
    book['id'], status_code
TypeError: string indices must be integers
```

Issue #344, opened 2016-08-17
Next links in contributor lane don't work


aslagle:
The contributor lane feeds have next links, but if you follow the next link you get a feed with the same books.

aslagle:
Same for the series lane.

Issue #343, opened 2016-08-17
Contributor lane returns 500 if loading facets returns a problem detail


aslagle:
It should return the problem detail.

Example: http://qa.circulation.openebooks.us/works/contributor/Tougas%2C%20Shelley?available=now&after=50&order=author&collection=main&size=50

This is the error:
{"host": "ip-10-0-3-239.ec2.internal", "name": "root", "level": "ERROR", "timestamp": "2016-08-17T19:31:38.273567", "app": "simplified", "traceback": "Traceback (most recent call last):\n File \"/var/lib/app/env/local/lib/python2.7/site-packages/flask/app.py\", line 1639, in full_dispatch_request\n rv = self.dispatch_request()\n File \"/var/lib/app/env/local/lib/python2.7/site-packages/flask/app.py\", line 1625, in dispatch_request\n return self.view_functions[rule.endpoint](**req.view_args)\n File \"/var/lib/app/env/local/lib64/python2.7/site-packages/newrelic-2.66.0.49/newrelic/hooks/framework_flask.py\", line 40, in _nr_wrapper_handler_\n return wrapped(_args, *_kwargs)\n File \"./core/app_server.py\", line 167, in decorated\n v = f(_args, *_kwargs)\n File \"./api/routes.py\", line 233, in contributor\n return app.manager.work_controller.contributor(contributor_name)\n File \"./api/controller.py\", line 825, in contributor\n pagination=load_pagination_from_request()\n File \"./api/opds.py\", line 130, in feed_url\n kwargs.update(dict(facets.items()))\nAttributeError: 'ProblemDetail' object has no attribute 'items'", "message": "Exception in web app: 'ProblemDetail' object has no attribute 'items'", "filename": "app_server.py"}

aslagle:
Also, I'm not sure why those facets were invalid.

aslagle:
I think the web client may be incorrectly using facets from a previous request.

aslagle:
The web client was getting them from the contributor lane feed - it has facet links that don't work.

PR #352, opened 2016-08-30
update_licensepool should handle unhelpful errors, and errors in building a series feed should be propagated

leonardr:
This branch fixes two unrelated issues: 

The overdrive collection reaper crash in https://github.com/NYPL-Simplified/circulation/issues/350

and https://github.com/NYPL-Simplified/circulation/issues/351

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -926,6 +926,7 @@ class WorkController(CirculationManagerController):
             return NO_SUCH_LANE.detailed(e.message)
 
         annotator = self.manager.annotator(lane)
+        facets = load_facets_from_request()
         if isinstance(facets, ProblemDetail):
             return facets
         pagination = load_pagination_from_request()
--- a/api/controller.py
+++ b/api/controller.py
@@ -952,10 +952,16 @@ class WorkController(CirculationManagerController):
 
         lane = SeriesLane(self._db, series_name)
         annotator = self.manager.annotator(lane)
+        facets = load_facets_from_request()
+        if isinstance(facets, ProblemDetail):
+            return facets
+        pagination = load_pagination_from_request()
+        if isinstance(pagination, ProblemDetail):
+            return pagination
         url = annotator.feed_url(
             lane,
-            facets=load_facets_from_request(),
-            pagination=load_pagination_from_request()
+            facets=facets,
+            pagination=pagination,
         )
 
         feed = AcquisitionFeed.page(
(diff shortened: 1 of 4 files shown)


tests/test_overdrive.py
```diff
@@ -323,12 +323,11 @@ def test_update_licensepool_error(self):
             "overdrive_availability_information.json"
         )
         api = DummyOverdriveAPI(self._db)
-        api.queue_response(response_code=500, content=availability)
+        api.queue_response(response_code=500, content="An error occured.")
```
leonardr: Our error handling only worked if you sent a non-200 response code but also provided the right data. That's the case we were testing but that's not actually how Overdrive deals with errors.

api/controller.py
```diff
@@ -952,10 +952,16 @@ def series(self, series_name):
 
         lane = SeriesLane(self._db, series_name)
         annotator = self.manager.annotator(lane)
+        facets = load_facets_from_request()
+        if isinstance(facets, ProblemDetail):
+            return facets
+        pagination = load_pagination_from_request()
+        if isinstance(pagination, ProblemDetail):
+            return pagination
```
aslagle: Cool, I had noticed this too. It's also a problem in the contributor lane.

aslagle: https://github.com/NYPL-Simplified/circulation/issues/343

leonardr: I'll take care of that one as well. Thanks for the link!

aslagle: Also the reason we're seeing this is that the contributor and series feeds have links with pagination and facets that don't work - I made an issue for that too https://github.com/NYPL-Simplified/circulation/issues/344

aslagle:
:+1:

api/controller.py
leonardr: Hmm, I won't fix that now, but I've fixed the `load_*` parts for the contributor and recommendation feeds.

api/controller.py
```diff
@@ -914,10 +926,15 @@ def related(self, data_source, identifier_type, identifier,
             return NO_SUCH_LANE.detailed(e.message)
 
         annotator = self.manager.annotator(lane)
+        if isinstance(facets, ProblemDetail):
```
aslagle: do you need to load the facets for this one?

leonardr: :blush:

aslagle:
:+1:

PR #355, opened 2016-09-01
Axis 360 reaper script now uses new API call

leonardr:
This branch changes the Axis 360 reaper script to use a new API call that returns all titles added to or removed from the collection as of a certain date. This is at the request of Baker & Taylor.

The old script, which checks every single book in the collection, is still around (renamed from `axis_reaper` to `axis_monitor`), and my experience with Axis 360 collections indicates it will still need to be run occasionally, but the new `axis_reaper` can take over the everyday job of seeing which books have been removed from the collection.

There's a lot of refactoring in this branch. Notably, I refactor the code that 'reaps' a LicensePool (zeroes out its `licenses_owned` etc.) so that it can be used by the reaper as well as by `update_licensepools_for_identifiers()` when it happens to discover that a book has been removed from the collection.

I also refactor `ResponseParser.exception_for_response_code`. The new API call returns JSON data rather than XML, so I had to reimplement a lot of the basic parsing code, but _hopefully_ the rules for error handling -- there's a numeric error code that corresponds to one of the exceptions in `circulation_exceptions.py` -- are unchanged.

I say "hopefully" because the design of this new API call is not finalized. The new API is not yet in production yet, and the documents I get from the QA server (including the format of the status messages) are inconsistent with the sample documents in the documentation. In the unit tests I use what I think the documents will end up looking like, but really, who knows? I might have to go back through this and change some things.

The good news is that I found a bug that would have prevented the old reaper script from working, so landing this branch won't break anything that wasn't already broken. However, unless someone needs to do other Axis 360 work in the near future, I think it's best we let this branch sit until we know what the final API looks like.


api/axis.py
```diff
@@ -197,29 +216,37 @@ def update_licensepools_for_identifiers(self, identifiers):
...
-                licenses_available=0,
-                licenses_reserved=0,
-                patrons_in_hold_queue=0,
+        pool = identifier.licensed_through
+        if not pool:
+            self.log.warn(
+                "Was about to reap %r but no local license pool.",
+                identifier
             )
-            availability.apply(pool, False)
```
leonardr: This is the bug that would have broken the reaper before. The second argument to `apply` is supposed to be a policy object, not a boolean.

aslagle:
:+1:

leonardr:
I checked in on this, and the Axis 360 endpoint is still returning status 500, so i've just merged in the new version of master and am letting it sit for a while longer.

leonardr:
It's been a long time, and I don't know if this API works yet, but I will say that we need a version of this script that works as is instead of using the new API. Sometimes Axis 360 staff changes things in their database and it doesn't update the internal timestamp, so our monitor script doesn't pick it up. We absolutely need a script that goes through and asks Axis about every single book in the collection. It's the only way to detect these changes.

leonardr:
I'm going to close this pull request because even when the new system works, the current system will be better for our purposes.

PR #356, opened 2016-09-06
Send an analytics event when the circulation manager initiates a loan.

leonardr:
This branch creates a new type of analytics event which is only sent out when a loan is initiated through the circulation manager. (As distinct from the 'checkout' event, which is sent out when the circulation manager discovers from the licensing authority that a book got checked out by _someone_.)

Tracking loans initiated by the circulation manager is important for NYPL in particular, so we can get statistics on usage of the SimplyE app vis-a-vis other ways of borrowing books.

The event is sent out when the CirculationAPI tells the licensing authority to borrow a book, and the licensing authority sends back a `LoanInfo` object (rather than raising an exception like `AlreadyCheckedOut`). Assuming the licensing authority interfaces are written correctly, this should track the stuff we want to track: new loans and loan renewals.

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -299,7 +299,7 @@ class CirculationAPI(object):
             __transaction.commit()
 
             if loan and new_loan:
-                # Send out a circulation event to record the fact that
+                # Send out an analytics event to record the fact that
                 # a loan was initiated through the circulation
                 # manager.
                 Analytics.collect_event(
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -8,6 +8,7 @@ import re
 import time
 from flask.ext.babel import lazy_gettext as _
 
+from core.analytics import Analytics
 from core.model import (
     get_one,
     Identifier,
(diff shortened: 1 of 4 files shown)


api/circulation.py
```diff
@@ -268,7 +282,7 @@ def parse_fines(fines):
             # We successfuly secured a loan.  Now create it in our
             # database.
             __transaction = self._db.begin_nested()
-            loan, is_new = licensepool.loan_to(
+            loan, new_loan_record = licensepool.loan_to(
```
leonardr: Renaming to avoid confusion between whether the loan itself is new, vs. whether the loan's data model object already exists in the database.

tests/test_circulationapi.py
```diff
@@ -75,6 +76,10 @@ def test_borrow_sends_analytics_event(self):
             }
         }
         with temp_config(config) as config:
+            provider = MockAnalyticsProvider()
+            analytics = Analytics.initialize(
+                ['core.mock_analytics_provider'], config
+            )
```
leonardr: The `Analytics` object is a singleton, so right now any test that wants to use a specific analytics provider needs to call its `initialize` method (not its constructor, as I saw in some unused code I removed).

What we really need is a `with temp_analytics_provider(analytics_class)` construct that handles the setup and teardown, but this is good enough for now.

courte:
👍

PR #357, opened 2016-09-06
Updated to the latest admin interface with a separate css file.

aslagle:
This uses https://github.com/NYPL-Simplified/circulation-web/pull/71

I need to publish circulation-web before it will work.


leonardr:
:+1:

PR #359, opened 2016-09-09
Remove references to Contributor.name; submodule update.

courte:
Not a lot of references to Contributor.name, but here they are -- corrected.


aslagle:
:+1:

PR #361, opened 2016-09-13
Implemented last reading position annotations.

aslagle:
This is the minimum viable product from https://github.com/NYPL-Simplified/Simplified/wiki/Annotations

Depends on https://github.com/NYPL-Simplified/server_core/pull/388

### DIFF
--- a/api/annotations.py
+++ b/api/annotations.py
@@ -1,0 +1,115 @@
+from nose.tools import set_trace
+from pyld import jsonld
+import json
+from datetime import datetime
+
+from core.model import (
+    Annotation,
+    Identifier,
+    get_one_or_create,
+)
+
+from core.app_server import (
+    url_for,
+)
+
+from problem_details import *
+
+class AnnotationWriter(object):
+
+    CONTENT_TYPE = 'application/ld+json; profile="http://www.w3.org/ns/anno.jsonld"'
+
+    JSONLD_CONTEXT = "http://www.w3.org/ns/anno.jsonld"
+    LDP_CONTEXT = "http://www.w3.org/ns/ldp.jsonld"
+
+    @classmethod
+    def annotation_container_for(cls, patron):
+        url = url_for("annotations", _external=True)
+        annotations = [annotation for annotation in patron.annotations if annotation.active]
+
+        container = dict()
+        container["@context"] = [cls.JSONLD_CONTEXT, cls.LDP_CONTEXT]
+        container["id"] = url
+        container["type"] = ["BasicContainer", "AnnotationCollection"]
+        container["total"] = len(annotations)
+        container["first"] = cls.annotation_page_for(patron, with_context=False)
+        return container
+        
+
+    @classmethod
+    def annotation_page_for(cls, patron, with_context=True):
+        url = url_for("annotations", _external=True)
+        annotations = [cls.detail(annotation, with_context=with_context) for annotation in patron.annotations if annotation.active]
+
+        page = dict()
+        if with_context:
+            page["@context"] = cls.JSONLD_CONTEXT
+        page["id"] = url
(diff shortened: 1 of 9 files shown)


api/annotations.py
```diff
@@ -0,0 +1,115 @@
...
+            return INVALID_ANNOTATION_TARGET
+
+        annotation, is_new = get_one_or_create(
+            _db, Annotation,
+            patron=patron,
+            identifier=identifier,
+            motivation=motivation,
+        )
+
+        annotation.target = json.dumps(target)
```
aslagle: I wasn't sure how to save the target. Because of the json-ld expansion, the target I'm saving here may be formatted differently than what the client provided. And I didn't compact it again in the AnnotationWriter, because I don't know how to get PyLD to compact using a context that's a url. It's valid json-ld but more verbose than necessary. I can keep working on that if it's a problem.

api/controller.py
```diff
@@ -820,6 +826,63 @@ def detail(self, data_source, identifier_type, identifier):
...
+
+        if flask.request.method=='HEAD':
+            return Response(status=200, headers=headers)
+
+        patron = flask.request.patron
+
+        if flask.request.method == 'GET':
+            headers['Link'] = ['<http://www.w3.org/ns/ldp#BasicContainer>; rel="type"',
+                               '<http://www.w3.org/TR/annotation-protocol/>; rel="http://www.w3.org/ns/ldp#constrainedBy"']
+            headers['Content-Type'] = AnnotationWriter.CONTENT_TYPE
```
aslagle: I think this complies with the Web Annotation Protocol spec, except that it does not have ETags.

leonardr: Since you have the timestamps you could do `Last-Modified`. Also see https://github.com/NYPL-Simplified/server_core/pull/388/files#r79421726 for my ETag implementation proposal. Both of these could go in a separate branch.

tests/test_annotations.py
```diff
@@ -0,0 +1,337 @@
...
+        eq_(self.default_patron.id, annotation.patron_id)
+        eq_(self.identifier.id, annotation.identifier_id)
+        eq_(Annotation.IDLING, annotation.motivation)
+        eq_(True, annotation.active)
+
+    def test_parse_jsonld_with_context(self):
+        self.pool.loan_to(self.default_patron)
+
+        data = dict()
+        data["@context"] = [AnnotationWriter.JSONLD_CONTEXT,
```
leonardr: Does it make sense to refactor this into a helper method? You generate some very similar documents here.

leonardr:
This looks good overall, :+1:  pending the changes to core and some kind of ETag implementation.

tests/test_annotations.py
aslagle: I'll do this for everything except the expanded/compacted ones.

api/controller.py
```diff
@@ -843,7 +854,13 @@ def container(self):
                                '<http://www.w3.org/TR/annotation-protocol/>; rel="http://www.w3.org/ns/ldp#constrainedBy"']
             headers['Content-Type'] = AnnotationWriter.CONTENT_TYPE
 
-            container = AnnotationWriter.annotation_container_for(patron)
+            container, timestamp = AnnotationWriter.annotation_container_for(patron)
+            etag = 'W/""'
+            if timestamp:
+                etag = 'W/"%s"' % timestamp
+                headers['Last-Modified'] = timestamp
```
leonardr: Is this timestamp going to be in the right format? I'm not sure and it's probably worth explicitly formatting the datetime object as a string. Search https://www.ietf.org/rfc/rfc2616.txt for "rfc1123-date" for the valid formats.

aslagle: Fixed.

leonardr:
:+1:

Issue #360, opened 2016-09-12
If the search engine is down, don't crash requests that have nothing to do with search


leonardr:
Currently all requests grab a connection to the Elasticsearch cluster. This is (slightly) inefficient, but the big problem happens when the cluster goes down. Then the request fails, even if the request wasn't going to use the Elasticsearch cluster in the first place.

leonardr:
This branch fixes bug #360

leonardr:
This branch fixes bug #360

PR #362, opened 2016-09-19
Connect to the search interface lazily.

leonardr:
This branch stops the `CirculationManager` constructor from connecting to the ElasticSearch server. Instead it happens the first time a search is about to be run. This means that if ElasticSearch goes down, searches will fail but other kinds of requests will still work.


leonardr:
This branch fixes bug #360

PR #363, opened 2016-09-20
Confirm that the Circulation Manager is capable of importing Bibliotheca covers

courte:
I wanted to confirm that the CirculationManager side of the equation wasn't part of the 3M cover import the problem. After looking at the logs and clearing this test, I'm pretty sure the issue was using the core URNLookupController instead of the Metadata Wrangler's version.

✔️ ✔️ ✔️

### DIFF
--- a/tests/test_coverage.py
+++ b/tests/test_coverage.py
@@ -23,6 +23,7 @@ from core.model import (
     CoverageRecord,
     DataSource,
     Edition,
+    Hyperlink,
     Identifier,
     LicensePool,
     Work,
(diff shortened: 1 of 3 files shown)


tests/test_coverage.py
```diff
@@ -331,6 +333,44 @@ def test_items_that_need_coverage_respects_count_as_covered(self):
...
+                           if l.rel in [Hyperlink.IMAGE, Hyperlink.THUMBNAIL_IMAGE]]
+        expected = [
+            'http://book-covers.nypl.org/Content%20Cafe/ISBN/9781594632556/cover.jpg',
+            'http://book-covers.nypl.org/scaled/300/Content%20Cafe/ISBN/9781594632556/cover.jpg'
+        ]
+
+        eq_(sorted(identifier_uris), sorted(expected))
+
+        # The ISBN doesn't get any information.
+        eq_(isbn.links, [])
```
courte: Is this the behavior we want?

aslagle: I think so - we look at recursively equivalent identifiers when choosing a cover so we don't need it on every identifier.

aslagle: Possibly it could go on the ISBN instead of the other identifier though, since that's what you found out about...

leonardr: Argument for: putting this information on the ISBN would be more accurate. It would let us get rid of the reverse mapping when data comes back from the metadata wrangler.

Counterargument: if core metadata comes back from the metadata wrangler (like a correction to the book's title or contributors), associating it with the ISBN won't be useful because we only find data associated with the ISBN when it's associated with the Identifier. That means classifications, descriptions, and cover images, but not title or contributors, which are associated with the Edition.

I think leave it alone for now.

PR #368, opened 2016-09-30
Fix username lookup

leonardr:
Username lookup is broken -- I'm not sure why it works for anyone now -- and this branch fixes it.


api/millenium_patron.py
```diff
@@ -245,7 +246,6 @@ def authenticated_patron(self, db, header):
                 # Sync our internal Patron record with what the API
                 # says.
                 self.update_patron(patron, identifier)
-                patron.last_external_sync = now
```
leonardr: Setting last_external_sync here instead of in update_patron ensures that a patron gets synced the first time we hear about them (when we use get_one_or_create), and then they also get synced on their second request.

PR #369, opened 2016-09-30
Made target in annotation detail json instead of string, and fixed tests.

aslagle:
This fixes a bug in annotations that Aferdita found - the target was a string of the json instead of json. 

I also noticed that some of my tests weren't running, so that's fixed too.

### DIFF
--- a/tests/test_annotations.py
+++ b/tests/test_annotations.py
@@ -6,6 +6,7 @@ import json
 import datetime
 
 from . import DatabaseTest
+from test_controller import ControllerTest
 
 from core.model import (
     Annotation,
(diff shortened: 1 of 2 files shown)


tests/test_annotations.py
```diff
@@ -18,112 +19,121 @@
 )
 from api.problem_details import *
 
-class AnnotationWriterTest(DatabaseTest):
+class TestAnnotationWriter(ControllerTest):
```
aslagle: This is why the tests weren't running - the class name didn't start with Test.

tests/test_annotations.py
```diff
@@ -18,112 +19,121 @@
...
     def test_annotation_container_for(self):
         patron = self._patron()
 
-        container, timestamp = AnnotationWriter.annotation_container_for(patron)
-        eq_(set([AnnotationWriter.JSONLD_CONTEXT, AnnotationWriter.LDP_CONTEXT]),
-            set(container['@context']))
-        assert "annotations" in container["id"]
-        eq_(set(["BasicContainer", "AnnotationCollection"]), set(container["type"]))
-        eq_(0, container["total"])
+        with self.app.test_request_context("/"):
```
aslagle: These tests needed a request context since they use url_for.

tests/test_annotations.py
```diff
@@ -18,112 +19,121 @@
...
-        assert "annotations/%i" % annotation.id in detail["id"]
-        eq_("Annotation", detail['type'])
-        eq_(Annotation.IDLING, detail['motivation'])
+        with self.app.test_request_context("/"):
+            detail = AnnotationWriter.detail(annotation)
+
+            assert "annotations/%i" % annotation.id in detail["id"]
+            eq_("Annotation", detail['type'])
+            eq_(Annotation.IDLING, detail['motivation'])
+            eq_(json.loads(target), detail["target"])
```
aslagle: This last line tests the change I made.

PR #371, opened 2016-10-03
Cache facets script args

leonardr:
This branch adds a number of command-line arguments to the `cache_opds_lane_facets` scripts. We disabled the script when we drastically increased the number of facets that needed to be cached. But most patrons don't go around changing the facets; they stick with a couple different sets of facets. This branch makes it possible to tell the cache script to only cache the most commonly used sets of facets, whatever that might be. This gives us almost all the benefit of the old script while drastically cutting down on the time it takes to run.

Just so you can keep the overall design in your head, here's the output of running the script with `--help`.

```
usage: cache_opds_lane_facets [-h] [--language LANGUAGE]
                              [--max-depth MAX_DEPTH] [--min-depth MIN_DEPTH]
                              [--order ORDER] [--availability AVAILABILITY]
                              [--collection COLLECTION] [--pages PAGES]

optional arguments:
  -h, --help            show this help message and exit
  --language LANGUAGE   Process only lanes that include books in this
                        language.
  --max-depth MAX_DEPTH
                        Stop processing lanes once you reach this depth.
  --min-depth MIN_DEPTH
                        Start processing lanes once you reach this depth.
  --order ORDER         Generate feeds for this ordering. Possible values:
                        author, title, added. Default: author
  --availability AVAILABILITY
                        Generate feeds for this availability setting. Possible
                        values: all, now, always. Default: all
  --collection COLLECTION
                        Generate feeds for this collection within each lane.
                        Possible values: full, main, featured. Default: main
  --pages PAGES         Number of pages to cache for each collection. Default:
                        2
```

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -268,7 +268,8 @@ class UpdateStaffPicksScript(Script):
 class LaneSweeperScript(Script):
     """Do something to each lane in the application."""
 
-    def __init__(self):
+    def __init__(self, _db=None):
+        super(LaneSweeperScript, self).__init__(_db)
         os.environ['AUTOINITIALIZE'] = "False"
         from api.app import app
         del os.environ['AUTOINITIALIZE']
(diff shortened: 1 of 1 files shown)


daryanypl:
👍

PR #372, opened 2016-10-04
Annotations for a work

aslagle:
This branch adds a link to get annotations for a work, and also fixes the annotation target so it's compacted jsonld.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -202,6 +202,13 @@ def annotations():
 def annotation_detail(annotation_id):
     return app.manager.annotations.detail(annotation_id)
 
+@app.route('/annotations/<identifier_type>/<path:identifier>/', methods=['GET'])
+@allows_patron_web()
+@requires_auth
+@returns_problem_detail
+def annotations_for_work(identifier_type, identifier):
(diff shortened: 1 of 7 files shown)


api/routes.py
```diff
@@ -202,6 +202,13 @@ def annotations():
 def annotation_detail(annotation_id):
     return app.manager.annotations.detail(annotation_id)
 
+@app.route('/annotations/<identifier_type>/<path:identifier>/', methods=['GET'])
```
aslagle: I originally wanted to put this under works, but I didn't want to have to give it a data source and without one it makes the route ambiguous.

leonardr:
This makes sense to me, except I don't understand why you compact the annotation target. Is the context not necessary because it's embedded in a larger document?

aslagle:
Right, the context is already in the document at the top-level so it doesn't need to be in the target again. 

Since Aferdita doesn't have a library for parsing JSON-LD it's easier for her to use the compacted version for both get and post, and it's also just a smaller thing to send the client.

leonardr:
:+1:

PR #373, opened 2016-10-04
Store clever token so users don't have to log in so often.

aslagle:


api/clever/__init__.py
```diff
@@ -56,8 +59,18 @@ class CleverAuthenticationAPI(OAuthAuthenticator):
...
+    TOKEN_TYPE = "Clever token"
+
     log = logging.getLogger('Clever authentication API')
 
+    def _data_source(self, _db):
+        data_source, is_new = get_one_or_create(
+            _db, DataSource, name="Clever",
+            offers_licenses=False,
+            primary_identifier_type=None
+        )
```
aslagle: This might be the first data source that's not in the well-known DataSources. I don't think it should be there - nothing outside this module should be using it and we don't want to have to change core for every new authentication provider.

api/clever/__init__.py
```diff
@@ -154,6 +162,13 @@ def oauth_callback(self, _db, params):
         )
         patron._external_type = external_type
 
+        credential, is_new = get_one_or_create(
+            _db, Credential, data_source=self._data_source(_db),
+            type=self.TOKEN_TYPE, patron=patron,
+        )
+        credential.credential = token
```
aslagle: This is the token that clever gives us. We never use it again, so we could also generate our own token and store that instead.

api/clever/__init__.py
```diff
@@ -154,6 +162,13 @@ def oauth_callback(self, _db, params):
...
+        credential.expires = datetime.datetime.utcnow() + datetime.timedelta(weeks=6)
```
aslagle: I don't know how long this should be. It should probably be configurable.

leonardr: 6 weeks is a good default but it should be configurable. (Even though it's easier for us to deploy a code change than a configuration change.)

leonardr:
This makes sense to me overall.

leonardr:
:+1:

PR #374, opened 2016-10-05
Introduce SIP2 client

leonardr:
This branch introduces a very basic client for the SIP2 protocol (http://multimedia.3m.com/mws/media/355361O/sip2-protocol.pdf). This is a standalone client which will eventually be in its own Python package on PyPI.

This client needs to do two things:

1. Make a Patron Information request to a SIP2 server and process the response. This response contains all sorts of information about the patron's name, email address, whether their password is valid, etc.
2. Make a Login request to a SIP2 server and process the response. Some SIP2 servers require a client to send a Login request as its first request upon opening a socket connection.

The Patron Information response contains a lot of different types of data, and not all of it is needed by the circulation manager. Since this is going to become a standalone library I've focused on parsing that data in a reliable way and putting it into a reasonable, generic data structure: a dictionary where the keys are names like 'patron_identifier', not two-letter SIP keys like 'AA'.

### DIFF
--- a/api/sip.py
+++ b/api/sip.py
@@ -1,0 +1,422 @@
+#!/usr/bin/env python
+
+"""A simple SIP2 client.
+
+This client implements a very small part of SIP2 but is easily extensible.
+
+This client is based on sip2talk.py. Here is the original licensing
+information for sip2talk.py:
+
+Copyright [2010] [Eli Fulkerson]
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+"""
+
+import datetime
+import logging
+from nose.tools import set_trace
+import socket
+import sys
+import time
+
+class fixed(object):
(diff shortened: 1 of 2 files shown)


api/sip.py
```diff
@@ -0,0 +1,596 @@
...
+        return parsed
+    
+    def consume_status_code(self, data, expected, in_progress):
+        """Pull the status code (the first two characters) off the 
+        given response string, and verify that it's as expected.
+        """
+        status_code = data[:2]
+        in_progress['_status'] = status_code
+        if status_code != expected:
+            raise ValueError(
```
daryanypl: Should it always be a value error?  Would we want to have different handling for different http statuses, or for when the SC got logged out, or when ACS asks for a message resend or responds to a previous message, rather than the current one?

api/sip.py
```diff
@@ -0,0 +1,596 @@
...
+            fixed.transaction_date,
+            fixed.hold_items_count,
+            fixed.overdue_items_count,
+            fixed.charged_items_count,
+            fixed.fine_items_count,
+            fixed.recall_items_count,
+            fixed.unavailable_holds_count,
+            named.institution_id.required,
+            named.patron_identifier.required,
+            named.personal_name.required,
```
daryanypl: I'm missing the bit of code where some of the named fields are made required?

leonardr: Maybe I'm being too clever here. `named.personal_name.required` calls the `named.required` property, which returns a `named` object that is "personal name, except required". So this _is_ the code where the named fields are made required.

api/sip.py
daryanypl: fair enough

daryanypl:
👍

PR #381, opened 2016-10-19
Authenticator refactor

leonardr:
This branch is a major refactoring of the authentication code to make different authentication mechanisms work on the same foundation. Each authentication mechanism is responsible for communicating with a source of truth and returning as much information as possible as a `PatronData` object. `PatronData` is a new class that can be automatically turned into a `Patron` object. It can also contain information that cannot be stored as a Patron object (such as the patron's email address), but which can be passed on to the client.

The new authentication system is designed to take whatever information the source of truth can provide, and maintain the corresponding set of Patron objects in the database. If the source of truth doesn't provide certain pieces of information (such as an unchanging permanent ID for the patron), patrons might discover that they suddenly get new accounts when their barcodes change. But we do the best we can. Most of this code is taken from the old Millenium Patron authentication system--I've just generalized it so that any new authentication system only has to focus on doing the lookup and stuffing all relevant information into a PatronData object.

Here are the major changes:

* All basic authentication mechanisms (FirstBook and Millenium Patron) derive from a common superclass and work the same way.
* Clever, the only OAuth authentication mechanism, is now a subclass of a generic OAuth authentication mechanism.
* Split out the Authenticator class (which handles authentication for the site as a whole) from the AuthenticationProvider classes (each of which communicates with a specific source of truth)
* I split out the OAuthAuthenticationProvider class (which communicates with a specific OAuth provider) from the OAuthController class (which handles the circ manager's implementation of the OAuth callbacks).
* I defined a new MockAuthenticationProvider, for use in tests (and when setting up a circ manager before connecting it to an ILS), and started using it in the unit tests, instead of using a hacked MilleniumPatronAuthenticationProvider.
* Took out the "account" link which goes to information about your user account. Clients weren't using it and it was easier to remove it than port it to the new system. (We can add it back later.)

Unrelated change: I added a previously missing file, `test_scripts.py`, which tests some of my recent new scripts. I had to change the script framework a bit to get the tests to pass.


api/circulation.py
```diff
@@ -202,7 +202,7 @@ def borrow(self, patron, pin, licensepool, delivery_mechanism,
         if patron.fines:
             def parse_fines(fines):
                 dollars, cents = re.match("\$([\d]+)\.(\d\d)", fines).groups()
-                return (dollars * 100) + cents
+                return (int(dollars) * 100) + int(cents)
```
leonardr: This was working most of the time when two amounts of fines were compared, but it worked by multiplying a string like "20" 100 times, "2020202020202020202020...", not by actually comparing two amounts of money as numeric values.

api/clever/__init__.py
```diff
@@ -44,11 +49,17 @@
...
+class CleverAuthenticationAPI(OAuthAuthenticationProvider):
 
-    NAME = 'Clever'
     URI = "http://librarysimplified.org/terms/auth/clever"
-    METHOD = "http://librarysimplified.org/authtype/Clever"
+
+    # TODO: I think this can be removed because
+    # "http://librarysimplified.org/authtype/OAuth-with-intermediary"
+    # should be good enough.
+    # METHOD = "http://librarysimplified.org/authtype/Clever"
```
leonardr: Let me know if you agree.

api/authenticator.py
```diff
@@ -232,69 +579,704 @@ def create_authentication_headers(self):
...
+        """
         method_doc = dict(links=dict(authenticate=self._internal_authenticate_url()))
         methods = {}
         methods[self.METHOD] = method_doc
         return dict(name=self.NAME, methods=methods)
+
+    def token_data_source(self, _db):
+        # TODO: This is one of those situations where we'd like to
+        # create the data source if it doesn't already exist.
+        return DataSource.lookup(_db, self.TOKEN_DATA_SOURCE_NAME)
```
leonardr: I need to address this TODO before landing the branch.

api/clever/__init__.py
leonardr: I've taken this out but we might need to add it back in if our deployed Open Ebooks client is looking for it.

api/authenticator.py
```diff
@@ -232,69 +579,704 @@ def create_authentication_headers(self):
...
+
+        :return: A Patron if one can be authenticated; a ProblemDetail
+        if an error occurs; None if the credentials are missing or
+        wrong.
+        """
+        patron = self.authenticate(_db, header)
+        if not isinstance(patron, Patron):
+            return patron
+        if patron.needs_external_sync:
+            self.update_patron_metadata(patron)
```
leonardr: That's true. I'm having trouble thinking of a good solution. The best I can come up with is to set the sync timeout for patrons without borrowing privileges to, like, five seconds, instead of zero.

api/authenticator.py
```diff
@@ -232,69 +579,704 @@ def create_authentication_headers(self):
...
+        self.authenticator = authenticator
+
+    def oauth_authentication_redirect(self, params):
+        """Redirect an unauthenticated patron to the authentication URL of the
+        appropriate OAuth provider.
+
+        Over on that other site, the patron will authenticate and be
+        redirected back to the circulation manager (the URL is stored
+        in params['redirect_uri']), ending up in
+        oauth_authentication_callback.
```
leonardr: I've changed the comment, and refactored the external_authenticate_url code in a way that I hope addresses this issue.

api/authenticator.py
```diff
@@ -232,69 +579,704 @@ def create_authentication_headers(self):
...
+
+        Then the redirect URI might be:
+
+            http://oauthprovider.org/success#access_token=1234&patron_info=%7B%22name%22%3A+%22Mary+Shell%22%7D
+
+        It's the client's responsibility to extract the access_token,
+        start using it as a bearer token, and make sense of the
+        patron_info.
+        """
+        code = params.get('code')
```
leonardr: Hmm, yeah, let's take another pass at this when we do a second OAuth integration.

api/authenticator.py
```diff
@@ -26,162 +36,505 @@
...
+        authentication_policy = Configuration.policy("authentication")
+        if not authentication_policy:
             raise CannotLoadConfiguration(
                 "No authentication policy given."
             )
+
+        if isinstance(authentication_policy, basestring):
+            authentication_policy = dict(providers=[authentication_policy])
+        bearer_token_signing_secret = authentication_policy.get(
+            'bearer_token_signing_secret'
```
aslagle: Before this, the same secret was used for signing bearer tokens and signing flask session cookies for the admin interface. Both of these secrets are only known to the circ manager and they're used for similar things, although one is handled through flask. There's no reason they need to be the same, I just wanted to point it out. We may want to move the other secret as well.

tests/test_authenticator.py
```diff
@@ -89,291 +336,1024 @@ def test_initialize(self):
...
+        self.basic = MockBasic()           
+        self.oauth1 = MockOAuthWithExternalAuthenticateURL(
+            self._db, "http://oauth1.com/", patron
+        )
+        self.oauth1.NAME = "Mock OAuth 1"
+        self.oauth2 = MockOAuthWithExternalAuthenticateURL(
+            self._db, "http://oauth2.org/", patron
+        )
+        self.oauth2.NAME = "Mock OAuth 2"
+        # Check that the correct auth provider is called.
```
aslagle: I think this comment was left over from the old code.

tests/test_authenticator.py
```diff
@@ -89,291 +336,1024 @@ def test_initialize(self):
...
+        params = dict(code="foo", state=json.dumps(dict(provider=self.oauth1.NAME)))
+        response = self.controller.oauth_authentication_callback(self._db, params)
+        eq_(302, response.status_code)
+        fragments = urlparse.parse_qs(urlparse.urlparse(response.location).fragment)
+        token = fragments.get("access_token")[0]
+        provider_name, provider_token = self.auth.decode_bearer_token(token)
+        eq_(self.oauth1.NAME, provider_name)
+        eq_(self.oauth1.token.credential, provider_token)
+        
+        # Successful callback through OAuth provider 1.
```
aslagle: This is OAuth provider 2.

tests/test_controller.py
```diff
@@ -114,9 +126,15 @@ def setup(self, _db=None):
         # were created in the test setup.
         app.config['PRESERVE_CONTEXT_ON_EXCEPTION'] = False
 
+        # Create the patron used by the dummy authentication mechanism.
+        self.default_patron, ignore = get_one_or_create(
+            _db, Patron, authorization_identifier="unittestuser",
+            create_method_kwargs=dict(external_identifier="unittestpassword")
```
aslagle: strange choice of external_identifier but ok

api/authenticator.py
leonardr: I changed core to go five seconds between syncs for patrons without borrowing privileges.

api/authenticator.py
leonardr: Since they started out as the same, should we keep them the same so that existing bearer tokens don't become invalid?

aslagle: The admin interface isn't on production yet, so we could move the existing secret key into the bearer_token_signing_secret and create a new one for the admin interface when we deploy it, or put the same value in both variables.

leonardr: Creating a new secret for the admin interface sounds like the right solution.

PR #383, opened 2016-10-21
Added overview stats for the admin dashboard.

aslagle:


api/admin/controller.py
```diff
@@ -73,6 +76,7 @@ def setup_admin_controllers(manager):
     manager.admin_work_controller = WorkController(manager)
     manager.admin_sign_in_controller = SignInController(manager)
     manager.admin_feed_controller = FeedController(manager)
+    manager.admin_dashboard_controller = DashboardController(manager)
```
aslagle: The circulation event methods were in the FeedController even though they don't return feeds. I made a new controller for those and the new stats endpoint.

api/admin/controller.py
```diff
@@ -610,6 +614,135 @@ def genres(self):
...
+
+        loan_count = self._db.query(
+            Loan
+        ).filter(
+            Loan.end >= datetime.now()
+        ).count()
+
+        hold_count = self._db.query(Hold).count()
+
+        overdrive_count = self._db.query(
```
leonardr: It would be more compact and more extensible to iterate over a list of (DataSource, key name) pairs and build the dictionary that way. You're doing the same query for each data source and I think you want to omit a value if there are no active loans, since the reason is most likely the library has no books from that data source.

tests/admin/test_controller.py
```diff
@@ -786,6 +788,119 @@ def test_bulk_circulation_events(self):
...
+            response, requested_date = self.manager.admin_dashboard_controller.bulk_circulation_events()
         rows = response[1::] # skip header row
         eq_(0, len(rows))
+
+    def test_stats_patrons(self):
+        with self.app.test_request_context("/"):
+
+            # At first, there's one patron in the database.
+            # TODO: when the authentication refactoring is done,
+            # we'll start with 0.
```
leonardr: Authentication refactoring should be done now...

leonardr:
This is looking good, I just have a couple minor comments.

tests/admin/test_controller.py
aslagle: You're right, I guess this didn't need to change after all.

leonardr:
:smile_cat:

PR #386, opened 2016-10-25
added checkout to oneclick

daryanypl:


api/oneclick.py
```diff
@@ -0,0 +1,214 @@
...
+        """
+        :param patron_id OneClick internal id
+        :param item_id isbn
+        :return information on the transaction or error status and message
+        """
+        url = "%s/libraries/%s/patrons/%s/checkouts/%s" % (self.base_url, str(self.library_id), patron_id, item_id)
+
+        method = "post"
+        if return_item:
+            method = "delete"
```
leonardr: This makes me think the method shouldn't be called `checkout_item`--that `checkout_item` and `return_item` should be wrappers around a method with a more generic name.

tests/test_oneclick.py
```diff
@@ -0,0 +1,175 @@
...
+        self.api.queue_response(status_code=200, content=datastr)
+        oneclick_patron_id = self.api.get_patron_internal_id(patron_cardno='9305722621')
+        eq_(None, oneclick_patron_id)
+
+        datastr, datadict = self.get_data("response_patron_internal_id_error.json")
+        self.api.queue_response(status_code=500, content=datastr)
+        assert_raises_regexp(
+            ValueError, "Got status code 500 from external server, cannot continue.", 
+            self.api.get_patron_internal_id, patron_cardno='130572262x'
+        )
```
leonardr: As an example of the `ValueError` thing I mentioned before, I would expect this to raise `RemoteInitiatedServerError`.

leonardr:
The part of this that communicates with One Click is solid but the part of it that feeds into our overall circulation code needs work. There's a standard circulation API that this class doesn't implement.

Unfortunately you didn't know this because I didn't define the standard in the appropriate place by having `BaseCirculationAPI` define abstract methods. The only way to know what methods you need is to look at one of the other implementations, like `OverdriveAPI`.

The methods you need to implement are: `checkout`, `checkin`, `fulfill`, `place_hold`, `release_hold`, and `patron_activity`. First, it would be great if you could define stubs for these methods in `BaseCirculationAPI` that raise `NotImplementedException`. (You can get the signatures and docstrings from `OverdriveAPI`). 

But my main point is these are the methods you need to implement. They need to return `LoanInfo` and `HoldInfo` objects, not dictionaries. When they raise an exception because of a problem with the API, it should be one of the exceptions in `circulation_exceptions.py`.

For testing purposes, you can also have methods that parse the output of the API and return dictionaries (for Overdrive I did this with `OverdriveRepresentationExtractor`, but these core methods need to be present in a subclass of `BaseCirculationAPI`.

Again, this is my fault for not making it clear in `BaseCirculationAPI` how a subclass actually gets used. You don't even need to change this in this branch, but to get One Click working alongside Overdrive in `CirculationAPI` (which is the next step) you will need to have it implement these methods.

api/oneclick.py
```diff
@@ -0,0 +1,214 @@
...
+            # consider raising an exception, since we should call methods with nice arguments
+            return None
+
+        url = "%s/rpc/libraries/%s/patrons/%s" % (self.base_url, str(self.library_id), patron_identifier)
+
+        try:
+            response = self.request(url)
+        except Exception, e:
+            self.log.error("Patron id call failed: %r", e, exc_info=e)
+            raise ValueError(e.message)
```
daryanypl: Yeah, OneClick handles a lot of responses through http errors.  Normally, an http request getting an exception means something dire has happened.  But with OneClick it could just mean a malformed identifier or a duplicate checkout request, etc..  Things that I'd put into the "oops" but not "stop the presses" category.  Don't know if ValueError is the precise right thing, but returning a None would make flow unintuitive, and leaving the original exception would cry "wolf" a bit much.

api/oneclick.py
daryanypl: ok

api/oneclick.py
daryanypl: It seems an appropriately less drastic level of "bad result".

leonardr: Pretty much anything OneClick might send you should be covered by the exceptions in api/circulation_exceptions.py. Malformed identifier == NotFoundOnRemote. Duplicate checkout request == AlreadyCheckedOut. This is the other part of the CirculationAPI contract -- if something goes wrong you're supposed to handle it in a generic way.

tests/test_oneclick.py
leonardr: It's for the third-party server breaking. It means we can't fulfill the patron's request (or do whatever we were trying to do), but there's absolutely nothing we can do on our end to fix the problem except bother the third party about it.

api/circulation.py
```diff
@@ -717,3 +717,64 @@ def default_notification_email_address(self, patron, pin):
...
+
+
+    def checkin(self, patron, pin, licensepool):
+        """  Return a book early.  
+
+        :param patron: a Patron object for the patron who wants
+        to check out the book.
+        :param pin: The patron's alleged password.
+        :param licensepool: Identifier of the book to be checked out is 
+        attached to this licensepool.
```
leonardr: It's not just the identifier -- the _LicensePool_ is the book that is going to be checked out. Right now there is only one LicensePool per Identifier, but that's going to change pretty soon.

api/circulation.py
```diff
@@ -717,3 +717,64 @@ def default_notification_email_address(self, patron, pin):
...
+        :param internal_format: Represents the patron's desired book format.
+
+        :return: a LoanInfo object.
+        """
+        raise NotImplementedException
+
+
+    def fulfill(self, patron, pin, licensepool, internal_format):
+        """ Get the actual resource file to the patron.
+        TODO:  to the patron?  or to some intermediary stage?
```
leonardr: This method's job is to create a FulfillmentInfo object; it will be delievered immediately to the patron, but it's not the repsonsibility of this method to care about that.

api/circulation.py
```diff
@@ -717,3 +717,64 @@ def default_notification_email_address(self, patron, pin):
...
+        """
+        raise NotImplementedException
+
+
+    def patron_activity(self, patron, pin):
+        """ Return a patron's current checkouts and holds.
+        TODO: current?  or also past history?
```
leonardr: Current only.

api/circulation.py
```diff
@@ -717,3 +717,64 @@ def default_notification_email_address(self, patron, pin):
...
+    def checkin(self, patron, pin, licensepool):
+        """  Return a book early.  
+
+        :param patron: a Patron object for the patron who wants
+        to check out the book.
+        :param pin: The patron's alleged password.
+        :param licensepool: Identifier of the book to be checked out is 
+        attached to this licensepool.
+        """
+        raise NotImplementedException
```
leonardr: All these need to raise `NotImplementedException()`, not the exception class.

leonardr:
:confused: Somehow this turned into a pull request to merge commits from `master` into `oneclick_talker` instead of the other way around. There are too many changes in revision b9d2489 for me to pull out which changes you made. Can you bring this branch up to date with `master`?

leonardr:
I checked out this branch and merged in master, which got rid of most of the extraneous changes. If you can do the same and then push the update I'll be able to add my comments through the Github interface.

api/circulation.py
```diff
@@ -95,11 +94,28 @@ def __repr__(self):
...
+    def __init__(self, oneclick_id=None, username=None, card_number=None, card_pin=None, 
+        email=None, first_name=None, last_name=None):
+        self.oneclick_id = oneclick_id
+        self.username = username
+        self.card_number = card_number
+        self.card_pin = card_pin
+        self.email = email
+        self.first_name = first_name
+        self.last_name = last_name
+
```
leonardr: Do you plan to use this information? Generally we get this stuff from the ILS. I see that this is created by `get_patron_information` but I'm not sure when we would need that information. If One Click has its own ID for patrons I could see storing that as a Credential associated with the Patron, but I feel like you cover that case with `get_patron_internal_id`.

If we need to use this information I'd suggest subclassing the new PatronData class (in authenticator.py) and storing it in there. But unless you have a use case in mind I'd say leave it alone until we determine whether we actually need it.

api/circulation.py
```diff
@@ -200,9 +216,14 @@ def borrow(self, patron, pin, licensepool, delivery_mechanism,
         internal_format = api.internal_format(delivery_mechanism)
 
         if patron.fines:
-            max_fines = Configuration.max_outstanding_fines()
-            if patron.fines >= max_fines.amount:
-                raise OutstandingFines()
+            def parse_fines(fines):
```
leonardr: I'm pretty sure this is a reversion of code I committed earlier. Calling `Configuration.max_outstanding_fines()` is the correct behavior.

api/circulation.py
```diff
@@ -304,7 +325,7 @@ def borrow(self, patron, pin, licensepool, delivery_mechanism,
                 # manager.
                 Analytics.collect_event(
                     self._db, licensepool,
-                    CirculationEvent.CM_CHECKOUT,
+                    self.CIRCULATION_MANAGER_INITIATED_LOAN_EVENT_TYPE
```
leonardr: Another reversion.

api/circulation.py
```diff
@@ -334,15 +355,6 @@ def borrow(self, patron, pin, licensepool, delivery_mechanism,
...
         )
-        if hold and is_new:
-            # Send out an analytics event to record the fact that
-            # a hold was initiated through the circulation
-            # manager.
-            Analytics.collect_event(
-                self._db, licensepool,
-                CirculationEvent.CM_HOLD_PLACE,
-            )
-
```
leonardr: Another reversion.

api/circulation.py
```diff
@@ -395,15 +407,6 @@ def fulfill(self, patron, pin, licensepool, delivery_mechanism, sync_on_failure=
...
                 raise NoAcceptableFormat()
-
-        # Send out an analytics event to record the fact that
-        # a fulfillment was initiated through the circulation
-        # manager.
-        Analytics.collect_event(
-            self._db, licensepool,
-            CirculationEvent.CM_FULFILL,
-        )
-
```
leonardr: Another.

api/circulation.py
```diff
@@ -457,14 +460,6 @@ def revoke_loan(self, patron, pin, licensepool):
...
             self._db.delete(loan)
             __transaction.commit()
-            # Send out an analytics event to record the fact that
-            # a loan was revoked through the circulation
-            # manager.
-            Analytics.collect_event(
-                self._db, licensepool,
-                CirculationEvent.CM_CHECKIN,
-            )
-
```
leonardr: Another.

api/oneclick.py
```diff
@@ -0,0 +1,453 @@
...
+        resp_dict = self.circulate_item(patron_id=patron_oneclick_id, item_id=item_oneclick_id, return_item=False)
+
+        if not resp_dict or ('error_code' in resp_dict):
+            return None
+
+        self.log.debug("Patron %s/%s checked out item %s with transaction id %s.", patron.authorization_identifier, 
+            patron_oneclick_id, item_oneclick_id, resp_dict['transactionId'])
+
+        today = datetime.datetime.now()
+        expires = today + self.EXPIRATION_DEFAULT
```
leonardr: How certain are we that loans are always 21 days long? Is this a policy across One Click?

api/oneclick.py
```diff
@@ -0,0 +1,453 @@
...
+            self.log.error("Item checkout/return failed: %r", e, exc_info=e)
+            raise RemoteInitiatedServerError(e.message)
+
+        resp_dict = {}
+        message = None
+        if response.text:
+            resp_dict = response.json()
+            message = resp_dict.get('message', None)
+
+        error_response = self.validate_response(response=response, message=message, action=action)
```
leonardr: I was confused by the fact that you assign the return value and don't use it. It's not a big deal but I would say just call `validate_response`--that makes it more clear what is going on.

api/oneclick.py
```diff
@@ -0,0 +1,453 @@
...
+
+        elif message:
+            # http code was OK, but info wasn't sucessfully read from db
+            self.log.warning("%s not retrieved: %s ", action, message)
+            raise CirculationException(action + ": " + message)
+
+        return None
+
+
+    def queue_response(self, status_code, headers={}, content=None):
```
leonardr: This looks like a mock method.

api/oneclick.py
```diff
@@ -0,0 +1,453 @@
...
+
+    # ------
+    def run_once(self, start, cutoff):
+        _db = self._db
+        added_books = 0
+        overdrive_data_source = DataSource.lookup(
+            _db, DataSource.OVERDRIVE)
+
+        total_books = 0
+        consecutive_unchanged_books = 0
```
leonardr: If the delta API is reliable, you don't need this whole `consecutive_unchanged_books` thing. That's taken from Overdrive, which does not have a reliable delta API.

api/oneclick.py
```diff
@@ -0,0 +1,453 @@
...
+                    self.log.info("Stopping at %d unchanged books.",
+                                  consecutive_unchanged_books)
+                    break
+
+        if total_books:
+            self.log.info("Processed %d books total.", total_books)
+
+
+
+class OneClickCollectionMonitor(OneClickCirculationMonitor):
```
leonardr: This is another thing that's not necessary if the delta API is reliable.

leonardr:
This is coming together overall! My two main concerns are the code copied from Overdrive for dealing with an unreliable delta API (hopefully the OneClick delta API is not so unreliable that we need that code) and the lack of clarity as to how a library's patrons get into the OneClick system in the first place. In conversation, you said there's an OneClick method for creating a patron on the remote side, so I'll wait for that.

api/oneclick.py
daryanypl: I think it's their ebook default.  moved it to config.

api/oneclick.py
daryanypl: cleaned

api/oneclick.py
daryanypl: Oh, it's quite serious.

api/circulation.py
```diff
@@ -77,7 +94,16 @@ def __repr__(self):
...
 class HoldInfo(CirculationInfo):
 
-    """A record of a hold."""
+    """A record of a hold.
+
+    :param identifier_type Ex.: Identifier.ONECLICK_ID.
+    :param identifier Expected to be the unicode string of the isbn, etc..
+    :param start_date When the patron made the reservation.
+    :param end_date When checked-out book is due.  Expected to be passed in 
+        date, not unicode format.
```
leonardr: The end_date of a hold is the date the book is estimated to be available to this patron.

api/circulation_exceptions.py
```diff
@@ -46,6 +46,12 @@ class AuthorizationFailedException(CirculationException):
 class PatronAuthorizationFailedException(AuthorizationFailedException):
     status_code = 400
 
+class PatronCreationFailedException(CirculationException):
+    status_code = 500
+
```
leonardr: I would call this something like `RemotePatronCreationFailedException` so it doesn't look like the problem is we couldn't create the patron's record _in the ILS_.

api/circulation_exceptions.py
```diff
@@ -46,6 +46,12 @@ class AuthorizationFailedException(CirculationException):
...
+class PatronNotFoundException(CirculationException):
+    status_code = 200
+
```
leonardr: Similarly I would call this PatronNotFoundOnRemote (see NotFoundOnRemote). Should its status_code be 404?

api/oneclick.py
```diff
@@ -155,16 +170,157 @@ def circulate_item(self, patron_id, item_id, return_item=False):
...
+
+
+    def place_hold(self, patron, pin, licensepool, notification_email_address):
+        """Place a book on hold.
+
+        :param patron: a Patron object for the patron who wants to check out the book.
+        :param pin: The patron's password (not used).
+        :param licensepool: The Identifier of the book to be checked out is 
+        attached to this licensepool.
+        :param internal_format: Represents the patron's desired book format.  Ignored for now.
```
leonardr: This isn't relevant to this method, and if you copied it from some other place_hold implementation, the other docstring is probably also wrong.

api/oneclick.py
```diff
@@ -155,16 +170,157 @@ def circulate_item(self, patron_id, item_id, return_item=False):
...
+
+        :return: A HoldInfo object on success, None on failure
+        """
+        patron_oneclick_id = self.validate_patron(patron)
+        (item_oneclick_id, item_media) = self.validate_item(licensepool)
+
+        resp_dict = self.circulate_item(patron_id=patron_oneclick_id, item_id=item_oneclick_id, hold=True, return_item=False)
+
+        if not resp_dict or ('error_code' in resp_dict):
+            return None
```
leonardr: Is this better than raising an exception?

api/oneclick.py
```diff
@@ -155,16 +170,157 @@ def circulate_item(self, patron_id, item_id, return_item=False):
...
+    ''' -------------------------- Patron Account Handling -------------------------- '''
+    def create_patron(self, patron):
+        """ Ask OneClick to create a new patron record.
+
+        :param patron: a Patron object which contains the permanent id to give to 
+        OneClick as the patron's library card number.
+        :return OneClick's internal patron id.
+        """
+
+        patron_cardno = patron.authorization_identifier
```
leonardr: This is not going to work in all libraries, but I don't have a solution in place yet, so it's fine for now.

api/oneclick.py
```diff
@@ -155,16 +170,157 @@ def circulate_item(self, patron_id, item_id, return_item=False):
...
+
+        url = "%s/libraries/%s/patrons/" % (self.base_url, str(self.library_id))
+        action="create_patron"
+        
+        args = dict()
+        args['libraryId'] = self.library_id
+        args['libraryCardNumber'] = patron.authorization_identifier
+        # generate indicative values for the account fields the patron has not supplied us with
+        args['userName'] = 'username' + patron.authorization_identifier
+        args['email'] = 'patron' + patron.authorization_identifier + '@librarysimplified.org'
```
leonardr: Add a random value here to avoid the possibility of a collision from patrons of different libraries who have the same barcode.

leonardr: Actually, just generate UUIDs and use them for username and email. There's no reason to give One Click any patron information at all here.

api/oneclick.py
```diff
@@ -185,12 +341,167 @@ def get_patron_internal_id(self, patron_email=None, patron_cardno=None):
...
+                expires = datetime.datetime.strptime(expires, self.EXPIRATION_DATE_FORMAT).date()
+
+            identifier, made_new = Identifier.for_foreign_id(self._db, 
+                    foreign_identifier_type=Identifier.ONECLICK_ID, 
+                    foreign_id=isbn, autocreate=False)
+
+            # Note: if OneClick knows about a patron's checked-out item that wasn't
+            # checked out through us, we ignore it
+            if not identifier:
+                continue
```
leonardr: This isn't consistent with how we treat books checked out via the Overdrive app--you can see those books but not necessarily fulfill them. I guess it doesn't matter because we're creating a totally new One Click account for the patron which is used solely by SimplyE, so it won't come up.

api/oneclick.py
```diff
@@ -185,12 +341,167 @@ def get_patron_internal_id(self, patron_email=None, patron_cardno=None):
...
+
+                # Note: download urls expire 15 minutes after being handed out
+                # in the checkouts call
+                download_url = file.get('downloadUrl', None)
+                # is included in the downloadUrl, actually
+                acs_resource_id = file.get('acsResourceId', None)
+
+            # TODO: For audio books, the downloads are done by parts, and there are 
+            # multiple download urls.  Need to have a mechanism for putting lists of 
+            # parts into fulfillment objects.
```
leonardr: :( We'll have to deal with this in general when we add audio support.

api/oneclick.py
leonardr: Also, this isn't going to catch all items checked out the OneClick, only ones that are not currently in the collection.

api/oneclick.py
```diff
@@ -211,105 +522,68 @@ def get_patron_information(self, patron_id):
...
+        """ Are we performing operations on a book that exists and can be 
+        uniquely identified? 
+        """
+        item_oneclick_id = None
+        media = None
 
+        identifier = licensepool.identifier
+        item_oneclick_id=identifier.identifier
+        if not item_oneclick_id:
+            raise InvalidInputException("Licensepool %r doesn't know its ISBN.", licensepool)
```
leonardr: This seems like a much lower-level test than anything you should need to worry about. Why would you end up with a Licensepool whose Identifier had a blank .identifier? And wouldn't its identifier be an internal OneClick ID, not an ISBN?

config.json.sample
```diff
@@ -18,6 +18,18 @@
...
+            "username" : "app-agent-username",
+            "password" : "app-agent-pwd",
+            "library_id" : "participating-library-id",
+            "remote_stage" : "qa", 
+            "url" : "https://api.oneclickdigital.us/", 
+            "basic_token" : "abc-def-ghi", 
+            "ebook_loan_length" : "integer-number-days", 
+            "eaudio_loan_length" : "integer-number-days"
+        },
+
```
leonardr: Rather than adding this to the sample config, you should use `temp_config()` to create a new Configuration object during any test that needs to have OneClick configured. 

See tests/test_novelist.py for a relevant example.

leonardr:
We're pretty close now I'd say.

config.json.sample
daryanypl: added temp_config to MockOneClickAPI

api/circulation.py
```diff
@@ -32,7 +32,16 @@ def fd(self, d):
...
 class FulfillmentInfo(CirculationInfo):
 
-    """A record of an attempt to fulfil a loan."""
+    """A record of an attempt to fulfill a loan.
+
+    :param identifier_type Ex: Third party provider, ISBN, URI, etc...
+    :param identifier Contains ISBN or third party item id, etc., and links to LicensePool, Work, etc..
+    :param content_link Either URL to download ACSM file from or URL to streaming content.
+    :param content_type Media type of the book version we're getting.  
+        Ex: "text/html" or Representation.TEXT_HTML_MEDIA_TYPE + DeliveryMechanism.STREAMING_PROFILE .
```
leonardr: This could be any media type. If you're giving examples, another example would be EPUB_MEDIA_TYPE.

api/circulation.py
```diff
@@ -78,7 +95,16 @@ def __repr__(self):
...
-    """A record of a hold."""
+    """A record of a hold.
+
+    :param identifier_type Ex.: Identifier.ONECLICK_ID.
+    :param identifier Expected to be the unicode string of the isbn, etc..
+    :param start_date When the patron made the reservation.
+    :param end_date When reserved book is expected to become available.  Expected to be passed in 
+        date, not unicode format.
+    :param hold_position  Patron's place in the hold line.  
+        When not available, default to be passed is 0.
```
leonardr: When hold_position is not available, the value passed in should be None, not 0. hold_position zero means that the patron is first in line to borrow the book and can check it out immediately.

api/oneclick.py
```diff
@@ -0,0 +1,707 @@
...
+
+        today = datetime.datetime.now()
+
+        hold = HoldInfo(
+            identifier_type=licensepool.identifier.type,
+            identifier=item_oneclick_id,
+            start_date=today,
+            # OneClick sets hold expirations to 2050-12-31, as a "forever"
+            end_date=None,
+            hold_position=0,
```
leonardr: This should be None as per above.

api/oneclick.py
```diff
@@ -0,0 +1,707 @@
...
+            # checked out through us, we ignore it
+            if not identifier:
+                continue
+
+            hold = HoldInfo(
+                Identifier.ONECLICK_ID,
+                isbn,
+                start_date=None,
+                end_date=expires,
+                hold_position=0
```
leonardr: Again, should be None.

api/oneclick.py
```diff
@@ -0,0 +1,707 @@
...
+
+class OneClickCirculationMonitor(Monitor):
+    """Maintain LicensePools for OneClick titles.
+
+    Bibliographic data isn't inserted into new LicensePools until
+    we hear from the metadata wrangler.
+    """
+    def __init__(self, _db, name="OneClick Circulation Monitor",
+                 interval_seconds=500,
+                 maximum_consecutive_unchanged_books=None):
```
leonardr: I don't think you do anything with maximum_consecutive_unchanged_books; I would take it out.

api/oneclick.py
```diff
@@ -0,0 +1,707 @@
...
+    def recently_changed_ids(self, start, cutoff):
+        return self.api.get_delta(start, cutoff)
+
+
+    def run(self):
+        self.api = OneClickAPI(self._db)
+        super(OneClickCirculationMonitor, self).run()
+
+    def run_once(self, start, cutoff):
+        pass
```
leonardr: I'm confused--what does this monitor do? Does get_delta do all the work?

api/oneclick.py
```diff
@@ -0,0 +1,707 @@
...
+
+
+
+class OneClickCollectionMonitor(OneClickCirculationMonitor):
```
leonardr: You only need this class if you are defining multiple strategies for keeping the One Click collection up to date. At any rate I don't see a different strategy implemented here.

leonardr:
Looks good with a couple minor changes and an update to the submodule.

api/circulation.py
daryanypl: added

api/circulation.py
daryanypl: added

api/oneclick.py
daryanypl: fixed

api/oneclick.py
daryanypl: yup

api/oneclick.py
daryanypl: out in oneclick_monitor branch

api/oneclick.py
daryanypl: monitor is started in the oneclick_monitor brach, pull request https://github.com/NYPL-Simplified/circulation/pull/421/files

leonardr:
:+1:

PR #387, opened 2016-10-26
Move authentication configuration

leonardr:
Currently you use the `authentication` section of the configuration to explain which modules to load. 

```
"authentication" : ["api.provider1", "api.provider2"]
```

Then you configure each module separately in the `integrations` section.

```
integrations : { 
"API Provider 1": { ... },
"API Provider 2": { ... },
}
```

This has a couple pitfalls. In particular, it's impossible to define two different authentication providers that happen to use the same module. You can put "api.module1" twice in `authentications`, but both providers will be initialized using the same settings.

This branch changes it so that an authentication provider is configured inside `authentication`:

```
"authentication" : {
 "providers": [
  { "module": "api.provider1", 
    "config_value": "1"
  },
  { "module": "api.provider2",
    "config_value": "2"
  }
 ]
}
```

This will probably not be necessary for a while, but since my big authentication refactor branch requires a change to the authentication configuration, I figured I would get this smaller branch in (which also changes the authentication configuration) before we actually have to deploy any changes. Also, I have to rewrite the documentation on how to configure your authentication settings, and I don't want to have to rewrite it twice.

The big code change here is that the contents of the authentication configuration (minus the `module` bit) are passed as keyword arguments into the `AuthenticationProvider` constructor. The `config_values` method is gone. The code that (for example) turned the config setting `authorization_identifier_blacklist` to the constructor argument `authorization_blacklist` is gone. I just renamed the constructor argument to match the name of the config setting.

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -768,16 +768,26 @@ class BasicAuthenticationProvider(AuthenticationProvider):
         """Load a BasicAuthenticationProvider from site configuration."""
         return cls(**config)
 
+    # Used in the constructor to signify that the default argument
+    # value for the class should be used (as distinct from None, which
+    # indicates that no value should be used.)
+    class_default = object()
+    
     def __init__(self,
-                 identifier_regular_expression=DEFAULT_IDENTIFIER_REGULAR_EXPRESSION,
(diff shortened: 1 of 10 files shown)


aslagle:
Looks like the config for travis needs an update, but the code looks good :)

PR #390, opened 2016-10-27
Fix authentication crash

leonardr:
This branch fixes a crash in local_patron_lookup when a patron's PatronData gave their  username as NO_VALUE. As a generic object, NO_VALUE evaluated to True, so we tried to use it in a query, which caused a crash.

This branch changes NO_VALUE to evaluate to False, so that it's correctly treated as the absence of a value and we don't try to do a lookup based on it.

### DIFF
--- a/api/adobe_vendor_id.py
+++ b/api/adobe_vendor_id.py
@@ -3,7 +3,6 @@ import uuid
 import base64
 import os
 import datetime
-import jwt
 
 import flask
 from flask import Response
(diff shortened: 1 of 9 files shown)


aslagle:
Did you mean to update core?

aslagle:
I guess it's updated to master so it should be fine.

leonardr:
Yeah, core update is incidental but should be OK.

PR #391, opened 2016-10-27
Since the redirect uri is quoted in oauth_authentication_redirect, it…

aslagle:
… needs to be unquoted in oauth_authentication_callback.


leonardr:
Makes sense. :+1:

PR #394, opened 2016-10-27
New patron analytics event

### LINKED ISSUE #382: Track analytics event when new patron is created

aslagle:
This fixes #382

### DIFF
--- a/tests/test_axis.py
+++ b/tests/test_axis.py
@@ -168,7 +168,15 @@ class TestCirculationMonitor(DatabaseTest):
     )
 
     def test_process_book(self):
+        config = {
+            Configuration.POLICIES: {
+                Configuration.ANALYTICS_POLICY: ["core.local_analytics_provider"]
+            }
+        }
         with temp_config() as config:
+            Analytics.initialize(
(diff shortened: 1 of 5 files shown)


tests/test_axis.py
```diff
@@ -168,7 +168,15 @@ class TestCirculationMonitor(DatabaseTest):
...
     def test_process_book(self):
+        config = {
+            Configuration.POLICIES: {
+                Configuration.ANALYTICS_POLICY: ["core.local_analytics_provider"]
+            }
+        }
         with temp_config() as config:
+            Analytics.initialize(
+                ['core.local_analytics_provider'], config
+            )
```
aslagle: I had to change this test because it was relying on local_analytics_provider, and now a previous test is initializing Analytics with a mock_analytics_provider first.

leonardr:
:+1:

leonardr:
I'm a little concerned about distinguishing between the first time we see a patron, and the actual creation of a patron, but since creating a patron is not the circ manager's job calling it "circulation_manager_new_patron" leaves it unambiguous what is happening.

PR #395, opened 2016-10-28
Implementation of core AuthdataUtility class

leonardr:
This branch introduces the `AuthdataUtility` class, which implements the new authdata generation system described by the Vendor ID Service spec (https://docs.google.com/document/d/1j8nWPVmy95pJ_iU4UTC-QgHK2QhDUSdQ0OQTFR2NE_0/edit#) and described in semi-technical terms on the Library Simplified wiki (https://github.com/NYPL-Simplified/Simplified/wiki/AdobeVendorID).

With the `AuthdataUtility` class, a library without a Vendor ID server can generate JWTs that can be validated by a library _with_ a Vendor ID server. We intend to use these JWTs in place of our existing one-off authdata implementation.

This will let us keep a single Vendor ID server for all libraries that use the SimplyE app, without having to plug NYPL's Vendor ID server into every individual ILS out there.

This branch focuses on implementing the algorithm described in the Vendor ID Service spec, and the settings you would put add to your site configuration to set it up. Future branches will focus on using this algorithm in the way described in the DRM Autodiscovery document (https://github.com/NYPL-Simplified/Simplified/wiki/DRM-Autodiscovery#adobe-acs).


api/adobe_vendor_id.py
```diff
@@ -305,3 +310,141 @@ def uuid(self):
...
+                "Unknown library: %s" % library_uri
+            )
+
+        # We know the secret for this library, so we can re-decode the
+        # secret and require signature valudation this time.
+        secret = self.secrets_by_library[library_uri]
+        decoded = jwt.decode(authdata, secret, algorithm=self.ALGORITHM)
+        if not 'sub' in decoded:
+            raise jwt.exceptions.DecodeError("No subject specified.")
+        return library_uri, decoded['sub']
```
aslagle: Should this method check the expiration?

api/adobe_vendor_id.py
```diff
@@ -305,3 +310,141 @@ def uuid(self):
...
+        by `secrets_by_library`.
+
+        :return: a 2-tuple (library_uri, patron_identifier)
+
+        :raise jwt.exceptions.DecodeError: When the JWT is not valid
+        for any reason.
+        """
+        # First, decode the authdata without checking the signature.
+        decoded = jwt.decode(
+            authdata, algorithm='HS256', options=dict(verify_signature=False)
```
aslagle: this could use the algorithm constant

api/adobe_vendor_id.py
aslagle: I guess jwt does this for you

tests/test_adobe_vendor_id.py
```diff
@@ -315,3 +333,169 @@ def test_handle_accountinfo_failure(self):
...
+    def test_cannot_decode_expired_token(self):
+        expires = datetime.datetime(2016, 1, 1, 12, 0, 0)
+        authdata = self.authdata._encode(
+            "Patron identifier", exp=expires
+        )
+        assert_raises(
+            ExpiredSignatureError, self.authdata.decode, authdata
+        )
+        
+    def test_cannot_encode_null_patron_identifier(self):
```
leonardr: I disagree. The first argument to `_encode` is the library ID but the first argument to `encode` (which calls `_encode`) is the patron identifier.

aslagle: yep, you're right

PR #396, opened 2016-10-31
Include DRM autodiscovery tags in OPDS feeds

leonardr:
This branch includes a <drm:licensor> tag beneath any appropriate <link> tag with rel="fulfill". This completes the circulation manager part of DRM autodiscovery (as opposed to the Vendor ID implementation, which also happens to be kept in the circulation manager code).

The DRM autodiscovery spec is here: https://github.com/NYPL-Simplified/Simplified/wiki/DRMAutodiscoverySpecs#drm-extensions-to-opds

This code will only work for Adobe; we'll need to handle LCP and (especially) URMS differently.

The <drm:licensor> tag includes a drm:vendor attribute, which explains who is providing the Adobe vendor ID. It also includes a <drm:clientToken> tag, which gives an authdata value that identifies and authenticates the currently logged in patron. The authdata value is a JWT created from site configuration and a new permanent Credential associated with the patron and used solely for this purpose. This lets us invalidate the Credential if the patron runs into a problem, and it stops the Adobe vendor ID server from having any information that can be tied to the patron.

The JWT is cached on the CirculationManagerAnnotator object so that we don't generate different JWTs for the same patron over the course of a single request. I've checked that CirculationManagerAnnotator is created per-request but I would appreciate it if you could confirm this.

### DIFF
--- a/api/opds.py
+++ b/api/opds.py
@@ -644,11 +644,10 @@ class CirculationManagerAnnotator(Annotator):
                 vendor_id, jwt = authdata.encode(patron_identifier)
 
                 drm_link = OPDSFeed.makeelement("{%s}licensor" % OPDSFeed.DRM_NS)
-                server_id = OPDSFeed.makeelement("{%s}vendor" % OPDSFeed.DRM_NS)
-                server_id.text = vendor_id
+                vendor_attr = "{%s}vendor" % OPDSFeed.DRM_NS
+                drm_link.attrib[vendor_attr] = vendor_id
                 patron_key = OPDSFeed.makeelement("{%s}clientToken" % OPDSFeed.DRM_NS)
                 patron_key.text = jwt
-                drm_link.append(server_id)
                 drm_link.append(patron_key)
                 cached = [drm_link]
             self._adobe_id_tags[patron_identifier] = cached
--- a/api/opds.py
+++ b/api/opds.py
@@ -603,20 +603,23 @@ class CirculationManagerAnnotator(Annotator):
         if not active_loan or not delivery_mechanism:
             return []
         
-        # Get an identifier for the patron that will be registered
-        # with the DRM server.
-        patron = active_loan.patron
-        _db = Session.object_session(active_loan)
-        internal = DataSource.lookup(_db, DataSource.INTERNAL_PROCESSING)
-        def refresh(credential):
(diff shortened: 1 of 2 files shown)


leonardr:
I tested this on my local machine and it looks good:

```
    <link href="http://localhost:6500/works/Overdrive/Overdrive%20ID/493a559d-8db7-4082-be73-37f8b0b7a5b7/fulfill/2" rel="http://opds-spec.org/acquisition" type="vnd.adobe/adept+xml">
      <opds:indirectAcquisition type="application/epub+zip"/>
      <opds:availability since="2016-10-26T20:23:47Z" status="available" until="2016-11-16T21:23:00Z"/>
      <opds:holds total="0"/>
      <opds:copies available="12" total="13"/>
      <drm:licensor drm:vendor="NYPL">
        <drm:clientToken>eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJodHRwOi8vbnlwbC5vcmcvIiwiaWF0IjoxNDc4MDEyNjY2Ljc1ODU3NCwic3ViIjoiZGE0MTM4MjYtYTA0My0xMWU2LWJiODQtNjg1YjM1YzEyOWZjIiwiZXhwIjoxNDc4MDE2MjY2Ljc1ODU3NH0._UgHy5_qWDuqlfst3KgNdtn2NMxdmVVsR2DPDBImABg</drm:clientToken>
      </drm:licensor>
    </link>
```

api/opds.py
```diff
@@ -568,8 +586,73 @@ def fulfill_link(self, data_source_name, identifier,
...
+        # with the DRM server.
+        patron = active_loan.patron
+        _db = Session.object_session(active_loan)
+        internal = DataSource.lookup(_db, DataSource.INTERNAL_PROCESSING)
+        def refresh(credential):
+            credential.credential = str(uuid.uuid1())
+        patron_identifier = Credential.lookup(
+            _db, internal, self.ADOBE_ID_PATRON_IDENTIFIER, patron,
+            refresher_method=refresh, allow_persistent_token=True
+        )
```
aslagle: This creates the Adobe ID even for fulfill links that don't have Adobe DRM?

leonardr: Good catch.

PR #398, opened 2016-11-02
Properly look up/generate Adobe IDs when given JWT authdata

leonardr:
This branch completes our implementation of the DRM autodiscovery features for ACS. You can now take a JWT from Library A's OPDS feed and send it to Library B's Adobe Vendor ID implementation as authdata. Library B will create a `DelegatedPatronIdentifier` that maps the identifier Library A used to represent you, to the Adobe ID that Library B created for you.

The old system (in which you used an undocumented feature to ask Library A to create a `Credential` with an `authdata` value, and then exchanged it for another `Credential` with an Adobe ID) still works. I've created a migration script that finds all patrons who have that magic `Credential`, and gives them a `DelegatedPatronIdentifier` that contains their existing Adobe ID. This way they won't get a new Adobe ID when their client stops using the undocumented feature and starts using DRM autodiscovery.

I'm a little concerned about someone who is using two devices, one of which uses the undocumented feature and one of which uses DRM autodiscovery. So long as that person's Adobe ID stays the same, everything will work fine. But if they get tech support to clear out their Adobe ID, they will end up with two different Adobe IDs.


api/adobe_vendor_id.py
```diff
@@ -442,10 +505,57 @@ def decode(self, authdata):
...
+        )
+        if not credential:
+            # This patron has no Adobe ID. Do nothing.
+            return None, None
+        adobe_id = credential.credential
+
+        # Create a new Credential containing an anonymized patron ID.
+        def refresh(credential):
+            credential.credential = str(uuid.uuid1())
+        data_source = DataSource.lookup(_db, DataSource.ADOBE)
```
leonardr: I think the data source here should be INTERNAL_PROCESSING and not ADOBE.

PR #400, opened 2016-11-03
Authdata lookup always ends up pulling the final value from a DelegatedPatronIdentifier

leonardr:
This branch changes the semi-deprecated `uuid_and_label(patron)` method so that it migrates a patron's Adobe account ID from a Credential to a DelegatedPatronIdentifier (though this shouldn't be necessary because of the migration script). Then it always uses the value found in the DelegatedPatronIdentifier, not the value found in the Credential.

If in the future a patron needs to have their Adobe ID changed, you just need to delete both their "Vendor ID UUID" Credential (the old style) and their "Identifier for Adobe account ID purposes" Credential (the new style). There's no need to make sure the two types of credentials are kept in sync. In fact, once the migration script runs we could in theory delete everyone's "Vendor ID UUID" Credentials--the same information will be kept in the DelegatedPatronIdentifiers.

### DIFF
--- a/api/adobe_vendor_id.py
+++ b/api/adobe_vendor_id.py
@@ -373,7 +373,7 @@ class AdobeVendorIDModel(object):
             DelegatedPatronIdentifier.ADOBE_ACCOUNT_ID, value_generator
         )
         return (identifier.delegated_identifier,
-                "Delegated account ID %s" % identifier.delegated_identifier)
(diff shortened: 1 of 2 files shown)


aslagle:
I think I would keep the "Vendor ID UUID" Credentials around for a little while after running the migration just in case something goes wrong. But once we're confident about it it might be better to delete them so we don't have to worry about deleting both to reset an Adobe ID.

PR #401, opened 2016-11-04
Reset patron adobe id script

leonardr:
This branch creates a `PatronInputScript` whose job is to delete all of a Patron's credentials that connect them to an Adobe account ID. Since this is very bad if it happens by accident, the script gives you lots of opportunities to back out of making a change.

I also moved the constant `CirculationManagerAnnotator.ADOBE_ACCOUNT_ID_PATRON_IDENTIFIER` to `AuthdataUtility`, since it's used from a lot of different pieces of code, not just the OPDS feed generation.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -586,33 +586,36 @@ class AdobeAccountIDResetScript(PatronInputScript):
     def arg_parser(cls):
         parser = PatronInputScript.arg_parser()
         parser.add_argument(
-            '--dry-run',
-            help="Show which credentials would be deleted, but don't delete them.",
-            action='store_true',
-            default=False
+            '--delete',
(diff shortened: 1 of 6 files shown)


scripts.py
```diff
@@ -574,6 +580,68 @@ def do_generate(self, lane):
...
+class AdobeAccountIDResetScript(PatronInputScript):
+
+    @classmethod
+    def arg_parser(cls):
+        parser = PatronInputScript.arg_parser()
+        parser.add_argument(
+            '--dry-run',
+            help="Show which credentials would be deleted, but don't delete them.",
+            action='store_true',
+            default=False
```
aslagle: I'd make this default to true just to be extra careful.

leonardr: What would the variable be called?

PR #406, opened 2016-11-10
Fixed a confusion between DeliveryMechanism and LicensePoolDeliveryMechanism in call to fulfill_open_access.

leonardr:
This stupid bug (caused by naming a variable `delivery_mechanism` when it actually holds a `LicensePoolDeliveryMechanism`) had the effect of making `fulfill_open_access` fulfill whatever DeliveryMechanism happened to be the first one created for that LicensePool, regardless of what the patron actually asked for.

I had to add a lot of testing for `fulfill_open_access` because it previously had no test coverage.

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -419,7 +419,7 @@ class CirculationAPI(object):
         DeliveryMechanism.
 
         :param licensepool: The title to be fulfilled.
-        :param requested_lpdm: A DeliveryMechanism.
+        :param delivery_mechanism: A DeliveryMechanism.
         """
         if isinstance(delivery_mechanism, LicensePoolDeliveryMechanism):
             self.log.warn("LicensePoolDeliveryMechanism passed into fulfill_open_access, should be DeliveryMechanism.")
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -14,6 +14,7 @@ from core.model import (
     CirculationEvent,
     Identifier,
     DataSource,
+    LicensePoolDeliveryMechanism,
     LicensePool,
     Loan,
     Hold,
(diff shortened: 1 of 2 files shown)


api/circulation.py
```diff
@@ -413,27 +415,32 @@ def fulfill(self, patron, pin, licensepool, delivery_mechanism, sync_on_failure=
...
-        # patron's desired delivery mechanism isn't available.
+        """Fulfill an open-access LicensePool through the requested
+        DeliveryMechanism.
+
+        :param licensepool: The title to be fulfilled.
+        :param requested_lpdm: A DeliveryMechanism.
+        """
+        if isinstance(delivery_mechanism, LicensePoolDeliveryMechanism):
+            self.log.warn("LicensePoolDeliveryMechanism passed into fulfill_open_access, should be DeliveryMechanism.")
+            delivery_mechanism = delivery_mechanism.delivery_mechanism
```
leonardr: Here's a hack that fixes the problem if anyone else ever calls fulfill_open_access with a LPDM instead of a DM.

api/circulation.py
```diff
@@ -413,27 +415,32 @@ def fulfill(self, patron, pin, licensepool, delivery_mechanism, sync_on_failure=
...
+                # be used for fulfillment.
                 continue
             if lpdm.delivery_mechanism == delivery_mechanism:
                 # We found it! This is how the patron wants
                 # the book to be delivered.
                 fulfillment = lpdm
                 break
-            elif not fulfillment:
-                # This will do in a pinch.
-                fulfillment = lpdm
```
leonardr: It will not do in a pinch. Either we fulfill the book in the format the patron asked for or we tell them we can't fulfill the book at all. Sending a random format just crashes the client.

api/circulation.py
```diff
@@ -413,27 +415,32 @@ def fulfill(self, patron, pin, licensepool, delivery_mechanism, sync_on_failure=
         return fulfillment
 
     def fulfill_open_access(self, licensepool, delivery_mechanism):
-        # Keep track of a default way to fulfill this loan in case the
-        # patron's desired delivery mechanism isn't available.
+        """Fulfill an open-access LicensePool through the requested
+        DeliveryMechanism.
+
+        :param licensepool: The title to be fulfilled.
+        :param requested_lpdm: A DeliveryMechanism.
```
aslagle: this param name changed

PR #414, opened 2016-11-15
Fixed self link pointing to /groups//All%20Books in cached feeds

### LINKED ISSUE #290: self link is broken in /groups feed
It links to `/groups//All%20Books`, which is broken. It should probably point to `/groups/`.

aslagle:
Fixes #290


leonardr:
:+1:

PR #415, opened 2016-11-16
Base64-encode JWTs so that Adobe doesn't choke on them.

leonardr:
This branch transparently base64-encodes JWTs on the way out and decodes them on the way in, in hopes that Adobe will let them pass unhindered.


aslagle:
Looks fine, but the tests are failing.

PR #417, opened 2016-11-16
Handle incomplete patron activity when syncing bookshelf

aslagle:
This branch fixes a bug where an error from one of the remote APIs would cause local loans and holds from that API to be deleted, so the patron would lose those books.

This change makes it so no local loans and holds will be deleted if any of the remote APIs raised an exception. This isn't ideal - it would be better to handle loans and holds separately for each API, so that APIs that do not raise an exception can still have local loans and holds removed. But this is a quick improvement. 

This branch doesn't have any tests for patron_activity yet. There weren't any before and we may want to release this right away and come back to that.

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -558,8 +558,12 @@ class CirculationAPI(object):
             thread.join()
         loans = []
         holds = []
+        complete = True
         for thread in threads:
             if thread.exception:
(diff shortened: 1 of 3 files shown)


leonardr:
:+1:

PR #420, opened 2016-11-17
Adobe vendor id resilience

leonardr:
This branch makes the JWT decoding code more resilient in the face of something weird is doing to the token before sending it to the circ manager. I also add debugging logs so that we can get a better picture of what exactly Adobe is doing.

As always, this needs to go into production because Adobe is hard-coded to talk to production.


api/threem.py
```diff
@@ -115,6 +115,7 @@ def patron_activity(self, patron, pin):
         patron_id = patron.authorization_identifier
         path = "circulation/patron/%s" % patron_id
         response = self.request(path)
+        print response.content
```
aslagle: looks like this was left in by mistake

PR #421, opened 2016-11-17
Oneclick monitor

daryanypl:


api/oneclick.py
```diff
@@ -272,6 +281,66 @@ def release_hold(self, patron, pin, licensepool):
...
+        If the book has never been seen before, a new LicensePool
+        will be created for the book.
+
+        The book's LicensePool will be updated with current approximate 
+        circulation information (we can tell if it's available, but 
+        not how many copies). 
+        Bibliographic coverage will be ensured for the OneClick Identifier. 
+        Work will be created for the LicensePool and set as presentation-ready.
+
+        :param isbn the identifier OneClick uses
```
leonardr: This is not necessarily an ISBN, right?

config.json.sample
```diff
@@ -20,12 +19,12 @@
     	"S3" : {
         }, 
 
-        "OneClick" : {
+        "ThirdPartyProvider" : {
```
leonardr: I don't think it's very useful to have a sample OneClick configuration (of all the things you might need a sample configuration for) in `config.json.sample`. But it makes sense to have examples, so I don't object very strongly to it going in. 

I do object to calling it "ThirdPartyProvider" instead of "OneClick". If you call it "OneClick" then it can be copied and modified by someone who wants to configure OneClick. If it's called "ThirdPartyProvider" I don't see it enlightening anyone.

tests/test_oneclick.py
```diff
@@ -394,4 +396,84 @@ def test_release_hold(self):
...
+                'password' : 'password_123',
+                'remote_stage' : 'qa', 
+                'base_url' : 'www.oneclickapi.test', 
+                'basic_token' : 'abcdef123hijklm', 
+                "ebook_loan_length" : '21', 
+                "eaudio_loan_length" : '21'
+            }
+            monitor = OneClickCirculationMonitor(self._db)
+            monitor.api = MockOneClickAPI.from_config(self._db)
+
```
leonardr: You don't need to create a temporary config and call `MockOneClickAPI.from_config()` unless you're specifically testing `from_config()`. In this case you can instantiate a `MockOneClickAPI` through its constructor. 

However, I think you should still create a temporary config so that you can have a mock analytics provider running when you call process_availability(). Here's an example I copied and modified from `test_circulationapi.py`:

```
            provider = MockAnalyticsProvider()
            analytics = Analytics.initialize(
                ['core.mock_analytics_provider'], config
            )
            ... call process_availability() here ...
            # An analytics event was created.
            mock = Analytics.instance().providers[0]
            eq_(1, mock.count)
            eq_(CirculationEvent.TITLE_ADD,
                mock.event_type)
```

This will let you confirm that the TITLE_ADD event is sent when a new LicensePool is created, but not when an existing LicensePool is updated.

leonardr:
Looks good with a few minor changes. Travis is still not picking up your core branch.

api/oneclick.py
daryanypl: for OneClick ebooks and eaudio, it's always an isbn.  The only other ids they have are for magazines, which aren't accessible to us right now.

config.json.sample
daryanypl: OK, I'll change it back.  I thought you wanted to take the OneClick-specific stuff out of the config file and into the wiki.

tests/test_oneclick.py
daryanypl: I'm not sure I fully grasp the temp_config usage.  Is it that the config files aren't being read during unit testing?  I'll ask in person.

daryanypl:
Closing, because got superceded by https://github.com/NYPL-Simplified/circulation/pull/423 .

PR #424, opened 2016-11-22
Sip authentication provider

leonardr:
This branch creates a BasicAuthenticationProvider that gets its patron information from the SIP client. This should work (more or less) with the SIP servers provided by Sierra and Polaris, as well as the SIPServer implementation used by Evergreen and Koha.

Polaris and SIPServer have custom fields which this branch exploits to get information not normally conveyed through SIP, such as the date a patron's authorization expires, and a permanent ID for them.

This authentication provider has a number of shortcomings when it comes to dealing with patrons whose borrowing privileges have been revoked, but that's mainly a shortcoming of our data model, which doesn't recognize the fact that you might lose borrowing privileges for reasons other than 'expired card' or 'excess fines'.


api/sip/client.py
```diff
@@ -174,14 +183,16 @@ class SIPClient(Constants):
     log = logging.getLogger("SIPClient")
    
     def __init__(self, target_server, target_port, login_user_id=None,
-                 login_password=None, separator='|'):
+                 login_password=None, location_code=None, separator=None):
```
daryanypl: I'm not sure what location_code is.  Could I have a method doc for it?

api/sip/__init__.py
```diff
@@ -0,0 +1,79 @@
...
+        if 'sipserver_patron_class' in info:
+            patrondata.external_type = info['sipserver_patron_class']
+        for expire_field in ['sipserver_patron_expiration', 'polaris_patron_expiration']:
+            if expire_field in info:
+                value = info.get(expire_field)
+                value = cls.parse_date(value)
+                if value:
+                    patrondata.authorization_expires = value
+                    break
+        return patrondata
```
leonardr: You would get an empty PatronData object. `BasicAuthenticationProvider.local_patron_lookup` would be unable to find an existing patron, and `PatronData.get_or_create_patron` would raise an exception.

api/sip/client.py
```diff
@@ -174,14 +183,16 @@ class SIPClient(Constants):
...
         self.target_server = target_server
         if target_port:
             self.target_port = int(target_port)
+        self.location_code = location_code
+        self.separator = separator or '|'
```
leonardr: When SIP2AuthenticationProvider is instantiated from config, if no field separator is provided in the config, the value is set to None. Since a value is available for field_separator, even though the value is None, the SIPClient constructor was being called with separator=None. This code makes means that you get the default separator whether you pass in separator=None or omit the separator altogether.

daryanypl:
👍   I still would like to have a docstring for the location_code, but it looks great.

leonardr:
The docstring for the `SIP2AuthenticationProvider` constructor describes `location_code`. Is that not where you wanted the docstring?

PR #427, opened 2016-12-01
Check whether a patron has borrowing privileges at the moment they try to borrow something

leonardr:
Currently, you cannot authenticate with the circulation manager if your library card has lost borrowing privileges (e.g. has expired). This branch changes it so that you can authenticate (meaning you can use the app and read books you've already checked out), but you can't borrow any new books or place any holds.

### DIFF
--- a/api/circulation_exceptions.py
+++ b/api/circulation_exceptions.py
@@ -79,7 +79,28 @@ class OutstandingFines(CannotLoan):
     """The patron has outstanding fines above the limit in the library's 
     policy."""
     status_code = 403
+    
+class AuthorizationExpired(CannotLoan):
+    """The patron's authorization has expired."""
+    status_code = 403
 
+    def as_problem_detail_document(self, debug=False):
+        """Return a suitable problem detail document."""
+        return EXPIRED_CREDENTIALS
+    
+class AuthorizationBlocked(CannotLoan):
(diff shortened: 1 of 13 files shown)


api/circulation_exceptions.py
```diff
@@ -79,7 +79,28 @@ class OutstandingFines(CannotLoan):
...
+    
+class AuthorizationExpired(CannotLoan):
+    """The patron's authorization has expired."""
+    status_code = 403
 
+    def as_problem_detail_document(self, debug=False):
+        """Return a suitable problem detail document."""
+        return EXPIRED_CREDENTIALS
+    
+class AuthorizationBlocked(CannotLoan):
```
leonardr: Currently we don't know how to read from the ILS whether the patron's card is blocked, but it's coming soon, so I figured I'd add the exception and the problem detail document as long as I'm dealing with reasons why the patron might not have borrowing privileges.

api/util/patron.py
```diff
@@ -0,0 +1,94 @@
+import datetime
+from api.config import Configuration
+from api.circulation_exceptions import *
+
+
+class PatronUtility(object):
```
leonardr: This code comes in from core basically unaltered except to become class methods of PatronUtility instead of instance methods of Patron.

PR #431, opened 2016-12-06
Added body to annotation parsing.

aslagle:
This is a small change to support last reading position in the mobile apps. We thought an idling annotation didn't need a body, but the apps would like to store arbitrary things in there, like a unique device id so they can tell if a synced annotation is an old one from the same device and shouldn't be shown to the patron. They're also considering storing an app timestamp, to handle the case of multiple devices being used offline and then syncing.


aslagle:
I just discovered that some of my tests don't run without network access, since the jsonld library has to get the context. I should fix that - the code also shouldn't depend on the context being available.

aslagle:
It looks like pyld does some in-memory caching, but the context still needs to be retrieved from the network the first time it's used.

leonardr:
This makes sense. Should be be concerned that the 'canonical' version of this document might change and we won't pick up the change? If so, then we should only replace the jsonld document loader during a test.

aslagle:
That could happen, but if it did I think we'd want to keep using the outdated version until we could make sure the changes don't break our code. This also ensures our app still works if the canonical version isn't available for some reason, but that's less likely.

leonardr:
:+1:

PR #437, opened 2016-12-16
Remove the final newline from Adobe vendor ID tokens; it confuses Adobe.

leonardr:


tests/test_adobe_vendor_id.py
```diff
@@ -924,14 +925,15 @@ def test_adobe_base64_encode_decode(self):
         """Test our special variant of base64 encoding designed to avoid
         triggering an Adobe bug.
         """
-        value = 'LbU}66%\\-4zt>R>_)\n2Q'
-
+        value = "!\tFN6~'Es52?X!#)Z*_S"
```
leonardr: I changed this to a value that contains both a : and a ; when encoded, so I could test both characters later on. Both the old value and the new one were randomly generated.

PR #438, opened 2016-12-16
Oneclick prod testing

daryanypl:


bin/oneclick_library_import
```diff
@@ -1,6 +1,10 @@
 #!/usr/bin/env python
 """Update the circulation manager server with new books from the
-OneClick content server."""
+OneClick content server.
+
+If you want to test with sample OneClick data on prod database, run like so:
```
aslagle: I don't think we'd ever want sample data on a prod database - do you mean a local database or a QA database?

daryanypl: Local, QA, or prod.  I had a situation where I wanted to make sure I was able to run a catalog populating script, and save the results to the database, and I didn't want to deal with the entire catalog.  I used the small sample catalog file that I created earlier for a unit test, and was able to troubleshoot my database connection and handling.

aslagle: That's fine, but I would take the word "prod" out of the comment. You definitely wouldn't want to run this with a production database that was actually in use.

daryanypl: done

leonardr:
The substantive commit in here looks good.

PR #440, opened 2017-01-06
Implement DRM Device Management Protocol

leonardr:
This branch implements the [DRM Device Management Protocol](https://github.com/NYPL-Simplified/Simplified/wiki/DRM-Device-Management) for Adobe ACS.

This is the first implementation of the protocol, so differences between the spec and the implementation might be a problem either in the spec or the implementation.

I have manually tested the code in `routes.py` and written unit tests for everything else.

### DIFF
--- a/api/adobe_vendor_id.py
+++ b/api/adobe_vendor_id.py
@@ -73,8 +73,13 @@ class AdobeVendorIDController(object):
     def status_handler(self):
         return Response("UP", 200, self.PLAIN_TEXT_HEADERS)
 
-    # Implementation of the DRM Device ID Management Protocol.
-    #
+
+class DeviceManagementProtocolController(object):    
+    """Implementation of the DRM Device ID Management Protocol.
+
+    The code that does the actual work is in DeviceManagementRequestHandler.
+    """
+
     @property
     def link_template_header(self):
         url = url_for("adobe_drm_device", device_id="{id}", _external=True)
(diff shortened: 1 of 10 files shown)


api/adobe_vendor_id.py
```diff
@@ -731,10 +846,11 @@ def decode_short_client_token(self, token):
 
         :raise ValueError: When the token is not valid for any reason.
         """
-        if not ' ' in token:
+        if not '|' in token:
             raise ValueError(
-                'Supposed client token "%s" does not contain a space.' % token
+                'Supposed client token "%s" does not contain a pipe.' % token
             )
+
```
leonardr: I changed the spec earlier, but not the code.

api/adobe_vendor_id.py
```diff
@@ -411,6 +523,9 @@ def to_delegated_patron_identifier_uuid(
             self._db, library_uri, foreign_patron_identifier,
             DelegatedPatronIdentifier.ADOBE_ACCOUNT_ID, value_generator
         )
+
+        if identifier is None:
+            return None, None
```
leonardr: This shouldn't happen, but it seems like a good way to be safe.

api/base_controller.py
```diff
@@ -0,0 +1,81 @@
+import flask
```
leonardr: All this code is refactored from `CirculationManagerController` in `controller.py` and moved to a new file to avoid a circular dependency. I wanted to keep all the Adobe-related controllers in `adobe_vendor_id.py`, so I needed a new place to put code that authenticates patrons according to the circulation manager's configuration (as opposed to the technique defined by Adobe's vendor ID spec).

api/adobe_vendor_id.py
```diff
@@ -147,6 +229,36 @@ def error_document(self, type, message):
...
+                x.device_identifier
+                for x in self.credential.drm_device_identifiers
+            )
+        )
+
+    def register_device(self, data):
+        device_ids = data.split("\n")
+        if len(device_ids) > 1:
+            return PAYLOAD_TOO_LARGE.detailed(
+                _("You may only register one device ID at a time.")
```
aslagle: I think this differs from the spec

leonardr: I would say the spec is silent on this issue. I can add something to the spec saying that the implementing server may put limits on how many IDs can be registered at once.

aslagle: sounds good

api/adobe_vendor_id.py
```diff
@@ -63,6 +68,83 @@ def status_handler(self):
...
+        return {"Link-Template": '<%s>; rel="item"' % url}
+
+    def _request_handler(self, patron):
+        """Create a DeviceManagementRequestHandler for the appropriate
+        Credential of the given Patron.
+       
+        :return: A DeviceManagementRequestHandler
+        """
+        if not patron:
+            return INVALID_CREDENTIALS.detailed("No authenticated patron")
```
aslagle: should this string be localized?

aslagle: there are a few other details that aren't as well, but they might not be shown to patrons

leonardr: Yes, in the sense that all our messages should be localized, but this should never be sent in real usage, let alone displayed to a patron.

api/adobe_vendor_id.py
```diff
@@ -63,6 +68,85 @@ def status_handler(self):
...
+                return output
+            headers = self.link_template_header
+            headers['Content-Type'] = device_ids
+            return Response(output, 200, headers)
+        elif flask.request.method=='POST':
+            # Add a device ID to the list.
+            incoming_media_type = flask.request.headers.get('Content-Type')
+            if incoming_media_type != device_ids:
+                return UNSUPPORTED_MEDIA_TYPE.detailed(
+                    _("Expected %s document.") % device_ids
```
aslagle: this won't work, you need something like `_("Integration error communicating with %(service_name)s", service_name=self.service_name)`

PR #441, opened 2017-01-06
Correct a call to the SeriesLane constructor

leonardr:
This branch fixes a problem with SeriesLane that causes me to get a lot of error notifications. I believe a similar problem with ContributorLane was fixed earlier.

Basically, positional methods are passed into the SeriesLane constructor, and there was a mismatch such that what was `languages` outside the constructor was `parent` inside the constructor.

I changed the constructor call to use positional arguments, and added a test.


courte:
:+1: JK, I looked now.

Was this working before because the routes assume english unless something else is specified?

leonardr:
It would have worked in situations where no language was specified, but near as I can tell the generated links always include a language.

PR #443, opened 2017-01-10
Generate short client tokens with blank expiration time and enforce blank expiration time on input.

leonardr:
Let's say you register your device with Adobe as a SimplyE device, using the short client token `NYNYPL|some-timestamp|some-uuid|some-signature`. The `some-timestamp` value there is the time the token expires, about one hour in the future.

Adobe asks the SimplyE authentication API whether username "NYNYPL|some-timestamp|some-uuid" password "some-signature" is valid, and the authentication API says yes. You're good to go.

Six months later, you decide to log out of SimplyE. Your device sends username "NYNYPL|some-timestamp|some-uuid" password "some-signature" to Adobe as part of a deactivation request. Adobe asks the authentication API if this username/password combination is valid, but it's no longer valid. `some-timestamp` was six months ago (minus one hour) and the token has expired.

Okay, no problem. The circ manager gives out new tokens all the time. Just use one of the new ones. They all authenticate the same Adobe ID.

This won't work either. Adobe will only deregister a device using the _exact same username and password_ that was used to register it. And that username doesn't work anymore.

This means that the username and password can never change. We can't use a patron's actual username and password (though we're long past that point), because both of those change routinely. We can't use a meaningless token that expires after an hour (that's what this branch fixes). We have to use a meaningless token that never, ever changes.

### DIFF
--- a/api/adobe_vendor_id.py
+++ b/api/adobe_vendor_id.py
@@ -881,9 +881,6 @@ class AuthdataUtility(object):
         library_short_name, expiration, patron_identifier = token.split("|", 2)
 
         library_short_name = library_short_name.upper()
-        # NOTE: See the note in _encode_short_client_token(). We no longer
-        # put expiration times in short client tokens. That field is blank and
-        # any value it might have is ignored.
         if expiration:
             raise ValueError('Expiration time must be blank.')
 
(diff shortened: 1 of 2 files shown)


aslagle:
I don't think we need to keep around the commented out code for the expiration token.

leonardr:
I can take it out but I'm holding out hope that Adobe changes the behavior and we can restore it. Just writing that down makes me feel foolish though. If Adobe wants to change something they can make deactivate work with authdata and we can avoid this whole dance.

PR #455, opened 2017-02-14
Move library and collection configuration from JSON into the database

leonardr:
This branch has a migration script that copies information about the library's collections, and about the library's relationship with the library registry, from the JSON configuration file into the database. (the new `libraries` and `collections` tables).

This branch changes `Authenticator` to take a `Library` object. This is the library people are authenticating with. The `Library` object is used to provide the name and ID for the Authentication for OPDS document.

The `Library` object is also used in `AuthdataUtility.from_config()`, which has been changed to pull the library's short name and shared secret (used when generating short client tokens) from the database instead of the JSON configuration file. Since some of the `AuthdataUtility` information now comes from the database, `from_config` now takes a database argument. There are a lot of small changes  to make sure a database argument was always available and that the configuration was always present when needed during a test.

This branch does not include any code that looks in the `Collection` table for collection configuration; that's coming up later.

I won't be merging this branch with `master` yet, but I'll be using it as a starting point for other branches.


api/controller.py
```diff
@@ -256,9 +257,10 @@ def setup_adobe_vendor_id(self):
             self.adobe_vendor_id = None
 
         # But almost all libraries will have this setup.
-        if adobe.get(AuthdataUtility.AUTHDATA_SECRET_KEY):
+        library = Library.instance(self._db)
+        if library.library_registry_shared_secret:
```
leonardr: This change had unexpected consequences; see below.

tests/test_controller.py
```diff
@@ -553,8 +554,9 @@ def test_borrow_and_fulfill_with_streaming_delivery_mechanism(self):
                     datetime.datetime.utcnow() + datetime.timedelta(seconds=3600),
                 )
             )
-            response = self.manager.loans.borrow(
-                data_source.name, identifier.type, identifier.identifier)
+            with self.temp_config():
```
leonardr: There are a number of places where I had to use `self.temp_config()` to create a configuration object where I didn't before. I think this is because of the change I mentioned earlier, on line 288 of `controller.py`. That code makes sure the AuthdataUtility is fully configured if the `library_registry_shared_secret` is set in the database. This always happens, since we call `MockAdobeConfiguration.initialize_library` at the start of these tests when setting up the `AuthdataUtility`. The configuration we create to set up the `AuthdataUtility` goes out of scope, but the change we made to the database persists throughout the test.

The rule is that the AuthdataUtility doesn't have to be configured if the configuration was completely absent. Most of the time, that's what used to happen. We would get to `CirculationManagerAnnotator._adobe_patron_identifier`, try to look up AuthdataUtility, notice that there is no configuration (since we're not testing that part of the code), and that part of the code would be skipped. But now, the config goes out of scope and the database change doesn't. We try to look up AuthdataUtility and get a crash, because the database part of the configuration is set up but the config part is not.

api/axis.py
```diff
@@ -244,12 +244,12 @@ def __init__(self, _db, name="Axis 360 Circulation Monitor",
         else:
             # This should only happen during a test.
             self.metadata_wrangler = None
+        self.api = api or Axis360API.from_environment(self._db)
         self.bibliographic_coverage_provider = (
-            Axis360BibliographicCoverageProvider(self._db)
```
leonardr: You can't just call these constructors anymore, you have to pass in a specific API objet.

api/overdrive.py
```diff
@@ -746,50 +749,23 @@ def json(self):
         return json.loads(self.content)
 
 
-class DummyOverdriveAPI(OverdriveAPI):
+class DummyOverdriveAPI(MockOverdriveAPI, OverdriveAPI):
```
leonardr: A lot of code in this class was a duplicate of code in core `MockOverdriveAPI`. Rather than change both classes to use the new `OverdriveAPI` technique, I just removed all the duplicate code. This had two main effects on the tests:

1. The circulation tests queue up mock responses by passing in Python dictionaries. The core tests queue up mock responses by passing in JSON strings. I had to change code in a number of places to treat a response body as either a JSON string or an object.

2. The circulation mock API incorrectly "queued" items by appending them to a list, instead of inserting them at the beginning of the list. The queue was actually a stack. This meant you had to "queue" responses from last to first. Now you queue responses in the same order as the mock HTTP requests are made.

api/threem.py
```diff
@@ -603,12 +603,12 @@ class ThreeMCirculationSweep(IdentifierSweepMonitor):
     count the same event.  However it will greatly improve our current
     view of our 3M circulation, which is more important.
     """
-    def __init__(self, _db, testing=False, api=None):
```
leonardr: All this `testing` stuff was unnecessary, since you can just pass in a mock API if you want. I took it out of all these classes.

tests/test_3m.py
```diff
@@ -208,30 +209,31 @@ def test_exception(self):
 
 class TestThreeMEventMonitor(DatabaseTest):
 
-  def test_default_start_time(self):
```
leonardr: This is almost entirely a change in indentation.

tests/test_overdrive.py
```diff
@@ -65,7 +65,7 @@ def test_default_notification_email_address(self):
             "patron_info.json"
         )
         api = DummyOverdriveAPI(self._db)
-        api.queue_response(content=patron_with_email)
+        api.queue_response(200, content=patron_with_email)
```
leonardr: The core `MockOverdriveAPI` implementation of `queue_response` has a non-optional argument for status code.

tests/test_overdrive.py
```diff
@@ -146,7 +146,7 @@ def test_place_hold_looks_up_notification_address(self):
         # And when we placed it on hold, we passed in foo@bar.com
         # as the email address -- not notifications@example.com.
         hold_request = api.requests[-1]
-        body = hold_request[4]
+        body = hold_request[3][1]
```
leonardr: The core `MockOverdriveAPI` doesn't put the request body in a special slot in the array it appends to `request`--it just puts all positional arguments to the "make a request" method into the fourth position of the array. The body happens to be the second positional argument.

api/overdrive.py
aslagle: It is confusing to have two mock API classes. Are you planning to combine them eventually?

migration/20170224-copy-collection-configuration-into-database.py
```diff
@@ -0,0 +1,119 @@
...
+    config = Configuration.integration('Axis 360')
+    if not config:
+        print u"No Axis 360 configuration, not creating a Collection for it."
+        return
+    print u"Creating Collection object for Axis 360 collection."
+    username = config.get('username')
+    password = config.get('password')
+    library_id = config.get('library_id')
+    # This is not technically a URL, it's u"production" or u"staging",
+    # but it's converted into a URL internally.
```
aslagle: You could have done that conversion here, instead of maintaining the URLs in the code. Is there an advantage to using the nicknames, like if axis decides to change their URLs?

migration/20170224-copy-collection-configuration-into-database.py
```diff
@@ -0,0 +1,119 @@
...
+    library.collections.append(collection)
+    collection.url = url
+    
+library = Library.instance(_db)
+copy_library_registry_information(_db, library)
+convert_overdrive(_db, library)
+convert_bibliotheca(_db, library)
+convert_axis(_db, library)
+convert_content_server(_db, library)
+set_trace()
```
aslagle: take this out sometime before you merge

tests/test_controller.py
aslagle: I think it's fine for now.

api/overdrive.py
leonardr: I don't have plans to do this. The mock API class in circulation and its superclass in core have the same relationship as the actual API class in circulation and its superclass in core. The circulation API does patron authentication to do stuff like borrow books, and the core API uses the library credentials to do stuff like check availability.

We could in theory merge the two classes but I think I've gotten rid of most of the confusion by removing the redundant code from the circulation mock API.

migration/20170224-copy-collection-configuration-into-database.py
leonardr: I was imagining in the admin interface, if someone needs to change from production to staging and back to test something, it would be better to change the nicknames rather than have to remember what the URLs are.

api/overdrive.py
aslagle: That makes sense, how about making the naming match the two OverdriveAPI classes, so it's more clear that they're doing the same thing? Import MockOverdriveAPI as BaseMockOverdriveAPI and name both classes MockOverdriveAPI.

migration/20170224-copy-collection-configuration-into-database.py
aslagle: :+1:

PR #456, opened 2017-02-15
Remove loans/holds known to have expired, and very old loans/holds with no expiration date.

leonardr:
This branch fixes https://github.com/NYPL-Simplified/server_core/issues/225.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -1,5 +1,8 @@
 from cStringIO import StringIO
-from datetime import timedelta
+from datetime import (
+    datetime,
+    timedelta,
+)
 from nose.tools import set_trace
 import csv
 import json
@@ -39,10 +42,12 @@ from core.model import (
     DataSource,
     DeliveryMechanism,
     Edition,
+    Hold,
     Hyperlink,
     Identifier,
     LicensePool,
     LicensePoolDeliveryMechanism,
+    Loan,
     Representation,
     Subject,
     Work,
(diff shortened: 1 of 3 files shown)


scripts.py
```diff
@@ -779,3 +784,50 @@ def __init__(self):
...
+                    obj.start < older_than).filter(
+                        LicensePool.open_access == False
+                    )
+            explain = "%s older than %s" % (
+                what, older_than.strftime("%Y-%m-%d")
+            )
+            self._reap(qu, explain)
+
+    def _reap(self, qu, what):
+        counter = 0
```
daryanypl: would you mind adding a doc string?  I'd like to see what's expected to come in as qu and what.

scripts.py
```diff
@@ -779,3 +784,50 @@ def __init__(self):
...
+        print "Reaping %d %s." % (qu.count(), what)
+        for o in qu:
+            self._db.delete(o)
```
daryanypl: Is there ever a case when the delete won't go through because there are dependent records?

scripts.py
```diff
@@ -779,3 +784,50 @@ def __init__(self):
...
+
+class LoanReaperScript(Script):
+    """Remove expired loans and holds whose owners have not yet synced
+    with the loan providers.
+
+    This stops the library from keeping a record of the final loans and
+    holds of a patron who stopped using the circulation manager.
+
+    If a loan or (more likely) hold is removed incorrectly, it will be
+    restored the next time the patron syncs their loans feed.
```
daryanypl: for my edification: is the patron's holds kept locally in SimplyE, and then merged over what the database knows about the patron's holds?

tests/test_scripts.py
```diff
@@ -211,7 +213,95 @@ def test_process_lane(self):
...
+class TestLoanReaperScript(DatabaseTest):
+
+    def test_reaping(self):
+
+        # This patron stopped using the circulation manager a long time
+        # ago.
+        inactive_patron = self._patron()
+
+        # This patron is still using the circulation manager.
+        current_patron = self._patron()
```
daryanypl: how do you know when someone is a current vs not-synched-in-a-long-time patron?

leonardr: We don't know that in general. For purposes of the test I'm setting up those two scenarios. If a patron has loans in the database that have expired, they clearly haven't synced since before those loans expired.

scripts.py
leonardr: The patron's holds are kept with whoever owns the DRM server. Overdrive, Bibliotheca, etc. The circulation manager tries to get a picture of a patron's holds through the distributor's API, but this picture is always incomplete because we don't know for sure when a book will be available. The SimplyE client learns about holds from the circulation manager. 

The SimplyE client has the ability to _create_ holds, but it does this by asking the circulation manager to do the work, and the circulation manager asks the distributor to do the work. The distributor is ultimately the one in charge of the holds.

scripts.py
leonardr: I don't think so. Loans are associated with LicensePool and LicensePoolDelivery, both of which can survive the deletion of an individual Loan.

daryanypl:
👍

PR #458, opened 2017-02-17
Move Lane-making to core

### LINKED ISSUE #457: Update submodule to include changes to Lane
Changes to `Lane` in NYPL-Simplified/server_core#457 break circulation, causing all `TestController` tests to fail.

courte:
Some elements of creating Lanes have been moved into core. This branch has circulation depend on those changes.

Fixes #457.

### DIFF
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -290,6 +290,7 @@ def lanes_for_large_collection(_db, languages):
              ),
             Lane(_db, full_name="Informational books", genres=None,
                  fiction=False, exclude_genres=[genres.Biography_Memoir],
+                 subgenre_behavior=Lane.IN_SAME_LANE,
                  **children_common_args
              )
         ],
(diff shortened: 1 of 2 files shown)


api/lanes.py
```diff
@@ -290,6 +289,7 @@ def lanes_for_large_collection(_db, languages):
              ),
             Lane(_db, full_name="Informational books", genres=None,
                  fiction=False, exclude_genres=[genres.Biography_Memoir],
+                 subgenre_behavior=Lane.IN_SAME_LANE,
```
courte: This was the cause of the errors that spurred #457. Because there were no genres, the `exclude_genres` set here was being ignored before (Biography/Memoir books can be see in the children's "Informational Books" lane at the moment). Now it's being properly excluded and the lane needs to be collapsed or else duplicate lane name errors ensue.

leonardr:
Awesome. :+1:

PR #459, opened 2017-02-22
Expose the User Profile Management Protocol through the circulation manager

leonardr:
This branch adds an endpoint on the circulation manager that implements the User Profile Management Protocol. It also adds a link into that endpoint from a patron's bookshelf feed.

Since the rules of the UPMP are implemented in core/user_profile.py, there's not much code in this branch.

In addition to the unit tests of the controller I've done a manual integration test to verify that the code in `routes.py` sends my request to the right place.

This branch also ensures that a patron cannot use the annotation controller to synchronize annotations without previously opting in to annotation sync.


api/annotations.py
```diff
@@ -166,9 +166,8 @@ def parse(cls, _db, data, patron):
         else:
             content = None
 
-        annotation, is_new = get_one_or_create(
-            _db, Annotation,
-            patron=patron,
+        annotation, is_new = Annotations.get_one_or_create(
+            _db, patron=patron,
```
leonardr: This makes sure that a poorly written client can't accidentally sync annotations when the patron has opted out.

api/overdrive.py
```diff
@@ -327,8 +327,10 @@ def get_fulfillment_link_from_download_link(self, patron, pin, download_link, fu
         # If this for Overdrive's streaming reader, and the link expires,
         # the patron can go back to the circulation manager fulfill url
         # again to get a new one.
-        if not fulfill_url:
+        if not fulfill_url and flask.request:
```
leonardr: This is an unrelated change to fix informational Overdrive scripts I sometimes write that run outside of a Flask environment.

api/annotations.py
```diff
@@ -166,9 +169,8 @@ def parse(cls, _db, data, patron):
         else:
             content = None
 
-        annotation, is_new = get_one_or_create(
-            _db, Annotation,
-            patron=patron,
+        annotation, is_new = Annotation.get_one_or_create(
+            _db, patron=patron,
```
leonardr: If the check on line 128 somehow doesn't happen, this will raise an exception.

api/annotations.py
```diff
@@ -125,6 +125,9 @@ class AnnotationParser(object):
 
     @classmethod
     def parse(cls, _db, data, patron):
+        if patron.synchronize_annotations != True:
```
leonardr: This stops annotation synchronization from proceeding unless the patron has opted in.

api/controller.py
```diff
@@ -1046,7 +1050,30 @@ def series(self, series_name, languages, audiences):
...
+    def protocol(self):
+        """Handle a UPMP request."""
+        controller = self._controller
+        if flask.request.method == 'GET':
+            result = controller.get()
+        else:
+            result = controller.put(flask.request.headers, flask.request.data)
+        if isinstance(result, ProblemDetail):
+            return result
+        return make_response(*result)
```
daryanypl: what does this do?

api/opds.py
```diff
@@ -781,13 +781,29 @@ def drm_device_registration_feed_tags(self, patron):
...
+        link = OPDSFeed.makeelement("link")
+        link.attrib['rel'] = 'http://librarysimplified.org/terms/rel/user-profile'
+        link.attrib['href'] = self.url_for(
+            'patron_profile', _external=True
+        )
+        return link
+        
     def annotate_feed(self, feed, lane):
-        """Add feed-level DRM device registration tags to the feed."""
+        """Annotate the feed with top-level DRM device registration tags
```
daryanypl: what's the relationship between drm and annotations?

api/opds.py
```diff
@@ -781,13 +781,29 @@ def drm_device_registration_feed_tags(self, patron):
...
+        and a link to the User Profile Management Protocol endpoint.
+        """
         super(CirculationManagerLoanAndHoldAnnotator, self).annotate_feed(
             feed, lane
```
daryanypl: What does lane parameter do?

api/controller.py
leonardr: If you're asking about line 1074, make_response() turns a 3-tuple (body, status code, headers) into a Flask `Response` object.

If you're asking about the method as a whole, it's a Flask-specific wrapper around the ProfileController.

api/opds.py
leonardr: There is no relationship between DRM and the database objects called Annotations.

"Annotate" here means to add extra information to an existing OPDS feed.

api/opds.py
leonardr: Lane is always going to be None in this situation because this is the patron's loans feed. But it's part of the method signature because annotate_feed() is part of the general OPDS feed interface, and most feeds are in the context of a specific lane.

leonardr:
This satisfies https://github.com/NYPL-Simplified/circulation/pull/462.

daryanypl:
👍

PR #468, opened 2017-03-01
Added open access dimension to google analytics. (Fixes #435)

### LINKED ISSUE #435: Add a circulation event to GA that captures the "checkouts" OA content server and associated dimension
Currently the circulation manager checkout events do not capture distribution of OA Server content in Google Analytics.  This makes it difficult to measure impact of those titles to overall circ or gauge its demand in SimplyE.  If there is any other tile of copyright information or license information that would be helpful to include (Public Domain, vs CC or Commercial Contract..

aslagle:


leonardr:
:+1:

aslagle:
Travis isn't working right now, but the tests pass locally and I also tested this with our google analytics QA account.

PR #471, opened 2017-03-15
Oneclick collection conversion

leonardr:
This branch makes a number of minor changes discovered when getting a test site working with Day 1.

The most obvious change is in the migration script that moves collections from JSON configuration to database-based configuration. It used to do nothing for OneClick collections, and now it covers them.

I also made a number of script files executable.


daryanypl:
👍  Looks good.

PR #472, opened 2017-03-16
Added admin controllers for editing libraries and collections.

aslagle:
This is the server component of https://github.com/NYPL-Simplified/circulation-web/issues/74


api/admin/controller.py
```diff
@@ -815,3 +823,181 @@ def result_to_row(result):
...
+            self._db.rollback()
+            return INCOMPLETE_COLLECTION_CONFIGURATION.detailed(_("An OPDS Import collection must have a url."))
+
+        if (collection.protocol == Collection.OVERDRIVE
+            and (not collection.username
+                 or not collection.password
+                 or not collection.external_account_id
+                 or not collection.setting("website_id").value)):
+            self._db.rollback()
+            return INCOMPLETE_COLLECTION_CONFIGURATION.detailed(_("An Overdrive collection must have username, password, external_account_id, and website_id."))
```
leonardr: This can wait for another branch, but these database fields have different human-readable names for different protocols. For example, the OneClick 'password' is called the 'token'. I would use the human-readable names everywhere, e.g. "website ID" versus "website_id".

api/admin/controller.py
```diff
@@ -815,3 +823,181 @@ def result_to_row(result):
...
+            and (not collection.username
+                 or not collection.password
+                 or not collection.external_account_id)):
+            self._db.rollback()
+            return INCOMPLETE_COLLECTION_CONFIGURATION.detailed(_("A OneClick collection must have username, password, and external_account_id."))
+
+        for short_name in libraries:
+            library = get_one(self._db, Library, short_name=short_name)
+            if not library:
+                return NO_SUCH_LIBRARY.detailed(_("%(library_short_name)s was added to the collection, but it does not exist.", library_short_name=short_name))
```
leonardr: This should make it clear that you _attempted_ to add a nonexistent library to the collection, not that it actually happened.

api/admin/routes.py
```diff
@@ -46,9 +46,10 @@ def decorated(*args, **kwargs):
 def requires_csrf_token(f):
     @wraps(f)
     def decorated(*args, **kwargs):
-        token = app.manager.admin_sign_in_controller.check_csrf_token()
-        if isinstance(token, ProblemDetail):
-            return token
+        if flask.request.method in ["POST", "PUT", "DELETE"]:
```
leonardr: I don't understand much about CRSF so it's not clear to me why GET requests don't need a token.

aslagle: cross-site request forgery is an attack where someone goes to a page that submits a form to another site, so it only applies to requests that might be modifying data. It doesn't let the attacker see the result of a request. The token is a way to ensure that the user submitted the form from your site.

This code wasn't here before because we didn't have any routes that allowed both GET and POST.

api/admin/controller.py
aslagle: I will need to change that on the client as well, so I guess it should use different form fields for each protocol.

api/admin/controller.py
```diff
@@ -815,3 +823,181 @@ def result_to_row(result):
...
+        if username:
+            collection.username = username
+        if password:
+            collection.password = password
+
+        for key, value in settings.items():
+            collection.setting(key).value = value
+        for setting in collection.settings:
+            if setting.key not in settings.keys():
+                self._db.delete(setting)
```
leonardr: I'm a little worried about this bit. I don't want the client to be able to set arbitrary settings. Each protocol has a limited number of settings that can be set, e.g. "website_id" for Overdrive. I'd prefer if only settings associated with collection.protocol could be set here, and all others had their values removed.

This might require changing core to add an inventory of allowable settings to each CirculationAPI subclass, so that it's kept in the same place as the code that implements the protocol.

Another problem with this: if you change the protocol, say from Bibliotheca to Overdrive, then a previously invalid setting becomes required. I think right now you won't be able to change the protocol from non-Overdrive to Overdrive. This needs to work somehow. Maybe you can change the protocol, put the collection into an unusable state, and then fix it by setting website_id. Maybe you can't change a protocol once you create the Collection.  

Or maybe maybe changing the protocol modifies the available fields in the interface, so that you can enter the new settings and then hit submit. Then this method would work fine, but I think you'd need a separate controller to explain what the interface should look like for the new protocol.

leonardr:
I have some concerns about what happens when you change a collection's protocol, and would like some minor wording changes, but everything else makes sense to me.

aslagle:
@leonardr I made the protocol immutable like you suggested, and added the fields for each protocol to the info that gets sent to the client.

leonardr:
:+1:

PR #473, opened 2017-03-17
Sip2 improvements

leonardr:
This branch makes miscellaneous improvements to support the Ferguson library deployment, primarily as it pertains to their SIP2 server and their Overdrive setup (which authenticates patrons using barcode only--no PIN).

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -18,6 +18,7 @@ from core.util.problem_detail import (
 )
 from core.util.opds_authentication_document import OPDSAuthenticationDocument
 from core.analytics import Analytics
+from sqlalchemy.ext.hybrid import hybrid_property
 from problem_details import *
 from util.patron import PatronUtility
 
@@ -139,8 +140,6 @@ class PatronData(object):
         self.username = username
         self.authorization_expires = authorization_expires
         self.external_type = external_type
-        if isinstance(fines, Money):
-            fines = fines.amount
         self.fines = fines
         self.block_reason = block_reason
         self.complete = complete
@@ -158,7 +157,20 @@ class PatronData(object):
             self.permanent_id, self.authorization_identifier,
             self.username
         )
-        
+
+    @hybrid_property
+    def fines(self):
+        return self._fines
+
+    @fines.setter
+    def _set_fines(self, value):
+        """When setting patron fines, only store the numeric portion of 
+        a Money object.
+        """
+        if isinstance(value, Money):
+            value = value.amount
+        self._fines = value
+    
     def apply(self, patron):
         """Take the portion of this data that can be stored in the database
         and write it to the given Patron record.
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -128,17 +128,27 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI):
         payload = dict(
             grant_type="password",
             username=patron.authorization_identifier,
-            password=pin,
             scope="websiteid:%s authorizationname:%s" % (
                 self.website_id, "default")
         )
+        if pin:
+            # A PIN was provided.
+            payload['password'] = pin
+        else:
+            # No PIN was provided. Depending on the library,
+            # this might be fine. If it's not fine, Overdrive will
+            # refuse to issue a token.
+            payload['password_required'] = 'false'
+            payload['password'] = '[ignore]'
         response = self.token_post(self.PATRON_TOKEN_ENDPOINT, payload)
         if response.status_code == 200:
             self._update_credential(credential, response.json())
         elif response.status_code == 400:
             response = response.json()
-            raise PatronAuthorizationFailedException(
-                response['error'] + "/" + response['error_description'])
+            message = response['error']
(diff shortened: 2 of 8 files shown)


api/authenticator.py
```diff
@@ -158,7 +157,20 @@ def __repr__(self):
...
+        return self._fines
+
+    @fines.setter
+    def _set_fines(self, value):
+        """When setting patron fines, only store the numeric portion of 
+        a Money object.
+        """
+        if isinstance(value, Money):
+            value = value.amount
+        self._fines = value
```
leonardr: This makes sure that the amount of the patron's fines is always set to a number. Previously it was being set to a string, which caused database problems.

api/overdrive.py
```diff
@@ -125,20 +125,35 @@ def refresh(credential):
...
         if response.status_code == 200:
             self._update_credential(credential, response.json())
         elif response.status_code == 400:
             response = response.json()
-            raise PatronAuthorizationFailedException(
-                response['error'] + "/" + response['error_description'])
+            message = response['error']
+            if 'error_description' in response:
+                message += '/' + response['error_description']
+            raise PatronAuthorizationFailedException(message)
```
leonardr: Sometimes against the Overdrive production server I would get an error message that included error but not error_description.

api/sip/__init__.py
```diff
@@ -13,7 +15,7 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
     DATE_FORMATS = ["%Y%m%d", "%Y%m%d%Z%H%M%S", "%Y%m%d    %H%M%S"]
 
     def __init__(self, server, port, login_user_id,
-                 login_password, location_code, field_separator,
+                 login_password, location_code, field_separator='|',
```
leonardr: It's almost always '|' and libraries generally don't even know it's configurable, so we shouldn't make libraries specify this parameter.

api/sip/__init__.py
```diff
@@ -86,7 +101,10 @@ def info_to_patrondata(cls, info):
         if 'personal_name' in info:
             patrondata.personal_name = info['personal_name']
         if 'fee_amount' in info:
-            patrondata.fines = info['fee_amount']
+            fines = info['fee_amount']
+        else:
+            fines = '0'
+        patrondata.fines = MoneyUtility.parse(fines)
```
leonardr: This is the code that actually makes sure that the incoming string from the SIP2 message is converted into an amount of money. The code above turns the amount of money into a number.

api/authenticator.py
daryanypl: Is there ever a difference in currency?  Like, do some libraries count in pennies, while others in dollars?

leonardr: I haven't seen an example of this and it seems unlikely. Even if it was true, this code would work so long as the maximum allowable fine was also counted in pennies.

api/overdrive.py
```diff
@@ -125,20 +125,35 @@ def refresh(credential):
             self._db, DataSource.OVERDRIVE, "OAuth Token", patron, refresh)
 
     def refresh_patron_access_token(self, credential, patron, pin):
+        """Request an OAuth bearer token that allows us to act on
+        behalf of a specific patron.
+
+        Documentation: https://developer.overdrive.com/apis/patron-auth
```
daryanypl: :)

api/sip/__init__.py
daryanypl: this is an informative comment.  would you mind adding it to the code?

api/sip/__init__.py
```diff
@@ -47,17 +49,30 @@ def __init__(self, server, port, login_user_id,
...
-            self.client = client
-        else:
-            self.client = SIPClient(
-                target_server=server, target_port=port,
-                login_user_id=login_user_id, login_password=login_password,
-                location_code=location_code, separator=field_separator
+        try:
+            if client:
+                if callable(client):
+                    client = client()
```
daryanypl: what if it's not callable?

api/sip/__init__.py
```diff
@@ -47,17 +49,30 @@ def __init__(self, server, port, login_user_id,
...
+        except IOError, e:
+            raise RemoteIntegrationException(
+                server or 'unknown server', e.message
             )
-            
+        self.client = client
+
     def remote_authenticate(self, username, password):
-        info = self.client.patron_information(username, password)
+        try:
```
daryanypl: whose username/password (the patron's or the library's)?

api/sip/__init__.py
leonardr: The docstring for this method already claims we use the pipe character by default. This just makes the code consistent with the docstring.

api/sip/__init__.py
leonardr: If it's not callable then it's the actual client object, not a client class that needs instantiation.

In most tests I pass in the actual mock client, but I need this feature to test the client that blows up when you call connect(). Since that's called during the constructor, I had to make sure that the constructor was called inside the try/catch block.

PR #476, opened 2017-03-27
Make SIP2 client handle IOErrors and socket errors 

leonardr:
Currently the SIP2 client opens a socket in its constructor and keeps that socket open forever. If the SIP2 server ever closes its socket, we never create a new one. 

In retrospect, this has obvious problems. SIP2 servers will close a socket after a certain amount of inactivity, rather than keep it open forever. Even if that wasn't the case, we need to be ready for socket connections that die due to network error.

This branch makes the smallest change I could make that has a shot at solving the problem. If an attempt to read or write to the socket fails, we create a new socket and try again. If it still fails, we reraise the exception.

I use an RLock to guarantee that only one thread is creating a socket at any given time, but I don't think this actually reduces the number of sockets that get created, so it might not be necessary. I could also use a normal Lock but RLock seems safer.

I've attached a test script I used to test the thread-safety and performance impact of this code. It creates a number of threads and has them all make SIP2 requests against a real server. Once in a while, a simulated error happens.

From my experiments with this script I think I've found that even though the client side code is thread-safe, it's not good to have two threads writing to the same socket simultaneously. This leads (I think) to garbled request messages and (definitely) timeouts on the server side.

One solution is to allow each thread of the app server to have its own `SIPClient`. A simpler solution might be to use the RLock to synchronize the entire request/response cycle, so that only one thread can be sending a request at a time. An even simpler solution would be to create a brand new socket connection for every single request. I've tested this and it doesn't seem to cause any problems on the server side.

[test_hammer_sip2_server.txt](https://github.com/NYPL-Simplified/circulation/files/873049/test_hammer_sip2_server.txt)

### DIFF
--- a/api/sip/client.py
+++ b/api/sip/client.py
@@ -238,15 +238,8 @@ class SIPClient(Constants):
     def connect(self):
         """Create a socket connection to a SIP server."""
         with self.socket_lock:
-            try:
-                sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            except socket.error, msg:
-                self.log.warn("Error initializing socket: %s", msg[1])
-
-            try:
-                sock.connect((self.target_server, self.target_port))
-            except socket.error, msg:
-                self.log.warn("Error connecting to SIP server: %s", msg)
+            sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            sock.connect((self.target_server, self.target_port))
 
             # Since this is a new socket connection, reset the message count
             # and, potentially, logged_in.
--- a/api/sip/client.py
+++ b/api/sip/client.py
@@ -191,6 +191,12 @@ class SIPClient(Constants):
         self.location_code = location_code
         self.separator = separator or '|'
 
+        # socket_lock controls access to the socket connection to the
+        # SIP2 server.
+        #
+        # We need to use an RLock here because both connect() and
+        # make_request() require the lock, and make_request() will end
+        # up calling connect() if there's an error.
         self.socket_lock = threading.RLock()
         self.connect()
 
@@ -252,10 +258,22 @@ class SIPClient(Constants):
         :param message_creator: A function that creates the message to send.
         :param parser: A function that parses the response message.
         """
+        with self.socket_lock:
+            return self._make_request(
+                message_creator, parser, *args, **kwargs
+            )
+
+    def _make_request(self, message_creator, parser, *args, **kwargs):
+        """Send a request to a SIP server and parse the response.
+        
+        :param message_creator: A function that creates the message to send.
+        :param parser: A function that parses the response message.
+        """
         if 'fail_on_network_error' in kwargs:
             fail_on_network_error = kwargs.pop('fail_on_network_error')
         else:
             fail_on_network_error = False
+
         if not self.logged_in and message_creator != self.login_message:
             # The first thing we need to do is log in.
             response = self.login(self.login_user_id, self.login_password,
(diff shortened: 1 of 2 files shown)


api/sip/client.py
```diff
@@ -230,15 +230,20 @@ def patron_information(self, *args, **kwargs):
...
-            sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-        except socket.error, msg:
-            self.log.warn("Error initializing socket: %s", msg[1])
-
-        try:
-            sock.connect((self.target_server, self.target_port))
-        except socket.error, msg:
-            self.log.warn("Error connecting to SIP server: %s", msg)
+        with self.socket_lock:
+            try:
```
leonardr: At this point it would be useful to check whether the socket was refreshed while we were waiting for `socket_lock`. Otherwise we will probably wastefully create a second socket. But I don't see how we can check that the socket was just refreshed _and is still known to be good_.

aslagle:
Using RLock to synchronize the request/response cycle or using a new socket connection for each request both seem reasonable. I'd like to know if there's any performance difference between those options.

leonardr:
Some SIP2 servers require that you send a special 'login' message immediately after creating a new socket. So for those servers it would be significantly slower to create a new socket every time.

I've got it working with the entire request-response cycle being controlled by a thread lock, and I think that's good enough for now. I looked into making it a per-thread thing, like the scoped database session, but it was way too confusing for me.

I ran my test script with 20 threads making 5 SIP2 requests each, with a sleep duration of less than a second between them. There were no deadlocks and no errors other than the random self-imposed errors.

leonardr:
My comment about servers that require login made me realize that this branch won't work for those servers. When you create a new socket connection, you reset the login status, and that wasn't happening here.

api/sip/client.py
```diff
@@ -682,10 +692,10 @@ def queue_response(self, response):
         self.responses.append(response)
 
     def connect(self):
-        # Since there is no socket, do nothing but reset the sequence
-        # number.
+        # Since there is no socket, do nothing but reset the local
+        # connection-specific variables.
         self.status.append("Creating new socket connection.")
-        self.sequence_number = 1
```
leonardr: The SIP2 spec doesn't say what number you're supposed to start the sequence number at, but I think 0 is a better choice than 1.

api/sip/client.py
```diff
@@ -587,11 +600,8 @@ def summary(self, hold_items=False, overdue_items=False,
             )
         return summary
     
-    def send(self, data, reset_sequence=False):
+    def send(self, data):
```
leonardr: reset_sequence was unused. I think it was left over from an earlier version of the code that resends a message if there's an error.

api/sip/client.py
```diff
@@ -230,37 +237,84 @@ def patron_information(self, *args, **kwargs):
...
-        except socket.error, msg:
-            self.log.warn("Error connecting to SIP server: %s", msg)
+        with self.socket_lock:
+            try:
+                sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
+            except socket.error, msg:
+                self.log.warn("Error initializing socket: %s", msg[1])
+
+            try:
+                sock.connect((self.target_server, self.target_port))
```
aslagle: Just noticed this will try to connect even if there was an error initializing the socket

leonardr: That's true. This structure is left over from sip2talk.py. I think if there's an exception at either step we should just let it happen.

PR #484, opened 2017-05-01
Fixes from 1.1.21 for master

courte:
Pairs with NYPL-Simplified/server_core#495 to incorporate some fixes from a recent deployment. Merging to master also required some test fixes.


leonardr:
:+1:

PR #485, opened 2017-05-02
Use new core

courte:
This branch goes with NYPL-Simplified/server_core#496. It fixes circulation to use the multi-tenant server_core branch, including:
- Updating CoverageProviders and Monitors to work with Collections
- Teaching a lot of controller methods that expected a single LicensePool to expect more from life and also themselves (multiple LicensePools, in other words)
- Fulfilling and revoking loans on particular LicensePools
- Using a collection for bibliographic APIs/parsers
- Removing references to "3M" in certain class and filenames
- Setting loans, holds, and fulfillment information using a collection


api/admin/controller.py
```diff
@@ -321,40 +321,51 @@ def edit(self, data_source, identifier_type, identifier):
...
             # Something went wrong.
-            return pool
-    
-        pool.suppressed = True
+            return pools
+
+        # Assume that the Work is being suppressed from the catalog, and
+        # not just the LicensePool.
+        # TODO: Suppress individual LicensePools when it's not that deep.
+        for pool in pools:
```
leonardr: I've filed https://github.com/NYPL-Simplified/server_core/issues/497 to address this.

api/axis.py
```diff
@@ -443,13 +456,17 @@ def process_one(self, e, namespaces):
             expiration_date = expiration_date.text
             expiration_date = datetime.strptime(
                 expiration_date, self.FULL_DATE_FORMAT)
-            
+
+        # WTH??? Why is identifier None? Is this ever used?
         fulfillment = FulfillmentInfo(
+            collection=None, data_source_name=DataSource.AXIS_360,
```
leonardr: It looks like you (or I) added `collection` to this class's constructor, so now we can set collection on all the *Info objects created in this class without relying on the calling code to do it.

leonardr: I know I originally wrote this code but it might be better to just change these ResponseParser subclasses to take a collection in their constructor. In the tests we can pass in `collection=None` and it should work. This would let us get rid of the code that fills in `collection` later.

api/controller.py
```diff
@@ -515,39 +559,30 @@ def sync(self):
...
-        mechanism = None
-        if mechanism_id:
-            mechanism = self.load_licensepooldelivery(pool, mechanism_id)
-            if isinstance(mechanism, ProblemDetail):
-                return mechanism
-
-        if not pool:
-            # I've never heard of this book.
+        # TODO: First make sure the patron does not already have this book
+        # on loan.
```
leonardr: This seems like an unusually important TODO. After landing this branch we should file an issue so that we don't lose track of it.

api/controller.py
```diff
@@ -619,7 +654,62 @@ def borrow(self, data_source, identifier_type, identifier, mechanism_id=None):
...
+            # Something went wrong.
+            return pools
+
+        best = None
+        mechanism = None
+        problem_doc = None
+        
+        # We found a number of LicensePools. Try to locate one that
+        # we can actually loan to the patron.
+        for pool in pools:
```
leonardr: Since we're iterating over all the pools, this would be a good place to check whether the patron already has this book out on loan (as per my previous comment). It wouldn't be too hard to take care of the TODO within this branch.

api/controller.py
```diff
@@ -634,17 +724,19 @@ def fulfill(self, data_source, identifier_type, identifier, mechanism_id=None, d
         header = self.authorization_header()
         credential = self.manager.auth.get_credential_from_header(header)
     
-        # Turn source + identifier into a LicensePool
-        pool = self.load_licensepool(data_source, identifier_type, identifier)
+        # Turn source + identifier into a set of LicensePools.
```
leonardr: I think I changed this comment and now it needs to be changed back.

api/controller.py
```diff
@@ -308,30 +305,55 @@ def load_lane(self, language_key, name):
...
 
-    def load_licensepool(self, data_source, identifier_type, identifier):
-        """Turn user input into a LicensePool object."""
-        if isinstance(data_source, DataSource):
-            source = data_source
-        else:
-            source = DataSource.lookup(self._db, data_source)
-        if source is None:
-            return INVALID_INPUT.detailed(_("No such data source: %(data_source)s", data_source=data_source))
+    def load_work(self, library, identifier_type, identifier):
```
leonardr: I know why you do this, but technically load_work() shouldn't require a Library as an argument. Not sure if it's worth changing in this branch though.

api/controller.py
```diff
@@ -1078,10 +1177,15 @@ def protocol(self):
 
 class AnalyticsController(CirculationManagerController):
 
-    def track_event(self, data_source, identifier_type, identifier, event_type):
+    def track_event(self, identifier_type, identifier, event_type):
+        # TODO: It usually doesn't matter, but there should be
+        # a way to distinguish between different LicensePools for the
+        # same book.
```
leonardr: I agree and I think it does matter, since different collections might be owned by different parties. Part of our work should be something like "add collection_id to CirculationEvent".

tests/test_coverage.py
```diff
@@ -109,53 +112,6 @@ def create_identifier_mapping(self, batch):
         eq_("Here's your title!", edition.title)
         eq_(id1, edition.primary_identifier)
 
-    def test_finalize_license_pool(self):
```
leonardr: It's been a while, but I think the code tested here has either been removed (the presentation_ready_on_success stuff) or it's tested in the new tests below.

api/admin/controller.py
```diff
@@ -321,40 +321,51 @@ def edit(self, data_source, identifier_type, identifier):
...
+            return pools
+
+        for pool in pools:
+            # Only unsuppress LicensePools that with Work-targetted,
+            # admin-editable complaints. Anything specific to a
+            # LicensePool is ignored until its specific complaints have
+            # been resolved.
+            ignore_pool = any(c.for_license_pool for c in pool.complaints)
+            if not ignore_pool:
+                pool.suppressed = False
```
aslagle: I have some concerns about this change. Unsuppressing is no longer the reverse of unsuppressing, so if an admin suppressed something accidentally it would be harder to undo. 

It might be a good idea to require admins to mark outstanding complaints resolved before unsuppressing, but if you try to resolve when a pool still has complaints you should get a problem detail. And I don't think it should be limited to license-pool-specific complaints, and we should also require the admin to add a complaint before suppressing a pool that doesn't have one yet. 

I think for now we should unsuppress all the pools regardless of whether there are complaints, and revisit this later.

api/controller.py
```diff
@@ -619,7 +651,69 @@ def borrow(self, data_source, identifier_type, identifier, mechanism_id=None):
...
+
+            # Beyond this point we have a license pool that we can
+            # actually loan or put on hold.
+
+            # But there might be many such LicensePools, and we want
+            # to pick the one that will get the book to the patron
+            # with the shortest wait.
+            if (not best
+                or pool.licenses_available > best.licenses_available
+                or pool.patrons_in_hold_queue < best.patrons_in_hold_queue):
```
aslagle: This could also take into account the total licenses owned. A pool with 10 holds but 10 licenses owned is better that a pool with 9 holds but 1 license owned.

aslagle: Also, this shouldn't be an or. If pool has fewer holds, but best still has more licenses available, best should win.

api/controller.py
aslagle: We probably need to start exposing the URL for this in OPDS. I think the mobile apps currently construct it by replacing part of the OPDS entry's URL.

api/google_analytics_provider.py
```diff
@@ -50,7 +51,8 @@ def collect_event(self, _db, license_pool, event_type, time, **kwargs):
                     'cd12': "true" if license_pool.open_access else "false",
                 })
         # urlencode doesn't like unicode strings so we convert them to utf8
-        fields = {k: unicode(v).encode('utf8') for k, v in fields.iteritems()}
+        fields = {k: unicodedata.normalize("NFKD", unicode(v)).encode("utf8") for k, v in fields.iteritems()}
```
aslagle: I'm curious what changed in this branch to require this....

leonardr: It's not related to the code in this branch. I think the problem was the tests of this method were giving different results on different systems, because `u"pi\u00F1ata"` was being converted to two different UTF-8 strings.

Courteney changed the tests in an earlier commit (324d9bbc77081868cdfa88165390c88318be51bb) which is in master already.

api/controller.py
leonardr: I filed https://github.com/NYPL-Simplified/circulation/issues/489 to deal with this.

api/controller.py
leonardr: Good point. I would like to deal with this later. I've filed https://github.com/NYPL-Simplified/circulation/issues/490.

PR #486, opened 2017-05-09
Patron has library

leonardr:
This branch ensures that AuthenticationProviders are always created in the context of a particular Library--the library whose patrons they are authenticating.

The majority of this work is changing all the AuthenticationProviders to take a library_id in their constructors, and then changing all the constructor calls in tests. We take a library_id instead of a Library because these objects are going to be instantiated in the context of a scoped session and used across requests. We can't store any data model objects inside them.

I'm a little concerned about performance implications--this might add an extra database lookup on every request, or even more than one--but data integrity is more important. We can profile the completed system and find the problems.

This is for https://github.com/NYPL-Simplified/circulation/issues/487.

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -452,11 +452,12 @@ class Authenticator(object):
         del config['module']
         provider_module = importlib.import_module(module_name)
         provider_class = getattr(provider_module, "AuthenticationProvider")
+        library = self.library(self._db)
         if issubclass(provider_class, BasicAuthenticationProvider):
-            provider = provider_class.from_config(self.library, config)
+            provider = provider_class.from_config(library, config)
             self.register_basic_auth_provider(provider)
         elif issubclass(provider_class, OAuthAuthenticationProvider):
-            provider = provider_class.from_config(self.library, config)
+            provider = provider_class.from_config(library, config)
             self.register_oauth_provider(provider)
         else:
             raise CannotLoadConfiguration(
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -678,11 +678,17 @@ class AuthenticationProvider(object):
     # different types of authentication.
     URI = None
 
-    # Each subclass MUST accept a library id in its constructor and
-    # MUST store it as .library_id. This allows the
-    # AuthenticationProvider to remember which library it manages
-    # without having to hold on to a Library object, which would be
-    # associated with a specific database session.
+    def __init__(self, library_id):
+        """Basic constructor.
+        
+        :param library_id: The database ID of the Library to be managed 
+        by this AuthenticationProvider.
+        """
+        if not isinstance(library_id, int):
+            raise Exception(
+                "Expected library_id to be an integer, got %r" % library_id
+            )
+        self.library_id = library_id
     
     def library(self, _db):
         return get_one(_db, Library, self.library_id)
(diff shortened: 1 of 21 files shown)


tests/test_controller.py
```diff
@@ -175,32 +186,28 @@ def setup(self, _db=None):
...
+        return self._default_library
 
-    def setup(self):
-        super(CirculationControllerTest, self).setup()
+    def make_default_collection(self, _db, library):
+        return self._default_collection
 
-        # TODO: This is a prime candidate for optimization. A lot of
-        # tests don't need these books, and they take over 1 second to
-        # create.
```
leonardr: It's not related, but this branch implements this TODO. On my computer, creating a Work takes about 1/3 of a second. Relatively few controller tests need more than one Work, and there are about 100 controller tests in this file and in admin/test_controller.py. So this change should reduce by an average of 1 minute the time it takes the test suite to run. I haven't verified this--runtimes on Travis are highly variable--but it certainly won't hurt.

tests/test_controller.py
```diff
@@ -1683,44 +1690,57 @@ def collection_link(entry):
             [link] = [l for l in entry['links'] if l['rel']=='collection']
             return link['title'], link['href']
 
-        # One book is in the recommendations feed.
-        [e1] = [e for e in feed['entries'] if e['title'] == self.english_2.title]
-        title, href = collection_link(e1)
-
-        eq_("Recommended Books", title)
+        # This feed contains five books: one recommended,
+        # one in the same series, and two by the same author.
```
leonardr: While fixing this test (which used to use a Work that's no longer created) I changed the way this test works in a way that hopefully makes it clearer where we are getting information about which books are in which recommendation feeds.

tests/test_controller.py
```diff
@@ -1863,6 +1883,11 @@ def test_series(self):
 
 class TestFeedController(CirculationControllerTest):
 
+    BOOKS = list(CirculationControllerTest.BOOKS) + [
+        ["english_2", "Totally American", "Uncle Sam", "eng", False],
+        ["french_1", u"Très Français", "Marianne", "fre", False],
+    ]
+
```
leonardr: These tests actually need those other two Works.

api/circulation.py
```diff
@@ -171,6 +171,11 @@ class CirculationAPI(object):
     def __init__(self, library, api_map=None):
         """Constructor.
 
+        # TODO: This needs to take a database connection as well so
+        # that we can keep the scoped session. It should only store
+        # library_id or patron_id. OR, we need to make sure we create
+        # a new CirculationAPI on each request.
+
```
leonardr: If we don't fix this, we'll have data integrity problems when we run in a multi-threaded environment.

api/authenticator.py
```diff
@@ -416,6 +433,10 @@ def __init__(self, library, basic_auth_provider=None,
                 self.oauth_providers_by_name[provider.NAME] = provider
         self.assert_ready_for_oauth()
 
+    @property
+    def library(self):
+        return get_one(self._db, Library, id=self.library_id)
```
aslagle: It doesn't look like this property gets used at all? Maybe I missed something.

aslagle: Or is that coming later?

leonardr: I think you're right, it looks like library_id is only passed on to the Authenticators. I'll rerun the full test suite to be sure.

api/authenticator.py
```diff
@@ -660,6 +681,15 @@ class AuthenticationProvider(object):
     # Authentication for OPDS document to distinguish between
     # different types of authentication.
     URI = None
+
+    # Each subclass MUST accept a library id in its constructor and
+    # MUST store it as .library_id. This allows the
+    # AuthenticationProvider to remember which library it manages
+    # without having to hold on to a Library object, which would be
+    # associated with a specific database session.
```
aslagle: Perhaps there should be a parent class constructor now.

leonardr: This is a great idea.

PR #492, opened 2017-05-22
Metadata wrangling for collections

### LINKED ISSUE #467: Circulation manager should authenticate with metadata wrangler when asking about a book
We have an authentication mechanism between the circ manager and the metadata wrangler which we don't use. Now that we're looking at a significant number of libraries using the metadata wrangler, this feature is becoming very important. It's a necessary prerequisite to creating a delta metadata API, which will stop every circ manager from constantly hammering the metadata wrangler asking for updates.

courte:
In conjunction with NYPL-Simplified/server_core#504, this branch moves away from `AdminAuthenticationService` and uses the more general `ExternalIntegration` changes instead, with a possible `type` setting to define how providers can be used.

It also adds some classes to manage remote Metadata Wrangler collections: adding to them, removing from them, getting ad-hoc updates as lookups take place.

The MetadataWranglerCoverageProvider itself, which just imports information, could probably be an IdentifierCoverageProvider at this point and avoid tangling with collections at all, except it inherits from a OPDSImporter, which is a CollectionCoverageProvider.

Fixes #467.


api/coverage.py
```diff
@@ -121,20 +125,11 @@ def create_identifier_mapping(self, batch):
         By default, no identifier mapping is needed.
         """
         return None
-    
-    def check_content_type(self, response):
```
courte: This was moved into `SimplifiedOPDSLookup` in core.

migration/20170328-2-copy-admin-authentication-into-database.py
```diff
@@ -1,43 +0,0 @@
-#!/usr/bin/env python
```
courte: Deleted this file because w'ere no longer using AdminAuthenticationService or its table. If the migration sticks around, it'll be a deadend point for the `migrate_database` command.

migration/20170522-3-move-configuration-to-external-integrations.py
```diff
@@ -0,0 +1,86 @@
...
+    if adobe_conf:
+        log_import(Configuration.ADOBE_VENDOR_ID_INTEGRATION)
+        adobe, ignore = get_one_or_create(
+            _db, ExternalIntegration, provider=ExternalIntegration.ADOBE_VENDOR_ID
+        )
+
+        adobe.username = adobe_conf.get(Configuration.ADOBE_VENDOR_ID)
+        node_value = adobe_conf.get(Configuration.ADOBE_VENDOR_ID_NODE_VALUE)
+        if node_value:
+            adobe.set_setting(u"node_value", node_value)
```
leonardr: :( but we'll get rid of this soon enough.

leonardr:
I've skimmed this and it looks solid except for the `CoverageRecord.collection_id` issue. Once that's changed I'll take a closer look.

leonardr:
:+1:

PR #494, opened 2017-05-23
Added email/password login for admins.

aslagle:
This is an initial implementation of the email/password admin login option.

There's a lot that can still be improved, but it works if you only need a few admins.

In future branches, I'd like to allow multiple admin auth services in the same circulation manager, and incorporate the password login form into the front-end interface instead of having a separate page that redirects. For both of those we would need something like the OPDS authentication document for admins. I'd also like some design input before going too much farther.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -44,6 +44,7 @@ from api.config import (
 )
 
 from oauth import GoogleAuthService
+from password_auth import PasswordAuthService
 
 from api.controller import CirculationManagerController
 from api.coverage import MetadataWranglerCoverageProvider
@@ -99,16 +100,21 @@ class AdminController(object):
     @property
     def auth(self):
         auth_service = get_one(self._db, AdminAuthenticationService)
-        if auth_service and auth_service.provider == AdminAuthenticationService.GOOGLE_OAUTH:
-            return GoogleAuthService(
-                auth_service,
-                self.url_for('google_auth_callback'),
-                test_mode=self.manager.testing,
-            )
+        if auth_service:
+            if auth_service.provider == AdminAuthenticationService.GOOGLE_OAUTH:
+                return GoogleAuthService(
+                    auth_service,
+                    self.url_for('google_auth_callback'),
+                    test_mode=self.manager.testing,
+                )
+            elif auth_service.provider == AdminAuthenticationService.LOCAL_PASSWORD:
+                return PasswordAuthService(
+                    auth_service,
+                )
         return None
 
     def authenticated_admin_from_request(self):
-        """Returns an authenticated admin or begins the Google OAuth flow"""
+        """Returns an authenticated admin or a problem detail."""
         if not self.auth:
             return ADMIN_AUTH_NOT_CONFIGURED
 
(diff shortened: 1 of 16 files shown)


api/admin/controller.py
```diff
@@ -183,25 +215,41 @@ def redirect_after_sign_in(self):
             return self.error_response(INVALID_ADMIN_CREDENTIALS)
         else:
             admin = self.authenticated_admin(admin_details)
-            flask.session["admin_email"] = admin_details.get("email")
-
-            # A permanent session expires after a fixed time, rather than
-            # when the user closes the browser.
-            flask.session.permanent = True
```
aslagle: This was moved into authenticated_admin since it's the same for both authentication mechanisms.

api/admin/controller.py
```diff
@@ -1068,6 +1116,15 @@ def admin_auth_services(self):
                         domains=json.loads(auth_service.external_integration.setting("domains").value),
                     )
                 ]
+            elif auth_service and auth_service.provider == AdminAuthenticationService.LOCAL_PASSWORD:
+                admins = self._db.query(Admin).filter(Admin.password_hashed != None)
+                auth_services = [
+                    dict(
+                        name=auth_service.name,
+                        provider=auth_service.provider,
+                        admins=[dict(email=admin.email) for admin in admins],
```
aslagle: When you edit the password service in the admin interface, you see the emails of current admins and you can can delete them from the list.

api/admin/controller.py
```diff
@@ -1101,30 +1158,54 @@ def admin_auth_services(self):
...
+            try:
+                admins = json.loads(admins)
+            except Exception:
+                self._db.rollback()
+                return INVALID_ADMIN_AUTH_ADMINS_LIST
+
+            for admin_details in admins:
+                admin, ignore = get_one_or_create(self._db, Admin, email=admin_details.get("email"))
+                if (admin_details.get("password")):
+                    admin.password = admin_details.get("password")
```
aslagle: Any newly created admins will have a password here. Admins that already existed will only have an email.

api/admin/controller.py
```diff
@@ -183,25 +215,41 @@ def redirect_after_sign_in(self):
...
+
+        if not isinstance(self.auth, PasswordAuthService):
+            return ADMIN_AUTH_MECHANISM_NOT_CONFIGURED
+
+        if flask.request.method == 'GET':
+            html = self.PASSWORD_SIGN_IN_TEMPLATE % dict(
+                password_sign_in_url=self.url_for("password_auth"),
+                redirect=flask.request.args.get("redirect"),
+            )
+            return Response(html, 200)
```
leonardr: This should include the mimetype, unless text/html is the default.

leonardr:
This basically looks good. Suggest creating an abstract base class that both `PasswordAuthService` and `GoogleAuthService` derive from. This class would define (at least) `active_credentials` and `auth_uri`.

It might also be good to rename these classes to make it clear they are for administrator auth and not patron auth, but I think being in the `admin` package will make it clear enough.

aslagle:
I tried to make the names a bit more like the patron auth code. Eventually I think there are some things that could be shared between patron and admin auth.

leonardr:
:+1:

PR #495, opened 2017-05-24
Use the SIP2 "patron_status" field to get more specific about why a patron is blocked

leonardr:
This branch comes out of some work I'm doing to understand how libraries actually use SIP2. There are some pieces that at first appeared to be weird library-specific quirks but upon closer inspection turned out to be generalizable to the SIP2 authenticator in general.

1. In general, if there are any 'Y' values in the 14-character "patron status" field, the patron should not be allowed to borrow books. The exception is the fourth field, "hold privileges denied", which only prevents the patron from putting books on hold. (We don't currently support limiting a patron's ability to put books on hold.) To manage this in an easily understandable way, I wrote code that parses the "patron status" field into a dictionary that contains 14 booleans.
2. Some SIP2 servers will send both the patron's current fine amount and their maximum fine amount in the authorization message. If a library has this type of SIP2 server, there's no need to configure the circulation manager with the maximum fine amount -- we can figure it out automatically. This branch implements the part where we figure it out automatically. We treat it as a reason why the patron is blocked.


tests/sip/test_client.py
```diff
@@ -240,3 +243,45 @@ def test_patron_password_is_optional(self):
...
+                'card reported lost',
+                #'too many items charged',
+                'too many items overdue',
+                #'too many renewals',
+                'too many claims of items returned',
+                #'too many items lost',
+                'excessive outstanding fines',
+                #'excessive outstanding fees',
+                'recall overdue',
+                #'too many items billed',
```
aslagle: why are half of these commented out?

aslagle: I guess those are the ones that aren't yes.

leonardr: Yes, it's a way to visualize the values I expect to be set to true versus the entire list of values.

leonardr:
I made a separate PR against master and merged it: https://github.com/NYPL-Simplified/circulation/pull/496

Issue #497, opened 2017-05-25
Related Works grouped feed breaks when Author name is listed twice in the feed.


jce1028:
![image](https://cloud.githubusercontent.com/assets/5263323/26459304/03096e44-4144-11e7-9e0d-9549ee3fd8ad.png)
![image](https://cloud.githubusercontent.com/assets/5263323/26459326/10d05f92-4144-11e7-80df-1309269510b1.png)

gioneill:
@leonardr has plans to update feeds to give each individual author a "related" ungrouped feed, in addition to one for the series, and one that will simply be "recommended".. I'm starting the client-side implementation this sprint, but this bug will likely be obsolete with these changes.

courte:
This was fixed in NYPL-Simplified/server_core#510 when started enforcing unique author names in the feed.

PR #499, opened 2017-05-30
Incorporate related lanes into entry #493

### LINKED ISSUE #493: Split out 'related books' feeds by type of relatedness
Currently we have a single feed for related books, and the books are grouped by how they're related to the target book (same author, same series, etc.). To improve the client-side user interface it's more useful to keep the books in separate feeds and retrieve them one at a time, as appropriate.

The new OPDS might look like this:

```
<entry>
<author>
 <link rel="related" href="https://.../author/{author ID}"/>
</author>

<schema:series name="A Series">
 <link rel="related" href="https://.../series/A%20Series"/>
</schema:series>

<link href="https://.../recommendations" type="application/atom+xml;profile=opds-catalog;kind=acquisition" rel="http://opds-spec.org/recommended" title="Recommended Works"/>
</entry>
```

I think we already build the Lanes that generate these feeds, we just need to give each Lane its own feed instead of putting them together into a grouped feed. It'll be a little tricky to find the <author> and <series> tags and insert the links, but we already insert links all over the place so it shouldn't be too bad.

Since a book can have more than one author, it's not sufficient to do this work on the client side -- it's too much work to associate the `collection` of a grouped feed with the appropriate part of the original OPDS entry. It's simpler to put the link near the text that will make it into a hyperlink.

courte:
This branch removes the `RelatedBooksLane` an its link, in preference of posting links for related Works directly in the OPDS feed for a book. It also removes a holdover in `ContributorLane` and keeps all of these related lanes from being searchable.

Fixes #493.


api/lanes.py
```diff
@@ -608,26 +529,19 @@ class ContributorLane(QueryGeneratedLane):
     ROUTE = 'contributor'
     MAX_CACHE_AGE = 48*60*60    # 48 hours
 
-    def __init__(self, _db, contributor_name, contributor_id=None,
```
courte: `contributor_id` isn't used on the endpoint, so it's not possible to use it for the lane.

tests/test_controller.py
```diff
@@ -181,7 +183,7 @@ def setup(self):
 
         # Create two English books and a French book.
         self.english_1 = self._work(
-            "Quite British", "John Bull", language="eng", fiction=True,
+            "Quite British", "Bull, John", language="eng", fiction=True,
```
courte: This didn't end up being the source of my test problem, but we're passing in the sort name now at this point, not the display name.

tests/test_opds.py
```diff
@@ -199,24 +203,6 @@ def test_lane_url(self):
         feed_url = self.annotator.lane_url(fantasy_lane_without_sublanes)
         eq_(feed_url, self.annotator.feed_url(fantasy_lane_without_sublanes))
 
-    def test_single_entry_no_active_license_pool(self):
```
courte: We're not giving a pool to AcquisitionFeed.single_entry at all anymore and the signature has changed since this test was created, so I removed it.

tests/test_opds.py
```diff
@@ -327,206 +313,251 @@ def test_adobe_id_tags_when_vendor_id_configured(self):
...
+
 class TestOPDS(WithVendorIDTest):
 
     def setup(self):
         super(TestOPDS, self).setup()
         parent = Lane(self._db, "Fiction", languages=["eng"], fiction=True)
-        fantasy_lane = Lane(self._db, "Fantasy", languages=["eng"], genres=[Fantasy], parent=parent)
-        self.lane = fantasy_lane
+        self.lane = Lane(self._db, "Fantasy", languages=["eng"], genres=[Fantasy], parent=parent)
+        self.annotator = CirculationManagerAnnotator(None, self.lane, test_mode=True)
```
courte: I refactored many of the tests below to use this annotator instead of creating one.

courte:
Also, I'm not sure if this will change #497 because I'm not sure why that error is being raised. I'd like to make the change and then check afterwards for the error @jce1028 listed there.

api/lanes.py
leonardr: Is it good to start using `contributor_id` now? Or should we wait until it's more likely that two contributors with the same name are different people? I feel like right now there are a lot of duplicate contributors, and searching by name is more reliable than searching by ID.

api/opds.py
```diff
@@ -286,14 +300,67 @@ def annotate_work_entry(self, work, active_license_pool, edition, identifier, fe
...
+        audience_key=None
+        if audiences:
+            audience_strings = [urllib.quote_plus(a) for a in sorted(audiences)]
+            audience_key = u','.join(audience_strings)
+
+        return language_key, audience_key
+
+    def add_author_links(self, work, feed, entry):
+        author_tag = '{%s}author' % OPDSFeed.ATOM_NS
+        author_entries = entry.findall(author_tag)
```
leonardr: As per my suggestion in your core branch, you would also need to do this for the atom:contributor tag.

tests/test_opds.py
```diff
@@ -327,206 +313,251 @@ def test_adobe_id_tags_when_vendor_id_configured(self):
...
-        [entry] = feed['entries']
-        [issues_link] = [x for x in entry['links'] if x['rel'] == 'issues']
-        assert '/report' in issues_link['href']
-
-    def test_acquisition_feed_includes_open_access_or_borrow_link(self):
-        w1 = self._work(with_open_access_download=True)
-        w2 = self._work(with_open_access_download=True)
-        w2.license_pools[0].open_access = False
-        w2.license_pools[0].licenses_owned = 1
-        self._db.commit()
```
leonardr: I don't understand why these two tests were removed.

leonardr:
This looks good, except, James has requested that we _also_ keep the "related works" feed and just display it differently on the client side. It shouldn't be too difficult to restore that code and also keep the author and series links.

api/lanes.py
courte: I agree, and I'd prefer to improve our Contributor data before we start treating identifiers as actually unique people. I'd like to leave it like this for the time being, but we can consider fixing it if we see problems.

leonardr:
Looks good :+1:

PR #500, opened 2017-05-31
Bring circulation up to date with recent core changes

leonardr:
This branch changes circulation to respect the merging of Collection.protocol and ExternalIntegration.provider into Collection.provider.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1074,23 +1074,23 @@ class SettingsController(CirculationManagerController):
                 providers=ExternalIntegration.ADMIN_AUTH_PROVIDERS,
             )
 
-        provider = flask.request.form.get("provider")
-        if not provider:
+        protocol = flask.request.form.get("provider")
+        if not protocol:
             return NO_PROVIDER_FOR_NEW_ADMIN_AUTH_SERVICE
 
-        if provider not in ExternalIntegration.ADMIN_AUTH_PROVIDERS:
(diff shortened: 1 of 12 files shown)


tests/test_config.py
```diff
@@ -32,6 +32,7 @@ def test_from_config(self):
             eq_(Configuration.DEFAULT_FACET, facet_config._default_facets)
 
         with temp_config() as config:
+            config[Configuration.POLICIES] = {}
```
leonardr: Unrelated change that has been annoying me for a while. I run tests with an extremely minimal configuration that doesn't define POLICIES at all, which is legal but which broke this test.

api/admin/controller.py
```diff
@@ -1071,26 +1072,26 @@ def admin_auth_services(self):
 
             return dict(
                 admin_auth_services=auth_services,
-                providers=ExternalIntegration.ADMIN_AUTH_PROVIDERS,
+                providers=ExternalIntegration.ADMIN_AUTH_PROTOCOLS,
             )
 
-        provider = flask.request.form.get("provider")
-        if not provider:
+        protocol = flask.request.form.get("provider")
```
leonardr: I didn't change `provider` to `protocol` in the web app because I think it makes sense to continue to call this a "provider" in user-facing applications. I can change it though if we want to be consistent.

aslagle: That's fine. This doesn't affect what's displayed to users, but if we did want to change it we'd also have to update the form in circulation-web.

PR #501, opened 2017-05-31
Library short name in urls

aslagle:
This branch adds the library short name to all urls that are library-specific. It works with either a subdomain prefix or a url path prefix, and if there's no library in the url, it can use a default library which is currently Library.instance(). This is done with a few new decorators in api/routes. Library.short_name is no longer nullable since it's used to create urls.

Most of the changes are adding the library short name to calls to url_for, and setting flask.request.library in tests. Maybe there's a better way to handle that so it doesn't need to happen in so many places.

The admin interface will need a new design for multiple libraries, but for now it shows you the default library.

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -443,6 +443,7 @@ class LibraryAuthenticator(object):
         self.library_id = library.id
         self.library_uuid = library.uuid
         self.library_name = library.name
+        self.library_short_name = library.short_name
         self.basic_auth_provider = basic_auth_provider
         self.oauth_providers_by_name = dict()
         self.bearer_token_signing_secret = bearer_token_signing_secret
@@ -652,7 +653,7 @@ class LibraryAuthenticator(object):
 
         library_name = self.library_name or unicode(_("Library"))
         doc = OPDSAuthenticationDocument.fill_in(
-            base_opds_document, list(self.providers), self._db,
+            base_opds_document, list(self.providers), short_name=self.library_short_name,
             name=library_name, id=self.library_uuid, links=links,
         )
         return json.dumps(doc)
@@ -782,7 +783,7 @@ class AuthenticationProvider(object):
             patron_or_patrondata
         )       
     
-    def authentication_provider_document(self, _db):
+    def authentication_provider_document(self, library_short_name):
         """Create a stanza for use in an Authentication for OPDS document.
 
         :return: A dictionary that can be associated with the
@@ -1084,7 +1085,7 @@ class BasicAuthenticationProvider(AuthenticationProvider):
     def authentication_header(self):
         return 'Basic realm="%s"' % self.AUTHENTICATION_REALM
     
-    def authentication_provider_document(self, _db):
+    def authentication_provider_document(self, library_short_name):
         """Create a stanza for use in an Authentication for OPDS document.
 
         Example:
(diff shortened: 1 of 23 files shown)


api/authenticator.py
```diff
@@ -639,7 +658,7 @@ def create_authentication_document(self):
 
         library_name = self.library_name or unicode(_("Library"))
         doc = OPDSAuthenticationDocument.fill_in(
-            base_opds_document, list(self.providers),
+            base_opds_document, list(self.providers), self._db,
```
aslagle: I don't like that I had to do this. The database is needed to look up the library to get its short name when creating urls.

api/base_controller.py
```diff
@@ -85,9 +75,21 @@ def authenticated_patron(self, authorization_header):
...
+        library_short_name = flask.request.library.short_name
+        if library_short_name not in self.manager.opds_authentication_documents:
+            self.manager.opds_authentication_documents[library_short_name] = self.manager.auth.create_authentication_document()
 
-        data = self.manager.opds_authentication_document
+        data = self.manager.opds_authentication_documents[library_short_name]
         headers = self.manager.auth.create_authentication_headers()
         return Response(data, 401, headers)
+
+    def library_from_request(self, library_short_name):
```
aslagle: Maybe I should rename this to library_from_short_name. It's not much like authenticated_patron_from_request.

api/authenticator.py
```diff
@@ -335,23 +335,48 @@ def set_authorization_identifier(self, authorization_identifier):
...
+        if short_name not in self.library_authenticators:
+            return LIBRARY_NOT_FOUND
+        return self.library_authenticators[short_name].create_authentication_headers()
+
+    def get_credential_from_header(self, header):
+        short_name = flask.request.library.short_name
+        if short_name not in self.library_authenticators:
+            return LIBRARY_NOT_FOUND
+        return self.library_authenticators[short_name].get_credential_from_header(header)
+
```
leonardr: I don't know if it's worth the effort, but you could remove a lot of this code with a helper method like this:

```
def invoke_authenticator_method(self, method_name, *args, **kwargs):
    short_name = flask.request.library.short_name
    if short_name not in self.library_authenticators:
        return LIBRARY_NOT_FOUND
    return getattr(self.library_authenticators[short_name], method_name)(*args, **kwargs)
```

api/authenticator.py
leonardr: Would it be better if you passed in the Library object or its short name directly?

api/base_controller.py
leonardr: I agree it's not an accurate name but since it modifies `flask.request` I like keeping 'request' in the name. Maybe library_for_request or add_library_to_request?

api/controller.py
```diff
@@ -478,8 +485,10 @@ def search(self, languages, lane_name):
         if isinstance(lane, ProblemDetail):
             return lane
         query = flask.request.args.get('q')
+        library_short_name = flask.request.library.short_name
         this_url = self.url_for(
             'lane_search', languages=languages, lane_name=lane_name,
+            library_short_name=library_short_name,
```
leonardr: I feel like there ought to be a way to make self.url_for automatically fill in `library_short_name=flask.request.library.short_name` if it's not provided. Most of the time, the library in the link will be the same library used in the request.

api/problem_details.py
```diff
@@ -253,3 +253,10 @@
     title=_("No annotation."),
     detail=_("The annotation you requested does not exist."),
 )
+
+LIBRARY_NOT_FOUND = pd(
+    "http://librarysimplified.org/terms/problem/library-not-found",
+    status_code=404,
+    title=_("Library not found."),
+    detail=_("The requested library short name does not exist."),
```
leonardr: I would say "No library with the requested name on this server."

tests/admin/test_controller.py
```diff
@@ -63,6 +63,7 @@ def test_details(self):
 
         lp.suppressed = False
         with self.app.test_request_context("/"):
+            flask.request.library = self._default_library
```
leonardr: We can get rid of all this code by subclassing test_request_context to set flask.request.library automatically.

leonardr:
In Slack I suggested making `text_request_context` set `flask.request.library = self._default_library`. This would let us get rid of nearly all the explicit sets of `flask.request.library`.

After that, I think this is ready to land. There's other work I'd like to do to reduce the number of times we have to explicitly think about the current library, but I think it's more important to continue to make progress towards the functionality we want.

leonardr:
For the record, some specific (difficult) things we could do:

* Have @library_route also call @has_library, so it doesn't have to be called explicitly
* Create a Controller object per request and give it the current Library to carry around as context.
* Then give Controller a `url_for` method that passes in the name of the current library so it can be used to generate appropriate URLs.

leonardr:
:+1:

PR #504, opened 2017-06-02
Merge master into multi tenancy

aslagle:
Most of this branch is changes that were already reviewed when they were merged to master, but I made a significant change to admin authentication to make it more compatible with the removal of the adminauthenticationservices table in favor of external integrations with an "admin_auth" goal. 

Managing admins that can log in with a password is no longer done via the "admin_auth_services" endpoint. I created a new "individual_admins" endpoint that creates or edits individual admins, one at a time. I think this is better anyway. 

If you have an google oauth external integration with the "admin_auth" goal, that is used to authenticate, and if you don't, passwords in the admins table are used. When you start a new circulation manager and open the admin interface, you get a page with two forms, one to create an admin auth service and one to create a single admin. If you choose to create a single admin, you can create additional admins or add an admin auth service after you log in as that admin. Once you add an admin auth service, the individual admin logins won't be usable anymore - you'll always get sent to google to log in. A future branch should enable having multiple auth services and password auth work at the same time - we might want something like the opds authentication document for admins. 

Another thing that needs to be done in a future branch is allow deleting admins, admin auth services, collections, and maybe libraries. You could delete admins in the original implementation of admin passwords, and now you can change their passwords but you can't delete them from the admin interface.


leonardr:
:+1:

PR #505, opened 2017-06-05
Library link configuration in database

aslagle:


api/opds.py
```diff
@@ -437,17 +452,12 @@ def annotate_feed(self, feed, lane):
...
-                ("privacy-policy", Configuration.privacy_policy_url()),
-                ("copyright", Configuration.acknowledgements_url()),
-                ("about", Configuration.about_url()),
-                ("license", Configuration.license_url()),
-        ):
-            if value:
-                d = dict(href=value, type="text/html", rel=rel)
+    def add_configuration_links(self, feed):
+        for rel in self.CONFIGURATION_LINKS:
+            _db = Session.object_session(self.library)
```
leonardr: Call this outside the for loop.

migration/20170605-move-configuration-links-into-db.py
```diff
@@ -0,0 +1,36 @@
...
+_db = production_session()
+library = Library.instance(_db)
+
+for rel, value in (
+        ("terms-of-service", Configuration.terms_of_service_url()),
+        ("privacy-policy", Configuration.privacy_policy_url()),
+        ("copyright", Configuration.acknowledgements_url()),
+        ("about", Configuration.about_url()),
+        ("license", Configuration.license_url()),
+):
```
leonardr: If you use the code these methods use, you can remove the methods themselves in this branch; otherwise we need to wait until everyone is migrated.

leonardr:
I have a couple suggestions but this is :+1: as is.

Looking at this code, I realize I was implicitly assuming that we would start allowing administrators to paste in the actual HTML they wanted to use as "about", etc., and then serve that HTML from special controllers. I still think we should allow that as an option, but it's a much bigger task than just moving the links out of the configuration, so let's not do it now.

aslagle:
I updated the migration and removed the old config methods in https://github.com/NYPL-Simplified/server_core/pull/516, and also fixed another place that used them.

PR #506, opened 2017-06-05
Millennium last name auth

datalogics-dans:
These changes will allow Millennium/III authentication using a patron's last name instead of PIN. This is specifically to accommodate one specific library's ILS implementation, but might end up being useful to others in the future.


api/millenium_patron.py
```diff
@@ -78,15 +86,26 @@ def remote_authenticate(self, username, password):
...
+                return PatronData(authorization_identifier=username, complete=False)
+            return False
+        elif self.auth_mode == "last_name":
+            path = "%(barcode)s/dump" % dict(barcode=username)
+            url = self.root + path
+            response = self.request(url)
+            data = dict(self._extract_text_nodes(response.content))
+            dump_name = data.get(self.PERSONAL_NAME_FIELD)
+            dump_name = dump_name.split(',')[0]
+            if dump_name == password:
```
leonardr: This needs to be a case-insensitive comparison, because the names are stored in uppercase and most people will enter their name in lowercase.

api/millenium_patron.py
```diff
@@ -78,15 +86,26 @@ def remote_authenticate(self, username, password):
...
+            path = "%(barcode)s/%(pin)s/pintest" % dict(
+                barcode=username, pin=password
+            )
+            url = self.root + path
+            response = self.request(url)
+            data = dict(self._extract_text_nodes(response.content))
+            if data.get('RETCOD') == '0':
+                return PatronData(authorization_identifier=username, complete=False)
+            return False
+        elif self.auth_mode == "last_name":
```
leonardr: Call this `family_name`, a la http://schema.org/familyName.

tests/test_millenium_patron.py
```diff
@@ -380,3 +380,22 @@ def test_verify_certificate(self):
...
+    def test_authorization_last_name_success(self):
+        """Test authenticating against the patron's last name, given the
+        correct name
+        """
+        self.api = MockAPI(auth_mode = "last_name")
+        self.api.enqueue("dump.success.html")
+        patrondata = self.api.remote_authenticate(
+            "44444444444447", "SHELDON"
+        )
+        eq_("44444444444447", patrondata.authorization_identifier)
```
leonardr: This is where to test that the comparison is case-insensitive.

PR #507, opened 2017-06-06
Patron auth mechanisms are externalintegrations

leonardr:
This branch stops looking in the JSON config for patron auth mechanisms and starts looking for ExternalIntegration objects with goal="patron_auth".

Existing auth mechanisms are inserted into the database with a migration script.


api/authenticator.py
```diff
@@ -369,53 +372,33 @@ class LibraryAuthenticator(object):
...
-        if (not isinstance(authentication_policy, dict)
-            or not 'providers' in authentication_policy):
-            raise CannotLoadConfiguration(
-                "Authentication policy must be a dictionary with key 'providers'."
-            )
-        bearer_token_signing_secret = authentication_policy.get(
-            'bearer_token_signing_secret'
+
+        bearer_token_signing_secret = OAuthAuthenticationProvider.bearer_token_signing_secret(
+            _db
```
leonardr: The bearer token signing secret is a site-wide configuration setting (currently the only one).

api/authenticator.py
```diff
@@ -369,53 +372,33 @@ class LibraryAuthenticator(object):
...
-        if (not authenticator.basic_auth_provider
-            and not authenticator.oauth_providers_by_name):
-            # TODO: This isn't unacceptable: a fully open-access
-            # collection doesn't need any authentication providers.
-            # But supporting that case requires specialized work, e.g.
-            # getting rid of all the links to controllers that require
-            # authentication.
-            raise CannotLoadConfiguration(
-                "No authentication provider configured"
-            )
```
leonardr: I had to take this code out, even though we still have those links to controllers, because when you've just created a library, there's going to be a period where your library has no authentication providers, and we need to treat that as a valid configuration, if only so the admin can use the interface to add a provider.

api/authenticator.py
```diff
@@ -1150,22 +1174,43 @@ class OAuthAuthenticationProvider(AuthenticationProvider):
...
+    BEARER_TOKEN_SIGNING_SECRET = "bearer_token_signing_secret"
+
+    @classmethod
+    def bearer_token_signing_secret(cls, _db):
+        """Find or generate the site-wide bearer token signing secret."""
+        secret = ConfigurationSetting.sitewide(
+            _db, cls.BEARER_TOKEN_SIGNING_SECRET
+        )
+        if not secret.value:
+            secret.value = os.urandom(24).encode('hex')
```
leonardr: This value comes from personal communication with Amy.

api/mock_authentication.py
```diff
@@ -1,77 +0,0 @@
-from nose.tools import set_trace
```
leonardr: I was able to get rid of this whole thing. Tests now use the new `api.simple_authentication` module which is simpler and doesn't have a name conflict with a mock class defined in `test_authenticator.py`.

This module doesn't work in the new system because it really needs to be configured with complicated JSON objects, and that doesn't fit with the web-based configuration we're going towards.

Many of the features in this module were overkill anyway and I've found simpler ways to test the same functionality. See `test_authenticated_patron_can_authenticate_with_expired_credentials` for an example.

api/simple_authentication.py
```diff
@@ -0,0 +1,62 @@
+from nose.tools import set_trace
```
leonardr: This is the replacement for mock_authentication.py. It can only authenticate a single patron -- the test patron. This can be used in pre-production to make sure your circulation manager is working without having to hook it up to your ILS.

tests/files/millenium_patron/dump.two_barcodes.html
```diff
@@ -33,8 +33,8 @@
 NOTICE PREF[p268]=z<BR>
 PATRN NAME[pn]=SHELDON, ALICE<BR>
 ADDRESS[pa]=111 MAIN STREET$NEW YORK, NY 10001<BR>
-P BARCODE[pb]=FIRST_barcode<BR>
-P BARCODE[pb]=SECOND_barcode<BR>
```
leonardr: These barcodes don't pass the default identifier regexp.

tests/sip/test_authentication_provider.py
```diff
@@ -34,12 +34,37 @@ class TestSIP2AuthenticationProvider(DatabaseTest):
     polaris_excess_fines = "64YYYY      Y   00120161121    144438000000000000000000000000AO3|AA25891000115879|AEFalk, Jen|BZ0050|CA0075|CB0075|BLY|CQY|BHUSD|BV11.50|CC9.99|BD123, Charlotte Hall, MD 20622|BE|BF501-555-1212|BC20140610    000000|PA1|PEHALL|PS|U1No|U2|U3|U4|U5|PZ20622|PX20170424    235959|PYN|FA0.00|AFPatron has blocks.|AGPatron has blocks.|AY2AZA27B"
 
     polaris_no_such_patron = "64YYYY          00120161121    143126000000000000000000000000AO3|AA1112|AE, |BZ0000|CA0000|CB0000|BLN|CQN|BHUSD|BV0.00|CC0.00|BD|BE|BF|BC|PA0|PE|PS|U1|U2|U3|U4|U5|PZ|PX|PYN|FA0.00|AFPatron does not exist.|AGPatron does not exist.|AY2AZBCF2"
-    
+
+    def test_initialize_from_integration(self):
```
leonardr: We never explicitly tested this before, because initializing a SIP2AuthenticationProvider would make it try to connect to the server.

tests/sip/test_authentication_provider.py
```diff
@@ -34,12 +34,37 @@ class TestSIP2AuthenticationProvider(DatabaseTest):
...
+        p = SIP2AuthenticationProvider
+        integration = self._external_integration(self._str)
+        integration.url = "server.com"
+        integration.username = "user1"
+        integration.password = "pass1"
+        integration.setting(p.FIELD_SEPARATOR).value = "\t"
+
+        provider = p(self._default_library.id, integration, connect=False)
```
leonardr: `connect=False` turns off "connect to the server automatically".

leonardr:
Here's the configuration I used to test the migration script. All this information should be read back to you from the database at the end of the script.

```
{
   "integrations" : {
      "Postgres" : {
         "production_url" : "postgres://simplified:[password]@localhost:5432/simplified_circulation_dev"
      }
   },
   "policies" : {
      "authentication" : {
	  "bearer_token_signing_secret" : "abcdefg",
          "providers" : [
            {
               "test_username" : "1234",
               "test_password" : "pass",
               "module" : "api.millenium_patron",
               "url" : "https://foo/PATRONAPI/",
               "authorization_identifier_blacklist" : [
                  "lost",
                  "inactive"
               ]
            },
            {
               "login_password" : "pass",
               "location_code" : "main",
               "port" : 2345,
               "field_separator" : "|",
               "server" : "server.com",
               "module" : "api.sip",
		"login_user_id" : "userid",
  	       "identifier_regular_expression" : "abcd",
	       "password_regular_expression" : "efgh"
            },
            {
               "test_password" : "firstbookpass",
               "module" : "api.firstbook",
               "key" : "akey",
               "url" : "https://foo/",
               "test_username" : "firstbookuser"
            },
            {
               "module" : "api.clever",
               "client_secret" : "cleverclientsecret",
               "client_id" : "cleverclientid"
            }
         ]
      }
   }
}
```

migration/20170606-1-move-authentication-configuration-to-external-integrations.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env python
```
courte: This filename doesn't need the `1-` on it. It will still run the same, but the counter isn't required.

leonardr: I was a little optimistic and thought I might get more than one of these done yesterday...

PR #508, opened 2017-06-07
Merge master

leonardr:
This branch brings the circulation multi-tenancy branch up to date with changes in master.


api/config.py
```diff
@@ -43,7 +43,9 @@ class Configuration(CoreConfiguration):
     OAUTH_TOKEN_EXPIRATION_DAYS = 'token_expiration_days'
     SECRET_KEY = "secret_key"
 
-    MILLENIUM_INTEGRATION = u"Millenium"
+    MILLENIUM_INTEGRATION = "Millenium"
```
courte: Are these two settings still required? The only places I see them used are here and -- for `MILLENIUM_INTEGRATION` in the integration_tests/__init__.py setup (which could be switched out for an external integration.

leonardr:
This branch has effectively already been merged.

PR #509, opened 2017-06-08
Added a message for when setting up an individual admin fails because pgcrypto is not installed.

aslagle:


aslagle:
I can't write a test for this since the test database needs pgcrypto to run other tests.

leonardr:
:+1:

I have a branch that detects the problem on startup and issues a warning via logging.error.

PR #510, opened 2017-06-08
Move (some) configuration settings from the JSON configuration and into the database

leonardr:
The focus of this branch is on configuration settings that are sitewide or library-specific. Courteney is handling settings that are related to configuring some external piece of software.

* The SECRET_KEY used to sign admin cookies has become a site-wide secret, just like the bearer token signing secret.
* INCLUDE_ADMIN_INTERFACE has been removed. The admin interface has become so important to site configuration that it's now required.
* I removed the preloaded content feature. We ended up not using it and it's easier to remove it than to convert its configuration.
* I removed the 'force language' setting, which as far as I can tell is not used.
* The default email address to use for patron book availability notifications is now a per-library setting.
* The maximum amount of fines before a patron loses lending privileges is now a per-library setting.
* The regular expression used to derive `Patron.external_integration` from `Patron.authorization_identifier` is now a per-library per-integration setting.

Since instantiating an AuthenticationProvider now requires both a Library and an ExternalIntegration, I changed the constructor to take a Library instead of a Library_id. This meant a lot of tiny changes to tests.

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -431,6 +431,7 @@ class LibraryAuthenticator(object):
         self.library_uuid = library.uuid
         self.library_name = library.name
         self.library_short_name = library.short_name
+
         self.basic_auth_provider = basic_auth_provider
         self.oauth_providers_by_name = dict()
         self.bearer_token_signing_secret = bearer_token_signing_secret
@@ -482,7 +483,7 @@ class LibraryAuthenticator(object):
             raise CannotLoadConfiguration(
                 "Loaded module %s but could not find a class called AuthenticationProvider inside." % module_name
             )
-        provider = provider_class.from_config(self.library_id, integration)
+        provider = provider_class(self.library, integration)
         if issubclass(provider_class, BasicAuthenticationProvider):
             self.register_basic_auth_provider(provider)
             # TODO: Run a self-test, or at least check that we have
@@ -579,7 +580,7 @@ class LibraryAuthenticator(object):
                 break
             
         return credential
-    
+
     def oauth_provider_lookup(self, provider_name):
         """Look up the OAuthAuthenticationProvider with the given name. If that
         doesn't work, return an appropriate ProblemDetail.
@@ -623,14 +624,14 @@ class LibraryAuthenticator(object):
         return jwt.encode(
             payload, self.bearer_token_signing_secret, algorithm='HS256'
         )
-    
+
     def decode_bearer_token_from_header(self, header):
         """Extract auth provider name and access token from an Authenticate
         header value.
         """
         simplified_token = header.split(' ')[1]
         return self.decode_bearer_token(simplified_token)
-    
+
     def decode_bearer_token(self, token):
         """Extract auth provider name and access token from JSON web token."""
         decoded = jwt.decode(token, self.bearer_token_signing_secret,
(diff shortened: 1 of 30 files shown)


api/authenticator.py
```diff
@@ -733,6 +768,36 @@ def update_patron_metadata(self, patron):
...
+            # regular expression at all. Leave their external_type
+            # alone.
+            return
+        groups = match.groups()
+        if not groups:
+            # The regular expression matched but didn't contain any groups.
+            # This is a configuration error; do nothing.
+            return
+            
+        patron.external_type = groups[0]
```
aslagle: Is it true that there doesn't need to be a migration to create external_type for patrons that were using the regexp before because this code will create it the next time it's needed?

aslagle:
The oneclick settings need to be removed from the admin collections controller method.

aslagle:
Looks good, I just had one question and a minor comment.

api/authenticator.py
leonardr: Yes, we don't need to set the value on patrons with a script. In fact we can't, because that information is kept in the ILS. However this makes me realize that I do need a migration script to move the regex, maximum fine level, and default email address from the config into the database.

leonardr:
Here's the config I used to test the migration script I just added:
```
{
   "integrations" : {
      "Postgres" : {
         "test_url" : "postgres://simplified_test:test@localhost:5432/simplified_circulation_test",
         "production_url" : "postgres://simplified:[password]@localhost:5432/simplified_circulation_dev"
      }
   },
   "secret_key": "abcde",
   "default_notification_email_address" : "foo@example.com",
   "policies" : {
       "external_type_regular_expression" : "^([HMEA])",
       "max_outstanding_fines" : "$15.00"
   }
}
```

Here's what it put into the `configurationsettings` table when I ran the script:

```
 id | external_integration_id | library_id |                key                 |      value      
----+-------------------------+------------+------------------------------------+-----------------
 21 |                         |            | secret_key                         | abcde
 22 |                         |          1 | default_notification_email_address | foo@example.com
 23 |                         |          1 | max_outstanding_fines              | $15.00
 24 |                       7 |          1 | external_type_regular_expression   | ^([HMEA])
```

migration/20170609-1-move-library-configuration-to-configurationsettings.py
```diff
@@ -0,0 +1,152 @@
...
+    # are pulled from the ExternalIntegration.
+    integration.url = provider.get('url')
+    auth_mode = provider.get('auth_mode')
+    blacklist = provider.get('authorization_identifier_blacklist')
+    if blacklist:
+        integration.setting(MilleniumPatronAPI.IDENTIFIER_BLACKLIST
+        ).value = json.dumps(blacklist)
+    if auth_mode:
+        integration.setting(MilleniumPatronAPI.AUTHENTICATION_MODE
+        ).value = auth_mode
```
aslagle: you're missing verify certificate. is that still needed?

leonardr: That whole thing shouldn't be in the file. I copied the old migration script, modified the end, and didn't change the beginning.

I think it's okay to omit verify_certificate, since it's only supposed to be used on a temporary basis until III fixes your ILS.

PR #512, opened 2017-06-12
Lanes are per-library

leonardr:
This branch applies the logic of https://github.com/NYPL-Simplified/server_core/pull/524 to the circulation manager.

Creating a lane now requires a library context.

Lots of configuration settings have been moved into the database.


api/config.py
```diff
@@ -112,40 +113,3 @@ def temp_config(new_config=None, replacement_classes=None):
         all_replacement_classes.extend(replacement_classes)
     with core_temp_config(new_config, all_replacement_classes) as i:
         yield i
-
-class FacetConfig(object):
```
leonardr: This class and its tests were moved to core.

tests/test_controller.py
```diff
@@ -461,32 +461,27 @@ def test_load_licensepooldelivery(self):
     def test_apply_borrowing_policy_when_holds_prohibited(self):
         with self.request_context_with_library("/"):
             patron = self.controller.authenticated_patron(self.valid_credentials)
-            with temp_config() as config:
```
leonardr: A lot of the changes to test code are indentation changes caused by the removal of a `with temp_config()` block.

PR #513, opened 2017-06-13
Added admin editing of patron auth services.

aslagle:


api/millenium_patron.py
```diff
@@ -63,6 +64,18 @@ class MilleniumPatronAPI(BasicAuthenticationProvider, XMLParser):
     AUTHENTICATION_MODES = [
         PIN_AUTHENTICATION_MODE, FAMILY_NAME_AUTHENTICATION_MODE
     ]
+
+    SETTINGS = [
+        { "key": VERIFY_CERTIFICATE, "label": _("Verify Certificate"), "optional": True },
```
aslagle: I'd like to remove this setting, or change it to "Don't verify certificate", so it makes more sense if you leave it blank.

leonardr: We can call it "Ignore certificate problems".

leonardr:
:+1:

api/millenium_patron.py
aslagle: I left it the same internally but changed the labels so it will hopefully be more clear in the admin interface.

leonardr: Yes, it looks good.

PR #514, opened 2017-06-13
Use external integrations

courte:
Supported by NYPL-Simplified/server_core#526, this branch imports third-party integrations from the configuration files and uses them wherever `Configuration` once roamed. Instead of setting ~EternalIntegrations~ ExternalIntegrations for "Circulation Manager" and "Patron Web Client" urls, they are each given a ConfigurationSetting `base_url` and `patron_web_client_url`, respectively.

There are a lot of changes present around the Adobe Vendor ID integration, a portion of the code base I'm not super-familiar with, so some extra 👀 there would be very helpful.

Lastly, many of the changes in `tests/` are largely indentation changes and slight signature changes (especially after `test_adobe_vendor_id.py`), but unfortunately not all.


migration/20170614-2-move-third-party-config-to-external-integrations.py
```diff
@@ -0,0 +1,140 @@
...
+        password = metadata_wrangler_conf.get('client_secret')
+
+        integration, is_new = get_one_or_create(
+            _db, EI, protocol=EI.METADATA_WRANGLER, goal=EI.METADATA_GOAL,
+            url=url, username=username, password=password
+        )
+        log_import(integration, is_new)
+
+    # Import Content Server configuration.
+    content_server_conf = Configuration.integration('Content Server')
```
courte: I'm not sure where the Content Server falls as an ExternalIntegration, since I think--best-case scenario--it's only used as a license source for certain Collections. I don't see us actually using it for metadata anywhere on circulation, despite the ContentServerCoverageProvider in core.

api/novelist.py
```diff
@@ -19,17 +19,21 @@
...
     Representation,
+    Session,
     Subject,
 )
 from core.util import TitleProcessor
 
 class NoveListAPI(object):
 
     IS_CONFIGURED = None
+    __library = None
```
leonardr: I don't see that you store a `Library` here, which means I'm not sure what you do with this in `is_configured` below. If you do need to keep track of which library is associated with these credentials (and I don't think you do), store the library ID and look it up when necessary. If you store the model object itself we'll run into problems with the scoped session.

migration/20170614-2-move-third-party-config-to-external-integrations.py
leonardr: I agree, let's remove it as a separate configuration setting.

leonardr:
This needs one big change: the Adobe Vendor ID config should be two different integrations.

One is the integration between a library and the library
registry. Every library has one of these integrations. It includes the
following information:

* The vendor ID of the library registry, issued by Adobe. This is
  "vendor_id" in JSON config.
* The short name used to identify this library with the registry. In
  master this is "library_short_name" in JSON config. In multi-tenant
  this is Library.library_registry_short_name.
* The shared secret with the library registry. In master this is
  "authdata_secret" in JSON config. In multi-tenant this is
  Library.library_registry_shared_secret.
* A URI identifying the library to external sources. This is
  "library_uri" in JSON config.

This integration is how libraries get their Adobe IDs issued. (Maybe
there should even be two different integrations here: one for
integration with the library registry as such, and one for integration
with the library registry as Adobe ID issuer.)

I'm not sure what the `goal` and `protocol` of this integration are,
since its theoretical purpose is to set up a line of communication
between this library and the registry, but its main practical purpose
is to get Adobe IDs from the registry. For the sake of argument let's
say `goal="registration"` and the protocol="Library Registry".

The second integration is for the thing that _issues_ the Adobe
IDs. This is a site-wide integration not associated with any
particular library. Only NYPL and Open Ebooks have this
integration.

This integration's `goal` is "DRM", its `protocol` is "Adobe Vendor ID"
and it will go away once we have the library registry. It's
effectively the configuration for the library registry itself. Now is
a good time to separate its configuration from the integration that's
going to stick around in the circ manager.

The library registry integration has the following configuration
settings:

* The vendor ID, issued by Adobe. This is "vendor_id" in JSON
  config.
* The node value, issued by Adobe. This is "node_value" in JSON config.
* A bunch of information about other libraries -- URI, short name,
  shared secret -- used when verifying the signatures before issuing
  Adobe IDs. This is "other_libraries" in JSON config.

Here's the changes I think we need. Since I have it in my head from
writing this, I can take a stab at implementing it on top of this
branch if you want, but feel free to take it yourself:

1. Create a per-library integration with goal="registration" and
protocol="Library Registry". Most of the stuff currently in
an "Adobe Vendor ID" goes here.  Library.library_registry_short_name
and Library.library_registry_shared_secret are also moved here.

2. Create a global integration with goal="DRM" and protocol="Adobe
ID". node_value and other_libraries go here.

courte:
This is the JSON object I've been using to test imports.

```js
{
  "integrations" : {
    "Postgres" : {
      "production_url" : "postgres://simplified:[password]@localhost:5432/simplified_multicirc_dev",
      "test_url" : "postgres://simplified_test:test@localhost:5432/simplified_multicirc_test"
    },

    "Adobe Vendor ID" : {
      "vendor_id" : "VID",
      "node_value" : "2349802592038478",
      "other_libraries" : {
        "http://your-library.org/": ["you", "Your library secret"] 
      }
    },

    "CDN" : {
      "opds" : "https://opds.cloudfront.net/",
      "book_covers" : "https://book-covers.cloudfront.net/",
      "open_access_books" : "https://open-access-books.cloudfront.net/"
    },

    "Circulation Manager" : {
      "url" : "http://localhost:4500/"
    },

    "Content Server" : {
      "url" : "http://content-server/"
    },

    "Elasticsearch" : {
      "url" : "http://elastic-search/",
      "works_index" : "es_works_index"
    },

    "Google OAuth" : {
      "web" : {
        "client_id" : "892347847-client-id.com",
        "project_id" : "g_project_id",
        "auth_uri":"https://accounts.google.com/o/oauth2/auth",
        "token_uri" : "https://accounts.google.com/o/oauth2/token",
        "auth_provider_x509_cert_url" : "https://cert-url",
        "client_secret" : "g_secret",
        "redirect_uris" : ["http://redirect-uri1", "http://redirect-uri2"]
      }
    },

    "loggly" : {
      "token" : "l_token",
      "url" : "http://l-url/"
    },

    "S3" : {
      "access_key" : "s3_account",
      "secret_key" : "s3_secret",
      "book_covers_bucket" : "http://book-covers.url",
      "open_access_content_bucket" : "oa-content.url",
      "static_feed_bucket" : "opds-feeds.url"
    },

    "Metadata Wrangler" : {
      "url" : "http://metadata-wrangler/",
      "client_id" : "mw_client_id",
      "client_secret" : "mw_client_secret"
    },

    "NoveList Select" : {
      "profile" : "ns_profile",
      "password" : "ns_password"
    },

    "New York Times" : {
      "best_sellers_api_key" : "nytblahblah:b:lah"
    },

    "Patron Web Client" : {
      "url" : "http://patron-web-client"
    }
  }
}
```

leonardr:
Apart from my issues with core, the only thing I think needs to be changed here is the name of the integration for generating Short Client Tokens. I would prefer to call the protocol SHORT_CLIENT_TOKEN rather than LIBRARY_REGISTRY. I find that name confusing because "Library registry" isn't really a protocol (Short Client Token totally is), and most of the information we'll be sending to the library registry will be kept in library configuration settings, not in a separate integration.

leonardr:
This is great! Saying "short client token" makes it so much clearer to me what's going on.

PR #515, opened 2017-06-14
Moved url, username, and password on ExternalIntegration to settings, and fixed a CSRF token problem

aslagle:
This simplifies the admin interface code using https://github.com/NYPL-Simplified/server_core/pull/528.

I also fixed an admin interface bug I noticed due to the CSRF token changing on every page load. When you opened a second tab, and then went back to the first tab and tried to submit a form, you'd get a CSRF token error since that tab's html still had the old token, but your cookie had the new token.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1008,16 +1008,16 @@ class SettingsController(CirculationManagerController):
             "fields": [
                 { "key": "external_account_id", "label": _("Library ID") },
                 { "key": "website_id", "label": _("Website ID") },
-                { "key": "username", "label": _("Client Key") },
-                { "key": "password", "label": _("Client Secret") },
+                { "key": ExternalIntegration.USERNAME, "label": _("Client Key") },
+                { "key": ExternalIntegration.PASSWORD, "label": _("Client Secret") },
             ],
         })
 
         protocols.append({
             "name": ExternalIntegration.BIBLIOTHECA,
             "fields": [
-                { "key": "username", "label": _("Account ID") },
-                { "key": "password", "label": _("Account Key") },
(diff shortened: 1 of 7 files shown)


api/admin/controller.py
```diff
@@ -1008,38 +1008,36 @@ def collections(self):
...
 
         protocols.append({
             "name": ExternalIntegration.ONE_CLICK,
             "fields": [
-                { "key": "password", "label": _("Basic Token") },
+                { "key": ExternalIntegration.PASSWORD, "label": _("Basic Token") },
                 { "key": "external_account_id", "label": _("Library ID") },
-                { "key": "url", "label": _("URL") },
-                { "key": "ebook_loan_length", "label": _("eBook Loan Length") },
-                { "key": "eaudio_loan_length", "label": _("eAudio Loan Length") },
```
aslagle: I believe these two settings were removed from oneclick in an earlier branch.

api/admin/routes.py
```diff
@@ -379,7 +379,7 @@ def admin_view(collection=None, book=None, **kwargs):
     else:
         home_url = None
 
-    csrf_token = app.manager.admin_sign_in_controller.generate_csrf_token()
+    csrf_token = flask.request.cookies.get("csrf_token") or app.manager.admin_sign_in_controller.generate_csrf_token()
```
aslagle: If you already have a CSRF token cookie, you keep it, but if not you get a new one.

api/admin/controller.py
leonardr: That's correct, but every Collection (regardless of its protocol) can now have a default_loan_period and a default_reservation period.

leonardr:
:+1:

PR #516, opened 2017-06-15
Added admin controller for editing sitewide settings.

aslagle:


tests/admin/test_controller.py
```diff
@@ -1122,6 +1123,11 @@ def setup(self):
          ):
             self._db.delete(auth_service)
 
+        # Delete any existing sitewide ConfigurationSettings.
+        for setting in self._db.query(ConfigurationSetting).filter(
+            ConfigurationSetting.library_id==None).filter(
+            ConfigurationSetting.external_integration_id==None):
+            self._db.delete(setting)
```
leonardr: I'm not sure about this. Is it okay to remove these values when a new value might not be forthcoming?

leonardr:
:+1:

PR #519, opened 2017-06-15
Fix to Millennium last name authentication

datalogics-dans:
Previously, a generic error was thrown when an invalid barcode was entered.
Now, the client will display the correct invalid barcode/PIN message.


datalogics-dans:
Closing this PR since most of this work has been done differently (and better) already.

api/millenium_patron.py
```diff
@@ -101,10 +101,15 @@ def remote_authenticate(self, username, password):
             url = self.root + path
             response = self.request(url)
             data = dict(self._extract_text_nodes(response.content))
-            dump_name = data.get(self.PERSONAL_NAME_FIELD)
-            dump_name = dump_name.split(',')[0]
-            if dump_name.upper() == password.upper():
-              return PatronData(authorization_identifier=username, complete=False)
+            if data.get('ERROR_MESSAGE_FIELD') is None:
```
leonardr: ERROR_MESSAGE_FIELD is a constant (self.ERROR_MESSAGE_FIELD), not a string. So I don't think the other branch of this code (the `return False`) will ever get run. The code still works because you don't test it with a document that uses both ERROR_MESSAGE_FIELD and PERSONAL_NAME_FIELD.

I don't think Millenium will ever send such a document, so that specific thing isn't a big deal, but the constant needs to be fixed and I don't think the weird situation you ran into (where the wrong error message was being sent out) has been tested.

leonardr:
:+1: 

Make sure this branch is also ported to https://github.com/NYPL-Simplified/circulation/tree/multi-tenancy.

PR #522, opened 2017-06-16
Moved constants for editing Collections and their ExternalIntegrations

aslagle:


api/admin/controller.py
```diff
@@ -987,59 +992,16 @@ def libraries(self):
...
+        for license_api in [OPDSImporter,
+                            OverdriveAPI,
+                            BibliothecaAPI,
+                            Axis360API,
+                            OneClickAPI,
+                           ]:
+            protocols.append({
+                "name": license_api.NAME,
+                "fields": license_api.FIELDS,
+            })
```
leonardr: I think this would be a better place to tack on BaseCirculationAPI.FIELDS (for all the APIs except OPDSImporter). Right now they're missing from `Axis360API.FIELDS` and `OneClickAPI.FIELDS`.

aslagle: Oops, maybe that would be better. I liked having the API classes get to decide whether to inherit the parent's fields though.

leonardr: It's hard to say so early how many fields there are going to be to inherit. At any rate, either way we'll have to make this change with every new content source. I'm fine with either solution.

aslagle: It's also possible that we'd add another content source like OPDS import that doesn't have them, and have to add a special case to this code to exclude them, so I think I'd rather leave them in the API classes. I'll fix the ones I missed.

PR #524, opened 2017-06-19
Stop assuming there's only one library

leonardr:
This branch brings the circulation branch up to date with https://github.com/NYPL-Simplified/server_core/pull/525, which removes the `Library.instance()` method that lets us assume there is only ever one library.

Within the CirculationManager, we need to keep a different set of Lanes for each library and a different CirculationAPI for each library.

There's a lot of work surrounding the Adobe Vendor ID integration. Many of the tests require that a test create two libraries and configure them differently. I spent some time eliminating bad interactions with other tests, by making sure that the Adobe Vendor ID integrations aren't initialized unless they're actually needed.


api/controller.py
```diff
@@ -142,29 +142,42 @@ def __init__(self, _db, lanes=None, testing=False):
...
+            self.circulation_apis[library.id] = self.setup_circulation(
+                library
+            )
+            self.setup_adobe_vendor_id(library)
         self.lending_policy = load_lending_policy(
             Configuration.policy('lending', {})
         )
 
         self.setup_controllers()
-        self.setup_adobe_vendor_id(library)
```
leonardr: It's simpler and faster to set up the Adobe Vendor ID configuration only when the test actually needs it.

api/testing.py
```diff
@@ -25,87 +26,75 @@
...
-    """Sets up an Adobe Vendor ID integration."""
+    """A DatabaseTest that knows how to set up an Adobe Vendor ID
+    integration.
+    """
 
-    TEST_NODE_VALUE = 114740953091845
     TEST_VENDOR_ID = u"vendor id"
-    TEST_LIBRARY_URI = u"http://me/"
-    TEST_OTHER_LIBRARY_URI = u"http://you/"
-    TEST_OTHER_LIBRARIES  = {TEST_OTHER_LIBRARY_URI: ("you", "secret2")}
```
leonardr: Rather than define constants, which will only work once or twice, we generate the appropriate values for an Adobe Vendor ID or Short Client Token integration based on the library's short_name, which is always different.

scripts.py
```diff
@@ -433,27 +428,16 @@ def process_lane(self, lane):
...
-        library = Library.instance(_db)
-        
-        enabled, default = cls.facet_settings(
-            library, Facets.ORDER_FACET_GROUP_NAME
-        )
-        order_help = 'Generate feeds for this ordering. Possible values: %s. Default: %s' % (
-            ", ".join(enabled), default
+        available = Facets.DEFAULT_ENABLED_FACETS[Facets.ORDER_FACET_GROUP_NAME]
+        order_help = 'Generate feeds for this ordering. Possible values: %s.' % (
+            ", ".join(available)
```
leonardr: We no longer know which facets are available, since it's a per-library setting and this script runs against all libraries.
Showing the general default list is the best we can do.

api/testing.py
```diff
@@ -25,87 +26,74 @@
...
 
+            # Each library will get a slightly different short
+            # name and secret for generating Short Client Tokens.
+            library_uri = self._url
+            short_name = library.short_name + "token"
+            secret = library.short_name + " token secret"
+            short_client_token.username = short_name
+            short_client_token.password = secret
+
+            library.setting(Library.WEBSITE_KEY).value = library_uri
```
aslagle: I guess this was from another branch, but I don't see what this setting is for?

leonardr: A library uses the URL of its website (e.g. https://nypl.org/) to generate JWTs. This URL is shared with the library registry -- it's the keys of the dictionary in the 'other_libraries' setting.

This URL is also going to be an important part of the Authentication for OPDS document, so Courteney and I decided it made more sense as a library-wide setting than as part of the Adobe Vendor ID and Short Client Token integrations.

scripts.py
```diff
@@ -536,27 +496,46 @@ def do_generate(self, lane):
...
+        for order in chosen_orders:
+            if order not in allowed_orders:
+                logging.warn("Ignoring unsupported ordering %s" % order)
+            for availability in chosen_availabilities:
+                if availability not in allowed_availabilities:
+                    logging.warn("Ignoring unsupported availability %s" % availability)
+
+                for collection in chosen_collections:
+                    if collection not in allowed_availabilities:
+                        logging.warn("Ignoring unsupported collection %s" % collection)
```
aslagle: shouldn't you continue if you hit an unsupported value?

PR #527, opened 2017-06-20
Allow creating multiple libraries in the admin interface

### LINKED ISSUE #525: Make it possible to create a new library through the admin interface
The admin interface currently uses a trick in the controller to prevent you from creating a new library by accident. This operation now needs to be explicitly supported through the interface.

aslagle:
This fixes #525

### DIFF
--- a/api/admin/problem_details.py
+++ b/api/admin/problem_details.py
@@ -100,13 +100,6 @@ INVALID_SERIES_POSITION = pd(
     detail=_("The series position must be a number or blank."),
 )
 
-LIBRARY_NOT_FOUND = pd(
-    "http://librarysimplified.org/terms/problem/library-not-found",
-    status_code=404,
-    title=_("Library not found"),
-    detail=_("The specified library uuid does not exist."),
-)
-
 MISSING_LIBRARY_SHORT_NAME = pd(
     "http://librarysimplified.org/terms/problem/missing-library-short-name",
     status_code=400,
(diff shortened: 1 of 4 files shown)


api/admin/problem_details.py
```diff
@@ -102,9 +102,23 @@
 
 LIBRARY_NOT_FOUND = pd(
     "http://librarysimplified.org/terms/problem/library-not-found",
+    status_code=404,
+    title=_("Library not found"),
+    detail=_("The specified library uuid does not exist."),
```
leonardr: I didn't notice this before, but can you combine the admin LIBRARY_NOT_FOUND and the api LIBRARY_NOT_FOUND? Not a big deal if that's annoying, since they use the same URI.

api/admin/routes.py
```diff
@@ -386,7 +386,9 @@ def admin_view(collection=None, book=None, **kwargs):
                     quoted_book,
                     quoted_book.replace("/", "%2F"))
             return redirect(app.manager.url_for('admin_sign_in', redirect=redirect_url))
-        library = Library.instance(app.manager._db)
+        # TODO: Design the admin interface for multiple libraries.
+        # Until then, pick some library so the admin interface can run.
+        library = app.manager._db.query(Library).all()[0]
```
leonardr: Can you sort this by Library.id? Since we'll be reintroducing the 'default library' concept soon it's OK if we wait and solve it there, but that's a small change.

leonardr:
:+1: with some nitpicks.

api/admin/problem_details.py
aslagle: good point, I didn't realize there were two.

PR #535, opened 2017-06-21
Metadata services in admin interface

aslagle:
This branch adds metadata service editing to the admin interface, with NYT and NoveList as the supported protocols. I also consolidated a bunch of similar problem details.

I tried to follow the pattern I started with patron auth services, where the constants that control the admin interface form are defined in the individual API classes. Unlike patron auth services, though, metadata services don't share a base class, so maybe it doesn't make as much sense. I wasn't sure what to do for the metadata wrangler protocol, since it doesn't have an API class. That still needs to be added. Other than metadata wrangler, I think the remaining protocols aren't used in the circ manager and don't need to be in the admin interface, but let me know if I missed anything.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -88,6 +88,9 @@ from api.bibliotheca import BibliothecaAPI
 from api.axis import Axis360API
 from api.oneclick import OneClickAPI
 
+from api.nyt import NYTBestSellerAPI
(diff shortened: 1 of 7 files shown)


aslagle:
Added in the metadata wrangler, using https://github.com/NYPL-Simplified/server_core/pull/540

PR #540, opened 2017-06-21
Fix missing config info

### LINKED ISSUE #528: Millenium Patron authentication form doesn't have a place to configure the URL

### LINKED ISSUE #529: Default loan period and reservation period should not be required fields
We know the defaults: 14 day loan period, 3 day reservation period.

aslagle:
Fixes #528 
Fixes #529


leonardr:
:+1:

Issue #533, opened 2017-06-21
Send a signal to reload all configuration settings


leonardr:
Currently you can modify site settings in the admin interface, but it takes a restart to make those changes take effect. It should be possible to tell the circ manager to reload its configuration without a restart.

The real challenge is propagating this message cross all circ managers in the cluster. Maybe the solution is to set a timestamp in the database which the circ manager checks periodically.

PR #541, opened 2017-06-21
Reload configuration on demand

leonardr:
This branch defines `CirculationManager.load_settings()`, which loads or reloads everything that a CirculationManager needs to load from database configuration such as ExternalIntegrations. The idea is that we can fix https://github.com/NYPL-Simplified/circulation/issues/533 by calling `load_settings()` whenever the configuration changes.

This branch also drastically reduces the time it takes for the `test_controller.py` tests to run by scaling down the number of lanes  created each time a `CirculationManager` is created. Creating a full compliment of Lane objects takes a long time, and only one of the tests actually needed deep-down lanes like "English / Fiction / Science Fiction".

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -166,6 +166,8 @@ class CirculationManager(object):
         # Create a CirculationAPI for each library.
         self.circulation_apis = {}
 
+        self.adobe_vendor_id = None
+        self.adobe_device_management = None
         for library in self._db.query(Library):
             lanes = make_lanes(library, self.lane_descriptions)
             
@@ -178,7 +180,12 @@ class CirculationManager(object):
             self.circulation_apis[library.id] = self.setup_circulation(
                 library
             )
-            self.setup_adobe_vendor_id(library)
+            authdata = self.setup_adobe_vendor_id(library)
+            if authdata and not self.adobe_device_management:
+                # There's at least one library on this system that
+                # wants Vendor IDs. This means we need to advertise support
+                # for the Device Management Protocol.
+                self.adobe_device_management = DeviceManagementProtocolController(self)
         self.lending_policy = load_lending_policy(
             Configuration.policy('lending', {})
         )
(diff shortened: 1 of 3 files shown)


api/controller.py
```diff
@@ -165,12 +180,17 @@ def __init__(self, _db, lanes=None, testing=False):
             self.circulation_apis[library.id] = self.setup_circulation(
                 library
             )
-            self.setup_adobe_vendor_id(library)
+            authdata = self.setup_adobe_vendor_id(library)
+            if authdata and not self.adobe_device_management:
+                # There's at least one library on this system that
+                # wants Vendor IDs. This means we need to advertise support
+                # for the Device Management Protocol.
+                self.adobe_device_management = DeviceManagementProtocolController(self)
```
leonardr: Previously, `self.adobe_device_management` was set within `setup_adobe_vendor_id`. This wasn't the best because it meant the controller could be set up multiple times if multiple libraries on the system had Short Client Token integrations.

api/controller.py
```diff
@@ -231,24 +251,37 @@ def setup_circulation(self, library):
...
         return cls(library)
         
-    def setup_controllers(self):
-        """Set up all the controllers that will be used by the web app."""
+    def setup_one_time_controllers(self):
+        """Set up all the controllers that will be used by the web app.
+
+        This method will be called only once, no matter how many times the
+        site configuration changes.
+        """
```
leonardr: I don't know how much time this saves, and it seems a little risky.

api/controller.py
```diff
@@ -267,6 +300,10 @@ def setup_adobe_vendor_id(self, library):
             vendor_id = adobe.username
             node_value = adobe.password
             if vendor_id and node_value:
+                if self.adobe_vendor_id:
+                    self.log.warn(
+                        "Multiple libraries define an Adobe Vendor ID integration. This is not supported and the last library seen will take precedence."
+                    )
```
leonardr: I'm not making a big deal out of this because the Adobe Vendor ID integration is going away in the medium term.

PR #542, opened 2017-06-22
Refactored admin controllers to share more code.

aslagle:
A lot of the admin controllers were pretty similar so I renamed things to be more consistent and pulled out a couple helper methods.


leonardr:
This makes sense. :+1:

PR #543, opened 2017-06-22
Autoreload configuration

leonardr:
This branch changes the `CirculationManager` to keep track of the last time it saw a site configuration change. If it detects a change, it calls `load_settings()` to reload the configuration. This should make sure any `CirculationManager` gets reloaded within 61 seconds of a change to any configuration item, even in a clustered scenario.

The check happens in `library_for_request`, which seems like the earliest code that runs on almost every request. The test illustrates a case where the name of a library changes immediately before a request comes in for the library under the new name.

### DIFF
--- a/api/bibliotheca.py
+++ b/api/bibliotheca.py
@@ -733,7 +733,7 @@ class BibliothecaEventMonitor(CollectionMonitor):
 
         The command line date argument should have the format YYYY-MM-DD.
         """
-        initialized = get_one(_db, Timestamp, self.service_name)
+        initialized = get_one(_db, Timestamp, service=self.service_name)
         default_start_time = datetime.datetime.utcnow() - self.DEFAULT_START_TIME
 
         if cli_date:
--- a/api/controller.py
+++ b/api/controller.py
@@ -154,7 +154,7 @@ class CirculationManager(object):
         """If the site configuration has been updated, reload the
         CirculationManager's configuration from the database.
         """
-        last_update = Configuration.site_configuration_last_update(self._db, timeout=0)
+        last_update = Configuration.site_configuration_last_update(self._db)
         if last_update > self.site_configuration_last_update:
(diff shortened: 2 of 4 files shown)


api/controller.py
```diff
@@ -143,10 +143,22 @@ def __init__(self, _db, lanes=None, testing=False):
...
+        )
         self.lane_descriptions = lanes
         self.setup_one_time_controllers()
         self.load_settings()
         
+    def reload_settings_if_changed(self):
+        """If the site configuration has been updated, reload the
+        CirculationManager's configuration from the database.
+        """
+        last_update = Configuration.site_configuration_last_update(self._db, timeout=0)
```
aslagle: why is the timeout 0 here?

leonardr: I figured when we are first initializing the CirculationManager, we'd want to go to the database no matter what. It shouldn't make a difference, though, since I don't see a case where you would call `site_configuration_last_update` before initializing the CM.

api/bibliotheca.py
```diff
@@ -733,7 +733,7 @@ def create_default_start_time(self, _db, cli_date):
 
         The command line date argument should have the format YYYY-MM-DD.
         """
-        initialized = get_one(_db, Timestamp, self.service_name)
+        initialized = get_one(_db, Timestamp, service=self.service_name)
```
leonardr: This ancient bug was uncovered by the fact that the Bibliotheca Timestamp is no longer the first one in the database when this code runs during `TestBibliothecaEventMonitor.test_default_start_time`.

api/controller.py
aslagle: That's fine, but this is the method that runs on every request, not initialization.

leonardr: Oops, I didn't look at the context. That's a copy-paste error.

PR #546, opened 2017-06-23
Made admin controllers more consistent

aslagle:
To support https://github.com/NYPL-Simplified/circulation-web/pull/91, I made all of the integration-editing controllers use ids, and put settings under a "settings" key for GET requests. I also added names, since they were added to the scripts for creating external integrations. 

@leonardr's fix for running the admin interface when there are no libraries is in here too.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1150,8 +1150,6 @@ class SettingsController(CirculationManagerController):
         if protocol and protocol not in [p.get("name") for p in protocols]:
             return UNKNOWN_PROTOCOL
 
-        name = flask.request.form.get("name")
-
         is_new = False
         collection = None
         if id:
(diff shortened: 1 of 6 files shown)


api/admin/controller.py
```diff
@@ -1147,15 +1150,30 @@ def collections(self):
         if protocol and protocol not in [p.get("name") for p in protocols]:
             return UNKNOWN_PROTOCOL
 
+        name = flask.request.form.get("name")
```
leonardr: It looks like this line is redundant.

api/admin/controller.py
```diff
@@ -1147,15 +1150,30 @@ def collections(self):
...
+                    return COLLECTION_NAME_ALREADY_IN_USE
+                
         else:
+            collection_with_name = get_one(self._db, Collection, name=name)
+            if collection_with_name:
+                return COLLECTION_NAME_ALREADY_IN_USE
+
             if protocol:
-                collection, is_new = get_one_or_create(
+                collection, is_new = create(
```
leonardr: Can you call get_one_or_create and then return COLLECTION_NAME_ALREADY_IN_USE if is_new is False?

leonardr:
:+1: with some minor tweaks.

leonardr:
Looks good!

PR #549, opened 2017-06-27
Make CirculationAPI database-session-safe

leonardr:
This branch changes CirculationAPI to take a database session as a constructor argument instead of deriving it from the Library object. In the web app, using Session.object_session will give a database session that's only good for the active request, and CirculationAPI needs to stick around for multiple requests.

This branch also changes `ServiceStatus` to take a `CirculationAPI` as an argument instead of trying to create one. I think `ServiceStatus` is going away soon but this is the easiest way to solve the `CirculationAPI` constructor problem.

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -20,6 +20,7 @@ from core.model import (
     ExternalIntegration,
     Identifier,
     DataSource,
+    Library,
     LicensePoolDeliveryMechanism,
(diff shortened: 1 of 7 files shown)


leonardr:
This fixes https://github.com/NYPL-Simplified/server_core/issues/548

PR #550, opened 2017-06-27
Use default library when no library is specified in incoming request

leonardr:
This branch uses `Library.default` to pick a library to use as the request library, instead of raising a ValueError when none is specified.

This will work for multi-library setups but it's mainly intended for single-library setups, where naming a library in the URL is redundant.


leonardr:
This is the backend work for https://github.com/NYPL-Simplified/server_core/issues/537.

aslagle:
Looks good.

migration/20170616-2-move-third-party-config-to-external-integrations.py still uses Library.instance and needs an update.

PR #551, opened 2017-06-27
Moved analytics config into the database and added it to the admin interface

### LINKED ISSUE #539: Add analytics integrations to admin interface

aslagle:
This fixes #539 
It uses https://github.com/NYPL-Simplified/circulation-web/pull/93 and https://github.com/NYPL-Simplified/server_core/pull/551

It moves the analytics configuration into the database, and changes Analytics to be initialized by the CirculationManager, instead of a global singleton. When scripts need it, they have to create their own. It also adds an admin route for editing analytics configuration.

### DIFF
--- a/api/google_analytics_provider.py
+++ b/api/google_analytics_provider.py
@@ -3,24 +3,45 @@ import uuid
 import unicodedata
 import urllib
 import re
+from flask.ext.babel import lazy_gettext as _
 from core.util.http import HTTP
+from core.model import (
+    ConfigurationSetting,
+    ExternalIntegration,
+    Session,
+    get_one,
+)
 
 class GoogleAnalyticsProvider(object):
-    INTEGRATION_NAME = "Google Analytics"
+    NAME = _("Google Analytics")
+
+    TRACKING_ID = "tracking_id"
+    DEFAULT_URL = "http://www.google-analytics.com/collect"
+
+    SETTINGS = [
+        { "key": ExternalIntegration.URL, "label": _("URL"), "default": DEFAULT_URL },
+    ]
+
+    LIBRARY_SETTINGS = [
+        { "key": TRACKING_ID, "label": _("Tracking ID") },
+    ]
     
-    @classmethod
-    def from_config(cls, config):
-        tracking_id = config[Configuration.INTEGRATIONS][cls.INTEGRATION_NAME]['tracking_id']
-        return cls(tracking_id)
+    def __init__(self, integration):
+        url_setting = ConfigurationSetting.for_externalintegration(ExternalIntegration.URL, integration)
+        self.url = url_setting.value or self.DEFAULT_URL
+        self.integration_id = integration.id
 
-    def __init__(self, tracking_id):
-        self.tracking_id = tracking_id
+    def collect_event(self, library, license_pool, event_type, time, **kwargs):
+        _db = Session.object_session(library)
+        integration = get_one(_db, ExternalIntegration, id=self.integration_id)
+        tracking_id = ConfigurationSetting.for_library_and_externalintegration(
+            _db, self.TRACKING_ID, library, integration,
+        ).value
 
-    def collect_event(self, _db, license_pool, event_type, time, **kwargs):
         client_id = uuid.uuid4()
         fields = {
             'v': 1,
-            'tid': self.tracking_id,
+            'tid': tracking_id,
             'cid': client_id,
             'aip': 1, # anonymize IP
             'ds': "Circulation Manager",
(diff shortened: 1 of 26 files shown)


migration/20170616-2-move-third-party-config-to-external-integrations.py
```diff
@@ -117,7 +117,7 @@ def log_import(integration_or_setting):
 
             library_url = adobe_conf.get('library_uri')
             ConfigurationSetting.for_library(
-                Library.WEBSITE_KEY, library).value = library_url
+                Configuration.WEBSITE_URL, library).value = library_url
```
aslagle: I noticed this wasn't working.

migration/20170616-2-move-third-party-config-to-external-integrations.py
```diff
@@ -145,6 +145,28 @@ def log_import(integration_or_setting):
...
+    policies = Configuration.get(u"policies", {})
+    analytics_modules = policies.get(u"analytics", ["core.local_analytics_provider"])
+
+    if "api.google_analytics_provider" in analytics_modules:
+        google_analytics_conf = Configuration.integration(u"Google Analytics Provider", {})
+        tracking_id = google_analytics_conf.get(u"tracking_id")
+
+        integration = EI(protocol=u"api.google_analytics_provider", goal=EI.ANALYTICS_GOAL)
+        _db.add(integration)
+        integration.url = "http://www.google-analytics.com/collect"
```
aslagle: The url wasn't part of the configuration before, but I think it should be.

api/google_analytics_provider.py
```diff
@@ -3,24 +3,45 @@
...
+        self.integration_id = integration.id
 
-    def __init__(self, tracking_id):
-        self.tracking_id = tracking_id
+    def collect_event(self, library, license_pool, event_type, time, **kwargs):
+        _db = Session.object_session(library)
+        integration = get_one(_db, ExternalIntegration, id=self.integration_id)
+        tracking_id = ConfigurationSetting.for_library_and_externalintegration(
+            _db, self.TRACKING_ID, library, integration,
+        ).value
```
aslagle: The tracking id is now a per-library setting, so different libraries can set up their own Google Analytics accounts if they want to. I have not yet done anything to add the library to the info in the analytics events.

leonardr: I think you'd get better performance (and simplify the API somewhat) if you followed the model we use for Authenticator. That is, create a separate LibraryAnalytics object for every Library and have Analytics keep them in a dictionary keyed by ID. Is that feasible?

migration/20170616-2-move-third-party-config-to-external-integrations.py
leonardr: Agreed. This would let us hook it up to the GA proxy or similar.

api/oneclick.py
```diff
@@ -800,8 +802,9 @@ def process_availability(self, media_type='ebook'):
             license_pool, is_new, is_changed = self.api.update_licensepool_for_identifier(isbn, available)
             # Log a circulation event for this work.
             if is_new:
-                Analytics.collect_event(
-                    self._db, license_pool, CirculationEvent.DISTRIBUTOR_AVAILABILITY_NOTIFY, license_pool.last_checked)
+                for library in self.collection.libraries:
+                    self.analytics.collect_event(
+                        library, license_pool, CirculationEvent.DISTRIBUTOR_TITLE_ADD, license_pool.last_checked)
```
aslagle: All these places that were calling Analytics.collect_event are missing test coverage.

Also, for some reason this one was using the wrong event type. I think AVAILABILITY_NOTIFY is when a patron has something on hold and it becomes available for checkout. I'm not completely sure what's going on here, but TITLE_ADD is usually the event type for a new license pool.

leonardr: Yes, this should be TITLE_ADD. Sounds like someone was thinking the "availability" meant availability _in general_.

api/google_analytics_provider.py
aslagle: It's a little different than Authenticator, because not all the Analytics integrations are associated with a Library. The LocalAnalyticsProvider is not - once you turn it on it stores events for all libraries in the db.

The comment also made me realize I'm currently calling all the analytics providers for every event, even though some of them might not be set up for every Library.

aslagle: I changed Analytics to keep a dictionary of providers by library id, as well as a list of sitewide providers. I didn't create a LibraryAnalytics class since there wasn't anything for it to do except call the provider.

leonardr:
:+1:

PR #554, opened 2017-06-29
Add a 'where are my books' script

leonardr:
This branch adds a script that explains why your books aren't showing up.


courte:
You have a test failure from the submodule update. It looks like a lingering use of `Library.instance`

PR #556, opened 2017-06-29
Short client token changes and drm services in admin interface

### LINKED ISSUE #538: Add DRM integrations to admin interface

### LINKED ISSUE #553: Make SHORT_CLIENT_TOKEN integration shared across libraries
The SHORT_CLIENT_TOKEN integration should be associated with multiple libraries.

The `vendor_id` is a configuration setting associated directly with the integration.

The `library_registry_short_name` and `library_registry_shared_secret` are configuration settings on the combination of the integration and a Library.

We should also get rid of `Library.library_registry_short_name` and `Library.library_registry_shared_secret` -- they are redundant.

aslagle:
This fixes #553 and fixes #538.

This uses https://github.com/NYPL-Simplified/circulation-web/pull/94.

### DIFF
--- a/api/admin/problem_details.py
+++ b/api/admin/problem_details.py
@@ -114,20 +114,6 @@ LIBRARY_SHORT_NAME_ALREADY_IN_USE = pd(
     detail=_("The library short name must be unique, and there's already a library with the specified short name."),
 )
 
-CANNOT_SET_BOTH_RANDOM_AND_SPECIFIC_SECRET = pd(
-    "http://librarysimplified.org/terms/problem/cannot-set-both-random-and-specific-secret",
-    status_code=400,
-    title=_("Cannot set both random and specific secret"),
-    detail=_("You can't set the shared secret to a random value and a specific value at the same time"),
-)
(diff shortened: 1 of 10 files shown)


api/admin/problem_details.py
```diff
@@ -114,20 +114,6 @@
...
-    title=_("Cannot set both random and specific secret"),
-    detail=_("You can't set the shared secret to a random value and a specific value at the same time"),
-)
-
-CANNOT_REPLACE_EXISTING_SECRET_WITH_RANDOM_SECRET = pd(
-    "http://librarysimplified.org/terms/problem/cannot-replace-existing-secret-with-random-secret",
-    status_code=400,
-    title=_("Cannot replace existing secret with random secret"),
-    detail=_("You can't overwrite an existing shared secret with a random value"),
-)
```
aslagle: Random secret generation is now handled on the client-side, so the server doesn't know if a secret was generated randomly or not.

leonardr:
Looks good. :+1:

PR #562, opened 2017-07-03
Made Lane constructor take db argument.

aslagle:
This goes with https://github.com/NYPL-Simplified/server_core/pull/560.

Having Lane._db come from the Library was causing a scoped session error when creating CachedFeeds, so Leonard suggested passing the db to the Lane constructor.

I'm not sure how to write a test for the problem I was seeing, but I verified that this fixed it. It happened when going to the root url and getting redirected after you'd already been to /groups.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -185,11 +185,11 @@ class CirculationManager(object):
         self.adobe_device_management = None
         self.short_client_token_initialization_exceptions = dict()
         for library in self._db.query(Library):
-            lanes = make_lanes(library, self.lane_descriptions)
+            lanes = make_lanes(self._db, library, self.lane_descriptions)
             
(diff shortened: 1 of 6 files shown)


leonardr:
:+1: 

If you want to test this, use `TestScopedSession.test_scoped_session` as a model. Maybe modify that method to call self.manager.index_controller() within its scoped request after checking that it gets its own database session.

aslagle:
I added a test, but it doesn't fail with the old code.

leonardr:
That's strange, but I do think there's a class of errors that this test will catch, so let's keep it.

aslagle:
I think this will have to do. The controller that gets the error works fine in a test, maybe because the transaction from previous requests aren't committed.

Issue #561, opened 2017-07-03
Store library configuration information as settings and display it in Authentication for OPDS document


leonardr:
A lot of information about how different libraries work is currently hard-coded in the client. It needs to be configured on the circulation manager and the client needs to program itself based on information it sees in the Authentication For OPDS document.

This issue is a mixed bag but it shouldn't be difficult to implement all this stuff at once.

These are configuration settings that apply to any Basic authentication mechanism:

* Which keyboard should the client use when asking the patron to type in their identifier? Options are "Full" and "Numeric".
* When a patron wants to scan their identifier rather than type it in, what kind of barcode scanner should the client use? Options are "CODABAR" and "None". In the future we can add "QR", maybe others.
* Which keyboard should the client use when asking the patron to type in their password? Options are "Full" and "Numeric". 
* What is the maximum length of the patron's identifier, if any?
* What is the term to use for the patron's identifier? Default is "Barcode".
* What is the term to use for the patron's password? Default is "PIN".
* What is the maximum length of the patron's password, if any? (0 means that authentication is handled entirely through the patron identifier, and password is always empty.)

Here's my proposed Authentication For OPDS extension for conveying information about the login and password. 

```
"inputs": {
  "login": {
    "keyboard": "full",
    "maximum_length": 20,
    "barcode_format": "CODABAR"
  },
  "password": {
    "keyboard": "numeric",
    "maximum_length": 4
  }
}
```

These are configuration settings that apply to a library and are not already being captured in the circulation manager:

* The library's logo. Given the small size and the fact that it will always be included inline rather than linked to, it's probably okay to store this in base64 format rather than as a `Representation`.
* The library's description. This would go in a different field than 'description', which is for a text prompt that's displayed when you want the user to authenticate.
* The URL of the library's web site. This is already stored as the per-library WEBSITE_URL setting. This would be represented as a link with `rel="home"`. (Not sure about this one -- we might want a custom URI relation.)
* The URL someone should use if they want to get a library card at this library. For NYPL this would be `simplified-card-creator:patrons.librarysimplified.org` and activate the native card creator. For Brooklyn it would be `https://bpl.bklynlibrary.org/patron_card/registration.aspx` and be handled in a web view. This would be represented as a link with `rel="register"`.
[...]

Issue #560, opened 2017-07-03
Use feature flags to request client features be turned on or off


leonardr:
Different circulation managers have different opinions on whether
certain client-side features should be enabled.

* Whether the 'reservations' tab should be shown. (Open Ebooks doesn't
  support putting books on hold at all; this is controlled by the
  Configuration.ALLOW_HOLDS library setting.)

* Whether a patron's barcode should be displayed in scannable
  form. (NYPL may have a policy against this, or at least may want to
  soft-pedal it for a bit. This would have to be controlled by a new
  library setting.)

Since these settings control the behavior of the OPDS client, it's not
enough to set them on the server -- the information has to be
communicated to the client. The Authentication For OPDS document is
the most convenient way to do this.

I consulted with Hadrien Gardeur to come up with a 'feature flags'
extension to Authentication For OPDS. Each feature has a URI, and you
can turn it on or off by including it in a dictionary that looks like
this:

```
"features": {
  "enabled" : [ "https://librarysimplified.org/rel/policy/reservations" ],
  "disabled" : [ "https://librarysimplified.org/rel/policy/display-scannable-barcode"]
}
```

The default value depends on the feature; this allows SimplyE to
always implement a sensible default.

I am confirming that the `https://librarysimplified.org/rel/policy/display-scannable-barcode` feature flag even needs to exist. If it does, the corresponding library setting needs to be created.

leonardr:
Feature flags are implemented; the barcode part is moved into https://github.com/NYPL-Simplified/circulation/issues/565.

Issue #559, opened 2017-07-03
Store color scheme as library setting, and advertise in Authentication for OPDS document


leonardr:
A SimplyE library can choose one of 7 color schemes:

* red
* blue
* gray
* yellow
* green
* teal
* purple

The specific colors used by these schemes are defined here: https://nypl.github.io/design-toolkit/sections/color.html

The admin interface should allow the user to select the color scheme, and provide examples. In fact, the selected color scheme could also be used to theme the admin interface itself.

The color scheme is inserted into the Authentication for OPDS document using the `color_scheme` extension, which I just made up:

`"color_scheme": "blue"`

PR #563, opened 2017-07-06
Settings for opds document

leonardr:
This branch adds a large number of configuration settings to each Library and to any ExternalIntegration that implements Basic Auth. These configuration settings are editable through the admin interface and are used to generate the Authentication for OPDS document, which I've extended to configure the behavior of a client in ways that are currently hard-coded in the SimplyE mobile app.

This fixes https://github.com/NYPL-Simplified/circulation/issues/561, https://github.com/NYPL-Simplified/circulation/issues/560, and https://github.com/NYPL-Simplified/circulation/issues/559. The only parts of those issues not addressed are the library icon and the barcode scanning format. I've split those issues into https://github.com/NYPL-Simplified/circulation/issues/566 and https://github.com/NYPL-Simplified/circulation/issues/565.

A library has the following new configuration options:

* Color scheme
* Description
* URL to the library web site (This is a preexisting configuration setting and all I did was start publishing it in the Authentication For OPDS feed.)
* Registration URL, i.e. where you go to get a library card.
* Feature flags. The only existing feature flag is "reservations".
* Three different types of "help" URL: an email address, a web page, and a custom integration using a custom URI scheme (designed for client-side API-based integrations such as HelpDesk).

An authentication mechanism has the following new configuration options:

* Keyboard to show when typing in patron identifier
* Keyboard to show when typing in patron password
* Maximum identifier length
* Maximum password length
* Label to use for identifier
* Label to use for password

The values of the labels are localized. It's a little unusual for us to localize the values of configuration settings, rather than constant strings, but I figure it will let us localize some common terms like "Barcode", "Library Card", "Email Address" and be able to handle them automatically. I don't know if there's a better way to do this, though.

Maximum identifier and password length are unique in this branch in that they also change the server-side behavior. `server_side_validation` enforces the maximum identifier and password lengths.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1379,7 +1379,6 @@ class SettingsController(CirculationManagerController):
                     return INTEGRATION_NAME_ALREADY_IN_USE
             auth_service.name = name
 
-        orig_protocol = protocol
         [protocol] = [p for p in protocols if p.get("name") == protocol]
         result = self._set_integration_settings_and_libraries(auth_service, protocol)
         if isinstance(result, ProblemDetail):
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1045,12 +1045,16 @@ class SettingsController(CirculationManagerController):
                     value = json.dumps(value)
             else:
                 value = flask.request.form.get(key)
-            if setting.get("options") and value not in [option.get("key") for option in setting.get("options")]:
-                self._db.rollback()
-                return INVALID_CONFIGURATION_OPTION.detailed(_(
-                    "The configuration value for %(setting)s is invalid.",
-                    setting=setting.get("label"),
-                ))
+            if value and setting.get("options"):
+                # This setting can only take on values that are in its
+                # list of options.
+                allowed = [option.get("key") for option in setting.get("options")]
+                if value not in allowed:
+                    self._db.rollback()
+                    return INVALID_CONFIGURATION_OPTION.detailed(_(
+                        "The configuration value for %(setting)s is invalid.",
+                        setting=setting.get("label"),
+                    ))
             if not value and not setting.get("optional"):
                 # Roll back any changes to the integration that have already been made.
                 self._db.rollback()
@@ -1058,7 +1062,7 @@ class SettingsController(CirculationManagerController):
                     _("The configuration is missing a required setting: %(setting)s",
                       setting=setting.get("label")))
             integration.setting(key).value = value
-
+                
         if not protocol.get("sitewide"):
             integration.libraries = []
 
@@ -1375,6 +1379,7 @@ class SettingsController(CirculationManagerController):
                     return INTEGRATION_NAME_ALREADY_IN_USE
             auth_service.name = name
 
+        orig_protocol = protocol
         [protocol] = [p for p in protocols if p.get("name") == protocol]
         result = self._set_integration_settings_and_libraries(auth_service, protocol)
         if isinstance(result, ProblemDetail):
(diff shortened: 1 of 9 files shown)


api/admin/controller.py
```diff
@@ -1045,20 +1045,24 @@ def _set_integration_settings_and_libraries(self, integration, protocol):
...
+            if value and setting.get("options"):
+                # This setting can only take on values that are in its
+                # list of options.
+                allowed = [option.get("key") for option in setting.get("options")]
+                if value not in allowed:
+                    self._db.rollback()
+                    return INVALID_CONFIGURATION_OPTION.detailed(_(
+                        "The configuration value for %(setting)s is invalid.",
+                        setting=setting.get("label"),
+                    ))
```
leonardr: I changed this so that an optional value with 'options' doesn't have to set a value.

api/authenticator.py
```diff
@@ -1528,15 +1667,18 @@ def remote_patron_lookup(self, access_token):
         """
         raise NotImplementedError()
     
-    def _internal_authenticate_url(self, library_short_name):
+    def _internal_authenticate_url(self, _db):
```
leonardr: This is how we get all the configuration settings when generating the authentication document. There used to be a single value that was passed in; now we get a database connection and look everything up.

api/authenticator.py
```diff
@@ -1010,17 +1057,72 @@ class BasicAuthenticationProvider(AuthenticationProvider):
...
+              { "key": NUMBER_PAD, "label": _("Number pad") },
+          ],
+          "default": DEFAULT_KEYBOARD
+        },
+        { "key": IDENTIFIER_MAXIMUM_LENGTH,
+          "label": _("Maximum identifier length"),
+          "optional": True,
+        },
+        { "key": PASSWORD_MAXIMUM_LENGTH,
+          "label": _("Maximum passowrd length"),
```
aslagle: you spelled password wrong

api/authenticator.py
```diff
@@ -1286,9 +1414,20 @@ def authentication_provider_document(self, library_short_name):
...
+
+        password_inputs = dict(keyboard=self.password_keyboard)
+        if self.password_maximum_length:
+            login_inputs['maximum_length'] = self.password_maximum_length
+
         method_doc = dict(
-            labels=dict(login=unicode(self.LOGIN_LABEL),
-                        password=unicode(self.PASSWORD_LABEL))
+            labels=dict(login=unicode(_(self.identifier_label)),
+                        password=unicode(_(self.password_label))),
```
aslagle: The localization here is a nice idea, but the translation files are generated from the code, so I don't know how this would work. It might ask the translator to translate "self.password_label". We should check what bin/util/generate_translation_template does with it.

If this like doesn't cause a problem, you could put constants like `COMMON_PASSWORD_LABELS = [_("Password"), _("PIN")]` somewhere so they'll be in the file.

aslagle:
I'm not sure about localizing the label values, but the rest of this looks great.

leonardr:
I changed it so that the most common values are localized ahead of time. If a label isn't in the pre-localized list, it doesn't get localized.

api/authenticator.py
```diff
@@ -1423,9 +1442,18 @@ def authentication_provider_document(self, _db):
...
+        )
+        localized_password_label = self.COMMON_PASSWORD_LABELS.get(
+            self.password_label,
+            self.password_label
+        )
         method_doc = dict(
-            labels=dict(login=unicode(_(self.identifier_label)),
-                        password=unicode(_(self.password_label))),
+            labels=dict(login=unicode(localized_identifier_label),
+                        password=unicode(_(localized_password_label))),
```
aslagle: This is a great solution. You missed removing the _ on this line though.

PR #564, opened 2017-07-06
Moved admin interface view to a controller

aslagle:
This branch updates the admin interface to a version that lets you select a library instead of keeping you on the default library. The code now redirects you to the default library if you're not on a library yet, and there's no longer a home url. I also updated the code to check if the local analytics provider is configured, and moved everything into a controller so I could test it.


leonardr:
:+1:

PR #567, opened 2017-07-06
Added support for setting a collection's parent in the admin interface.

aslagle:
I also fixed a bug from my last pr where you would sometimes get redirected to the default library from a config page in the admin interface.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -190,7 +190,7 @@ class AdminController(object):
         
 
 class ViewController(AdminController):
-    def __call__(self, collection, book):
+    def __call__(self, collection, book, path=None):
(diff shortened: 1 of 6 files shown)


api/admin/controller.py
```diff
@@ -208,7 +208,7 @@ def __call__(self, collection, book):
                         quoted_book.replace("/", "%2F"))
                 return redirect(self.url_for('admin_sign_in', redirect=redirect_url))
 
-            if not collection and not book:
+            if not collection and not book and not path:
```
leonardr: Note to self: what's path used for?

leonardr:
This is a great design! :+1:

PR #569, opened 2017-07-07
Allow variations in family name format

datalogics-dans:
In addition to "Last, First", the Millennium API can now recognize
"First Last" formatted names.

This change is due to the way certain libraries store patron info.
I have been told that we don't need to worry about "First Middle Last"
or other such variations.


datalogics-dans:
This is no good anymore.

PR #570, opened 2017-07-07
Language policy admin config

leonardr:
This branch moves into the database the most basic lane configuration: which languages in a library's should get the multi-lane "large collection" treatment, which should get the "small collection" treatment which distinguishes only between fiction and nonfiction, and which are 'tiny collections' that should be contained within a single lane.

Unlike other configuration settings, these values are set automatically the first time they're needed. I use the `estimated_holdings_by_language` method to get the raw data, and a new method that uses semi-arbitrary cutoffs for "large", "small", and "tiny".

Right now, this is serious overkill. The first time this value is needed, there will almost certainly be no titles in the collection, so the default (English gets a "large" collection and no other languages are presumed to exist) will always apply to start with. However in the future we can add a button to the admin interface that lets you reset the language settings based on the current collection, or shows you what they _would_ be at the moment and lets you change them yourself.

If you give an invalid value to one of these configuration settings, it redoes the estimate. This ensures that there's always a valid value and we can always create a set of lanes that make sense. I'm not really happy about this, but I figure it's a good backstop for whatever editor we put into place in the admin interface.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1043,10 +1043,15 @@ class SettingsController(CirculationManagerController):
 
         for setting in Configuration.LIBRARY_SETTINGS:
             if setting.get("type") == "list":
-                value = []
-                for option in setting.get("options"):
-                    if setting["key"] + "_" + option["key"] in flask.request.form:
-                        value += [option["key"]]
+                if setting.get('options'):
+                    # Restrict to the values in 'options'.
+                    value = []
+                    for option in setting.get("options"):
+                        if setting["key"] + "_" + option["key"] in flask.request.form:
+                            value += [option["key"]]
+                else:
+                    # Allow any entered values.
+                    value = [item for item in flask.request.form.getlist(key) if item]
                 value = json.dumps(value)
             else:
                 value = flask.request.form.get(setting['key'], None)
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1051,7 +1051,7 @@ class SettingsController(CirculationManagerController):
                             value += [option["key"]]
                 else:
                     # Allow any entered values.
-                    value = [item for item in flask.request.form.getlist(key) if item]
+                    value = [item for item in flask.request.form.getlist(setting.get('key')) if item]
                 value = json.dumps(value)
             else:
                 value = flask.request.form.get(setting['key'], None)
--- a/api/app.py
+++ b/api/app.py
@@ -23,10 +23,15 @@ app = Flask(__name__)
 
 testing = 'TESTING' in os.environ
 db_url = Configuration.database_url(testing)
-SessionManager.initialize(db_url)
+# Initialize a new database session unless we were told not to
+# (e.g. because a script already initialized it).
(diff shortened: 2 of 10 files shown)


aslagle:
Does this need a migration, or does using the estimate cover that?

api/config.py
```diff
@@ -151,6 +164,18 @@ class Configuration(CoreConfiguration):
...
+            "label": _("The primary languages represented in this library's collection")
+        },
+        {
+            "key": SMALL_COLLECTION_LANGUAGES,
+            "label": _("Other major languages represented in this library's collection")
+        },        
+        {
+            "key": TINY_COLLECTION_LANGUAGES,
+            "label": _("Other languages in this library's collection")
+        },
```
aslagle: You can add "type": "list" to these. The UI still needs work, but it lets you add each language you want in its own input box and end up with a JSON list in the db. Without that, you'd have to put a JSON string in the single input box.

leonardr: I think "type": "list" doesn't work here because that type is also expected to have an exhaustive list of "options" defined. I don't want to put up all the ISO-639-2 languages in a big list. (Although I can if it won't make the page unusable.)

aslagle: There's "type": "select" which has fixed options and "type": "list" where the UI lets you add your own things. The only other thing "type": "list" is used for is the allowed domains for admin authentication with google. Type "list" seems like the best option for now.

aslagle: Sorry, that's not quite right. Type "list" can have fixed options, or not, and if it doesn't have them you get to add your own. Type "select" only lets you pick one thing from fixed options.

api/app.py
```diff
@@ -23,10 +23,15 @@
...
+# Initialize a new database session unless we were told not to
+# (e.g. because a script already initialized it).
+autoinitialize = os.environ.get('AUTOINITIALIZE') != 'False'
+if autoinitialize:
+    SessionManager.initialize(db_url)
 session_factory = SessionManager.sessionmaker(db_url)
 _db = flask_scoped_session(session_factory, app)
-SessionManager.initialize_data(_db)
+if autoinitialize:
+    SessionManager.initialize_data(_db)
```
leonardr: At some point we took out the AUTOINITIALIZE hack (script tests turn off database initialization code because otherwise there's deadlock) because it didn't make a difference. Something in this branch (or possibly my recent core branch that changes how we get the database URL) jiggled it, and deadlock in script tests came back, so I restored the AUTOINITIALIZE hack.

api/admin/controller.py
```diff
@@ -1043,10 +1043,15 @@ def libraries(self):
...
-                        value += [option["key"]]
+                if setting.get('options'):
+                    # Restrict to the values in 'options'.
+                    value = []
+                    for option in setting.get("options"):
+                        if setting["key"] + "_" + option["key"] in flask.request.form:
+                            value += [option["key"]]
+                else:
+                    # Allow any entered values.
+                    value = [item for item in flask.request.form.getlist(key) if item]
```
aslagle: key isn't defined here, it should be setting.get("key")

PR #571, opened 2017-07-07
Fixed error when collection in CirculationManager.api_for_collection is not bound to session

aslagle:
This branch fixes the error
```
  File "/Users/amyslagle/dev/circulation/api/circulation.py", line 295, in borrow
    api.SET_DELIVERY_MECHANISM_AT == BaseCirculationAPI.BORROW_STEP)
AttributeError: 'NoneType' object has no attribute 'SET_DELIVERY_MECHANISM_AT'
```

I got this error when I was setting up a new circ manager, and trying to look at api_for_collection gave me ```*** DetachedInstanceError: Instance <Collection at 0x10bc20390> is not bound to a Session; attribute refresh operation cannot proceed```. Storing the collection id instead of the Collection seems to work.

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -221,7 +221,7 @@ class CirculationAPI(object):
                     )
                     self.initialization_exceptions[collection.id] = e
                 if api:
-                    self.api_for_collection[collection] = api
+                    self.api_for_collection[collection.id] = api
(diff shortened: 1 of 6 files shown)


leonardr:
Yep, this is exactly the kind of error we get and this is exactly the right fix. :+1:

PR #572, opened 2017-07-10
Feature add content source enki

datalogics-tarar:
Added support to ingest books from the Enki content provider via bin/enki_import and set the number of licenses for deleted books to 0 via bin/enki_reaper


api/enki.py
```diff
@@ -0,0 +1,638 @@
...
+    lib = 1
+
+    SET_DELIVERY_MECHANISM_AT = BaseCirculationAPI.BORROW_STEP
+    SERVICE_NAME = "Enki"
+
+    # may or may not be useful
+    DATE_FORMAT = "%m-%d-%Y %H:%M:%S"
+
+    log = logging.getLogger("Enki API")
+    # TODO: make sure this logger exists :-)
```
leonardr: A logger that doesn't exist will be created on request, so this is fine.

api/enki.py
```diff
@@ -0,0 +1,638 @@
...
+    PRODUCTION_BASE_URL = "http://enkilibrary.org/API/"
+    availability_endpoint = "ListAPI"
+    item_endpoint = "ItemAPI"
+    lib = 1
+
+    SET_DELIVERY_MECHANISM_AT = BaseCirculationAPI.BORROW_STEP
+    SERVICE_NAME = "Enki"
+
+    # may or may not be useful
+    DATE_FORMAT = "%m-%d-%Y %H:%M:%S"
```
leonardr: Looks like you ended up not needing this.

api/enki.py
```diff
@@ -0,0 +1,638 @@
...
+                 base_url=None):
+        self._db = _db
+        (env_library_id, env_username,
+         env_password, env_base_url) = self.environment_values()
+        self.library_id = library_id or env_library_id
+        self.username = username or env_username
+        self.password = password or env_password
+        self.base_url = base_url or env_base_url
+        if self.base_url == 'qa':
+            self.base_url = self.QA_BASE_URL
```
leonardr: You never define QA_BASE_URL.

api/enki.py
datalogics-tarar: I'll delete that section.

api/enki.py
datalogics-tarar: I'll delete that section.

.gitmodules
```diff
@@ -1,3 +1,4 @@
 [submodule "core"]
 	path = core
-	url = git@github.com:NYPL-Simplified/server_core.git
+	url = git@github.com:datalogics-tarar/server_core.git
+	branch = feature-add-content-source-enki-core
```
leonardr: I understand why you did this during development, but it needs to switch back before merging into master.

api/enki.py
```diff
@@ -0,0 +1,633 @@
...
+            self.token = None
+            return self.request(
+                url=url, method=method, extra_headers=extra_headers,
+                data=data, params=params, exception_on_401=True
+            )
+        else:
+            return response
+
+    def availability(self, patron_id=None, since=None, title_ids=[], strt=0, qty=2000):
+        print "requesting : "+ str(qty) + " books starting at econtentRecord" +  str(strt)
```
leonardr: It's best to either take out print statements before merging or convert them to `self.log.debug` statements.

api/enki.py
```diff
@@ -0,0 +1,633 @@
...
+        url = str(self.base_url) + str(self.item_endpoint)
+        args = dict()
+        args['method'] = "getItem"
+        args['recordid'] = identifier
+        args['size'] = "small"
+        args['lib'] = self.lib
+        response = self.request(url, method='get', params=args)
+        if not(response.content.startswith("{\"result\":{\"id\":\"")):
+            response = None
+            print "This book is no longer available."
```
leonardr: This should either be removed or become a `self.log.info` statement that mentions which book is no longer available.

api/enki.py
```diff
@@ -0,0 +1,633 @@
...
+
+        # Create a CoverageFailure object for each original identifier
+        # not mentioned in the results.
+        for identifier_string in identifier_strings:
+            if identifier_string not in seen_identifiers:
+                identifier, ignore = Identifier.for_foreign_id(
+                    self._db, Identifier.ENKI_ID, identifier_string
+                )
+                result = CoverageFailure(
+                    identifier, "Book not in collection", data_source=self.output_source, transient=False
```
leonardr: This should be treated as a transient error, since the book might be absent from Enki temporarily.

I would say "Book not found in ENKI" rather than "Book not in collection", since "not in collection" sounds like the book isn't in the collection on the local side -- the problem here is that the book _is_ in the collection on the local side but not on the remote.

api/enki.py
```diff
@@ -0,0 +1,633 @@
...
+            metadata_replacement_policy=metadata_replacement_policy,
+            **kwargs
+        )
+
+    def process_batch(self, identifiers):
+        identifier_strings = self.api.create_identifier_strings(identifiers)
+        response = self.api.availability(title_ids=identifier_strings)
+        seen_identifiers = set()
+        batch_results = []
+        for metadata, availability in self.parser.process_all(response.content, "//enki:title"):
```
leonardr: I don't think you need `//enki:title` since you're parsing JSON, not XML.

api/enki.py
```diff
@@ -0,0 +1,633 @@
...
+	    data = self.process_one(book, namespaces)
+            if data:
+                yield data
+
+    DELIVERY_DATA_FOR_AXIS_FORMAT = {
+        "Blio" : None,
+        "Acoustik" : None,
+        "ePub" : (Representation.EPUB_MEDIA_TYPE, DeliveryMechanism.ADOBE_DRM),
+        "PDF" : (Representation.PDF_MEDIA_TYPE, DeliveryMechanism.ADOBE_DRM),
+    }
```
leonardr: It looks like you don't use this at all.

datalogics-tarar:
@leonardr I'll make those changes and also implement some changes Enki just made to their ingestion API

api/enki.py
```diff
@@ -0,0 +1,633 @@
...
+    role_abbreviation_to_role = dict(
+        INT=Contributor.INTRODUCTION_ROLE,
+        EDT=Contributor.EDITOR_ROLE,
+        PHT=Contributor.PHOTOGRAPHER_ROLE,
+        ILT=Contributor.ILLUSTRATOR_ROLE,
+        TRN=Contributor.TRANSLATOR_ROLE,
+        FRW=Contributor.FOREWORD_ROLE,
+        ADP=generic_author, # Author of adaptation
+        COR=generic_author, # Corporate author
+    )
```
leonardr: Does ENKI happen use the same format for describing contributor roles as Axis does, or is this an artifact of copy-and-paste?

api/enki.py
```diff
@@ -0,0 +1,633 @@
...
+        self.include_availability = include_availability
+        self.include_bibliographic = include_bibliographic
+
+    def extract_availability(self, circulation_data, element, ns):
+	primary_identifier = IdentifierData(Identifier.ENKI_ID, element["id"])
+        if not circulation_data:
+            circulation_data = CirculationData(
+                data_source=DataSource.ENKI,
+                primary_identifier=primary_identifier,
+            )
```
leonardr: When you create a `CirculationData` object you need to include `FormatData` objects that make it clear what media types/DRM schemes the book is actually available in. Otherwise it will be treated as a book that's licensed but not available in any format. 

The last time I dealt with Enki it seemed like all books were in EPUB format with the Adobe DRM scheme, but I don't know if that's always the case.

api/enki.py
```diff
@@ -0,0 +1,633 @@
...
+
+    """Helper function to parse JSON"""
+    def process_all(self, json_data, xpath, namespaces=None, handler=None, parser=None):
+        data = json.loads(json_data)
+	returned_titles = data["result"]["titles"]
+	titles = returned_titles
+	for book in returned_titles:
+            print "A book titled '%s'" % book["title"]
+            print book
+            print "\n"
```
leonardr: Again, print statements need to go or be turned into log statements.

api/enki.py
datalogics-tarar: That is the case.

api/enki.py
```diff
@@ -0,0 +1,633 @@
...
+    def run(self):
+        super(EnkiImport, self).run()
+
+    def run_once(self, start, cutoff):
+        # Give us five minutes of overlap because it's very important
+        # we don't miss anything.
+        since = start-self.FIVE_MINUTES
+        x=0
+        step=2000
+        while x < 80000:
```
leonardr: Why 80,000? Can you keep going until `self.api.availability` stops returning results?

api/enki.py
```diff
@@ -0,0 +1,633 @@
...
+	    status_code = availability.status_code
+            content = availability.content
+            count = 0
+            for bibliographic, circulation in BibliographicParser().process_all(
+                    content, "//enki:title"):
+                self.process_book(bibliographic, circulation)
+                count += 1
+                if count % self.batch_size == 0:
+                    self._db.commit()
+            x += step
```
leonardr: `step` and `self.batch_size` are redundant. Just use self.batch_size and set the default batch size to 2000.

api/enki.py
```diff
@@ -0,0 +1,633 @@
...
+        self.batch_size = batch_size
+        #line 83-90 should be removable during refactoring
+        metadata_wrangler_url = Configuration.integration_url(
+                Configuration.METADATA_WRANGLER_INTEGRATION
+        )
+        if metadata_wrangler_url:
+            self.metadata_wrangler = SimplifiedOPDSLookup(metadata_wrangler_url)
+        else:
+            # This should only happen during a test.
+            self.metadata_wrangler = None
```
leonardr: You never use `self.metadata_wrangler`, so you should be able to remove the setup altogether.

api/enki.py
```diff
@@ -0,0 +1,633 @@
...
+    def __init__(self, _db, api=None, interval_seconds=3600*4):
+        super(EnkiCollectionReaper, self).__init__(_db, "Enki Collection Reaper", interval_seconds)
+        self._db = _db
+        if not api:
+            api = EnkiAPI.from_environment(_db)
+        self.api = api
+        self.data_source = DataSource.lookup(self._db, DataSource.ENKI)
+
+    def run(self):
+        self.api = EnkiAPI.from_environment(self._db)
```
leonardr: You shouldn't need to set self.api again--it was set in the constructor, potentially to a mocked value.

api/enki.py
```diff
@@ -0,0 +1,633 @@
...
+
+        identifiers_not_mentioned_by_enki= set(identifiers)
+        now = datetime.datetime.utcnow()
+
+        for identifier in identifiers:
+            result = self.api.reaper_request(identifier.identifier)
+            if not result:
+                print "skipping this deleted book"
+                continue
+            print "keeping this existing book"
```
leonardr: These print statements should be removed IMO. `reaper_request` contains the informatino you would want to keep in the logs.

api/enki.py
```diff
@@ -0,0 +1,633 @@
...
+            pool.licenses_owned = 0
+            pool.licenses_available = 0
+            pool.licenses_reserved = 0
+            pool.patrons_in_hold_queue = 0
+            pool.last_checked = now
+
+class ResponseParser(BibliographicParser):
+    id_type = Identifier.ENKI_ID
+
+    SERVICE_NAME = "Enki"
```
leonardr: I think this can go into BibliographicParser, you don't need a subclass.

tests/files/enki/empty_item_circulation.xml
```diff
@@ -0,0 +1 @@
+<ArrayOfItemCirculation xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" />
```
leonardr: This is a Bibliotheca response, which makes me think this possibility isn't tested.

datalogics-tarar: Yeah, we're in the process of writing tests. I'll delete that file.

tests/test_enki.py
```diff
@@ -0,0 +1,158 @@
...
+class TestEnkiAPI(DatabaseTest, BaseEnkiTest):
+
+    def setup(self):
+        super(TestEnkiAPI, self).setup()
+        self.api = MockEnkiAPI(self._db)
+
+    def test_create_identifier_strings(self):
+        identifier = self._identifier()
+        values = EnkiAPI.create_identifier_strings(["foo", identifier])
+        eq_(["foo", identifier.identifier], values)
```
leonardr: Also check `identifier.type`.

tests/test_enki.py
```diff
@@ -0,0 +1,158 @@
...
+        error = parser.process_all(self.TRIED_TO_RETURN_UNLOANED_BOOK)
+        assert isinstance(error, NotCheckedOut)
+
+        error = parser.process_all(self.TRIED_TO_HOLD_LOANABLE_BOOK)
+        assert isinstance(error, CurrentlyAvailable)
+
+        # This is such a weird case we don't have a special
+        # exception for it.
+        error = parser.process_all(self.TRIED_TO_HOLD_BOOK_ON_LOAN)
+        assert isinstance(error, CannotHold)
```
leonardr: These all look like Bibliotheca errors.

api/enki.py
datalogics-tarar: Question. Where do info and debug logs go? I noticed warn logs show up in the console but debug and info don't and I also don't see a log file.

datalogics-dans: @datalogics-tarar They go to the uwsgi.log file.

api/enki.py
```diff
@@ -0,0 +1,556 @@
...
+
+        # Create a CoverageFailure object for each original identifier
+        # not mentioned in the results.
+        for identifier_string in identifier_strings:
+            if identifier_string not in seen_identifiers:
+                identifier, ignore = Identifier.for_foreign_id(
+                    self._db, Identifier.ENKI_ID, identifier_string
+                )
+                result = CoverageFailure(
+                    identifier, "Book not found in Enki", data_source=self.output_source, transient=False
```
leonardr: This should be a transient failure because the book might show up in this collection later.

api/enki.py
```diff
@@ -0,0 +1,556 @@
...
+    def process_item(self, identifier):
+        results = self.process_batch([identifier])
+        return results[0]
+
+class BibliographicParser(object):
+
+    """Helper function to parse JSON"""
+    def process_all(self, json_data):
+        data = json.loads(json_data)
+	returned_titles = data["result"]["titles"]
```
leonardr: I don't know how big a problem this will cause, but there's an indentation mismatch here.

api/enki.py
```diff
@@ -0,0 +1,556 @@
...
+
+    def extract_bibliographic(self, element):
+        identifiers = []
+        contributors = []
+        identifiers.append(IdentifierData(Identifier.ISBN, element["isbn"]))
+        sort_name = element["author"]
+        if not sort_name:
+            sort_name = "Unknown"
+ 	contributors.append(ContributorData(sort_name=sort_name))
+        primary_identifier = IdentifierData(Identifier.ENKI_ID, element["id"])
```
leonardr: Another indentation mismatch.

api/enki.py
```diff
@@ -0,0 +1,556 @@
...
+        self.include_bibliographic = include_bibliographic
+
+    def extract_bibliographic(self, element):
+        identifiers = []
+        contributors = []
+        identifiers.append(IdentifierData(Identifier.ISBN, element["isbn"]))
+        sort_name = element["author"]
+        if not sort_name:
+            sort_name = "Unknown"
+ 	contributors.append(ContributorData(sort_name=sort_name))
```
leonardr: The contributor should have a role, probably `Contributor.PRIMARY_AUTHOR_ROLE`.

api/enki.py
```diff
@@ -0,0 +1,556 @@
...
+
+    def process_one(self, element):
+        if self.include_bibliographic:
+            bibliographic = self.extract_bibliographic(element)
+        else:
+            bibliographic = None
+
+        passed_availability = None
+        if bibliographic and bibliographic.circulation:
+            passed_availability = bibliographic.circulation
```
leonardr: It doesn't look like this code does anything.

api/enki.py
```diff
@@ -0,0 +1,556 @@
...
+        self.bibliographic_coverage_provider = (
+            EnkiBibliographicCoverageProvider(self._db, enki_api=api)
+        )
+
+    def run_once(self, start, cutoff):
+        # Give us five minutes of overlap because it's very important
+        # we don't miss anything.
+        since = start-self.FIVE_MINUTES
+        id_start = 0
+        while True:
```
leonardr: This part needs work. I won't comment here because it feels unfinished but let me know if you want comments.

api/enki.py
```diff
@@ -0,0 +1,556 @@
...
+        self.data_source = DataSource.lookup(self._db, DataSource.ENKI)
+
+    def run(self):
+        super(EnkiCollectionReaper, self).run()
+
+    def identifier_query(self):
+        return self._db.query(Identifier).filter(
+            Identifier.type==Identifier.ENKI_ID)
+
+    def process_batch(self, identifiers):
```
leonardr: Since you're asking about one item at a time, I recommend copying `OverdriveCollectionReaper`. It'll let you get rid of a lot of code and also make this code easier to port to the 2.0 monitoring system. All this batch handling code is for Bibliotheca, where you can ask about 25 items at a time.

api/enki.py
datalogics-dans: Fixed the few things I'm assuming you're talking about in a new commit.

api/enki.py
```diff
@@ -0,0 +1,554 @@
...
+                    "Could not contact Enki server for content availability. Status: %d",
+                    availability.status_code
+                )
+            content = availability.content
+            count = 0
+            for bibliographic, circulation in BibliographicParser().process_all(content):
+                self.process_book(bibliographic, circulation)
+                count += 1
+            if count == 0:
+                break
```
leonardr: You should break after you commit to the database.

datalogics-dans: For my own understanding, does it make a difference? If process_book never runs, is there anything to commit? I figured I'd save on an extra commit operation if it wasn't going to do anything.

leonardr: You're right, no work happens on this loop so the commit won't do anything. This code is fine as is.

datalogics-dans: 👍

api/enki.py
```diff
@@ -218,8 +218,23 @@ def reaper_request(self, identifier):
         args['lib'] = self.lib
         response = self.request(url, method='get', params=args)
         if not(response.content.startswith("{\"result\":{\"id\":\"")):
-            response = None
-        return response
+            pool = identifier.licensed_through
+            if pool & (pool.licenses_owned > 0):
```
datalogics-dans: This should be "and", not "&".

api/enki.py
```diff
@@ -218,8 +218,23 @@ def reaper_request(self, identifier):
...
+                else:
+                    self.log.warn(
+                        "Removing unknown work %s from circulation.",
+                        identifier.identifier
+                    )
+            pool.licenses_owned = 0
+            pool.licenses_available = 0
+            pool.licenses_reserved = 0
+            pool.patrons_in_hold_queue = 0
+            pool.last_checked = now
```
datalogics-dans: Now needs to be defined, e.g. "datetime.datetime.utcnow()".

tests/test_enki.py
```diff
@@ -106,3 +106,39 @@ def test_process_item_creates_presentation_ready_work(self):
...
+        # We have never checked the circulation information for this
+        # LicensePool. Put some random junk in the pool to verify
+        # that it gets changed.
+        pool.licenses_owned = 10
+        pool.licenses_available = 5
+        pool.patrons_in_hold_queue = 3
+        pool.identifier.identifier = nonexistent_id
+        eq_(None, pool.last_checked)
+
+        # Modify the data so that it appears to be talking about the
```
datalogics-dans: Yeah, I should probably clean up this copy/paste job a bit...

datalogics-tarar:
@leonardr I'm not sure what to test for EnkiAPI.availability since all that function does it make an API call. The response is parsed elsewhere. I see that the other content provider tests are reading data from files rather than actually making API calls. Thoughts?

api/enki.py
```diff
@@ -0,0 +1,521 @@
...
+
+    def availability(self, patron_id=None, since=None, title_ids=[], strt=0, qty=2000):
+        self.log.debug ("requesting : "+ str(qty) + " books starting at econtentRecord" +  str(strt))
+        url = str(self.base_url) + str(self.availability_endpoint)
+        args = dict()
+	args['method'] = "getAllTitles"
+	args['id'] = "secontent"
+        args['strt'] = strt
+        args['qty'] = qty
+        args['lib'] = self.lib
```
leonardr: Availability builds a URL and parameters for an HTTP request. Some parameters, like `method`, are set to constant strings, and some are taken from input. This method contains knowledge about how the Enki API works and that knowledge--not the results of making the request--is what I'd like to see tested.

It's not as important as testing other stuff, but if you were to test it that's what I suggest you test.

datalogics-tarar: sounds good

leonardr: Again, it's not super important that this be tested, but in this situation I find it's useful to create a helper method that creates the data structures (URL and args, in this case). Have the method that makes the request call the helper method, test the helper method and leave the request method alone.

This lets you encode information about how the Enki API works, so that you'll find out immediately if another argument is ever introduced in the code.

datalogics-dans:
@leonardr I'm not sure what else to test for the Reaper class either. Do you know of any existing tests for the other sources that test the run() or process_batch() functions that I could look at?

leonardr:
It's still not complete, but 2.0 (current HEAD) has much better test coverage of monitors than 1.x. I would take a look at https://github.com/NYPL-Simplified/server_core/blob/master/tests/test_monitor.py and https://github.com/NYPL-Simplified/circulation/blob/master/tests/test_monitor.py.

That said, you probably won't have much code to test once you start using the new system. Hopefully you'll only have to test `process_item`. This can probably wait until you port to 2.0, since you'll have to port both the code and the tests.

datalogics-dans:
Pretty sure we can close this PR now that we're just working on the port to 2.0.

PR #573, opened 2017-07-10
multi tenancy fixes

aslagle:
This fixes a bunch of problems I found when setting up a circ manager with a couple libraries.


aslagle:
Actually just one problem, but it also brings in https://github.com/NYPL-Simplified/server_core/pull/567 and https://github.com/NYPL-Simplified/circulation-web/pull/98

leonardr:
:+1:

PR #574, opened 2017-07-10
Allow family name auth to parse "First Last"

datalogics-dans:
Patron names won't always be stored as "Lastname, Firstname" so we need to account for the major alternative, "Firstname Lastname".

### DIFF
--- a/api/millenium_patron.py
+++ b/api/millenium_patron.py
@@ -171,7 +171,10 @@ class MilleniumPatronAPI(BasicAuthenticationProvider, XMLParser):
         """Does `supposed_family_name` match `actual_name`?"""
         if actual_name is None or supposed_family_name is None:
             return False
-        actual_family_name = actual_name.split(',')[0]
+        if acutal_name.find(',') != -1:
+            actual_family_name = actual_name.split(',')[0]
+        else:
+            actual_family_name = actual_name.split(' ')[1]
         if actual_family_name.upper() == supposed_family_name.upper():
             return True
         return False
--- a/api/millenium_patron.py
+++ b/api/millenium_patron.py
@@ -171,7 +171,7 @@ class MilleniumPatronAPI(BasicAuthenticationProvider, XMLParser):
         """Does `supposed_family_name` match `actual_name`?"""
         if actual_name is None or supposed_family_name is None:
             return False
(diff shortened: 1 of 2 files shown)


leonardr:
:+1:

api/millenium_patron.py
```diff
@@ -171,7 +171,10 @@ def family_name_match(self, actual_name, supposed_family_name):
         """Does `supposed_family_name` match `actual_name`?"""
         if actual_name is None or supposed_family_name is None:
             return False
-        actual_family_name = actual_name.split(',')[0]
+        if acutal_name.find(',') != -1:
```
leonardr: Typo for `actual_name` is causing test failures.

api/millenium_patron.py
```diff
@@ -171,7 +171,10 @@ def family_name_match(self, actual_name, supposed_family_name):
         """Does `supposed_family_name` match `actual_name`?"""
         if actual_name is None or supposed_family_name is None:
             return False
-        actual_family_name = actual_name.split(',')[0]
+        if actual_name.find(',') != -1:
+            actual_family_name = actual_name.split(',')[0]
+        else:
+            actual_family_name = actual_name.split(' ')[1]
```
leonardr: This fails on the empty string; it will also fail the "CHER" test where the patron has only one name. Check for a space before you split, similar to how you check for a comma before you split.

api/millenium_patron.py
```diff
@@ -171,7 +171,11 @@ def family_name_match(self, actual_name, supposed_family_name):
         """Does `supposed_family_name` match `actual_name`?"""
         if actual_name is None or supposed_family_name is None:
             return False
-        actual_family_name = actual_name.split(',')[0]
+        if actual_name.find(',') != -1:
+            actual_family_name = actual_name.split(',')[0]
+        else:
+            actual_name_split = actual_name.split(' ')
+            actual_family_name = actual_name_split[len(actual_name_split)-1]
```
leonardr: This makes sense; you should use the shorthand `actual_name_split[-1]` here and add a test for a name like "Anthony Michael Hall".

datalogics-dans: That's a good tip for someone who's rusty at Python.

PR #575, opened 2017-07-12
Added descriptions and improved wording and input types for the admin interface.

aslagle:
This adds more documentation of settings into the admin interface itself, mostly copied from the wiki or code comments, and fixes types and defaults for some of the settings.

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -784,10 +784,18 @@ class AuthenticationProvider(object):
     LIBRARY_SETTINGS = [
         { "key": EXTERNAL_TYPE_REGULAR_EXPRESSION,
           "label": _("External Type Regular Expression"),
+          "description": _("Derive a patron's external type from their identifier."),
           "optional": True,
         },
         { "key": PATRON_IDENTIFIER_RESTRICTION,
           "label": _("Patron identifier prefix"),
+          "description": _("<p>When multiple libraries share an ILS, a person may be able to " +
+                           "authenticate with the ILS but not be considered a patron of " +
+                           "<i>this</i> library. This setting contains the rule for determining " +
+                           "whether an identifier is valid for a specific library.</p>" +
+                           "<p>Usually this is a prefix string which is compared against the " +
+                           "patron's identifiers, but if the string starts with a carat (^) " +
+                           "it will be interpreted as a regular expression."),
           "optional": True,
         }
     ]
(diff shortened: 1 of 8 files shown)


api/authenticator.py
```diff
@@ -784,10 +784,18 @@ class AuthenticationProvider(object):
     LIBRARY_SETTINGS = [
         { "key": EXTERNAL_TYPE_REGULAR_EXPRESSION,
           "label": _("External Type Regular Expression"),
+          "description": _("Derive a patron's external type from their identifier."),
```
aslagle: Maybe this should explain what an external type is. Since the setting to map external types to lanes isn't in the admin interface yet there's nowhere else you would see it.

api/config.py
```diff
@@ -151,18 +152,22 @@ class Configuration(CoreConfiguration):
         {
             "key": REGISTER,
             "label": _("Patron registration URL"),
+            "description": _("A URL where someone who doesn't have a library card yet can sign up for one."),
         },
         {
             "key": HELP_EMAIL,
             "label": _("Patron support email address"),
+            "description": _("An email address a patron can use if they need help, e.g. 'mailto:simplyehelp@nypl.org'."),
```
aslagle: I'm not sure whether or not this is supposed to include the 'mailto'.

api/authenticator.py
leonardr: Maybe we can call this a "patron type" in user-facing documentation.

api/millenium_patron.py
```diff
@@ -79,8 +79,14 @@ class MilleniumPatronAPI(BasicAuthenticationProvider, XMLParser):
...
         },
-        { "key": BLOCK_TYPES, "label": _("Block types"), "optional": True },
-        { "key": IDENTIFIER_BLACKLIST, "label": _("Identifier Blacklist"), "optional": True },
+        { "key": BLOCK_TYPES, "label": _("Block types"),
+          "description": _("Values of MBLOCK[p56] which mean a patron is blocked. By default, any value other than '-' indicates a block."),
+          "optional": True },
+        { "key": IDENTIFIER_BLACKLIST, "label": _("Identifier Blacklist"),
+          "type": "list",
+          "description": _("Identifiers containing any of these strings are ignored when finding the 'correct' " +
+                           "identifier for a patron's record, even if it means they end up with no identifier at all."),
```
aslagle: This might need a better description of when you would want to use it.

api/authenticator.py
```diff
@@ -784,10 +784,18 @@ class AuthenticationProvider(object):
...
           "optional": True,
         },
         { "key": PATRON_IDENTIFIER_RESTRICTION,
           "label": _("Patron identifier prefix"),
+          "description": _("<p>When multiple libraries share an ILS, a person may be able to " +
+                           "authenticate with the ILS but not be considered a patron of " +
+                           "<i>this</i> library. This setting contains the rule for determining " +
+                           "whether an identifier is valid for a specific library.</p>" +
```
leonardr: "for this specific library" rather than "for a specific library"

api/authenticator.py
```diff
@@ -1104,12 +1112,21 @@ class BasicAuthenticationProvider(AuthenticationProvider):
...
-        { "key": TEST_PASSWORD, "label": _("Test Password") },
-        { "key": IDENTIFIER_REGULAR_EXPRESSION, "label": _("Identifier Regular Expression"), "optional": True },
-        { "key": PASSWORD_REGULAR_EXPRESSION, "label": _("Password Regular Expression"), "optional": True },
+        { "key": TEST_IDENTIFIER,
+          "label": _("Test Identifier"),
+          "description": _("A valid identifier that can be used to test that patron authentication is working.") },
+        { "key": TEST_PASSWORD, "label": _("Test Password"), "description": _("The password for the test identifier.") },
+        { "key": IDENTIFIER_REGULAR_EXPRESSION,
+          "label": _("Identifier Regular Expression"),
+          "description": _("Identifiers will be rejected if they don't match this regular expression."),
```
leonardr: "Patron identifiers"

api/authenticator.py
```diff
@@ -1104,12 +1112,21 @@ class BasicAuthenticationProvider(AuthenticationProvider):
...
+          "optional": True },
+        { "key": PASSWORD_REGULAR_EXPRESSION,
+          "label": _("Password Regular Expression"),
+          "description": _("Passwords will be rejected if they don't match this regular expression."),
```
leonardr: "A patron's password will be immediately rejected if it doesn't match..."

api/millenium_patron.py
leonardr: Add "If librarians invalidate library cards by adding strings like "EXPIRED" or "INVALID" on to the beginning of the card number, put those strings here so the Circulation Manager knows they do not represent real card numbers."

leonardr:
:+1: with some minor wording changes.

PR #587, opened 2017-07-14
Fix migrations for 2.0.0

courte:
This includes changes from master and resorts / fixes migrations to make sure they run in the switch from proposed 1.1.23 to multi-tenancy releases.

Supported by NYPL-Simplified/server_core#573


api/admin/password_admin_authentication_provider.py
```diff
@@ -19,12 +19,12 @@ def sign_in(self, _db, request={}):
         password = request.get("password")
```
courte: Changes from here to my next comment have been merged into master.

migration/20170713-11-move-configuration-links-into-db.py
```diff
@@ -29,7 +29,7 @@
         ("license", Configuration.get('links', {}).get('license', None)),
```
courte: New changes begin here.

tests/admin/test_controller.py
```diff
@@ -84,7 +84,7 @@ def setup(self):
     def test_setting_up(self):
```
courte: All changes in this file are from master.

PR #588, opened 2017-07-14
Added library logo uploading to the admin interface.

aslagle:
This doesn't have any validation of the image yet, or even check that the file is an image. I'm not sure what the requirements are.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1053,6 +1053,15 @@ class SettingsController(CirculationManagerController):
                     # Allow any entered values.
                     value = [item for item in flask.request.form.getlist(setting.get('key')) if item]
                 value = json.dumps(value)
+            elif setting.get("type") == "image":
+                image_file = flask.request.files.get(setting.get("key"))
+                if not image_file and not setting.get("optional"):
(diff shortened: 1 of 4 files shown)


leonardr:
Requirements: GIF/PNG/JPG, must be approximately square, no larger than 135x135 pixels, look good on a white background.

Here are some examples I got from Connecticut libraries.

image
image

api/admin/controller.py
```diff
@@ -1053,6 +1053,15 @@ def libraries(self):
...
                     value = [item for item in flask.request.form.getlist(setting.get('key')) if item]
                 value = json.dumps(value)
+            elif setting.get("type") == "image":
+                image_file = flask.request.files.get(setting.get("key"))
+                if not image_file and not setting.get("optional"):
+                    return INCOMPLETE_CONFIGURATION.detailed(_(
+                            "The library is missing a required setting: %s." % setting.get("key")))
+                if image_file:
+                    type = image_file.headers.get("Content-Type")
+                    b64 = base64.b64encode(image_file.read())
```
leonardr: It looks like this would be a good place to enforce the media type restriction.

PR #592, opened 2017-07-17
Stabilize 2.0 migration scripts

leonardr:
This branch fixes two errors I found while migrating Open Ebooks from 1.1.23 to 2.0.

### DIFF
--- a/migration/20170713-15-move-library-configuration-to-configurationsettings.py
+++ b/migration/20170713-15-move-library-configuration-to-configurationsettings.py
@@ -80,15 +80,14 @@ try:
         for k, v in default.items():
             library.default_facet_setting(k).value = v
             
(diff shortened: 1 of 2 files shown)


migration/20170713-15-move-library-configuration-to-configurationsettings.py
```diff
@@ -80,15 +80,14 @@
         for k, v in default.items():
             library.default_facet_setting(k).value = v
             
-    # Copy external type regular expression into each collection for each
-    # library.
+    # Copy external type regular expression into each authentication
+    # mechanism for each library.
```
leonardr: The external type regular expression is attached to the library's authentication mechanism, not to the collections. I only found this because it failed in an unexpected way trying to set this value on the Open Access Content Server collection.

PR #595, opened 2017-07-18
Warn on missing series tag

leonardr:
This branch fixes a crash during OPDS entry annotation that I don't think will happen on a real site. If a Work has a series but its OPDS entry doesn't have a <schema:Series> tag, `add_series_link()` will crash when it can't find that tag.

This should only happen if the OPDS entry hasn't been rewritten for a very long time (like the ones on Open Ebooks QA) so I think this is a good enough fix.

### DIFF
--- a/api/opds.py
+++ b/api/opds.py
@@ -407,7 +407,7 @@ class CirculationManagerAnnotator(Annotator):
     def add_series_link(self, work, feed, entry):
         series_tag = OPDSFeed.schema_('Series')
         series_entry = entry.find(series_tag)
-        if not series_entry:
+        if series_entry is None:
             # There is no <series> tag, and thus nothing to annotate.
             # This probably indicates an out-of-date OPDS entry.
             if isinstance(work, Work):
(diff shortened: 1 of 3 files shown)


api/opds.py
```diff
@@ -401,22 +407,37 @@ def add_author_links(self, work, feed, entry):
...
+                work_id = work.id
+                work_title = work.title
+            else:
+                work_id = work.works_id
+                work_title = work.sort_title
+            self.log.error(
+                'add_series_link() called on work %s ("%s"), which has no <schema:Series> tag in its OPDS entry.',
+                work_id, work_title
+            )
+            return
```
courte: Is the return here to overlook it until the OPDS entries are recalculated?

leonardr: Yeah, since this should never happen I decided it was simpler to leave the entry alone rather than try to regenerate the entry on the spot.

If it does happen, there's no guarantee that regenerating the OPDS entry will fix the problem.

PR #599, opened 2017-07-18
Add more return statements so unnecessary integrations aren't created.

leonardr:
Same code as https://github.com/NYPL-Simplified/circulation/pull/598, just trying to fix it everywhere so it doesn't crop up again.

### DIFF
--- a/migration/20170713-13-move-authentication-configuration-to-external-integrations.py
+++ b/migration/20170713-13-move-authentication-configuration-to-external-integrations.py
@@ -104,8 +104,8 @@ def convert_clever(_db, integration, provider):
         integration.setting(OAuthAuthenticationProvider.TOKEN_EXPIRATION_DAYS
         ).value = expiration_days
     
-configuration = Configuration.load()
-if not configuration:
+Configuration.load()
+if not Configuration.instance:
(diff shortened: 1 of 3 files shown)


migration/20170713-13-move-authentication-configuration-to-external-integrations.py
```diff
@@ -104,8 +104,8 @@ def convert_clever(_db, integration, provider):
         integration.setting(OAuthAuthenticationProvider.TOKEN_EXPIRATION_DAYS
         ).value = expiration_days
     
-configuration = Configuration.load()
-if not configuration:
+Configuration.load()
+if not Configuration.instance:
```
leonardr: A separate problem I found when migrating NYPL's database and discovering the authentication didn't get migrated. `Configuration.load()` doesn't return anything.

courte: This works fine, but on core at this point, `Configuration.load()` is supposed to be returning the configuration instance. Is that not happening?

leonardr:
core's `Configuration.load()` returns the object, but circulation's `Configuration.load()` doesn't. I agree it should be consistent so I'll make the change in this branch.

PR #603, opened 2017-07-21
Merge with master

leonardr:
This branch fixes merge conflicts with master in preparation for merging multi-tenant into master.


.travis.yml
```diff
@@ -41,4 +41,4 @@ before_script:
 env:
  - SIMPLIFIED_TEST_DATABASE="postgres://simplified_test:test@localhost:5432/simplified_circulation_test"
 
-script: ./test
```
courte: If you add `$1` to the end of the `test` script, you won't need to make another file and also specific test requests or other options can be passed in. It won't be as clear why you're doing it, though.

Issue #577, opened 2017-07-12
Manual registration with library registry


leonardr:
The admin interface should have a controller that registers this library with the currently configured library registry and shows the results of the registration, whether it succeeds or fails.

Until we get a better feel for the rules around registration and updates I don't think we should automatically re-register.

PR #606, opened 2017-07-21
Added admin controller for registering libraries with a discovery service.

aslagle:
This is for https://github.com/NYPL-Simplified/circulation/issues/577.

It doesn't do a great job of showing the results yet - it will tell you if there was an error, but I think eventually the circ manager should keep track of registration status for all libraries and registries and return that info in response to a GET.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -18,6 +18,7 @@ from flask.ext.babel import lazy_gettext as _
 from sqlalchemy.exc import ProgrammingError
 from PIL import Image
 from StringIO import StringIO
+import feedparser
 
 from core.model import (
     create,
@@ -47,6 +48,7 @@ from core.model import (
     WorkGenre,
 )
 from core.util.problem_detail import ProblemDetail
+from core.util.http import HTTP
 from problem_details import *
 
 from api.config import (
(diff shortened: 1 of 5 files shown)


api/admin/controller.py
```diff
@@ -1906,4 +1908,34 @@ def discovery_services(self):
...
+                                  goal=ExternalIntegration.DISCOVERY_GOAL,
+                                  id=integration_id)
+            if not integration:
+                return MISSING_SERVICE
+
+            library = get_one(self._db, Library, short_name=library_short_name)
+            if not library:
+                return NO_SUCH_LIBRARY
+
+            auth_document_url = self.url_for("acquisition_groups", library_short_name=library.short_name)
```
leonardr: Is it necessary to make a request and parse out the registration link? I would expect to be able to look up this library's registration URL as a ConfigurationSetting.

aslagle: We could store the registration link instead of the main registry link, or we could get the feed and find the register link when we first create the integration.. Would either of those be better?

leonardr: Ah, I misread the code. I thought this was the URL you GET the next line, but this is the URL you're POSTing as data on line 1938.

Issue #580, opened 2017-07-12
Automatically negotiate Short Client Token shared secret/library name with library registry


leonardr:
A library that wants to be able to generate Short Client Tokens needs to have a shared secret with, and a short library name assigned by, a library registry capable of validating them.

Here's how it could work when an admin pushes the "library sync" button:

* The library generates a public/private key pair.
* The library inserts the public key into its authentication document.
* The library contacts the registry: "hey, give me my shared secret and short library name"
* The registry responds with two pieces of information:
   -  A short library name
   -  A shared secret, encrypted with the library's public key
* The library decrypts the shared secret and stores it locally along with the short library name.
* The library starts using shared secret and short library name to build Short Client Tokens.
* (Optional) The library destroys its public/private key pair.

Once a shared secret is in place, changing it will disrupt peoples' ability to get their Adobe IDs, but only until they reload their loans page and get a new Short Client Token signed with the new secret.

To change the shared secret, a library repeats this process, but it provides the existing shared secret (in plaintext) as evidence that it is authorized to change the secret.

PR #608, opened 2017-07-27
Registry shared secret

aslagle:
This branch implements the initial part of #580 - when you register a library with the registry, the circ manager generates a public key, and obtains the short name and shared secret from the registry's response. It also obtains the registry's vendor id. 

This only generates new secrets, there's no controller for changing the secret yet.

I got rid of the short client token integration and moved the vendor id and short client token settings into the registry integration instead, and they are no longer editable in the admin interface.


aslagle:
The SHORT_CLIENT_TOKEN and DRM_GOAL constants in ExternalIntegration are no longer used. I could remove those as well, but I'm not sure if we'll want them for something else.

leonardr:
I think you should definitely get rid of SHORT_CLIENT_TOKEN and DRM_GOAL. If we need DRM_GOAL in the future we can add it back, but we don't need it for URMS or LCP, so we'll hopefully never need it.

PR #609, opened 2017-07-31
Changed id in authentication for opds document to be the library's url instead of a uuid.

aslagle:


api/authenticator.py
```diff
@@ -692,9 +692,10 @@ def create_authentication_document(self):
             links["logo"] = dict(type="image/png", href=logo)
                 
         library_name = self.library_name or unicode(_("Library"))
+        library_url = url_for("acquisition_groups", library_short_name=library.short_name, _external=True)
```
leonardr: I would expect this be the link to `index`, not `acquisition_groups`. I'm thinking specifically of the Open Ebooks case where `index` gives you a 401 error but `acquisition_groups` gives you the feed.

Issue #610, opened 2017-07-31
Gather Overdrive `ils_name` and use it when doing patron authentication


leonardr:
Currently we always send `default` as the `authorizationname` when setting up a Patron authentication token. It turns out that although this works for NYPL, Brooklyn, and Ferguson, it doesn't always work. In general, the `ils_name` needs to be set to a special value obtained from Overdrive. We need to capture this value as a configuration setting and use it when appropriate.

leonardr:
This issue may come back. I've seen some stuff in Wisconsin that makes me think the `ils_name` is a property of the relationship between the library and the Overdrive account, not a property of the Overdrive account. Wisconsin has a consortium in which many libraries with different ILS systems share an Overdrive account. The `ils_name` might be used to tell Overdrive which ILS to check a patron's credentials with.

PR #611, opened 2017-07-31
Use the .ils_name instead of the literal string 'default'

leonardr:
This branch pulls the Overdrive ILS name from the `ils_name` configuration setting instead of assuming it's always the literal string 'default'.

This fixes https://github.com/NYPL-Simplified/circulation/issues/610


api/overdrive.py
```diff
@@ -55,6 +55,7 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI):
         { "key": BaseOverdriveAPI.WEBSITE_ID, "label": _("Website ID") },
         { "key": ExternalIntegration.USERNAME, "label": _("Client Key") },
         { "key": ExternalIntegration.PASSWORD, "label": _("Client Secret") },
+        { "key": BaseOverdriveAPI.ILS_NAME, "label": _("ILS Name") },
```
aslagle: This is fine, but you could also put `"optional": True` or `"default": "default"` in here, up to you.

PR #616, opened 2017-08-01
Run docker builds

### LINKED ISSUE #612: Run tests in docker when a tag is pushed
This will make sure that a container can be built from a given tag so adjustments can be made before a container is released.

courte:
This branch adds a run of the tests in a docker container environment. Some small errors came up, corrected here and in NYPL-Simplified/server_core#591.

As I mention there, I'm a little worried that this will raise strange, annoying-to-diagnose errors just before (or after) we've launched something. I'm hoping that the use of release candidate tags will avoid this by causing the tests before anything is on production. I don't want to add the length of this test sequence to development pushes and PRs, so I think this is the next best option.

Fixes #612.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-sudo: false
+sudo: required
 
 addons:
   postgresql: "9.3"
@@ -8,6 +8,7 @@ addons:
 
 services:
   - postgresql
+  - docker
 
 language: python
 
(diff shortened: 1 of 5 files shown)


courte:
We can, but yeah not through Travis. It's something that we'd have to build and run for ourselves, similar to the bot that parses our Loggly errors. I've been wanting to do some GitHub hooks around the release process, so maybe it's time for it? Especially since we'd need an independent process to automate building the containers (my next goal), too.

PR #624, opened 2017-08-07
Finalize authentication for opds format

leonardr:
This branch completes (hopefully) the work on https://github.com/NYPL-Simplified/library_registry/issues/43.

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -364,9 +364,6 @@ class Authenticator(object):
     def authenticated_patron(self, _db, header):
         return self.invoke_authenticator_method("authenticated_patron", _db, header)
 
-    def create_authentication_document(self):
-        return self.invoke_authenticator_method("create_authentication_document")
-
     def create_authentication_headers(self):
         return self.invoke_authenticator_method("create_authentication_headers")
 
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -364,6 +364,9 @@ class Authenticator(object):
     def authenticated_patron(self, _db, header):
         return self.invoke_authenticator_method("authenticated_patron", _db, header)
 
+    def create_authentication_document(self):
+        return self.invoke_authenticator_method("create_authentication_document")
+
     def create_authentication_headers(self):
         return self.invoke_authenticator_method("create_authentication_headers")
 
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -19,7 +19,10 @@ from core.util.problem_detail import (
     ProblemDetail,
     json as pd_json,
 )
-from core.util.opds_authentication_document import OPDSAuthenticationDocument
+from core.util.authentication_for_opds import (
+    AuthenticationForOPDSDocument,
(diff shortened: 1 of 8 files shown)


api/authenticator.py
```diff
@@ -649,12 +652,10 @@ def decode_bearer_token(self, token):
...
 
     def create_authentication_document(self):
-        """Create the OPDS authentication document to be used when
+        """Create the Authentication For OPDS document to be used when
         a request comes in with no authentication.
         """
-        base_opds_document = Configuration.base_opds_authentication_document()
-
-        links = {}
+        links = []
```
leonardr: Links are now objects stored in a list with `rel` as one of the properties of the link object, analogous to HTML <a> tags.

api/authenticator.py
```diff
@@ -665,38 +666,42 @@ def create_authentication_document(self):
...
+            links.append(dict(rel="logo", type="image/png", href=logo))
                 
         library_name = self.library_name or unicode(_("Library"))
-        library_url = url_for("index", library_short_name=library.short_name, _external=True)
-        doc = OPDSAuthenticationDocument.fill_in(
-            base_opds_document, list(self.providers),
-            _db=self._db, name=library_name, id=library_url, links=links
+        auth_doc_url = url_for(
+            "authentication_document", library_short_name=library.short_name,
+            _external=True
```
leonardr: The `id` of an A4OPDS document is the URL to the document itself, not to the root of the OPDS server.

api/authenticator.py
```diff
@@ -1484,15 +1483,13 @@ def authentication_provider_document(self, _db):
             self.password_label,
             self.password_label
         )
-        method_doc = dict(
+        return dict(
+            description=unicode(self.DISPLAY_NAME),
```
leonardr: We should probably change this wording to be more imperative, like "Log in using your library card" or something.

tests/test_controller.py
```diff
@@ -780,6 +781,19 @@ def test_authenticated_patron_root_lane(self):
                 eq_(302, response.status_code)
                 eq_("http://cdn/default/groups/", response.headers['location'])
 
+    def test_authentication_document(self):
+        """Test the ability to retrieve an Authentication For OPDS document."""
+        with self.request_context_with_library(
+                "/", headers=dict(Authorization=self.invalid_auth)):
```
leonardr: Note that we get the authentication document even though we provided invalid authentication credentials. The authentication document itself does not require credentials to access.

api/controller.py
```diff
@@ -496,6 +497,17 @@ def __call__(self):
         # their type, and redirect them to an appropriate feed.
         return self.appropriate_index_for_patron_type()
 
+    def authentication_document(self):
+        """Serve this library's Authentication For OPDS document."""
+        content = self.manager.auth.create_authentication_document()
```
aslagle: This could check if the document is already cached in self.manager.authentication_for_opds_documents, but it won't be called very often so I guess it doesn't matter much.

PR #625, opened 2017-08-08
Feature enki 2.0 port

datalogics-dans:
PR opened for review.


datalogics-dans:
Note: this needs to incorporate a small change in server_core. https://github.com/NYPL-Simplified/server_core/pull/603

datalogics-dans:
I was hoping the addition of Enki to Datasources would add it to the database on a clean deployment, but I just started with a fresh DB and found out that's not the case. How can I make sure that an Enki datasources entry is created alongside all the others?

api/enki.py
```diff
@@ -209,17 +210,17 @@ def _make_request(self, url, method, headers, data=None, params=None,
...
         args = dict()
         args['method'] = "getItem"
-        args['recordid'] = identifier
+        args['recordid'] = identifier.identifier
         args['size'] = "small"
         args['lib'] = self.lib
         response = self.request(url, method='get', params=args)
         if not(response.content.startswith("{\"result\":{\"id\":\"")):
-            pool = identifier.licensed_through
+            pool = identifier.licensed_through[0]
```
datalogics-dans: I'm pretty sure this isn't the right way to do this, but am curious about how to do it better. What's the easiest way to find the licensepool that has X as its datasource?

leonardr: `[x for x in identifier.licensed_through if x.data_source==correct_data_source][0]`

api/enki.py
```diff
@@ -0,0 +1,542 @@
...
+    SETTINGS = [
+        { "key": Collection.EXTERNAL_ACCOUNT_ID_KEY, "label": _("Library ID") },
+        { "key": ExternalIntegration.URL, "label": _("URL") },
+    ] + BaseCirculationAPI.SETTINGS
+
+    PRODUCTION_BASE_URL = "http://enkilibrary.org/API/"
+    availability_endpoint = "ListAPI"
+    item_endpoint = "ItemAPI"
+    #TODO: get this from configurationsettings. Also make sure the value actually gets to the DB.
+    lib = 1
```
leonardr: What's this? If it's the library ID you should be able to get it from ConfigurationSetting now.

datalogics-dans: We are a bit further down. This code can be removed.

api/enki.py
```diff
@@ -0,0 +1,542 @@
...
+        if not self.library_id or not self.base_url:
+            raise CannotLoadConfiguration(
+                "Enki configuration is incomplete."
+            )
+        self.enki_bibliographic_coverage_provider = (
+            EnkiBibliographicCoverageProvider(
+                _db, collection, api_class=self
+            )
+        )
+        self.token = "mock_token"
```
leonardr: This should probably be in a `MockEnkiAPI` subclass.

api/enki.py
datalogics-tarar: Some of Enki's API's take a parameter called "id" while others take one called "lib," so we need both.

api/enki.py
```diff
@@ -0,0 +1,542 @@
...
+
+    @property
+    def authorization_headers(self):
+        # I don't think Enki needs authorization headers...
+        #authorization = u":".join([self.username, self.password, self.library_id])
+        authorization = u":".self.library_id
```
leonardr: This will cause a crash, but it's called by `refresh_bearer_token`, which makes me think the whole token code is broken.

api/enki.py
```diff
@@ -0,0 +1,542 @@
...
+        authorization = authorization.encode("utf_16_le")
+        authorization = base64.standard_b64encode(authorization)
+        return dict(Authorization="Basic " + authorization)
+
+    def refresh_bearer_token(self):
+        url = self.base_url + self.access_token_endpoint
```
leonardr: I don't think you define access_token_endpoint. Is this code left over from Overdrive or does Enki do authentication this way?

api/enki.py
```diff
@@ -0,0 +1,542 @@
...
+            )
+        else:
+            return response
+
+    def availability(self, patron_id=None, since=None, title_ids=[], strt=0, qty=2000):
+        self.log.debug ("requesting : "+ str(qty) + " books starting at econtentRecord" +  str(strt))
+        url = str(self.base_url) + str(self.availability_endpoint)
+        args = dict()
+	args['method'] = "getAllTitles"
+	args['id'] = "secontent"
```
leonardr: There's a mixture of tabs and spaces here, which can cause subtle bugs. We follow PEP 8 which recommends using spaces instead of tabs, and prohibits mixing tabs and spaces. In another language this wouldn't be a big deal but in Python it's very important. You should be able to configure your editor to use spaces instead of tabs.

datalogics-dans:
@leonardr All that auth_token stuff was a relic of an old copy-paste job some time ago. There are probably a lot of those sorts of things floating around that'll we'll need to get to pruning eventually.

api/enki.py
```diff
@@ -0,0 +1,542 @@
...
+        self.log.debug ("Checking availability for " + str(identifier.identifier))
+        now = datetime.datetime.utcnow()
+        url = str(self.base_url) + str(self.item_endpoint)
+        args = dict()
+        args['method'] = "getItem"
+        args['recordid'] = identifier.identifier
+        args['size'] = "small"
+        args['lib'] = self.lib
+        response = self.request(url, method='get', params=args)
+        if not(response.content.startswith("{\"result\":{\"id\":\"")):
```
leonardr: You can make this easier to read by formatting it with single quotes.

`'{"result":{"id":""`

But it would make the code more reliable to parse it as JSON and check the JSON object. Otherwise small server-side changes that don't affect the meaning of the document (such as putting the ID elsewhere in the nested object) can make it look like all the books have disappeared.

api/enki.py
```diff
@@ -0,0 +1,542 @@
...
+                else:
+                    self.log.warn(
+                        "Removing unknown work %s from circulation.",
+                        identifier.identifier
+                    )
+            pool.licenses_owned = 0
+            pool.licenses_available = 0
+            pool.licenses_reserved = 0
+            pool.patrons_in_hold_queue = 0
+            pool.last_checked = now
```
leonardr: Calling CirculationData.apply() should set these values, so you shouldn't need to set them inside this method.

api/enki.py
```diff
@@ -0,0 +1,542 @@
...
+                last_checked = now
+            )
+
+            circulationdata.apply(
+                self._db,
+                self.collection,
+                replace=ReplacementPolicy.from_license_source(self._db)
+            )
+
+            self._db.commit()
```
leonardr: You shouldn't need to commit here; the calling code should take care of it.

api/enki.py
datalogics-tarar: It's left over from Overdrive. We just put it in to get the API calls to run. We'll have to implement proper authentication before checkout. I was actually going to ask what's the best way to do this and also what's the best way to do checkouts eg. if there's a certain content provider you'd recommend basing our checkout functionality off of.

api/enki.py
```diff
@@ -0,0 +1,542 @@
...
+    log.setLevel(logging.DEBUG)
+
+    @classmethod
+    def parse_list(self, l):
+        """Turn strings like this into lists:
+
+        FICTION / Thrillers; FICTION / Suspense; FICTION / General
+        Ursu, Anne ; Fortune, Eric (ILT)
+        """
+        return [x.strip() for x in l.split(";")]
```
leonardr: You don't use this method.

api/enki.py
```diff
@@ -0,0 +1,542 @@
...
+        self.include_availability = include_availability
+        self.include_bibliographic = include_bibliographic
+
+    def extract_bibliographic(self, element):
+        identifiers = []
+        contributors = []
+        identifiers.append(IdentifierData(Identifier.ISBN, element["isbn"]))
+        sort_name = element["author"]
+        if not sort_name:
+            sort_name = "Unknown"
```
leonardr: We define Edition.UNKNOWN_AUTHOR for this purpose.

api/enki.py
```diff
@@ -0,0 +1,542 @@
...
+        contributors.append(ContributorData(sort_name=sort_name))
+        primary_identifier = IdentifierData(EnkiAPI.ENKI_ID, element["id"])
+        metadata = Metadata(
+        data_source=DataSource.ENKI,
+        title=element["title"],
+        language="ENGLISH",
```
leonardr: This needs to be a 3-letter ISO language code, so `"eng"`.

api/enki.py
```diff
@@ -0,0 +1,542 @@
...
+            contributions=True,
+            formats=True,
+        )
+        availability.apply(
+            self._db,
+            license_pool.collection,
+            replace=policy,
+        )
+        if new_edition:
+            bibliographic.apply(edition, self.collection, replace=policy)
```
leonardr: This is also where you'd call self.analytics.collect_event to reflect that a book was just added to the collection.

leonardr:
The code here is pretty close to being ready but it's missing test coverage. I'm OK with merging this code when it doesn't have all the functionality (like checkouts or authentication) but not with merging code that doesn't have test coverage.

I wouldn't worry too much about this branch getting out of sync with master. For the next couple of weeks we're focusing on bug fixes. There won't be another big change like you saw with multi-tenancy any time soon.

api/enki.py
datalogics-tarar: So delete lines 407-408?

aslagle: I think he means change line 408 to sort_name = Edition.UNKNOWN_AUTHOR

api/enki.py
```diff
@@ -233,8 +228,6 @@ def reaper_request(self, identifier):
                 replace=ReplacementPolicy.from_license_source(self._db)
             )
 
-            self._db.commit()
```
datalogics-dans: @leonardr Just to confirm, the commit should be called once we've iterated over every ID, right? I'm assuming that's the case since I see no change to the DB if I kill the script after only a few dozen iterations.

aslagle: A monitor calls commit after every batch it processes (which is every 100 ids by default). 

(Leonard's out until Monday so I'm trying to answer his questions)

datalogics-tarar:
@leonardr Would it be adequate to copy over the 1.x tests from here minus those that are obsolete? https://github.com/NYPL-Simplified/circulation/blob/c06e290834011bec68d694765e210041838533ce/tests/test_enki.py

api/enki.py
datalogics-tarar: @leonardr, do you mean only set new_license_pool to availability.license_pool(self._db, self.collection) if it's not already set to something?

aslagle: The way you're setting new_license_pool is good. He means you can add some code here like
```
if new_license_pool:
    self.analytics.collect_event(...)
```

aslagle:
@datalogics-tarar I think bringing over the 1.x tests and updating them for 2.0 makes sense. I'd have to go through the code to see if those tests cover everything on this branch. If you've added new code for 2.0, you should probably add tests for it as well.

datalogics-tarar:
Wounds good. We just refactored the same functionality to work with 2.0 and didn't add any new classes or significant methods.

api/enki.py
datalogics-tarar: @aslagle I added the following code: "self.analytics.collect_event(self._db, license_pool, CirculationEvent.DISTRIBUTOR_TITLE_ADD, now)" and got the following error:

2017-08-11 10:14:09,321:Enki Circulation Monitor:ERROR:scripts.py:Error running monitor Enki Circulation Monitor for collection Enki: 'Session' object has no attribute 'id'
  File "/var/www/circulation/core/scripts.py", line 224, in do_run
    monitor.run()
  File "/var/www/circulation/core/monitor.py", line 155, in run
    new_timestamp_value = self.run_once(start, cutoff) or cutoff
  File "/var/www/circulation/api/enki.py", line 484, in run_once
    self.process_book(bibliographic, circulation)
  File "/var/www/circulation/api/enki.py", line 519, in process_book
    self.analytics.collect_event(self._db, license_pool, CirculationEvent.DISTRIBUTOR_TITLE_ADD, now)
  File "/var/www/circulation/core/analytics.py", line 40, in collect_event
    for provider in (self.sitewide_providers + self.library_providers[library.id]):
AttributeError: 'Session' object has no attribute 'id'

Thoughts?

aslagle: The first argument to collect_event is supposed to be a Library, not _db. I see some of the other APIs have this wrong too - they must be missing test coverage so we didn't notice it before.

datalogics-tarar: @aslagle I copied Bibliotecha. I see the other libraries use "for library in self.collection.libraries:." What does a collection and a library refer to?

aslagle: When there are multiple libraries on a circ manager, you can configure which libraries have access to which collections. So you would configure the Enki collection's settings first, and then configure which libraries have access to it. You should probably use "for library in self.collection.libraries" here too, even if you don't have any Enki collections with multiple libraries yet.

datalogics-tarar:
@aslagle @leonardr I copied over the tests from 1.x and ran into the following error. Thoughts?

  File "/var/www/circulation/env/local/lib/python2.7/site-packages/nose/case.py", line 381, in setUp
    try_run(self.inst, ('setup', 'setUp'))
  File "/var/www/circulation/env/local/lib/python2.7/site-packages/nose/util.py", line 471, in try_run
    return func()
  File "/var/www/circulation/tests/test_enki.py", line 49, in setup
    self.api = MockEnkiAPI(self._db)
  File "/var/www/circulation/api/enki.py", line 265, in __init__
    library = Library.instance(_db)
AttributeError: type object 'Library' has no attribute 'instance'

aslagle:
@datalogics-tarar In 2.0 there is no longer a single library instance, so you can't use Library.instance. You can use `self._default_library` to get a library for tests.

datalogics-tarar:
@aslagle Can I just get rid of this test?

    def test_script_instantiation(self):
        """Test that RunCoverageProviderScript can instantiate
        the coverage provider.
        """
        script = RunCoverageProviderScript(
            EnkiBibliographicCoverageProvider, self._db, [],
            enki_api=self.api
        )
        assert isinstance(script.provider,
                          EnkiBibliographicCoverageProvider)
        eq_(script.provider.api, self.api)

aslagle:
@datalogics-tarar That test looks like it could still be useful. What's the problem with it? Did we get rid of similar tests for the other coverage providers in 2.0?

datalogics-tarar:
@aslagle @leonardr I keep getting this error: "AttributeError: type object 'Library' has no attribute 'instance'" even though I'm not defining a variable called "instance" anywhere. Thoughts?

aslagle:
@datalogics-tarar That's because the `Library.instance` method no longer exists in 2.0. You'll have to change any code that calls it.

datalogics-tarar:
New error: "AttributeError: type object 'Collection' has no attribute 'ENKI'" Any idea what's causing this?

aslagle:
@datalogics-tarar You probably need to change `Collection.ENKI` to `EnkiAPI.ENKI` somewhere in the code.

datalogics-tarar:
@aslagle good catch. Now I'm getting this error: "InvalidRequestError: Entity '<class 'core.model.Collection'>' has no property 'PROTOCOL'" regardless of whether protocol is caps or lower case.

aslagle:
@datalogics-tarar Are you using Collection.PROTOCOL as a constant? If you are, you probably want to use EnkiAPI.ENKI as the protocol instead. 

If you have a collection object and you want to see what its protocol is, collection.protocol should work if you use it on the object instead of the class.

datalogics-tarar:
@aslagle I have "protocol=EnkiAPI.ENKI" set in enki.py but I still get that error.

aslagle:
@datalogics-tarar I think I found your problem. In 2.0, Collection no longer has a protocol field, it has an ExternalIntegration which has a protocol. So if you call get_one_or_create for a Collection, you can't pass in the protocol. After you've created the collection, you can use `collection.protocol = EnkiAPI.ENKI` to set the protocol on the collection's ExternalIntegration.

datalogics-tarar:
Thanks @aslagle. That fixed a lot of errors.

datalogics-tarar:
Thanks, @aslagle. Since I'm using EnkiAPI.ENKI instead of ExternalIntegration.ENKI, how do I combat this error?

ERROR: tests.test_enki.TestBibliographicCoverageProvider.test_create_identifier_strings
  File "/var/www/circulation/env/local/lib/python2.7/site-packages/nose/case.py", line 381, in setUp
    try_run(self.inst, ('setup', 'setUp'))
  File "/var/www/circulation/env/local/lib/python2.7/site-packages/nose/util.py", line 471, in try_run
    return func()
  File "/var/www/circulation/tests/test_enki.py", line 49, in setup
    self.api = MockEnkiAPI(self._db)
  File "/var/www/circulation/api/enki.py", line 274, in __init__
    collection.protocol=EnkiAPI.ENKI
  File "/var/www/circulation/env/local/lib/python2.7/site-packages/sqlalchemy/ext/hybrid.py", line 747, in __set__
    self.fset(instance, value)
  File "/var/www/circulation/core/model.py", line 9736, in set_protocol
    self.external_integration.protocol = new_protocol
  File "/var/www/circulation/core/model.py", line 9826, in external_integration
    "No known external integration for collection %s" % self.name
ValueError: No known external integration for collection Test Enki Collection

aslagle:
@datalogics-tarar I guess the integration doesn't automatically get created when you set the protocol directly. You can create the collection with `Collection.by_name_and_protocol(_db, name, protocol)` instead of `get_one_or_create` and that should set it up correctly.

datalogics-tarar:
@aslagle, that seems to work so far. Thoughts on this error?

  File "/var/www/circulation/env/local/lib/python2.7/site-packages/nose/case.py", line 381, in setUp
    try_run(self.inst, ('setup', 'setUp'))
  File "/var/www/circulation/env/local/lib/python2.7/site-packages/nose/util.py", line 471, in try_run
    return func()
  File "/var/www/circulation/tests/test_enki.py", line 49, in setup
    self.api = MockEnkiAPI(self._db)
  File "/var/www/circulation/api/enki.py", line 273, in __init__
    _db, collection, *args, **kwargs
  File "/var/www/circulation/api/enki.py", line 112, in __init__
    self.library_id = collection.external_account_id.encode("utf8")
AttributeError: 'NoneType' object has no attribute 'encode'

aslagle:
@datalogics-tarar collection.external_account_id needs to be set. It looks like you were setting it as part of the get_one_or_create call before, so now you'll need to set it after the collection is created.

datalogics-tarar:
@aslagle Is this syntax correct?

        script = RunCoverageProviderScript(
            EnkiBibliographicCoverageProvider, self._db,
            api_class=MockEnkiAPI
        )

I'm getting this error:

ERROR: Test that RunCoverageProviderScript can instantiate
  File "/var/www/circulation/env/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/var/www/circulation/tests/test_enki.py", line 66, in test_script_instantiation
    api_class=MockEnkiAPI
  File "/var/www/circulation/core/scripts.py", line 629, in __init__
    **kwargs
TypeError: __init__() takes at least 3 arguments (3 given)

aslagle:
@datalogics-tarar I think the syntax is correct, but you should use RunCollectionCoverageProviderScript instead of RunCoverageProviderScript. 

You might also need to update the constructor of the EnkiBibliographicCoverageProvider, if you haven't already.

datalogics-tarar:
Thanks @aslagle. Any idea what to do about this error in test_enki.py?

ERROR: Failure: ImportError (cannot import name RunCollectionCoverageProviderScript)

aslagle:
@datalogics-tarar The import should be `from core.scripts import RunCollectionCoverageProviderScript`. Do you have that already?

datalogics-tarar:
@aslagle I left out "core." Now I'm back to this error:

ERROR: Test that RunCollectionCoverageProviderScript can instantiate
  File "/var/www/circulation/env/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/var/www/circulation/tests/test_enki.py", line 66, in test_script_instantiation
    api_class=MockEnkiAPI
  File "/var/www/circulation/core/scripts.py", line 271, in __init__
    providers += self.get_providers(_db, provider_class, **kwargs)
  File "/var/www/circulation/core/scripts.py", line 276, in get_providers
    return list(provider_class.all(_db, **kwargs))
  File "/var/www/circulation/core/coverage.py", line 770, in all
    yield cls(collection, **kwargs)
TypeError: __init__() takes at least 3 arguments (3 given)

aslagle:
Okay, this is where you need to update the constructor of EnkiBibliographicCoverageProvider. The first argument should be a collection.

aslagle:
I think the problem is that yours has _db as a required argument and RunCollectionCoverageProviderScript doesn't give it a database.

datalogics-tarar:
I'm running into this error: "CollectionMissing: EnkiBibliographicCoverageProvider must be instantiated with a Collection." and I wonder if it's because of this snippet:        

        script = RunCollectionCoverageProviderScript(
            EnkiBibliographicCoverageProvider, self._db,
            api_class=MockEnkiAPI
        )

Do I need to pass in collection after EnkiBibliographicCoverageProvider?

aslagle:
@datalogics-tarar I think the problem is that RunCollectionCoverageProviderScript calls __init__ with a collection as the first argument, but EnkiBibliographicCoverageProvider expects _db. So in EnkiBibliographicCoverageProvider.__init__, `_db` ends up being a Collection and `collection` is something else. I would recommend making the `__init__` arguments match the ones used by the other BibliographicCoverageProviders.

datalogics-tarar:
@aslagle I fixed a couple more errors and now I'm running into this error:

ERROR: Test that RunCollectionCoverageProviderScript can instantiate
  File "/var/www/circulation/env/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/var/www/circulation/tests/test_enki.py", line 66, in test_script_instantiation
    api_class=MockEnkiAPI
  File "/var/www/circulation/core/scripts.py", line 271, in __init__
    providers += self.get_providers(_db, provider_class, **kwargs)
  File "/var/www/circulation/core/scripts.py", line 276, in get_providers
    return list(provider_class.all(_db, **kwargs))
  File "/var/www/circulation/core/coverage.py", line 770, in all
    yield cls(collection, **kwargs)
TypeError: __init__() takes at least 3 arguments (3 given)

I checked that EnkiBibliographicCoverageProvider's init arguments are self, collection, _db, api_class=EnkiAPI, **kwargs

aslagle:
@datalogics-tarar RunCollectionCoverageProviderScript doesn't provide the `_db` argument to CoverageProvider classes. You'll need to remove that argument from EnkiBibliographicCoverageProvider or make it optional. If it needs the db session, you can get it from the collection with `_db = Session.object_session(collection)`.

datalogics-tarar:
@aslagle Sounds good. I made the argument optional in EnkiBibliographicCoverageProvider's constructor (set the default to None.) Would I also need to get rid of _db in EnkiImport constructor?

    def __init__(self, _db, collection, api_class=EnkiAPI):
        """Constructor."""
        super(EnkiImport, self).__init__(_db, collection)
        self.analytics = Analytics(_db)
        self.bibliographic_coverage_provider = (
            EnkiBibliographicCoverageProvider(collection, _db, api_class=self.api)
        )

aslagle:
@datalogics-tarar I think you should leave EnkiImport the way it is - it doesn't have the same problem. EnkiImport's parent class is CollectionMonitor, and if you look at CollectionMonitor in core/monitor.py, it's `__init__` has _db and collection, so EnkiImport already uses the same arguments.

datalogics-tarar:
Thanks @aslagle. Making progress! Now I'm getting a different error.

ERROR: Test that RunCollectionCoverageProviderScript can instantiate
  File "/var/www/circulation/env/local/lib/python2.7/site-packages/nose/case.py", line 197, in runTest
    self.test(*self.arg)
  File "/var/www/circulation/tests/test_enki.py", line 66, in test_script_instantiation
    api_class=MockEnkiAPI
  File "/var/www/circulation/core/scripts.py", line 271, in __init__
    providers += self.get_providers(_db, provider_class, **kwargs)
  File "/var/www/circulation/core/scripts.py", line 276, in get_providers
    return list(provider_class.all(_db, **kwargs))
  File "/var/www/circulation/core/coverage.py", line 770, in all
    yield cls(collection, **kwargs)
  File "/var/www/circulation/api/enki.py", line 333, in __init__
  File "/var/www/circulation/api/enki.py", line 266, in __init__
    library = DatabaseTest.make_default_library(_db)
  File "/var/www/circulation/core/testing.py", line 802, in make_default_library
    ), short_name="default"
  File "/var/www/circulation/core/model.py", line 417, in get_one_or_create
    one = get_one(db, model, **kwargs)
  File "/var/www/circulation/core/model.py", line 394, in get_one
    q = db.query(model).filter_by(**kwargs)
AttributeError: 'NoneType' object has no attribute 'query'

aslagle:
@datalogics-tarar that's probably from making the default for `_db` None. If it's set to None, you'll need to get the database session from the collection with `Session.object_session` before anything needs to use it.

api/enki.py
```diff
@@ -0,0 +1,539 @@
...
+        headers = dict(extra_headers)
+        if exception_on_401:
+            disallowed_response_codes = ["401"]
+        else:
+            disallowed_response_codes = None
+            response = self._make_request(
+            url=url, method=method, headers=headers,
+            data=data, params=params,
+            disallowed_response_codes=disallowed_response_codes
+        )
```
aslagle: The indentation looks off here.. I'm surprised it's not a syntax error.

api/enki.py
```diff
@@ -0,0 +1,539 @@
...
+    PROTOCOL = EnkiAPI.ENKI
+    INPUT_IDENTIFIER_TYPES = EnkiAPI.ENKI_ID
+
+    def __init__(self, collection, api_class=EnkiAPI, **kwargs):
+        """Constructor.
+
+        :param collection: Provide bibliographic coverage to all
+            Axis 360 books in the given Collection.
+        :param api_class: Instantiate this class with the given Collection,
+            rather than instantiating Axis360API.
```
aslagle: This comment should be updated to say EnkiAPI instead of Axis360.

api/enki.py
```diff
@@ -0,0 +1,539 @@
...
+    DEFAULT_BATCH_SIZE = 100 
+    FIVE_MINUTES = datetime.timedelta(minutes=5)
+    
+    def __init__(self, _db, collection, api_class=EnkiAPI):
+        """Constructor."""
+        super(EnkiImport, self).__init__(_db, collection)
+        self.api = api_class(_db, collection)
+        self.analytics = Analytics(_db)
+        self.bibliographic_coverage_provider = (
+            EnkiBibliographicCoverageProvider(_db, collection, api_class=self.api)
```
aslagle: Looks like you removed the _db argument from `EnkiBibliographicCoverageProvider.__init__` but forgot to change it here. That makes me think this code doesn't have test coverage.

tests/test_enki.py
```diff
@@ -0,0 +1,127 @@
...
+        path = os.path.join(cls.resource_path, filename)
+        return open(path).read()
+
+
+class TestEnkiAPI(DatabaseTest, BaseEnkiTest):
+
+    def setup(self):
+        super(TestEnkiAPI, self).setup()
+        self.collection = self._collection(protocol=EnkiAPI.ENKI)
+        self._collection(protocol=EnkiAPI.ENKI)
```
aslagle: You can delete these two lines since you didn't end up using self.collection.

api/enki.py
datalogics-dans: Ah, yes, apparently you'd be right. We test creating license pools, but not via EnkiImport. That'll be something to add.

tests/test_enki.py
datalogics-tarar: I tried and it caused an error. We are using the collection.

aslagle: Oh you're right. You can still remove the second line though - you only need the first one that sets self.collection.

api/enki.py
datalogics-dans: How would you recommend testing this? Would we need to create a new mock API for EnkiImport?

aslagle: I think you could use the same MockEnkiAPI that the other tests use. To catch this bug all you'd need to do is instantiate EnkiImport, but it would be nice to have tests for the other methods as well. I think the Axis360CirculationMonitorTest is a good example. https://github.com/NYPL-Simplified/circulation/blob/master/tests/test_axis.py#L142

datalogics-dans: Added just the init test for now.

Issue #627, opened 2017-08-09
Most content provider scripts are failing in 2.0


datalogics-dans:
Four scripts, overdrive_reaper, oneclick_monitor_availability, axis_monitor, and axis_reaper have the same error:
```2017-08-09 10:00:32,862:root:ERROR:scripts.py:Fatal exception while running script: __init__() got an unexpected keyword argument '_db'
Traceback (most recent call last):
  File "/var/www/circulation/core/scripts.py", line 147, in run
    self.do_run()
  File "/var/www/circulation/core/scripts.py", line 216, in do_run
    for monitor in self.monitor_class.all(self._db, **self.kwargs):
  File "/var/www/circulation/core/monitor.py", line 241, in all
    yield cls(_db=_db, collection=collection, **constructor_kwargs)
TypeError: __init__() got an unexpected keyword argument '_db'
Traceback (most recent call last):
  File "./bin/overdrive_reaper", line 10, in <module>
    RunCollectionMonitorScript(OverdriveCollectionReaper).run()
  File "/var/www/circulation/core/scripts.py", line 153, in run
    raise e
TypeError: __init__() got an unexpected keyword argument '_db'`
```

Another, oneclick_library_delta, has a different one, possibly just due to something wrong with our system. This is the only place I've seen this error:
```2017-08-09 10:00:23,406:Configuration file loader:INFO:config.py:Loading configuration from /etc/circulation/config.json
Applying catalog delta of library #175
Traceback (most recent call last):
  File "./bin/oneclick_library_delta", line 15, in <module>
    OneClickDeltaScript().run()
  File "/var/www/circulation/core/scripts.py", line 1457, in __init__
    _db = Session.object_session(collection)
  File "/var/www/circulation/env/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 77, in object_session
    return object_session(instance)
  File "/var/www/circulation/env/lib/python2.7/site-packages/sqlalchemy/orm/session.py", line 2764, in object_session
    raise exc.UnmappedInstanceError(instance)
sqlalchemy.orm.exc.UnmappedInstanceError: Class '__builtin__.NoneType' is not mapped
```

And "threem*" scripts are failing because cron tries to run them even though they've been renamed to "bibliotheca*". I have a separate circulation-docker issue open for that: https://github.com/NYPL-Simplified/circulation-docker/issues/45

leonardr:
[...]

PR #628, opened 2017-08-09
Collection monitor takes redundant database connection

leonardr:
This branch makes sure that certain `CollectionMonitor`s take a database connection in their constructor. Even though it's redundant, the `CollectionMonitor` class takes a database connection as well as a `Collection`, for the sake of preserving a single constructor signature between `Monitor` and `CollectionMonitor`.

I looked into removing the database connection from the argument, but that single constructor signature is really important. In particular, some subclasses use`IdentifierSweepMonitor` as a `CollectionMonitor` and some use it as a regular `Monitor.


leonardr:
This is the first half of https://github.com/NYPL-Simplified/circulation/issues/627 (the rest is an unrelated issue).

PR #633, opened 2017-08-14
Fix session problem

leonardr:
This branch fixes a problem where a non-scoped database session was used to get a LicensePool before creating a loan or hold. This conflicted with the scoped database session used to get the Patron whose loan or hold it was.

Amy wrote the code, I'm just making a clean PR for review.


leonardr:
I made a tiny additional change but this makes sense and I think it's good to go.

PR #636, opened 2017-08-16
Performance improvements

leonardr:
This branch represents the fruits of my work profiling typical circulation manager usage and optimizing both the initial startup time (which is vastly improved but could use a little more work) and the typical request (which is improved a fair amount and is now dominated by blocking on external services).

Most of the work went into my earlier core branches, so this is mostly using the `HasFullTableCache` caches instead of hitting the database to look things up by their database IDs.


tests/admin/test_controller.py
```diff
@@ -2389,6 +2389,7 @@ def test_sitewide_settings_get(self):
             assert AcquisitionFeed.NONGROUPED_MAX_AGE_POLICY in keys
             assert Configuration.SECRET_KEY in keys
 
+        set_trace()
```
aslagle: looks good once you take out this set_trace

Issue #638, opened 2017-08-17
Opportunities for improving performance of loan feeds


leonardr:
I've made some improvements to the performance of the loan feed. Here are some more changes that are theoretically possible but which I'm not going to make right now because I think they're too much work from a cost-benefit perspective:

* We look up Loan and Hold objects individually based on their LicensePool, rather than pulling them in with whatever query obtains the LicensePools.

* We can't keep a complete table cache of Identifiers the way we can with Collections, but Identifiers never change or get deleted once they're created, so there's no reason we can't keep a big in-memory LRU cache of N thousand identifiers and have `Identifier.for_foreign_identifier` use that LRU to save a database query here and there.

* We get a patron's Overdrive credential twice, once to get their loans and once to get their holds.

PR #639, opened 2017-08-17
When the remote mentions a loan we already have loaded from the database, modify it in place rather than load it again.

leonardr:
Part of https://github.com/NYPL-Simplified/circulation/issues/638 turned out to be a lot easier than I thought, so I implemented the fix. When a patron loads their loans feed, this saves one database request per loan or hold.

While I'm at it, I handled the (unusual) case where the source of truth modifies the term of a loan during the loan.

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -771,9 +771,22 @@ class CirculationAPI(object):
             # This is a remote loan. Find or create the corresponding
             # local loan.
             pool = loan.license_pool(self._db)
-            start = loan.start_date or now
+            start = loan.start_date
             end = loan.end_date
-            local_loan, new = pool.loan_to(patron, start, end)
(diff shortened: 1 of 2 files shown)


api/circulation.py
```diff
@@ -771,9 +771,22 @@ def sync_bookshelf(self, patron, pin):
             # This is a remote loan. Find or create the corresponding
             # local loan.
             pool = loan.license_pool(self._db)
-            start = loan.start_date or now
```
leonardr: If there is no local loan and the remote has no opinion about the start date, the start date will be set to `now` inside `loan_to()`, so this isn't necessary. If there is a local loan and the remote has no opinion about the start date, we want to leave the existing start date alone rather than update it to the current date.

PR #640, opened 2017-08-17
Update licenses owned and licenses available

leonardr:
This branch fixes a bug that stopped OneClick books from being available. `licenses_available` was being updated but `licenses_owned` was always staying at zero.

Maybe we should also put some stuff in core that freaks out if `licenses_available` is set to a value higher than `licenses_owned`, but this fixes the problem for OneClick.


leonardr:
I don't think this will solve the whole problem because the books are also missing delivery mechanisms.

PR #641, opened 2017-08-17
Fix OneClick integration to the point where an end-to-end test works.

leonardr:
This work is not done; I'm creating a PR so I can hand off to Amy while I'm on vacation.


api/oneclick.py
```diff
@@ -343,14 +345,35 @@ def update_licensepool_for_identifier(self, isbn, availability):
...
+        delivery_type = None
+        medium = medium.lower()
+        if medium == 'ebook':
+            delivery_type = Representation.EPUB_MEDIA_TYPE
+        elif medium == 'eaudio':
+            # A questionable assumption.
+            delivery_type = Representation.MP3_MEDIA_TYPE
+
+        if delivery_type:
+            formats.append(FormatData(delivery_type, None))
```
leonardr: I imported data from a OneClick collection and all the books were missing their delivery mechanisms. This fixes the problem but `isbn_info_to_metadata` (in core/oneclick.py) does create a CirculationData object with the appropriate FormatData, so it shouldn't have been a problem. I wasn't able to investigate why the FormatData isn't created initially. The code in that method like `cls.oneclick_formats.get("audiobook-mp3-oneclick")` seems copied and pasted  from Overdrive code and can be simplified quite a bit.

api/oneclick.py
```diff
@@ -369,6 +392,13 @@ def update_availability(self, licensepool):
         pass
 
 
+    def internal_format(self, delivery_mechanism):
+        """We don't need to do any mapping between delivery mechanisms and
+        internal formats, because each title is only available in one
+        format.
+        """
+        return delivery_mechanism
+
```
leonardr: The superclass implementation of this method, which is always called, assumes the existence of a mapping that isn't necessary for OneClick.

api/oneclick.py
```diff
@@ -530,16 +560,32 @@ def get_patron_checkouts(self, patron_id):
...
+            # We need to use requests instead of self.request because
+            # it's a different server that will reject the credentials
+            # we use for the API. This needs to be mocked separately.
+            access_document = requests.get(download_url)
+            data = json.loads(access_document.content)
+            content_link = data['url']
+            content_type = data['type']
+            if content_type == 'application/vnd.adobe':
+                # The manifest spells the media type wrong. Fix it.
+                content_type = DeliveryMechanism.ADOBE_DRM
```
leonardr: At this point there's a whole other document--the manifest--between us and the ACSM file.

api/oneclick.py
```diff
@@ -530,16 +560,32 @@ def get_patron_checkouts(self, patron_id):
...
+            # For ebooks, the manifest includes a single link to an ACSM
+            # file.
+            #
+            # We can't do anything about audiobook manifests right
+            # now, but we can turn an ebook manifest into a
+            # FulfillmentInfo.
+
+            # We need to use requests instead of self.request because
+            # it's a different server that will reject the credentials
+            # we use for the API. This needs to be mocked separately.
```
leonardr: The previous code assumed that download_url pointed to the ACSM file. But at this point there's a whole other document--the manifest--between us and the ACSM file.

api/oneclick.py
```diff
@@ -670,12 +716,12 @@ def validate_item(self, licensepool):
             raise InvalidInputException("Licensepool %r doesn't know its ISBN.", licensepool)
 
         if licensepool.work and licensepool.work.presentation_edition:
-            media = licensepool.work.presentation_edition.media
+            media = licensepool.work.presentation_edition.medium
```
leonardr: This typo indicates this method does not have code coverage.

api/oneclick.py
```diff
@@ -670,12 +716,12 @@ def validate_item(self, licensepool):
...
 
         return item_oneclick_id, media
 
 
-    def validate_patron(self, patron, create=False):
+    def validate_patron(self, patron, create=True):
```
leonardr: validate_patron was never being called with create=True. It was impossible to ever borrow a book because you'd never get a OneClick account. I don't see any reason to _ever_ call this without create=True, since we always want any interested patron to have a OneClick account, so I fixed it by changing the default.

api/oneclick.py
```diff
@@ -688,7 +734,11 @@ def validate_patron(self, patron, create=False):
         if not patron_cardno:
             raise InvalidInputException("Patron %r has no card number.", patron)
 
-        patron_oneclick_id = self.get_patron_internal_id(patron_cardno=patron_cardno)
+        try:
+            patron_oneclick_id = self.get_patron_internal_id(patron_cardno=patron_cardno)
+        except NotFoundOnRemote, e:
+            patron_oneclick_id = None
+
```
leonardr: This was raising an exception when the patron had no OneClick account, and it was halting the request even when `create=True`, which was supposed to detect this situation and fix it.

leonardr:
This isn't relevant yet, but the DeliveryMechanism for audio books created in core is media type "vnd.librarysimplified/obfuscated-one-click", DRM scheme "OneClick DRM", but the DeliveryMechanism used in circulation is media type "audio/mpeg" with no DRM scheme. I don't know which one is right but they should be consistent.

api/oneclick.py
leonardr: I wiped the database and tried again from scratch, and this time the books were created with the proper delivery mechanisms (the ones defined in core). So this might have been fixed by some change I made later in this branch.

leonardr:
This looks good. I'm OK with landing it if you are.

datalogics-bhaugen:
In the next day or so we will be able to try this out end to end if you want verification before merging ;)

aslagle:
I'm fine with merging this now. It was broken before so we aren't making it worse :)

If we find more issues we can have a follow-up branch.

Issue #605, opened 2017-07-21
Implement OPDS import with authentication


leonardr:
Plympton and Biblioboard have a system where books are made available through password-protected OPDS feeds. The password is given to libraries, not to patrons. The password grants access to the books, but more importantly, it also grants access to a service that issues OAuth Client Credential tokens. These time-limited tokens can be given to patrons to let them access books.

This overall protocol needs a name.

The circulation manager needs to implement a new collection integration with goal=LICENSE_GOAL, to implement this protocol. It handles the OPDS import using Basic Auth and also knows how to generate OAuth Client Credential tokens in response to a patron's fulfillment request.

We need to design a document that conveys the OAuth Client Credential and the target URL as part of the response to the fulfillment request. This document serves the same logical purpose as an ACSM file. An HTTP redirect will _almost_ work for this purpose but not quite, because you need to pass the OAuth Client Credential in the Authorization header.

Finally, the client needs to be modified to know what to do when it gets one of these documents.

leonardr:
I wrote up the protocol here: https://github.com/NYPL-Simplified/Simplified/wiki/BearerTokenPropagation

As a first step, it would be easier to avoid sending the bearer token altogether. We would gulp up a copy of the book every time it's requested and send it to the patron. From the client's perspective, the EPUB would be sent directly in response to the `fulfill` request.

leonardr:
We've got this implemented so far as downloading the EPUB and sending it directly to the client, and for compatibility reasons that's where it's going to stay for a while, so I'm closing the issue.

PR #642, opened 2017-08-18
OPDS for distributors

aslagle:
This implements #605, but with getting the book and sending it on to the patron instead of putting the bearer token in OPDS.

I've tested with Biblioboard. Plympton doesn't work yet because it doesn't have an authentication document.


api/opds_for_distributors.py
```diff
@@ -0,0 +1,329 @@
...
+            if not auth_links:
+                raise LibraryAuthorizationFailedException()
+            self.auth_url = auth_links[0].get("href")
+
+        headers = dict()
+        auth_header = "Basic %s" % base64.b64encode("%s:%s" % (self.username, self.password))
+        headers['Authorization'] = auth_header
+        headers['Content-Type'] = "application/x-www-form-urlencoded"
+        token_response = self._request_with_timeout('POST', self.auth_url, data={}, headers=headers)
+        token = json.loads(token_response.content).get("access_token")
```
leonardr: This doesn't need to go in this branch, but we can store the token in a `Credential` rather than fetching a new one every time. This will speed up both metadata imports and fulfillment.

tests/test_opds_for_distributors.py
```diff
@@ -0,0 +1,358 @@
...
+        # it adds delivery mechanisms for books with epub acquisition links
+        # and sets pools' licenses_owned and licenses_available.
+
+        # Both works were created, since we can use their acquisition links
+        # to give copies to patrons.
+        [camelot, southern] = sorted(imported_works, key=lambda x: x.title)
+
+        # Each work has a license pool.
+        [camelot_pool] = camelot.license_pools
+        [southern_pool] = southern.license_pools
```
leonardr: It would be good to see what happens with a book that has an OPDS entry but no acquisition link. Not that important since that's not how Biblioboard gates a library's access to books.

aslagle: I think what happens is that the edition and pool get created without delivery mechanisms.

aslagle:
@leonardr I've updated the authenticate request to have the grant type in the body, which works for Biblioboard now that it's not in the authenticate url as well. This should also work for Plympton once they have an authentication document.

leonardr:
Cool, let's merge it.

PR #646, opened 2017-08-24
Helpful scripts

leonardr:
This branch gets two helpful scripts off of my hard drive and into the codebase where they can be used by others.

One script prints out information about all the Overdrive Advantage accounts associated with the site's Overdrive collections. The other takes a Short Client Token and verifies that NYPL's circulation manager can validate it.

I also added a previously missing script file that invokes an existing script to reset a patron's Adobe ID.

### DIFF
--- a/bin/informational/adobe-id-for-short-client-token
+++ b/bin/informational/adobe-id-for-short-client-token
@@ -1,0 +1,60 @@
+#!/usr/bin/env python
+"""Retrieve an Adobe ID from a Short Client Token."""
+# This script doesn't import any code from circulation or core. It
+# tests an integration based on HTTP.
+
+import requests
+import base64
+import re
+import sys
+
+base_url = "https://circulation.librarysimplified.org"
+
+if len(sys.argv) != 2:
+    print "Usage: %s [Short Client Token]" % sys.argv[0]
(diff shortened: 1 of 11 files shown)


bin/informational/adobe-id-for-short-client-token
```diff
@@ -0,0 +1,60 @@
...
+print " Timestamp: %s" % timestamp
+print " Patron identifier: %s" % identifier
+print " Signature: %s" % signature
+print
+
+username, password = client_token.rsplit("|", 1)
+print "Using SignIn to log in via Short Client Token."
+print "Username: %s" % username
+print "Password: %s" % password
+print
```
aslagle: Is this meant to be used if the short client token didn't look good above? How does this work?

leonardr: The Short Client Token has four parts, and all four are necessary to _validate_ the SCT, but the two pieces that go through the SignIn API are username and password. I'll change the wording to make this more clear.

aslagle: That makes sense. I didn't get why you needed to split the token again, but you're not getting anything different from it.

PR #650, opened 2017-08-25
Oneclick fixes

leonardr:
This branch fixes the OneClick branch to bring it into sync with how the API actually works.

* When there are no available copies of a book, we raise NoAvailableCopies instead of CannotLoan. This puts the book on hold automatically. Previously the "no available copies" case was not handled specially.
* When revoking a loan or hold, the API returns `{"message": "success"}`. The code assumes it returns no entity-body, which may have been true earlier.

It might be the case that we should never put a OneClick book on hold, but if things turn out that way, it's an easy change to make.


aslagle:
While you're in here you could add the missing PROTOCOL to OneClickCirculationMonitor.

PR #654, opened 2017-08-28
Add Enki checkout/download functionality

datalogics-dans:
We still have to implement the ability to place holds and check books back in, but with this PR I believe we have the majority of Enki functionality done.


api/enki.py
```diff
@@ -233,11 +249,40 @@ def reaper_request(self, identifier):
...
+        enki_id = identifier.identifier
+        response = self.loan_request(patron.authorization_identifier, pin, enki_id)
+        if response.status_code != 200:
+            raise CannotLoan(response.status_code)
+        result = json.loads(response.content)['result']
+        if not result['success']:
+            message = result['message']
+            if "That record is already checked out to you" in message:
+                # this might be okay
+                pass
```
leonardr: This is fine, raise AlreadyCheckedOut() and the caller will handle it.

api/enki.py
```diff
@@ -233,11 +249,40 @@ def reaper_request(self, identifier):
...
+            elif "There are no available copies" in message:
+                # TODO: raise no copies error
+                self.log.error("There are no copies of this book available.")
+                raise CirculationException()
```
leonardr: Raise NoAvailableCopies() here.

api/enki.py
```diff
@@ -233,11 +249,40 @@ def reaper_request(self, identifier):
...
+            elif "Login unsuccessful" in message:
+                # TODO: raise invalid barcode/password error
+                self.log.error("User validation against Enki server was unsuccessful.")
+                raise CirculationException()
```
leonardr: Raise AuthorizationFailedException() here. All these exceptions are in circulation_exceptions.py.

api/enki.py
```diff
@@ -249,17 +294,136 @@ def checkout(self, patron, pin, licensepool, internal_format):
...
+        book_id = licensepool.identifier.identifier
+        response = self.loan_request(patron.authorization_identifier, pin, book_id)
+        if response.status_code != 200:
+            raise CannotFulfill(response.status_code)
+        result = json.loads(response.content)['result']
+        if not result['success']:
+            message = result['message']
+            if "That record is already checked out to you" in message:
+                # this might be okay
+                pass
```
leonardr: This isn't okay. Here you know you have a loan and you're asking for a copy of the book. If the error is "the book is already checked out" that indicates the server thinks you're trying to create another loan. Hopefully this is copy-and-paste from the checkout() method that can be removed.

api/enki.py
```diff
@@ -249,17 +294,136 @@ def checkout(self, patron, pin, licensepool, internal_format):
...
+            elif "There are no available copies" in message:
+                # TODO: raise no copies error
+                self.log.error("There are no copies of this book available.")
+                raise CirculationException()
+            elif "Login unsuccessful" in message:
+                # TODO: raise invalid barcode/password error
+                self.log.error("User validation against Enki server was unsuccessful.")
+                raise CirculationException()
```
leonardr: Again, AuthorizationFailedException() here.

api/enki.py
```diff
@@ -249,17 +294,136 @@ def checkout(self, patron, pin, licensepool, internal_format):
...
+        media_type = self.get_enki_media_type(book_id)
+        url, item_type, expires = self.parse_fulfill_result(result)
+
+        if not media_type and item_type == 'epub':
+            media_type = self.epub
```
leonardr: Just want to confirm that the URL here leads to an EPUB and not an ACSM file.

api/enki.py
```diff
@@ -249,17 +294,136 @@ def checkout(self, patron, pin, licensepool, internal_format):
...
+    def patron_activity(self, patron, pin):
+        response = self.patron_request(patron.authorization_identifier, pin)
+        if response.status_code != 200:
+            raise PatronNotFoundOnRemote(response.status_code)
+        result = json.loads(response.content)['result']
+        if not result['success']:
+            message = result['message']
+            if "That record is already checked out to you" in message:
+                # this might be okay
+                pass
```
leonardr: Why would either of these messages happen? You're not trying to create a loan or get a book. If this is copy-and-paste I think you can remove it.

api/enki.py
```diff
@@ -420,7 +584,7 @@ def extract_bibliographic(self, element):
         licenses_available=element["availability"]["availableCopies"]
         hold=element["availability"]["onHold"]
         formats = []
-        formats.append(FormatData(content_type=Representation.EPUB_MEDIA_TYPE, drm_scheme=DeliveryMechanism.ADOBE_DRM))
+        formats.append(FormatData(content_type=Representation.EPUB_MEDIA_TYPE, drm_scheme=EnkiAPI.adobe_drm))
```
leonardr: You should be able to change this back now.

api/enki.py
```diff
@@ -249,17 +294,136 @@ def checkout(self, patron, pin, licensepool, internal_format):
...
+                pass
+            elif "There are no available copies" in message:
+                # TODO: raise no copies error
+                self.log.error("There are no copies of this book available.")
+                raise CirculationException()
+            elif "Login unsuccessful" in message:
+                # TODO: raise invalid barcode/password error
+                self.log.error("User validation against Enki server was unsuccessful.")
+                raise CirculationException()
+        media_type = self.get_enki_media_type(book_id)
```
leonardr: I would call this get_enki_drm_type, since you calculate the media type separately.

api/enki.py
datalogics-dans: My reasoning here is that we should always get either "free" or "acs" for media_type, and in either of those cases we'll have the right type string to return in FulfillmentInfo. If we get something unexpected there (either because we don't have all the information now or because something changes later) we'll assume it's an EPUB and handle it then if that breaks things.

api/enki.py
datalogics-dans: This error message is, I believe, a relic of how I was doing things before. The request we're sending to Enki will never return this message. On the first request, the book will be checked out if it isn't already. On subsequent requests, it will still return that book's information without creating any new loan records.

Issue #632, opened 2017-08-11
Bibliotheca API is using analytics.collect_event incorrectly


aslagle:
The first argument should be a Library, but the BibliothecaAPI is passing it a database session.

It's also missing test coverage, since this didn't break anything.

We should check the other APIs as well.

PR #657, opened 2017-08-30
Properly test the case where BibliothecaCirculationSweep discovers a new work.

leonardr:
This fixes https://github.com/NYPL-Simplified/circulation/issues/632.

I examined all other uses of collect_event() in circulation and they looked fine.

### DIFF
--- a/api/bibliotheca.py
+++ b/api/bibliotheca.py
@@ -648,34 +648,13 @@ class BibliothecaCirculationSweep(IdentifierSweepMonitor):
         identifiers_not_mentioned_by_bibliotheca = set(identifiers)
         now = datetime.datetime.utcnow()
 
-        collection = self.api.collection
         for circ in self.api.get_circulation_for(bibliotheca_ids):
             if not circ:
                 continue
-            bibliotheca_id = circ[Identifier][Identifier.BIBLIOTHECA_ID]
-            identifier = identifiers_by_bibliotheca_id[bibliotheca_id]
-            identifiers_not_mentioned_by_bibliotheca.remove(identifier)
-            pools = [lp for lp in identifier.licensed_through
-                     if lp.data_source.name==DataSource.BIBLIOTHECA
-                     and lp.collection == collection]
-            if not pools:
-                # We don't have a license pool for this work. That
-                # shouldn't happen--how did we know about the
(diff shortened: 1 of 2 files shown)


api/bibliotheca.py
```diff
@@ -700,6 +679,44 @@ def process_batch(self, identifiers):
                 pool.update_availability(0, 0, 0, 0, self.analytics)
                 pool.last_checked = now
 
+    def _process_circulation_data(
```
leonardr: It turns out I didn't really need this helper method, but there's still some things in this method not tested so it seems useful to have it.

api/bibliotheca.py
```diff
@@ -700,6 +679,44 @@ def process_batch(self, identifiers):
...
+            # create one.
+            pool, ignore = LicensePool.for_foreign_id(
+                self._db, self.collection.data_source, identifier.type,
+                identifier.identifier, collection=self.collection
+            )
+
+            # Bibliotheca books are never open-access.
+            pool.open_access = False
+
+            for library in self.collection.libraries:
```
leonardr: This is the line that actually fixes the bug.

tests/test_bibliotheca.py
```diff
@@ -219,6 +220,57 @@ def test_place_hold_fails_if_exceeded_hold_limit(self):
...
+        
+        # Three circulation events were created for this license pool,
+        # marking the creation of the license pool, the addition of
+        # licenses owned, and the making of those licenses available.
+        circulation_events = self._db.query(CirculationEvent).join(LicensePool).filter(LicensePool.id==pool.id)
+        eq_(3, circulation_events.count())
+        types = [e.type for e in circulation_events]
+        eq_(sorted([CirculationEvent.DISTRIBUTOR_LICENSE_ADD,
+                    CirculationEvent.DISTRIBUTOR_TITLE_ADD,
+                    CirculationEvent.DISTRIBUTOR_CHECKIN
```
aslagle: Where did the checkin come from? Is that because the number of available licenses went up and we interpreted it as a checkin?

leonardr: Exactly, that's how we interpret "available licenses went up"

PR #661, opened 2017-09-05
Make registration work in a real test

leonardr:
This branch makes the changes necessary to make a circulation manager able to register itself with the library registry in a real test.

The `Link` header is not used by the library registry, but I'll start picking it up in my follow-up branch that verifies the two-way connection between the Authentication For OPDS document and the OPDS document at the other end of its `start` link.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1909,7 +1909,10 @@ class SettingsController(CirculationManagerController):
             # OPDS Authentication document.
             self._db.commit()
 
-            library_url = self.url_for("index", library_short_name=library.short_name)
+            library_url = self.url_for(
+                "authentication_document", 
+                library_short_name=library.short_name
+            )
             response = do_post(
                 register_url, dict(url=library_url), 
                 allowed_response_codes=["2xx"], timeout=60
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1050,9 +1050,10 @@ class SettingsController(CirculationManagerController):
         if short_name:
             library.short_name = short_name
 
+        NO_VALUE = object()
         for setting in Configuration.LIBRARY_SETTINGS:
             # Start off by assuming the value is not set.
-            value = ""
+            value = NO_VALUE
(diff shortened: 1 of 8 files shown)


api/admin/controller.py
```diff
@@ -1051,6 +1051,8 @@ def libraries(self):
             library.short_name = short_name
 
         for setting in Configuration.LIBRARY_SETTINGS:
+            # Start off by assuming the value is not set.
+            value = ""
```
leonardr: I'm not sure how to test this. The problem I found was that when I didn't upload a logo, the value of the 'logo' configuration option was set to the value I'd set for "Color scheme", the previous configuration option.

leonardr: This code might introduce another problem I've noticed: if I upload a logo, and then make a change, the logo disappears.

aslagle: Good catch. Maybe the value should be None instead of empty string. You could change tests/admin/test_controller.py:TestSettingsController.test_libraries_post_edit to test this but it depends on the order of the settings so it'd be hard to cover everything.

aslagle: The second problem is interesting. Usually if you make a change and don't include things you previously set, that's interpreted as removing them. That doesn't work for images though.

leonardr: See if the change I made works for you.

aslagle: Yeah, that's fine, although we might want to add a way to delete your image without replacing it as well.

tests/admin/test_controller.py
```diff
@@ -1442,6 +1445,10 @@ def test_libraries_post_edit(self):
                 Configuration.ENABLED_FACETS_KEY_PREFIX + FacetConstants.ORDER_FACET_GROUP_NAME,
                 library).value)
 
+        # A library-wide setting that was not updated has been left alone.
```
aslagle: I would clarify this comment since it only applies to images. Other setting types will be removed if they're not included in the edit request.

PR #666, opened 2017-09-07
Fix help links

leonardr:
This branch fixes an error that stopped help links from showing up in feeds (and sometimes caused crashes), and adds the test coverage that would have caught the problem.

### DIFF
--- a/api/opds.py
+++ b/api/opds.py
@@ -499,7 +499,7 @@ class CirculationManagerAnnotator(Annotator):
                 _add_link(d)
 
         for type, value in Configuration.help_uris(self.library):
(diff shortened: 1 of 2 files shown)


api/opds.py
```diff
@@ -499,7 +499,7 @@ def _add_link(l):
                 _add_link(d)
 
         for type, value in Configuration.help_uris(self.library):
-            d = dict(href=setting.value, rel="help")
+            d = dict(href=value, rel="help")
```
leonardr: `setting.value` is the value of the last setting that was considered above, in the previous `for` loop.

PR #667, opened 2017-09-08
Registration with the Metadata Wrangler

courte:
This branch adds routes to register a circulation manager with the metadata wrangler, including the temporary creation of a sitewide public key document.

I'm not sure about the changes that need to happen to the admin interface to create a "register" button. That will be the last step of this change.


api/admin/controller.py
```diff
@@ -1942,3 +1942,67 @@ def library_registrations(self, do_get=HTTP.get_with_timeout, do_post=HTTP.post_
...
+            return MISSING_SERVICE
+
+        # Get the catalog for this service.
+        try:
+            response = do_get(integration.url, allowed_response_codes=['2xx', '3xx'])
+        except Exception as e:
+            return REMOTE_INTEGRATION_FAILED
+
+        catalog = response.json()
+        links = catalog.get('links', [])
```
leonardr: Before looking at the links I would check the media type of the document to make sure we're not registering with a library registry or something.

api/admin/controller.py
```diff
@@ -1942,3 +1942,67 @@ def library_registrations(self, do_get=HTTP.get_with_timeout, do_post=HTTP.post_
...
+
+        # Get the link for registration from the catalog.
+        register_urls = filter(lambda l: l.get('rel')=='register', links)
```
leonardr: I would also check the `type` of the link to make sure it's not offering to register with a library registry or something. Should be looking for `application/opds+json;profile=https://librarysimplified.org/rel/profile/metadata-service`.

api/admin/controller.py
```diff
@@ -1942,3 +1942,67 @@ def library_registrations(self, do_get=HTTP.get_with_timeout, do_post=HTTP.post_
...
+            )
+
+        # Get the full registration url.
+        register_url = register_urls[0].get('href')
+        if not register_url.startswith('http'):
+            # We have a relative path. Create a full registration url.
+            base_url = catalog.get('id')
+            if base_url.endswith('/'):
+                base_url = base_url[:-1]
+            register_url = base_url + register_url
```
leonardr: You should be able to replace this with a call to `urlparse.urljoin`.

leonardr:
Looks good, just needs a little extra checking to make sure it's talking to the right kind of server.

PR #668, opened 2017-09-08
Fix an error in BibliothecaEventMonitor.handle_event and test that method so the problem doesn't recur.

leonardr:


api/bibliotheca.py
```diff
@@ -742,7 +742,11 @@ class BibliothecaEventMonitor(CollectionMonitor):
 
     def __init__(self, _db, collection, api_class=BibliothecaAPI, cli_date=None):
         super(BibliothecaEventMonitor, self).__init__(_db, collection)
-        self.api = api_class(_db, collection)
+        if isinstance(api_class, BibliothecaAPI):
+            # We were given an actual API object. Just use it.
+            self.api = api_class
+        else:
+            self.api = api_class(_db, collection)
```
leonardr: Test support.

api/bibliotheca.py
```diff
@@ -837,7 +841,9 @@ def handle_event(self, bibliotheca_id, isbn, foreign_patron_id,
                      start_time, end_time, internal_event_type):
         # Find or lookup the LicensePool for this event.
         license_pool, is_new = LicensePool.for_foreign_id(
-            self._db, self.api.source, Identifier.BIBLIOTHECA_ID, bibliotheca_id)
+            self._db, self.api.source, Identifier.BIBLIOTHECA_ID, 
+            bibliotheca_id, collection=self.collection
+        )
```
leonardr: The bug fix -- `collection` was null, which isn't allowed.

tests/test_bibliotheca.py
```diff
@@ -61,7 +61,7 @@ def setup(self):
 
     @classmethod
     def sample_data(self, filename):
-        return sample_data(filename, 'threem')
+        return sample_data(filename, 'bibliotheca')
```
leonardr: Remove a place where the old company name was still being used.

tests/test_bibliotheca.py
```diff
@@ -658,3 +658,30 @@ def test_default_start_time(self):
         proper_args = ['2013-04-02']
         default_start_time = monitor.create_default_start_time(self._db, proper_args)
         eq_(datetime.datetime(2013, 4, 2), default_start_time)
+
+    def test_handle_event(self):
```
leonardr: Add test coverage for previously untested method.

leonardr:
I fixed another problem discovered by the same tester and tested `run_once` as well.

api/bibliotheca.py
```diff
@@ -833,7 +833,6 @@ def run_once(self, start, cutoff):
                         exc_info=e
                     )
                 raise e
-            self.timestamp.timestamp = most_recent_timestamp
```
leonardr: This code doesn't work anymore, because `self.timestamp` is now a method. It's no longer necessary because `Monitor.run` takes the return value and sets the `Timestamp` object appropriately.

courte:
👍

PR #670, opened 2017-09-11
Submodule update.

leonardr:


greysteil:
@leonardr - would you be up for automatically creating PRs to update submodules? I just added that functionality to [Dependabot](https://dependabot.com) and would love some beta testers for it.

PR #671, opened 2017-09-12
Admin MW registration

courte:
This change quietly registers a circulation manager with the metadata wrangler without an overhaul to the admin interface forms. Instead, the shared secret won't be shown at all. On submitting the metadata wrangler URL, the shared secret will be generated.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1700,7 +1700,6 @@ class SettingsController(CirculationManagerController):
                 allowed_response_codes=['2xx'], headers=headers
             )
         except Exception as e:
-            set_trace()
             public_key_setting.value = None
             return REMOTE_INTEGRATION_FAILED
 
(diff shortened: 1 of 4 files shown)


api/admin/controller.py
```diff
@@ -1614,11 +1616,106 @@ def metadata_services(self):
...
 
+    def sitewide_registration(self, integration, do_get=HTTP.get_with_timeout,
+        do_post=HTTP.post_with_timeout, key=None
+    ):
+        """Performs a sitewide registration for a particular service, currently
+        only the Metadata Wrangler.
+
+        :return: A ProblemDetail or, if successful, None
+        """
+        if not integration:
```
courte: This method is largely the same as it was below. It takes an integration object now, which it will update with a shared secret if the exchange is successful. In the case of success, it returns `None` -- otherwise a `ProblemDetail` is passed back to be returned to the admin interface.

PR #674, opened 2017-09-13
Fix several errors exposed by using an Authenticator instead of a LibraryAuthenticator as the OAuthController's .auth

leonardr:
In real usage, `OAuthController.auth` is an `Authenticator` object and the methods called on it (e.g. `create_bearer_token` are supposed to be delegated corresponding `LibraryAuthenticator` object. Three of these methods were defined on `LibraryAuthenticator` but the delegate methods on `Authenticator` were not implemented. Tests didn't catch this because our test code populated `OAuthController.auth` with a `LibraryAuthenticator` object, not an `Authenticator`.

This branch adds the missing delegate methods. It introduces a `MockAuthenticator` which makes it easy to test `Authenticator` code outside of a request context, and changes the `OAuthController` tests to use that instead of a `LibraryAuthenticator`. This will catch any future errors of this type.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -23,6 +23,8 @@ from opds import (
     CirculationManagerAnnotator,
 )
 from controller import CirculationManager
+from problem_details import REMOTE_INTEGRATION_FAILED
+from flask.ext.babel import lazy_gettext as _
 
 # TODO: Without the monkeypatch below, Flask continues to process
 # requests while before_first_request is running. Those requests will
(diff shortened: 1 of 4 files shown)


api/routes.py
```diff
@@ -366,6 +368,10 @@ def track_analytics_event(identifier_type, identifier, event_type):
 @requires_auth
 @returns_problem_detail
 def adobe_vendor_id_get_token():
+    if not app.manager.adobe_vendor_id:
+        return REMOTE_INTEGRATION_FAILED.detailed(
+            _("This server does not have an Adobe Vendor ID server configured.")
+        )
```
leonardr: This is an unrelated change to replace a 500 error I found when getting Open Ebooks QA working with a more helpful error. REMOTE_INTEGRATION_FAILED isn't quite right but it's close enough, I think.

api/routes.py
```diff
@@ -403,7 +409,7 @@ def adobe_drm_device(device_id):
 @has_library
 @returns_problem_detail
 def oauth_authenticate():
-    return app.manager.oauth_controller.oauth_authentication_redirect(flask.request.args)
+    return app.manager.oauth_controller.oauth_authentication_redirect(flask.request.args, app.manager._db)
```
leonardr: Another problem I found in real usage.

PR #676, opened 2017-09-14
setup_adobe_vendor_id needs to be given a scoped session or the objects it creates will be given a bad session.

leonardr:
My previous branch didn't go far enough. The database session passed into the AdobeVendorIDController was not the scoped session. I had to go up a level to make sure the scoped session was available to pass in in the first place.

### DIFF
--- a/migration/20170713-13-move-authentication-configuration-to-external-integrations.py
+++ b/migration/20170713-13-move-authentication-configuration-to-external-integrations.py
@@ -109,8 +109,8 @@ if not Configuration.instance:
     # No need to import configuration if there isn't any.
     sys.exit()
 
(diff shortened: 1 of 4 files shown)


migration/20170713-13-move-authentication-configuration-to-external-integrations.py
```diff
@@ -109,8 +109,8 @@ def convert_clever(_db, integration, provider):
     # No need to import configuration if there isn't any.
     sys.exit()
 
+_db = production_session()
 try:
-    _db = production_session()
```
leonardr: An unrelated change that stops this migration script from crashing awkwardly if it's going to crash.

PR #677, opened 2017-09-14
Fixed CORS decorator to handle config in the db.

aslagle:
The CORS headers to support the patron web client stopped working when we moved the configuration into the database.

It's no longer possible to look up a ConfigurationSetting when the routes are created, so I couldn't use the cross_origin decorator from flask_cors and had to write my own based on it. 

As a bonus, this means you can edit the patron web client url in the admin interface and changes will take effect without restarting the app.

Unfortunately I don't think there's any way to have an automated test for this.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -1,5 +1,5 @@
 from nose.tools import set_trace
-from functools import wraps, partial
+from functools import wraps, update_wrapper
 import os
 
 import flask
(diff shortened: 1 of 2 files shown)


api/routes.py
```diff
@@ -94,33 +95,40 @@ def decorated(*args, **kwargs):
             return f(*args, **kwargs)
     return decorated
 
-patron_web_url = None
-if (hasattr(app, 'manager') and hasattr(app.manager, '_db')):
```
aslagle: This was the problem - app.manager wasn't created yet when this code ran.

PR #682, opened 2017-09-15
Fix NoveListAPI

courte:
Some embarrassing bugfixes for NoveListAPI -- this version actually sends the authentication details, helpfully enough.

Because I don't want to hold the authentication details in the database, I try to find a cached version of the recommendations outside of `Representation.get` and scrub the authentication details out of the URL. I also cut the max cache age down to one week.

Fixes #682.


courte:
I made one more change while you were reviewing--just in the method name on core & a logic misstep. Are those good to go, too?

leonardr:
Yes!

PR #692, opened 2017-09-21
Reduce the risk of race conditions in CirculationManager.load_settings

leonardr:
load_settings() had a lot of code that put the CirculationManager into an invalid state and then fixed it. Fixing it sometimes took several seconds, and in those intervening seconds incoming requests would fail. This branch tries to make sure CirculationManager is always in a valid (if not self-consistent) state by calculating the new values first and then assigning them to CirculationManager instance variables.

The only thing that still gets cleared out is `authentication_for_opds_documents`, which was generated on demand to begin with. It's not a problem if an A4OPDS document gets generated twice, since it'll get the same document each time.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -173,29 +173,24 @@ class CirculationManager(object):
         self.analytics = Analytics(self._db)
         self.auth = Authenticator(self._db, self.analytics)
 
-        self.__external_search = None
-        self.external_search_initialization_exception = None
+        self.setup_external_search()
         
         # Track the Lane configuration for each library by mapping its
         # short name to the top-level lane.
-        self.top_level_lanes = {}
-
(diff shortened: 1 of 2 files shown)


leonardr:
I'm currently working on figuring out this test failure:

```
FAIL: tests.test_controller.TestCirculationManager.test_load_settings

  File "/home/travis/virtualenv/python2.7.13/lib/python2.7/site-packages/nose/case.py", line 197, in runTest

    self.test(*self.arg)

  File "/home/travis/build/NYPL-Simplified/circulation/tests/test_controller.py", line 351, in test_load_settings

    DeviceManagementProtocolController)

AssertionError: 
```

I can't duplicate it locally and it doesn't happen all the time on Travis. Any help is appreciated.

PR #694, opened 2017-09-21
Moved metadata uploading from OPDS for distributors importer to a coverage provider.

### LINKED ISSUE #686: OPDS for Distributors import for a book should not be complete until its data has been sent to MW
Currently OPDSForDistributorsImporter calls `MetadataWranglerOPDSLookup.add_with_metadata` on a full feed of information, after the OPDS import has completed. If the `add_with_metadata` call fails (which it generally does, due to a timeout) there's no further opportunity to send that information over, because all the entries in that feed have CoverageRecords indicating that the import is complete.

I can think of three possible solutions. 

1. Call `add_with_metadata` _first_, before doing the OPDS import. Currently we send a generated OPDS feed obtained after running the original OPDS feed through the metadata layer. Since that happens during the import, doing this before the import means sending the original feed as obtained from the distributor. However, this doesn't solve the timeout problem.

1a. So we'd also need to split the original OPDS feed into entries and send each entry individually, rather than sending the original feed.

1b. OR, we can send the original feed and have the metadata wrangler store it for later processing, rather than parsing and processing it while the circulation manager is waiting on the other end of the line.

2. Make `add_with_metadata` part of `Metadata.apply()`, just like uploading images to S3. If `OPDSImporter` was initialized with a `metadata_client`, call `add_with_metadata` individually on every entry it processes, and if the call fails, create a CoverageFailure for a new monitor to clean up by calling `add_with_metadata` again.

This has the advantage that we start sending the metadata wrangler the metadata we get from _all_ sources, not just OPDS For Distributors feeds.

aslagle:
This fixes #686.

Instead of uploading to the metadata wrangler at the end of an OPDS for distributors import, the metadata layer records a transient failure with the upload operation, and the MetadataUploadCoverageProvider does the uploading later.

It made sense to me to associate the CoverageRecords with the Collection, but none of the other CoverageProviders do that so I'm not sure.


api/admin/controller.py
```diff
@@ -1636,6 +1636,7 @@ def metadata_services(self, do_get=HTTP.get_with_timeout,
                 service, do_get=do_get, do_post=do_post, key=key
             )
             if problem_detail:
+                self._db.rollback()
```
aslagle: I noticed this while testing - if you're setting up a Metadata Wrangler integration and registering fails, we should rollback the creation of that integration so you can try the registration again.

leonardr: Not for this branch, but it would also be useful to see the problem detail document the metadata wrangler sent.

api/coverage.py
```diff
@@ -400,6 +403,87 @@ def finalize_batch(self):
...
+        small batch. Not efficient, but it works.
+        """
+        [result] = self.process_batch([identifier])
+        return result
+
+    def add_coverage_record_for(self, item):
+        """Record this CoverageProvider's coverage for the given identifier.
+        Unlike other CoverageProviders, the CoverageRecord will be associated
+        with the collection. This is because different collections may have
+        different information about the same book that needs to be uploaded.
```
leonardr: I don't think this is true given that the thing we're uploading is the Work's OPDS entry. All versions of the book are going to have the same Work.

aslagle: Cool, I can probably get rid of these then.

PR #696, opened 2017-09-21
Merge hotfixes

leonardr:
This branch merges into the main branch some hotfixes I applied today to the NYPL circulation manager.

### DIFF
--- a/api/adobe_vendor_id.py
+++ b/api/adobe_vendor_id.py
@@ -185,6 +185,7 @@ class AdobeVendorIDRequestHandler(object):
         try:
             data = parser.process(data)
         except Exception, e:
+            logging.error("Error processing %s", data, exc_info=e)
             return self.error_document(self.AUTH_ERROR_TYPE, str(e))
         user_id = label = None
         if not data:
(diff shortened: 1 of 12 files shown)


api/adobe_vendor_id.py
```diff
@@ -185,6 +185,7 @@ def handle_signin_request(self, data, standard_lookup, authdata_lookup):
         try:
             data = parser.process(data)
         except Exception, e:
+            logging.error("Error processing %s", data, exc_info=e)
```
leonardr: Without this, we know when there are problems processing Adobe Vendor ID SignIn requests but we don't know the payloads that caused the problems.

api/controller.py
```diff
@@ -327,16 +327,17 @@ def setup_adobe_vendor_id(self, _db, library):
...
         )
 
+        new_adobe_vendor_id = None
         if adobe:
             # Relatively few libraries will have this setup.
             vendor_id = adobe.username
             node_value = adobe.password
             if vendor_id and node_value:
-                if self.adobe_vendor_id:
+                if new_adobe_vendor_id:
```
leonardr: Sometimes `self.adobe_vendor_id` wasn't set and this was crashing. I couldn't find the reason for this, but the new idiom is safer and avoids rapid changes to `self.adobe_vendor_id` while the configuration is loading.

PR #697, opened 2017-09-22
Fix scripts

leonardr:
This branch fixes some minor issues that prevented scripts from running, as well as two bits of code that would cause problems if one collection lost access to a given book while another collection still had access to that book.

### DIFF
--- a/api/axis.py
+++ b/api/axis.py
@@ -217,30 +217,23 @@ class Axis360API(BaseAxis360API, Authenticator, BaseCirculationAPI):
                     "Was about to reap %r but no local license pool in this collection.",
                     removed_identifier
                 )
-            self.reap(pool)
+                continue
+            if pool.licenses_owned == 0:
+                # Already reaped.
+                continue
+            self.log.info(
+                "Reaping %r", removed_identifier
+            )
 
-    def reap(self, pool):
-        """Actually reap a LicensePool."""
-        if pool.collection != self.collection:
-            # Safeguard to stop Collection A's API from reaping
-            # Collection B's licenses.
-            return
-        if pool.licenses_owned == 0:
-            # Already reaped.
-            return
-        self.log.info("Reaping %r", pool.identifier)
-
-        availability = CirculationData(
-            data_source=pool.data_source,
-            primary_identifier=pool.identifier,
-            licenses_owned=0,
-            licenses_available=0,
-            licenses_reserved=0,
-            patrons_in_hold_queue=0,
-        )
-        availability.apply(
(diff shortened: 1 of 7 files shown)


api/axis.py
```diff
@@ -336,7 +336,7 @@ def __init__(self, _db, collection, api_class=Axis360API):
         else:
             self.api = api_class(_db, collection)
 
-    def process_batch(self, identifiers):
+    def process_items(self, identifiers):
```
leonardr: This is an API change. process_batch still exists but it takes an offset now. We want to use the default process_batch which finds the actual identifiers and calls process _items.

api/enki.py
```diff
@@ -221,8 +221,8 @@ def reaper_request(self, identifier):
             data = json.loads(response.content)
             self.log.debug ("Keeping existing book: " + str(identifier))
         except:
-            # Get the license pool for the ID, but make sure it's the one belonging to Enki
-            pool = [x for x in identifier.licensed_through if x.data_source.name == DataSource.ENKI][0]
+            # Get this collection's license pool for this identifier.
+            pool = identifier.licensed_through_collection(self.collection)
```
leonardr: If two libraries have two Enki collections it's possible for the old code to return more than one LicensePool, and it might not be the one we're supposed to reap.

api/routes.py
```diff
@@ -433,7 +433,8 @@ def oauth_callback():
 def heartbeat():
     return app.manager.heartbeat.heartbeat()
 
-@app.route('/service_status')
+@library_route("/service_status")
+@has_library
```
leonardr: service_status shows the status of the integrations for one specific library.

bin/update_nyt_best_seller_lists
```diff
@@ -5,7 +5,7 @@ import sys
 bin_dir = os.path.split(__file__)[0]
 package_dir = os.path.join(bin_dir, "..")
 sys.path.append(os.path.abspath(package_dir))
-from core.scripts import (
+from scripts import (
```
leonardr: This script moved from core into circulation in 2.0.

api/axis.py
```diff
@@ -211,10 +211,10 @@ def update_licensepools_for_identifiers(self, identifiers):
         # k books are the identifiers in `remainder`. These books have
         # been removed from the collection without us being notified.
         for removed_identifier in remainder:
-            pool = removed_identifier.licensed_through
+            pool = identifier.licensed_through_collection(self.collection)
```
leonardr: removed_identifier.licensed_through now returns a list. We only want the license for this identifier through this collection.

PR #699, opened 2017-09-25
The Axis 360 availability parser needs to set collection_id, not collection (which is a method).

leonardr:
This branch fixes an error that is preventing Axis 360 loans from showing up in the circulation manager. There's old code that sets `CirculationInfo.collection` to a Collection object, overwriting the `CirculationInfo.collection()` _method_ that `CirculationAPI` expects to be able to call. Removing this code solves the problem.


api/controller.py
```diff
@@ -706,7 +706,9 @@ def sync(self):
             except Exception, e:
                 # If anything goes wrong, omit the sync step and just
                 # display the current active loans, as we understand them.
-                self.manager.log.error("ERROR DURING SYNC: %r", e, exc_info=e)
+                self.manager.log.error(
+                    "ERROR DURING SYNC for %s: %r", patron.id, e, exc_info=e
+                )
```
leonardr: Indicating in the error message which patron had a sync problem in the error message will make future problems of this sort easier to debug. Using the internal database ID instead of the patron's actual identifier reduces the risk of leaking information about the patron.

PR #708, opened 2017-09-26
Logging update

leonardr:
This branch updates circulation to use the new logging configuration code.

Logging is re-configured inside `load_settings`, but I haven't written a configuration interface to manage logging. I filed https://github.com/NYPL-Simplified/circulation/issues/709 to deal with that, and I think we can work on that later.

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -16,6 +16,7 @@ from core.model import (
     ConfigurationSetting,
     SessionManager,
 )
+from core.log import LogConfiguration
 from core.util import LanguageCodes
 from flask.ext.babel import Babel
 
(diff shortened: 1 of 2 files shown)


api/app.py
```diff
@@ -40,7 +41,9 @@
 import routes
 import admin.routes
 
-debug = Configuration.logging_policy().get("level") == 'DEBUG'
+log_level = LogConfiguration.initialize(_db, testing=testing)
```
leonardr: I don't really like calling `initialize` again just to get the current log level, but I don't think it hurts performance all that much.

PR #717, opened 2017-10-02
Turn off gzip encoding for Enki

datalogics-dans:
Full disclaimer: this might be addressing a symptom rather than the
disease, but it works. If the CM needs to get the content for an Enki
book that is not DRM-protected, it will explicitly ask for it without
any content encoding. The clients should be able to handle gzip
encoding, but for some reason iOS was unable to download these books.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -941,6 +941,11 @@ class LoanController(CirculationManagerController):
             )
 
         headers = dict()
+        encoding_header = dict()
+        if (fulfillment.data_source_name == DataSource.ENKI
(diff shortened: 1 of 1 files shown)


datalogics-dans:
Just some testing notes here...
* Confirmed the new code block isn't hit for Overdrive books
* Confirmed it isn't hit for Bibliotheca books
* Confirmed it isn't hit for DRM-protected Enki books
* Confirmed it _is_ hit for non-DRM-protected Enki books
  * These books are downloadable and readable on both Android and iOS, with the original problem being that iOS couldn't download them

PR #720, opened 2017-10-03
Enki thumbnails

datalogics-tarar:
Updated Enki API to use thumbnail images. The small images provided by Enki are used as thumbnails and the large images are used as cover images since the large images don't show up as thumbnails.


datalogics-tarar:
@leonardr I tried using the large images instead and the book covers didn't show up. The Enki import script was replacing covers fine so I didn't think I'd need to wipe the DB first. Also, the small images looked fine when I tested.

leonardr:
What do you mean when you say the book covers didn't show up? I can think of three things this might mean:

1. The URLs to the book covers were not recorded in `editions.cover_thumbnail_url` and `editions.cover_full_url`.
2. The images were not properly included as `<link>` tags in the OPDS feeds served by the circulation manager.
3. The images did not show up when the OPDS feed was rendered in a mobile client.

If the data is correct in the database (i.e. we can confirm that the problem is not #1) then any other problem should be fixable by clearing or repopulating caches. There are enough caches in between `editions.cover_thumbnail_url` and the screen rendered by a mobile phone that I would like to eliminate #1 as a potential problem before moving forward.

The unit test I outlined in my original comment should be enough to confirm or eliminate problem #1. Then we can investigate what needs to change on a preexisting site before the value in `editions.cover_thumbnail_url` ends up in an OPDS feed that is rendered in a mobile client. If the data is not correct in the database, (i.e. the problem is #1) then there is a problem in core which I need to fix before we can proceed.

What device did you use to test the legibility of the sample images? Eyeballing the 'small'-sized Enki images, I think they would appear okay on a mobile phone but unacceptably small on a high-definition tablet. A problem with the data model (assuming there is one) is not a good reason to accept images that are significantly smaller than our standard size -- the underlying problem needs to be fixed.

PR #721, opened 2017-10-03
Docker submodule update

courte:
This submodule change removes the use of a base image, which will keep the automatically generated deploy and scripts development containers in sync with the code base.


courte:
Thank you for the thorough review!

PR #724, opened 2017-10-04
Added an endpoint for editing custom lists.

aslagle:


api/admin/controller.py
```diff
@@ -805,6 +807,81 @@ def genres(self):
...
+        library = flask.request.library
+
+        data_source = DataSource.lookup(self._db, DataSource.LIBRARY_STAFF)
+
+        if flask.request.method == "GET":
+            custom_lists = []
+            for list in library.custom_lists:
+                entries = []
+                for entry in list.entries:
+                    entries.append(dict(id=entry.edition.primary_identifier.urn,
```
aslagle: Is there something better I could be using to identify editions? It needs to be something the client can get from the opds entry.

api/admin/controller.py
```diff
@@ -805,6 +807,81 @@ def genres(self):
...
+                    Collection.libraries
+                ).filter(
+                    Identifier.id==identifier.id
+                ).filter(
+                    Library.id==library.id
+                ).all()
+                if not pools:
+                    return CANNOT_ADD_CUSTOM_LIST_ENTRY_WITHOUT_LICENSEPOOL
+
+                edition = pools[0].presentation_edition
```
aslagle: I'm not sure if using the first pool is the best way to do this...

api/admin/controller.py
leonardr: I think it would be better to identify the list items by their work (through permanent work ID). CustomListEntry.editions is used for situations like the NYT best-sellers, where we know that a book with certain bibliographic information is on a list, but we don't know if we have any licenses for that book and we're not convinced that the bibliographic information is even accurate. When you're editing a list by choosing books from search results, you know that the books are already in the system as Works.

It's currently possible for multiple Works to have the same permanent work ID, but the same is true of an Edition's primary identifier, so this change shouldn't cause a new problem.

api/admin/controller.py
```diff
@@ -805,6 +807,81 @@ def genres(self):
...
+            else:
+                list, is_new = create(self._db, CustomList, name=name, data_source=data_source)
+                list.created = datetime.now()
+                list.library = library
+
+            list.updated = datetime.now()
+            list.name = name
+
+            if entries:
+                entries = json.loads(entries)
```
leonardr: The process of turning the JSON payload into a list of Works can go into a separately testable method.

api/admin/controller.py
leonardr: This is where switching to Works will pay off.

api/admin/problem_details.py
```diff
@@ -253,3 +253,24 @@
...
+    title=_("Cannot change library for custom list"),
+    detail=_("A custom list's associated library cannot be changed once it is set.."),
+)
+
+CANNOT_ADD_CUSTOM_LIST_ENTRY_WITHOUT_LICENSEPOOL = pd(
+    "http://librarysimplified.org/terms/problem/cannot-add-entry-without-licensepool",
+    status_code=400,
+    title=_("Cannot add custom list entry without license pool."),
+    detail=_("You can't add an entry without a license pool to a custom list."),
+)
```
leonardr: Will you want to prevent the addition of a Work to a list if the library has no LicensePool for it? It's possible for a library to lose Works because it loses access to a collection, so we can't stop this from happening in all cases, but it might make sense to make the editor filter out attempts to add Works to a list when we know they won't show up.

aslagle: I think I should remove this restriction - editors will be adding to the list from search results, and works without a pool won't show up in the search results. If a work loses its pool after being added, this restriction could make it impossible to save the list until you remove that work. I think it's better to leave the work in the list and it will still show in the admin interface but it won't show up in lanes. And if the pool comes back, it can start showing up again.

aslagle:
@leonardr I made the change to works instead of editions.

I also changed the create and edit responses to include the list id. The UI only uses it for create, so that it can put you on the edit page for the list once it is created.

And I added a route for deleting a list. I made both of these put the CSRF token in a header instead of form data.

I want to do a follow-up branch to make other things in the admin interface consistent with this - add deletion, use the CSRF token header, and have the create and edit methods return the id.

api/admin/controller.py
```diff
@@ -852,29 +852,16 @@ def custom_lists(self):
...
-                edition = pools[0].presentation_edition
-                list.add_entry(edition, featured=True)
-
-            new_urns = [entry.get("id") for entry in entries]
+                pwid = entry.get("pwid")
+                edition = get_one(self._db, Edition,
+                                  on_multiple='interchangeable',
+                                  permanent_work_id=pwid)
+                if edition and edition.work:
+                    list.add_entry(edition.work, featured=True)
```
leonardr: This code could be a little nicer if you brought back the joins and got one Work that has an Edition with the given permanent work ID.

PR #726, opened 2017-10-10
Use index coverageprovider

leonardr:
This branch creates a new script that incrementally recreates the search engine to reflect changes in availability.


aslagle:
Should we delete the old bin/repair/search_index script, or maybe change it to delete CoverageRecords and use the new coverage provider?

leonardr:
I've changed the search index Monitor to create new CoverageRecords as necessary. Rather than figure out all the potential migration paths, I'd like to spend some time making it really fast to create/update a bunch of CoverageRecords. Then it won't matter what happened in the past.

PR #728, opened 2017-10-10
Optional fields

leonardr:
This branch adds the 'optional' flag to as many configuration items as possible, so that someone who sets a value by accident can generally unset it.

Note that both the test username and test password are optional, even though I'd like the test username to be required in almost all cases, just because it doesn't make sense to set username for OAuth-style authentication mechanisms.


api/authenticator.py
```diff
@@ -1194,16 +1194,22 @@ class BasicAuthenticationProvider(AuthenticationProvider):
     SETTINGS = [
         { "key": TEST_IDENTIFIER,
           "label": _("Test Identifier"),
-          "description": _("A valid identifier that can be used to test that patron authentication is working.") },
-        { "key": TEST_PASSWORD, "label": _("Test Password"), "description": _("The password for the test identifier.") },
+          "description": _("A valid identifier that can be used to test that patron authentication is working."),
+          "optional": True,
```
aslagle: This is already the BasicAuthenticationProvider, so you don't need to make it optional for OAuth providers...

aslagle:
I don't think the library code in the admin interface is set up to handle the 'optional' setting yet - it treats all the fields as optional if I remember correctly.

aslagle:
The `libraries` method in `api/admin/controller.py:SettingsController` would need an update as well.

leonardr:
I've updated SettingsController but I don't understand how that's different from your earlier comment about "the library code in the admin interface".

aslagle:
Sorry, those comments were about the same thing - I was just being more specific in the second one.

tests/admin/test_controller.py
```diff
@@ -1459,21 +1463,35 @@ def test_libraries_post_edit(self):
...
-                library).value)
+            val(Configuration.ENABLED_FACETS_KEY_PREFIX + FacetConstants.ORDER_FACET_GROUP_NAME)
+        )
 
         # The library-wide logo was not updated and has been left alone.
         eq_("A tiny image", 
             ConfigurationSetting.for_library(Configuration.LOGO, library).value
         )
+
+    def test_libraries_post_errors(self):
```
aslagle: There was already a test_libraries_post_errors above the previous test, you might want to add on to that.

PR #731, opened 2017-10-13
Use search code from core

leonardr:
This branch updates the core submodule to take advantage of the new search indexing code (which is about ten times faster), and remove code that's been copied to core.


leonardr:
I'll be updating this branch to use the submodule from https://github.com/NYPL-Simplified/server_core/pull/681, once both branches are approved.

scripts.py
```diff
@@ -60,6 +60,10 @@
     PatronInputScript,
     RunMonitorScript,
 )
+# TODO: This class used to be defined in this file and was moved into core.
+# The import statement can be removed once we're sure all references to it
+# have been changed.
+from core.scripts import UpdateSearchIndexScript
```
aslagle: I can't think of any reason something would be referencing this, other than the script file you already changed or a test.

PR #735, opened 2017-10-20
CSRF token consistency and other admin interface fixes

aslagle:
This branch changes the admin interface to only accept CSRF tokens in an "X-CSRF-Token" header - form data no longer works. This change is *not* backwards-compatible with the old version of the front-end code, so everyone will have to re-run npm install when updating or the admin interface will no longer work.

In addition, I found a couple other problems with the admin interface while testing this and fixed them here.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -207,9 +207,8 @@ class AdminController(object):
         data.
         """
         cookie_token = self.get_csrf_token()
-        form_token = flask.request.form.get("csrf_token")
         header_token = flask.request.headers.get("X-CSRF-Token")
-        if not cookie_token or cookie_token != (form_token or header_token):
+        if not cookie_token or cookie_token != header_token:
             return INVALID_CSRF_TOKEN
         return cookie_token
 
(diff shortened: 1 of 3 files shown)


api/admin/controller.py
```diff
@@ -540,8 +544,8 @@ def refresh_metadata(self, identifier_type, identifier, provider=None):
 
         if record.exception:
             # There was a coverage failure.
-            if (isinstance(record.exception, int)
-                and record.exception in [201, 202]):
+            if (str(record.exception).startswith("201") or
+                str(record.exception).startswith("202")):
```
aslagle: I'm not sure about this change. I was getting "202: I'm working to locate a source for this identifier." instead of an int when I tested it.

api/admin/opds.py
```diff
@@ -19,16 +19,11 @@ def annotate_work_entry(self, work, active_license_pool, edition, identifier, fe
...
-        if isinstance(work, BaseMaterializedWork):
-            data_source_name = work.name
-        else:
-            data_source_name = active_license_pool.data_source.name
-
         feed.add_link_to_entry(
             entry,
             rel="http://librarysimplified.org/terms/rel/refresh",
             href=self.url_for(
-                "refresh", data_source=data_source_name,
```
aslagle: These were adding an unnecessary query parameter now that data source is no longer part of these routes.

api/admin/controller.py
leonardr: I think the new behavior is right. `record.exception` is created from `OPDSMessage.status_code` and `OPDSMessage.message`, in `OPDSImporter.coveragefailure_from_message`

PR #737, opened 2017-10-22
ODL with consolidated copies

aslagle:
This pull request adds support for ODL with consolidated copies for the DPLA exchange.

ODL (Open Distribution to Libraries) is a specification that allows libraries to manage their own loans and holds. It offers a deeper level of control to the library, but implementing full ODL support will require changing the circulation manager to keep track of individual copies rather than license pools, and manage its own holds queues.

'ODL With Consolidated Copies' builds on ODL to provide an API that is more consistent with what other distributors provide. In addition to an ODL feed, the 'ODL With Consolidated Copies' distributor provides an endpoint to get a consolidated copies feed. Each consolidated copy has the total number of licenses owned and available across all the library's copies. In addition, the distributor provides an endpoint to create a loan for a consolidated copy, rather than an individual copy. That endpoint returns an License Status Document (https://readium.github.io/readium-lsd-specification/) and can also be used to check the status of an existing loan.

Holds are not supported yet.

When the circulation manager has full ODL support, the consolidated copies code can be removed.

This depends on https://github.com/NYPL-Simplified/server_core/pull/687.


api/circulation.py
```diff
@@ -585,6 +591,15 @@ def revoke_loan(self, patron, pin, licensepool):
...
         )
         if loan:
+            if not licensepool.open_access:
+                api = self.api_for_license_pool(licensepool)
+                try:
+                    api.checkin(patron, pin, licensepool)
+                except NotCheckedOut, e:
+                    # The book wasn't checked out in the first
+                    # place. Everything's fine.
+                    pass
```
aslagle: I changed this to call the API's `checkin` before deleting the local Loan, because for ODL the Loan's external identifier is needed to return a loan with the API.

aslagle:
After discussing with Hadrien, I may need to change this to store a License Status Document URI instead of the UUID as the external identifier for a Loan. But there's currently a bug in the LSD self links in Feedbooks' consolidated copies API.

aslagle:
Storing the status document URI would also mean I don't need to store a UUID for the patron - I can generate a random one for every loan, and never need it again.

api/odl.py
```diff
@@ -0,0 +1,598 @@
...
+        else:
+            checkout_id = str(uuid.uuid1())
+            expires = datetime.datetime.utcnow() + datetime.timedelta(days=self.default_loan_period)
+            # The patron UUID is generated randomly on each loan, so the distributor
+            # doesn't know when multiple loans come from the same patron.
+            patron_id = str(uuid.uuid1())
+            notification_url = self._url_for(
+                "odl_notify",
+                library_short_name=patron.library.short_name,
+                loan_external_identifier=checkout_id,
```
aslagle: This will no longer work due to my change to make the external identifier the URL instead of the UUID. I'll need to come up with something else for the notify route.

leonardr: You get to choose this URL, right? Can you use the loan's internal database ID?

aslagle: I have to send the URL as part of the request to create the loan, so the loan isn't in the db at this point. I could create it and then delete it after if it didn't work.

api/odl.py
```diff
@@ -0,0 +1,598 @@
...
+        loan = loan.one()
+
+        doc = self.get_license_status_document(licensepool, patron, loan)
+        status = doc.get("status")
+        if status in [self.REVOKED_STATUS, self.RETURNED_STATUS, self.CANCELLED_STATUS, self.EXPIRED_STATUS]:
+            # This loan was already returned early or revoked by the distributor, or it expired.
+            raise NotCheckedOut()
+        elif status == self.ACTIVE_STATUS:
+            # This loan has already been fulfilled, so it needs to be returned through the DRM system.
+            raise CannotReturn()
```
leonardr: Generally, in cases where the web-based return doesn't actually do any returning, it's a no-op rather than an exceptional condition. I think the client always knows to do both a web-based return and a DRM return.

api/odl.py
```diff
@@ -0,0 +1,598 @@
...
+        return_url = doc.get("links", {}).get("return", {}).get("href")
+        if return_url:
+            # Hit the distributor's return link.
+            self._get(return_url)
+            # Get the status document again to make sure the return was successful,
+            # and if so update the pool availability and delete the local loan.
+            self.update_loan(loan)
+        else:
+            # The distributor didn't provide a link to return this loan.
+            raise CannotReturn()
```
leonardr: Whereas this I think is a good place to raise `CannotReturn` -- there's something wrong with the remote server.

api/odl.py
```diff
@@ -0,0 +1,598 @@
...
+            # Something went wrong with this loan and we don't actually
+            # have the book checked out. This should never happen.
+            raise CannotLoan()
+
+        # We have successfully borrowed this book.
+        licensepool.licenses_available -= 1
+
+        expires = doc.get("potential_rights", {}).get("end")
+        expires = datetime.datetime.strptime(expires, self.TIME_FORMAT)
+        external_identifier = doc.get("links", {}).get("self", {}).get("href")
```
leonardr: Given the importance of the external_identifier I think the lack of one should be treated as an error condition.

tests/test_controller.py
```diff
@@ -2572,6 +2573,48 @@ def test_device_id_handler_bad_method(self):
...
+               "id": loan.external_identifier,
+               "status": "revoked",
+            })
+            response = self.manager.odl_notification_controller.notify(
+                loan.external_identifier)
+            eq_(200, response.status_code)
+
+            # The pool's availability has been updated.
+            api = self.manager.circulation_apis[self._default_library.id].api_for_license_pool(loan.license_pool)
+            eq_([loan.license_pool], api.availability_updated_for)
```
leonardr: It should also check the effect on the Loan object, right?

leonardr: `update_loan` is tested separately, but I don't think you have a test that `notify` calls `update_loan`, and this would be the place to put it.

api/odl.py
leonardr: I think that's better than creating another place to store the UUID, because once you have two extra pieces of information associated with each loan, we need to talk about reusing the `resources` table or something. 

Maybe create the Loan object with as little information as possible just to get the ID, and then fill in the rest later.

But we'll need to make sure that everything works in case someone gets an extra Loan with no associated LicensePool.

leonardr: I should say I'm not opposed to creating a new database table for keeping track of information that dies with the loan. Reusing `resources` is probably not a good idea since that's full of bibliographic information that never dies.

tests/test_controller.py
aslagle: I implemented `update_loan` in the mock api class and had it set `api.availability_updated_for`, so this line verifies that it was called.

api/odl.py
aslagle: This is a little tricky because the local loan gets deleted, but the local license pool availability doesn't get updated. If I use the loan id or something else associated with the loan for notifications, it won't be there anymore when the book does get returned. But the license pool will update the next time the consolidated copies monitor runs. That's probably good enough. Worst case is a book is actually available for a while before the circ manager finds out.

api/odl.py
aslagle: I changed it to create the loan object with just the patron and pool and use its id, then delete the loan if the checkout fails.

tests/test_controller.py
leonardr: Ok, cool.

api/odl.py
```diff
@@ -280,19 +287,26 @@ def checkout(self, patron, pin, licensepool, internal_format):
...
+        loan, ignore = get_one_or_create(_db, Loan, patron=patron, license_pool_id=licensepool.id)
+
+        doc = self.get_license_status_document(loan)
         status = doc.get("status")
 
         if status not in [self.READY_STATUS, self.ACTIVE_STATUS]:
             # Something went wrong with this loan and we don't actually
             # have the book checked out. This should never happen.
+            # Remove the loan we created.
+            _db.delete(loan)
```
leonardr: If there's a preexisting loan, and then `checkout` is called, and it's revealed that the status is not 'ready' or 'active', then the loan will be deleted. I think this is correct, or at least not a problem, but I want to confirm because it's not what I was expecting to happen here.

aslagle: If there's a preexisting loan, checkout will raise AlreadyCheckedOut() before it gets here.

leonardr:
I like where this is going and I'd like to merge it to minimize the risk of conflicts later on.

PR #740, opened 2017-10-23
Added deletion to config pages in the admin interface.

aslagle:


api/admin/routes.py
```diff
@@ -288,6 +288,18 @@ def libraries():
...
+@requires_admin
+@requires_csrf_token
+def library(library_uuid):
+    data = app.manager.admin_settings_controller.library(library_uuid)
+    if isinstance(data, ProblemDetail):
+        return data
+    if isinstance(data, Response):
+        return data
+    return flask.jsonify(**data)
+
```
leonardr: There are so many copies of this code -- do a thing, check what was returned, jsonify it -- that it might be worth creating yet another decorator and putting the code in there.

aslagle: Done.

leonardr: This looks great!

PR #742, opened 2017-10-24
Move database initialization code into a hook method that runs before the first request.

leonardr:
Currently database initialization code runs when `app` is imported. This requires a special step for scripts (which may want to import `app` but which are initialized with their own database connection) and it results in problems when combined with the reloader, which we use in development.

The first thing the reloader does is restart the server in a subprocess. Any database connections that were in the middle of a session when the reloader strikes are left hanging. This can cause problems in edge cases but it _always_ causes problems the first time you run a development system, because the main process dies in the middle of database initialization.

This branch moves the database initialization code so that it happens the first time a request comes in, not before.

My main concern is what this will do in production, when multiple threads all call `initialize_database` at once and a number of `flask_scoped_sessions` are created in rapid succession. Only one of them will end up as `app._db`, but I'm not sure what will happen to the others. This problem (assuming there is a problem) wouldn't happen in the current master branch, because the scoped session is created when `app` is imported by UWSGI.


api/app.py
```diff
@@ -20,37 +20,37 @@
...
-autoinitialize = os.environ.get('AUTOINITIALIZE') != 'False'
-if autoinitialize:
-    SessionManager.initialize(db_url)
-session_factory = SessionManager.sessionmaker(db_url)
-_db = flask_scoped_session(session_factory, app)
-if autoinitialize:
-    SessionManager.initialize_data(_db)
 
+app = Flask(__name__)
+app._db = None
```
leonardr: app._db needs to be initialized before it can handle requests, but it doesn't have to happen right now.

api/app.py
```diff
@@ -20,37 +20,37 @@
...
-logging.getLogger().info("Application debug mode==%r" % debug)
-app.config['DEBUG'] = debug
-app.debug = debug
+import routes
+import admin.routes
 
 def run(url=None):
-    base_url = ConfigurationSetting.sitewide(_db, Configuration.BASE_URL_KEY)
-    base_url = url or base_url.value or u'http://localhost:6500/'
+    base_url = url or u'http://localhost:6500/'
```
leonardr: We can no longer use the database BASE_URL_KEY as the URL, because that would require initializing a database connection before calling app.run(), which is what causes the problem. This could be a big problem or it could be nothing; I'm not sure.

api/admin/routes.py
```diff
@@ -32,19 +32,20 @@
...
 @app.before_first_request
-def setup_admin():
+def setup_admin(_db=None):
     if getattr(app, 'manager', None) is not None:
         setup_admin_controllers(app.manager)
+    _db = _db or app._db
+    # The secret key is used for signing cookies for admin login
+    app.secret_key = ConfigurationSetting.sitewide_secret(
+        _db, Configuration.SECRET_KEY
+    )
```
aslagle: Just want to make sure you manually tested the admin interface with this. This secret key is used internally by flask and I don't know at what point it needs to be set.

api/app.py
```diff
@@ -20,37 +20,37 @@
...
+    _db = flask_scoped_session(session_factory, app)
+    app._db = _db
+    if autoinitialize:
+        SessionManager.initialize_data(_db)
 
-log_level = LogConfiguration.initialize(_db, testing=testing)
-debug = log_level == 'DEBUG'
+    log_level = LogConfiguration.initialize(_db, testing=testing)
+    debug = log_level == 'DEBUG'
+    app.config['DEBUG'] = True
```
aslagle: This should be `debug`, not `True`, right?

api/admin/routes.py
leonardr: After some travails we confirmed that the key is still available to Flask soon enough.

courte:
I'm not certain how multiple uwsgi flask_scoped_sessions will work out either, especially since you're only dealing with DataSources and genres here. I think because we've added database initialization options to startup and existing partners are more familiar with the containers, we've seen fewer instances of uwsgi workers fighting each other to create the schema.

I can run tests with the container generated from this PR. If we see overwrite problems, we can brush off `InstanceInitializationScript` and start using it in circulation-docker again.

leonardr:
Thanks, Courteney, give it a try and let's see what happens.

courte:
I tried this branch in a container with and without using the preferred `SIMPLIFIED_DB_TEST="init"` value, and both were able to start the app without problem.

leonardr:
OK, let's try it out and see what happens.

PR #743, opened 2017-10-24
Set default API URLs for the collection integrations that have them.

leonardr:
This branch eliminates the need to remember the standard API endpoints for Axis 360, ENKI, and RBDigital when configuring collections.

### DIFF
--- a/api/enki.py
+++ b/api/enki.py
@@ -83,13 +83,14 @@ from core.testing import DatabaseTest
 
 class EnkiAPI(BaseCirculationAPI):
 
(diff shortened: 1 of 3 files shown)


api/enki.py
```diff
@@ -83,13 +83,14 @@
 
 class EnkiAPI(BaseCirculationAPI):
 
+    PRODUCTION_BASE_URL = "http://enkilibrary.org/API/"
```
aslagle: The production url doesn't use https?

aslagle: I see this was already in the code, but it seems questionable.

Issue #568, opened 2017-07-06
Circ manager should be able to run with an invalid SIP2 integration


aslagle:
Currently the SIP2 auth provider tries to connect when it's initialized, and raises a RemoteIntegrationException if it can't. This means if you configure an invalid SIP2 auth provider and restart the circ manager, you won't be able to use the admin interface to fix it.

The exception should be stored somewhere, but shouldn't break the rest of the application.

datalogics-cgreen:
Also applies to running circulation managers that suddenly cannot connect to their ILS for various reasons.

PR #745, opened 2017-10-24
Dont crash on unconnectable patron authentication providers

leonardr:
This branch fixes https://github.com/NYPL-Simplified/circulation/issues/568 by turning an authentication provider's `RemoteIntegrationException` (which is treated as fatal) into a `CannotLoadConfiguration` (which merely results in an authentication provider not being registered).

### DIFF
--- a/api/sip/client.py
+++ b/api/sip/client.py
@@ -295,7 +295,14 @@ class SIPClient(Constants):
         """Create a socket connection to a SIP server."""
         with self.socket_lock:
             sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            sock.connect((self.target_server, self.target_port))
+            try:
+                sock.connect((self.target_server, self.target_port))
+            except TypeError:
+                raise IOError(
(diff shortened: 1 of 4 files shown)


api/sip/client.py
```diff
@@ -295,7 +295,14 @@ def connect(self):
...
             sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            sock.connect((self.target_server, self.target_port))
+            try:
+                sock.connect((self.target_server, self.target_port))
+            except TypeError:
+                raise IOError(
+                    "Could not connect to %s:%s" % (
+                        self.target_server, self.target_port
+                    )
+                )
```
leonardr: This was necessary for the test to work without a lot of extra scaffolding, but it also handles the case where you've created a SIP2 server and given it absolutely no configuration. Any IOError is turned into a RemoteIntegrationException in SIP2AuthenticationProvider.__init__, but TypeError is such a generic error that I don't want to catch all of them.

PR #747, opened 2017-10-25
Rbdigital improvements

leonardr:
This branch incorporates submodules:
https://github.com/NYPL-Simplified/server_core/pull/690
https://github.com/NYPL-Simplified/circulation-docker/pull/62

It also dramatically improves the performance of the RBdigital availability monitor by not bothering to call CirculationData.apply() if it's obvious that nothing about a LicensePool's availability has changed. When there are no changes to be processed, this takes the run time on my computer from about 5 minutes down to 1 minute.


api/oneclick.py
```diff
@@ -299,8 +299,19 @@ def release_hold(self, patron, pin, licensepool):
...
 
+    @property
+    def default_circulation_replacement_policy(self):
+        return ReplacementPolicy(
+            identifiers=False,
+            subjects=True,
+            contributions=True,
+            formats=True,
+            analytics=Analytics(self._db),
+        )
```
leonardr: When there _are_ changes to be made, reusing this object instead of creating a new one for every LicensePool also saves a little bit of time.

bin/rbdigital_availability_monitor
```diff
@@ -14,12 +14,5 @@ sys.path.append(os.path.abspath(package_dir))
 
 from core.scripts import RunCollectionMonitorScript
 from api.oneclick import OneClickCirculationMonitor
-from api.oneclick import MockOneClickAPI
-
-from api.config import Configuration
-Configuration.load()
-
-library_id = Configuration.integration(Configuration.ONECLICK_INTEGRATION).get("library_id", None)
-print "Updating availability data of library #%s" % library_id
```
leonardr: This code was left over from when there could only be one RBdigital collection on a server, and it no longer worked.

PR #748, opened 2017-10-26
A patron's RBdigital account is now associated with a single-purpose custom identifier

leonardr:
This branch makes two major improvements to the RBdigital integration:

1. Instead of creating an RBdigital account for a patron using their authorization identifier, use a custom identifier which never changes and is only used to identify that patron to RBdigital. This custom identifier is stored in a persistent Credential object, as is the patron's RBdigital account ID.

2. Since the patron's RBdigital account ID is stored in a persistent Credential, there's no need to send an API call to look it up every time we're about to make a request on behalf of the patron. This cuts the number of RBdigital API calls we need to make approximately in half.

Since no one is using RBdigital collections in production, I haven't made any attempt to migrate the system, but anyone who had RBdigital loans before is going to lose them because a new account will be created for them.

### DIFF
--- a/api/oneclick.py
+++ b/api/oneclick.py
@@ -210,11 +210,7 @@ class OneClickAPI(BaseOneClickAPI, BaseCirculationAPI):
         patron_oneclick_id = self.patron_remote_identifier(patron)
         (item_oneclick_id, item_media) = self.validate_item(licensepool)
 
-        # find patron's checkouts
         checkouts_list = self.get_patron_checkouts(patron_id=patron_oneclick_id)
-        if not checkouts_list:
-            raise NoActiveLoan("Cannot fulfill %s - patron %s/%s has no checkouts.", item_oneclick_id, 
-                patron.authorization_identifier, patron_oneclick_id)
 
         # find this licensepool in patron's checkouts
         found_checkout = None
--- a/api/oneclick.py
+++ b/api/oneclick.py
@@ -66,6 +66,8 @@ class OneClickAPI(BaseOneClickAPI, BaseCirculationAPI):
     
     EXPIRATION_DATE_FORMAT = '%Y-%m-%d'
 
+    REMOTE_PATRON_IDENTIFIER_CREDENTIAL_TYPE = "Remote Patron Identifier"
+
     log = logging.getLogger("OneClick Patron API")
    
     def __init__(self, *args, **kwargs):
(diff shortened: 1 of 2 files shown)


api/oneclick.py
```diff
@@ -206,14 +207,10 @@ def fulfill(self, patron, pin, licensepool, internal_format):
...
 
-        patron_oneclick_id = self.validate_patron(patron)
+        patron_oneclick_id = self.patron_remote_identifier(patron)
         (item_oneclick_id, item_media) = self.validate_item(licensepool)
 
-        # find patron's checkouts
         checkouts_list = self.get_patron_checkouts(patron_id=patron_oneclick_id)
-        if not checkouts_list:
-            raise NoActiveLoan("Cannot fulfill %s - patron %s/%s has no checkouts.", item_oneclick_id, 
-                patron.authorization_identifier, patron_oneclick_id)
```
leonardr: This is unnecessary because if you have no checkouts, you don't have the checkout we're looking for, so you fall into the case below, and that error message is more helpful.

api/oneclick.py
```diff
@@ -419,49 +417,78 @@ def internal_format(self, delivery_mechanism):
...
-        # generate random values for the account fields the patron has not supplied us with
-        patron_uuid = str(uuid.uuid1())
-        post_args['userName'] = 'username_' + patron_uuid
-        post_args['email'] = 'patron_' + patron_uuid + '@librarysimplified.org'
+        post_args['libraryCardNumber'] = patron_identifier
+
+        # Generate meaningless values for account fields that are not
+        # relevant to our usage of the API.
+        post_args['userName'] = 'username_' + patron_identifier
+        post_args['email'] = 'patron_' + patron_identifier + '@librarysimplified.org'
```
leonardr: I want to incorporate the default notification email address here.

api/oneclick.py
```diff
@@ -419,49 +417,78 @@ def internal_format(self, delivery_mechanism):
...
         resp_dict = {}
         message = None
-        try:
-            response = self.request(url=url, data=json.dumps(post_args), method="post")
-            if response.text:
-                resp_dict = response.json()
-                message = resp_dict.get('message', None)
-        except Exception, e:
-            self.log.error("Patron create failed: %r", e, exc_info=e)
-            raise RemoteInitiatedServerError(e.message, action)
```
leonardr: I'm not sure about this but I think it's more useful to raise the actual exception here rather than catching it and reraising a different one.

api/oneclick.py
```diff
@@ -740,36 +767,6 @@ def validate_item(self, licensepool):
 
         return item_oneclick_id, media
 
-
-    def validate_patron(self, patron, create=True):
```
leonardr: This whole thing was replaced by patron_remote_identifier.

tests/test_oneclick.py
```diff
@@ -216,27 +331,21 @@ def test_checkin(self):
...
-        datastr, datadict = self.api.get_data("response_patron_internal_id_found.json")
-        self.api.queue_response(status_code=200, content=datastr)
         # queue checkin success
         self.api.queue_response(status_code=200, content='{"message": "success"}')
 
         success = self.api.checkin(patron, None, pool)
         eq_(True, success)
 
-        # queue patron id
-        self.api.queue_response(status_code=200, content=datastr)
```
leonardr: These are the extra API requests that don't happen anymore.

PR #753, opened 2017-10-27
Serve a bearer token so that the client can download the book, instead of downloading the book and passing it along

leonardr:
This branch changes the way OPDS For Distributors collections are fulfilled. Instead of downloading the EPUB file ourselves and sending it to the client, we send the client the bearer token we use to make  requests to the OPDS For Distributors site. This allows the client to make requests on the library's behalf, for a short period of time.

In addition to changing the unit tests I've tested this on a Biblioboard collection and I was able to get all the way to a bearer token document with the correct URL in place.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -487,7 +487,7 @@ class CirculationManagerController(BaseCirculationManagerController):
         license_pool = get_one(self._db, LicensePool, id=license_pool_id)
         if not license_pool:
             return INVALID_INPUT.detailed(
-                _("License Pool #%d does not exist.") % license_pool_id
+                _("License Pool #%s does not exist.") % license_pool_id
             )
         return license_pool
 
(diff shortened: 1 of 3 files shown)


api/controller.py
```diff
@@ -487,7 +487,7 @@ def load_licensepool(self, license_pool_id):
         license_pool = get_one(self._db, LicensePool, id=license_pool_id)
         if not license_pool:
             return INVALID_INPUT.detailed(
-                _("License Pool #%d does not exist.") % license_pool_id
+                _("License Pool #%s does not exist.") % license_pool_id
```
leonardr: Unrelated change to stop an ISE I randomly caused while testing.

api/opds_for_distributors.py
```diff
@@ -119,12 +119,14 @@ def refresh(credential):
...
-            credential.expires = datetime.datetime.utcnow() + datetime.timedelta(seconds=expires_in)
+            # We'll avoid edge cases by assuming the token expires 75%
+            # into its useful lifetime.
+            credential.expires = datetime.datetime.utcnow() + datetime.timedelta(seconds=expires_in*0.75)
         return Credential.lookup(_db, self.data_source_name,
                                  "OPDS For Distributors Bearer Token",
                                  patron=None,
                                  refresher_method=refresh,
-                                 ).credential
+                                 )
```
leonardr: We return the Credential object instead of its string because we need all the information in the Credential object to generate the token document.

PR #754, opened 2017-10-30
OdiloAPI Integration.

ejcepas:
Odilo Connector API Integration (discovery and circulation API), first version commit.


api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+from core.util.http import HTTP
+
+from circulation_exceptions import *
+
+
+class OdiloAPI(BaseOdiloAPI, BaseCirculationAPI):
+    NAME = ExternalIntegration.ODILO
+    DESCRIPTION = _("Integrate an Odilo library collection.")
+    SETTINGS = [
+                   {"key": BaseOdiloAPI.LIBRARY_API_BASE_URL, "label": _("Library API base URL")},
```
leonardr: This is where you would set DEFAULT_API_BASE_URL as a `default`.

api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+    delivery_mechanism_to_internal_format = {
+        (pdf, adobe_drm): 'ACSM',
+        (pdf, streaming_text): 'PDF',
+        (epub, streaming_text): 'EPUB',
+        (mp3, no_drm): 'MP3',
+        (mp4, streaming_text): 'MP4',
+        (wmv, streaming_text): 'WMV',
+        (jpg, no_drm): 'JPG',
+        (scorm, no_drm): 'SCORM',
+    }
```
leonardr: You can get this by inverting `OdiloRepresentationExtractor.format_data_for_odilo_format`, and I recommend doing that rather than keep two copies of this information.

api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+        # Data just as 'x-www-form-urlencoded', no JSON
+        payload = dict(patronId=patron, format=internal_format)
+
+        response = self.patron_request(
+            patron, pin, self.CHECKOUT_ENDPOINT.replace("{recordId}", record_id),
+            extra_headers={'Content-Type': 'application/x-www-form-urlencoded'},
+            data=payload)
+
+        response_json = response.json()
+        if response.status_code == 404:
```
leonardr: Putting this next bit of code into a helper method makes it easier to test.

api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+            Identifier.ODILO_ID,
+            checkout['id'],
+            start_date,
+            end_date,
+            checkout['downloadUrl']
+        )
+
+    def checkin(self, patron, pin, licensepool):
+        checkout_id = licensepool.identifier.identifier
+        url = self.CHECKIN_ENDPOINT.replace("{checkoutId}", checkout_id) + "?patronId=" + patron
```
leonardr: I would put `?patronId={patron}` into the CHECKIN_ENDPOINT constant, since it is always part of the URL.

api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+
+        response = self.patron_request(patron, pin, url, method='POST')
+        if response.status_code == 404:
+            data = response.json()
+            if data and data['errors'] and len(response.json()['errors']) > 0:
+                error = data['errors'][0]
```
leonardr: This whole thing of checking all the different ways the errors might or might not be present can be refactored, as well as the part that turns an error message into a circulation exception.

api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+        else:
+            return response
+
+    @classmethod
+    def extract_date(cls, data, field_name):
+        if field_name not in data:
+            d = None
+        else:
+            # OdiloAPI dates are timestamps in milliseconds
+            d = datetime.datetime.fromtimestamp(float(data[field_name]) / 1000.0)
```
leonardr: fromtimestamp converts a time to local time, and we use UTC exclusively. When we need to convert a timestamp to milliseconds, we create an 'epoch' datetime for January 1, 1970, and then add `datetime.timedelta(milliseconds=timestamp)` to it.

api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+        response = self.patron_request(patron, pin, url)
+        if response.status_code == 200 and response.content:
+            data = response.json()
+            self.raise_exception_on_error(data)
+            return data
+        else:
+            return None
+
+    def get_hold(self, patron, pin, hold_id):
+        url = self.HOLD_GET.replace("{holdId}", hold_id)
```
leonardr: I'm not sure why you don't stick `patronId` on to this URL like you do with others.

api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+
+    def get_loan(self, patron, pin, checkout_id):
+        url = self.CHECKOUT_GET.replace("{checkoutId}", checkout_id) + '?patronId=' + patron
+        response = self.patron_request(patron, pin, url)
+        if response.status_code == 200 and response.content:
+            data = response.json()
+            self.raise_exception_on_error(data)
+            return data
+        else:
+            return None
```
leonardr: If you're unable to get a specific loan, I think raising an exception is appropriate.

api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+
+    def get_fulfillment_link(self, patron, pin, checkout_id, format_type):
+        """Get the link corresponding to an existing loan.
+        """
+        # Retrieve loan with its download_ulr. It is necessary to generate a download token in our API
+        loan = self.get_loan(patron, pin, checkout_id)
+        if not loan:
+            raise NoActiveLoan("Could not find active loan for %s" % checkout_id)
+
+        if format_type and loan['format'] and format_type == loan['format']:
```
leonardr: This whole thing would also benefit from being made into an independently testable method.

api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+            code = error['id']
+            description = ''
+            if 'description' in error:
+                description = error['description']
+
+            if code == 'LOAN_ALREADY_RESERVED':
+                raise AlreadyOnHold(description)
+            elif code == 'ERROR_DATA_NOT_FOUND':
+                raise NotFoundOnRemote(code + ', record: ' + record_id)
+            elif code == 'PatronExceededHoldLimit':
```
leonardr: This seems to be a leftover constant from copy-and-pasted Overdrive code.

api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+            if response_json and response_json['errors'] and len(response_json['errors']) > 0:
+                error = response_json['errors'][0]
+                if error['id'] == 'ERROR_DATA_NOT_FOUND':
+                    raise NotFoundOnRemote('Record %s' % record_id)
+
+                elif error['id'] == 'NoCopiesAvailable':
+                    # Clearly our info is out of date.
+                    # self.update_licensepool(identifier.identifier)
+                    raise NoAvailableCopies()
+                elif error['id'] == 'PatronHasExceededCheckoutLimit':
```
leonardr: These constants are leftover from copy-and-pasted Overdrive code.

api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+
+    def recently_changed_ids(self, start):
+        modification_date = None
+        if start:
+            if isinstance(start, datetime.date):
+                modification_date = start.strftime('%Y-%m-%d')  # Format YYYY-MM-DD
+            elif isinstance(start, basestring):
+                modification_date = start
+
+        self.all_ids(modification_date)
```
leonardr: You can just call `all_ids` and let it parse the modification date.

api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+
+    def all_ids(self, modication_date=None):
```
leonardr: "modification" is misspelled.

api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+
+                parsed += 1
+
+            # Persist each bunch of retrieved records
+            self._db.commit()
+
+            # Retrieve next group of records
+            url = "%s?limit=%i&offset=%i" % (self.api.ALL_PRODUCTS_ENDPOINT, limit, offset)
+            if modication_date:
+                url = "%s&modificationDate=%s" % (url, modication_date)
```
leonardr: You have this code twice and its complexity makes it a good candidate for refactoring.

api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+            status_code, headers, content = self.api.get(url)
+            content = json.loads(content)
+
+        if status_code >= 400:
+            self.log.error('ERROR: Fail while retrieving data from remote source: HTTP ' + status_code)
+            if content:
+                self.log.error('ERROR response content: ' + str(content))
+        else:
+            self.log.info('Retrieving all ids finished ok. Retrieved %i records!!' % retrieved)
+            self.log.info('New records: %i' % new)
```
leonardr: I would combine these two info statements.

api/odilo.py
```diff
@@ -0,0 +1,532 @@
...
+class RecentOdiloCollectionMonitor(OdiloCirculationMonitor):
+    """Monitor recently changed books in the Odilo collection."""
+
+    SERVICE_NAME = "Odilo Collection Recent Monitor"
+    INTERVAL_SECONDS = 60
+
+
+class MockOdiloAPI(BaseMockOdiloAPI, OdiloAPI):
+    def patron_request(self, patron, pin, *args, **kwargs):
+        pass
```
leonardr: Generally our mock APIs record the fact that a request was made and return a previously queued-up response.

tests/test_odilo.py
```diff
@@ -0,0 +1,250 @@
...
+db_url = Configuration.database_url(testing)
+# Initialize a new database session unless we were told not to
+# (e.g. because a script already initialized it).
+autoinitialize = os.environ.get('AUTOINITIALIZE') != 'False'
+if autoinitialize:
+    SessionManager.initialize(db_url)
+session_factory = SessionManager.sessionmaker(db_url)
+_db = flask_scoped_session(session_factory, app)
+if autoinitialize:
+    SessionManager.initialize_data(_db)
```
leonardr: You should not need to set up a fake Flask application or do any of this stuff. Everything you need to test should be manageable with a regular `DatabaseTest`.

tests/test_odilo.py
```diff
@@ -0,0 +1,250 @@
...
+
+
+class OdiloAPITest(DatabaseTest, unittest.TestCase):
```
leonardr: I don't think you need to subclass `unittest.TestCase`. We generally don't do this I'm not opposed to doing so but I'd want to understand the reasoning.

tests/test_odilo.py
```diff
@@ -0,0 +1,250 @@
...
+    def setUp(self):
+        self.setUp_common()
+        self._db = _db
+
+        # Load from DB or fake collection
+        self.collection = None
+        self.fake_collection()
+        self.api = OdiloAPI(self._db, self.collection)
+
+    def fake_collection(self):
```
leonardr: This seems redundant given that you create a mock_collection in the `setup` method.

tests/test_odilo.py
```diff
@@ -0,0 +1,250 @@
...
+        self.api = self.circulation.api_for_collection[self.collection.id]
+
+    def setUp(self):
+        self.setUp_common()
+        self._db = _db
+
+        # Load from DB or fake collection
+        self.collection = None
+        self.fake_collection()
+        self.api = OdiloAPI(self._db, self.collection)
```
leonardr: I believe this method is redundant with `setup`.

tests/test_odilo.py
```diff
@@ -0,0 +1,250 @@
...
+        Odilo given a certain error condition.
+        """
+        message = message or self._str
+        token = token or self._str
+        data = dict(errorCode=error_code, message=message, token=token)
+        return json.dumps(data)
+
+
+if __name__ == '__main__':
+    unittest.main()
```
leonardr: I think this answers my 'how are you running the tests' question. We run tests with `nosetests` so this code is unnecessary.

tests/test_odilo.py
```diff
@@ -0,0 +1,250 @@
...
+# licensepool.collection = 'Odilo Test Collection'
+# licensepool.collection.id = 1
+licensepool.data_source.name = DataSource.ODILO
+licensepool.identifier.type = Identifier.ODILO_ID,
+licensepool.identifier.identifier = record_id  # Record_id
+
+
+# licensepool.identifier.identifier = '00010953'  # Record_id 2
+# licensepool.identifier.identifier = '00011052'  # Record_id 2
+# licensepool.identifier.identifier = '0001098345342'  # fake Record_id
```
leonardr: I'm not sure what this is for.

api/odilo.py
ejcepas: yes, I realized

api/odilo.py
leonardr: Now that I know that a format like "EPUB" has two different delivery mechanisms, it might not work to simply invert the dictionary, but something similar should still work.

ejcepas: I have also joined together both 'EPUB' and 'PDF' format constants just in one 'EBOOK_STREAMING' as they behave the same, and are basically the same thing for the final users.

api/odilo.py
ejcepas: ok

api/odilo.py
ejcepas: ok, changed!

api/odilo.py
ejcepas: ok, I refactorized where I considered it useful.

api/odilo.py
ejcepas: ok, changed. What about using ```datetime.datetime.utcfromtimestamp``` ??

leonardr: Good catch. I think `utcfromtimestamp` will work fine.

api/odilo.py
ejcepas: I agree, I've moved the exception I already raised where the method was used, within that method.

api/odilo.py
ejcepas: Ok, I've made what you say first.
Related to the format of the modification date, it is enough for us to use just the date and not the time.

api/odilo.py
ejcepas: thank you, I didn't realize!

api/odilo.py
leonardr: Does the API show the modification date but not the time?

I'm concerned that every time this runs over the course of a day, it's going to pick up more and more books, until the stroke of midnight, at which point it will pick up zero books and the process will repeat. This shouldn't hurt the integrity of the data--like I said, I'm just concerned about performance.

api/odilo.py
ejcepas: Ok, I refactorized it.

api/odilo.py
ejcepas: Ok, changed.

api/odilo.py
ejcepas: It was a mistake that has now been corrected, thank you !!

api/odilo.py
ejcepas: We use just the modification date and not the time due to internal restrictions of the DB and the app.

api/odilo.py
```diff
@@ -0,0 +1,495 @@
...
+    def get_patron_access_token(self, credential, patron, pin):
+        """Request an OAuth bearer token that allows us to act on
+        behalf of a specific patron.
+        """
+
+        self.client_key = patron
+        self.client_secret = pin
+        message = self.refresh_creds(credential)
+
+        if 'OK' == message:
```
leonardr: See my core comment https://github.com/NYPL-Simplified/server_core/pull/695/files#r150567737

api/odilo.py
```diff
@@ -0,0 +1,495 @@
...
+
+        :return: a LoanInfo object.
+        """
+        record_id = licensepool.identifier.identifier
+
+        # We have 2 formats, 'EBOOK_STREAMING' could be in our API ('PDF', 'EPUB'):
+        ebook_streaming = False
+        if internal_format == 'EBOOK_STREAMING':
+            ebook_streaming = True
+            internal_format = 'EPUB'  # Try first ebook_streaming as EPUB
```
leonardr: I'm not happy about this but I don't see a good solution to it except for storing custom media types in the delivery mechanisms. 

Can you change the Odilo API to understand 'EBOOK_STREAMING'? If it really doesn't matter to the client whether the underlying format is PDF or EPUB, there's no reason the client should have to keep track of the distinction and repeat it back to the server.

api/odilo.py
```diff
@@ -0,0 +1,495 @@
...
+
+        # Data just as 'x-www-form-urlencoded', no JSON
+        payload = dict(patronId=patron, format=internal_format)
+
+        response = self.patron_request(
+            patron, pin, self.CHECKOUT_ENDPOINT.replace("{recordId}", record_id),
```
leonardr: Another place to use `format` instead of `replace`.

api/odilo.py
```diff
@@ -0,0 +1,495 @@
...
+        if response.content:
+            response_json = response.json()
+            if response.status_code == 404:
+                error_id, error_description = self.raise_exception_on_error(response_json)
+                raise CannotLoan('Error id: %s, description: %s' % (error_id, error_description))
+            else:
+                return self.loan_info_from_odilo_checkout(licensepool.collection, response_json)
+
+        elif response.status_code == 400 and ebook_streaming:
+            internal_format = 'PDF'  # Retry now with ebook_streaming as PDF
```
leonardr: If you decide to keep going down this path, I think you should check the error message to make sure switching to PDF is really the solution. But maybe this is covered in your TODO on line 161.

api/odilo.py
```diff
@@ -0,0 +1,495 @@
...
+        end_date = self.extract_date(checkout, 'endTime')
+
+        return LoanInfo(
+            collection,
+            DataSource.ODILO,
+            Identifier.ODILO_ID,
+            checkout['id'],
+            start_date,
+            end_date,
+            checkout['downloadUrl']
```
leonardr: I just want to confirm that this downloadUrl is the place you go to get (e.g.) the ACSM file. Is this field always present (but possibly null), even in a streaming context?

api/odilo.py
```diff
@@ -0,0 +1,495 @@
...
+    def extract_date(cls, data, field_name):
+        if field_name not in data or not data[field_name]:
+            d = None
+        else:
+            # OdiloAPI dates are timestamps in milliseconds
+            d = datetime.datetime.utcfromtimestamp(float(data[field_name]) / 1000.0)
+        return d
+
+    @classmethod
+    def raise_exception_on_error(cls, data, custom_error_to_exception={}):
```
leonardr: A method called raise_exception_on_error should raise an exception if there's an error, and its return value should be irrelevant. You have code that calls this method and then raises an exception based on the return value. Either rename the method or (my preference) have it raise the exceptions itself.

api/odilo.py
```diff
@@ -0,0 +1,495 @@
...
+
+        return error_code, message
+
+    def get_checkout(self, patron, pin, record_id):
+        patron_checkouts = self.get_patron_checkouts(patron, pin)
+        for checkout in patron_checkouts:
+            if checkout['recordId'] == record_id:
+                return checkout
+
+        raise NotFoundOnRemote("Could not find active for patron %s, record %s" % (patron, record_id))
```
leonardr: Saying 'active loan' will make it clear what couldn't be found.

api/odilo.py
```diff
@@ -0,0 +1,495 @@
...
+                        raise CannotFulfill(error_id + ': ' + error_description)
+
+                return content_link, content, content_type
+
+        raise CannotFulfill("Cannot obtain a download link for patron[%r], record_id[%s], format_type[%s].", patron,
+                            record_id, format_type)
+
+    def get_patron_checkouts(self, patron, pin):
+        data = self.patron_request(patron, pin, self.PATRON_CHECKOUTS_ENDPOINT.format(patronId=patron)).json()
+        self.raise_exception_on_error(data)
```
leonardr: This assumes raise_exception_on_error will raise an exception, which isn't the case for your implementation. Same comment for line 286.

api/odilo.py
```diff
@@ -0,0 +1,495 @@
...
+            patron, pin, self.PLACE_HOLD_ENDPOINT.format(recordId=record_id),
+            extra_headers={'Content-Type': 'application/x-www-form-urlencoded'},
+            data=payload)
+
+        data = response.json()
+        if response.status_code == 200:
+            return self.hold_from_odilo_hold(licensepool.collection, data)
+
+        error_code, error_description = self.raise_exception_on_error(data)
+        raise CannotHold(error_code + ': ' + error_description)
```
leonardr: Your use of this pattern makes me think it might be useful to pass a default_exception_class into raise_exception_on_error, and change that method to raise an instance of that class if the error message doesn't result in a more specific exception.

tests/test_odilo.py
```diff
@@ -0,0 +1,342 @@
...
+        return json.dumps(data)
+
+
+class TestOdiloCirculationAPI(OdiloAPITest):
+    #################
+    # General tests
+    #################
+
+    # Test 404 Not Found --> patron not found --> 'patronNotFound'
+    def test_01_patron_not_found(self):
```
leonardr: It shouldn't be necessary to number the test cases.

tests/test_odilo.py
```diff
@@ -0,0 +1,342 @@
...
+        self.api.queue_response(200, content=checkout_json)
+
+        self.checkout('EBOOK_STREAMING')
+
+    def checkout(self, internal_format):
+        print('Checkout test ' + internal_format)
+
+        loan_info = self.api.checkout(self.PATRON, self.PIN, self.licensepool, internal_format)
+        ok_(loan_info, msg="LoanInfo null --> checkout failed!")
+        print 'Loan ok: %s' % loan_info.identifier
```
leonardr: All the print statements should be removed or converted to log statements, but this one in particular seems misleading.

tests/test_odilo.py
```diff
@@ -0,0 +1,342 @@
...
+
+    def test_12_checkout_acsm(self):
+        checkout_data, checkout_json = self.sample_json("checkout_acsm_ok.json")
+        self.api.queue_response(200, content=checkout_json)
+        self.checkout('ACSM')
+
+    def test_13_checkout_ebook_streaming_epub(self):
+        checkout_data, checkout_json = self.sample_json("checkout_ebook_streaming_ok_epub.json")
+        self.api.queue_response(200, content=checkout_json)
+        self.checkout('EBOOK_STREAMING')
```
leonardr: I would call this helper method something like `validate_checkout` so that it's clear that it contains assertions.

tests/test_odilo.py
```diff
@@ -0,0 +1,342 @@
...
+        patron_checkouts_data, patron_checkouts_json = self.sample_json("patron_checkouts.json")
+        patron_holds_data, patron_holds_json = self.sample_json("patron_holds.json")
+        self.api.queue_response(200, content=patron_checkouts_json)
+        self.api.queue_response(200, content=patron_holds_json)
+
+        loans_and_holds = self.api.patron_activity(self.PATRON, self.PIN)
+        if loans_and_holds:
+            print 'Found: %i loans and holds' % len(loans_and_holds)
+        else:
+            print 'No loans or holds found'
```
leonardr: You know the outcome here -- either there are loans and holds or there aren't. The case you're expecting should be asserted.

tests/test_odilo.py
```diff
@@ -0,0 +1,342 @@
...
+        checkin_data, checkin_json = self.sample_json("checkin_ok.json")
+        self.api.queue_response(200, content=checkin_json)
+
+        response = self.api.checkin(self.PATRON, self.PIN, self.licensepool)
+        eq_(response.status_code, 200,
+            msg="Response code != 200, cannot perform checkin for record: " + self.licensepool.identifier.identifier
+                + " patron: " + self.PATRON)
+
+        checkout_returned = response.json()
+        print 'Checkout returned: %s' % checkout_returned['id']
```
leonardr: If this is important enough to print, its (known) value should be asserted.

tests/test_odilo.py
```diff
@@ -0,0 +1,342 @@
...
+
+        monitor = RecentOdiloCollectionMonitor(self._db, self.collection, api_class=MockOdiloAPI)
+        ok_(monitor, 'Monitor null !!')
+        eq_(ExternalIntegration.ODILO, monitor.protocol, 'Wat??')
+
+        records_metadata_data = self.sample_data("records_metadata.json")
+        monitor.api.queue_response(200, content=records_metadata_data)
+
+        availability_data = self.sample_data("record_availability.json")
+        monitor.api.queue_response(200, content=availability_data)
```
leonardr: This would be clearer and take less code if you called this once for every item in `records_metadata_data`.

leonardr:
This is really good, the test coverage is good, my concerns are mainly related to our coding practices.

api/odilo.py
ejcepas: Yes, I've already changed it.

api/odilo.py
ejcepas: Ok, changed !

api/odilo.py
ejcepas: Yes, you are right. This URL is the place where you go to get an ACSM file, the streaming contents, etc...
This field is always present, and it shouldn't be null if all it's ok. It is specially necessary in a streaming context, if not, you couldn't get the actual streaming resource.

I mean, if you call that URL for a checkout with format:
'ACSM' --> you get the .acsm file to be downloaded.
'EBOOK_STREAMING'  --> you get redirected to the online viewer and get the streaming contents themselves of the resource.
Same for formats video, audio, etc...

api/odilo.py
ejcepas: Yes, it's a write error, thank you!!

api/odilo.py
ejcepas: Good idea, I've added that 'default_exception_class', thank you.

tests/test_odilo.py
ejcepas: Yes, I know, and it isn't "necessary", it's just "useful" for me to have them numbered, in order to ensure myself that I have tested all things.
The tests are independent among them and could be run in any order.

tests/test_odilo.py
ejcepas: It works because it isn't a "test" method, is an auxiliary method called by the 3 tests above, which queue up the corresponding checkout response each.

tests/test_odilo.py
ejcepas: Ok, I've renamed it to 'perform_and_validate_checkout'.

tests/test_odilo.py
ejcepas: Right, I've already changed it.

tests/test_odilo.py
ejcepas: ok, asserted!

tests/test_odilo.py
ejcepas: I couldn't agree more, what would I be thinking...?, changed !

api/odilo.py
```diff
@@ -200,26 +185,27 @@ def extract_date(cls, data, field_name):
         return d
 
     @classmethod
-    def raise_exception_on_error(cls, data, custom_error_to_exception={}):
+    def raise_exception_on_error(cls, data, default_exception_class=None, ignore_exception_code=None):
```
leonardr: Since your'e changing the API, ignore_exception_code should be a list of exceptions to ignore, not a single item.

ignore_exception_code is unusual enough that I would value a description of it in the docstring. Something like "These error codes do not represent exceptional conditions and will be ignored rather than transformed into exceptions."

api/odilo.py
```diff
@@ -255,21 +241,20 @@ def get_fulfillment_link(self, patron, pin, record_id, format_type):
         loan_format = checkout['format']
         if format_type and loan_format and (
                         format_type == loan_format or
-                        (format_type == 'EBOOK_STREAMING' and loan_format in ('EPUB', 'PDF'))
+                        (loan_format == 'ACSM' and format_type in ('ACSM_EPUB', 'ACSM_PDF'))
```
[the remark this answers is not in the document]
ejcepas: Ok, changed !!

api/odilo.py
ejcepas: ok, changed!

PR #755, opened 2017-10-30
Handle an error in the final POST to metadata wrangler initialization

leonardr:
This branch fixes the handling of the case where all the GET requests to the metadata wrangler succeed, but the final POST request to actually create the integration returns a ProblemDetail. Previously we tried to call `.json()` on the ProblemDetail, causing a crash.


leonardr:
Yes, but only because of the core bug I identified earlier. I figured I'd put this in since it's obviously the right way to handle a situation that shouldn't happen normally.

leonardr:
Thinking about this overnight, I think this could also happen if the POST request times out.

PR #757, opened 2017-11-02
Port Bibliotheca dump scripts

leonardr:
This branch ports the informational scripts that dump bibliographic and circulation data for Bibliotheca so that they work in the new collection-based system. I had to use these scripts and figured I'd port them while I was at it. One of these scripts I never actually put under version control, so this is the first time you're seeing it.


aslagle:
You have a submodule problem here.

aslagle:
Except for that, looks good.

PR #758, opened 2017-11-02
Database lanes

aslagle:
This branch moves the lane configuration into the database, using the `database-lanes` branch in core.

It changes routes to use a single lane identifier that includes language rather than languages and lane name.

There is a script in `bin/repair/reset_lanes` that deletes existing lanes and regenerates them based on library language settings or estimates of the collection size.

This doesn't handle NYPL's current Staff Picks lanes, but since that's only for NYPL we could create them in the database ourselves.


aslagle:
We'll need more preparation before merging this, but I'd like to start the review now.

api/admin/controller.py
```diff
@@ -1452,6 +1452,7 @@ def collections(self):
             else:
                 return NO_PROTOCOL_FOR_NEW_SERVICE
 
+        collection.name = name
```
aslagle: This is an unrelated bug I found - if you changed a Collection name in the admin interface it didn't actually change.

api/lanes.py
```diff
@@ -1,91 +1,140 @@
...
+    if not large and not small and not tiny:
+        estimates = library.estimated_holdings_by_language()
+        [(ignore, largest)] = estimates.most_common(1)
+        for language, count in estimates.most_common():
+            if count > largest * 0.1:
+                large.append(language)
+            elif count > largest * 0.01:
+                small.append(language)
+            else:
+                tiny.append(language)
```
aslagle: These cutoffs are pretty arbitrary.

api/lanes.py
```diff
@@ -94,308 +143,623 @@ def lanes_from_genres(_db, library, genres, **extra_args):
         "Periodicals" : dict(invisible=True)
```
aslagle: Should probably take this out - invisible is no longer a thing.

api/controller.py
```diff
@@ -416,36 +399,21 @@ def circulation(self):
...
-            name = name.replace("__", "/")
-
-        lanes = lanelist.by_languages[language_key]
-
-        if not name:
-            defaults = [x for x in lanes.values() if x.default_for_language]
-            if len(defaults) == 1:
-                # This language has one, and only one, default lane.
-                return defaults[0]
+        lane = get_one(self._db, Lane, identifier=lane_identifier, library_id=library_id)
```
leonardr: This can wait for a future branch, but you have a great opportunity to turn Lane into a HasFullTableCache, with (lane_identifier, library_id) as the cache key.

api/lanes.py
leonardr: I think I would prefer cutoffs based on absolute numbers but I also like always treating the largest collection as a 'large' and once you do that you can't do the cutoffs with absolute numbers. So this is good for now.

api/lanes.py
```diff
@@ -1,401 +1,769 @@
...
+                         identifier=ya_fiction_identifier + " " + genres.Science_Fiction.name,
+                         priority=ya_fiction_priority, **ya_common_args))
+    ya_fiction_priority += 1
+    ya_fiction.sublanes.append(
+        lane_from_genres(_db, library, [genres.Steampunk],
+                         identifier=ya_fiction_identifier + " " + genres.Steampunk.name,
+                         priority=ya_fiction_priority, **ya_common_args))
+    ya_fiction_priority += 1
+    # TODO:
+    # Paranormal -- what is it exactly?
```
leonardr: I think we can get rid of this TODO. We have not had a groundswell of complaints that there's no separate "Paranormal" lane.

api/controller.py
```diff
@@ -552,12 +520,11 @@ def authenticated_patron_root_lane(self):
             return patron
 
         policy = Configuration.root_lane_policy()
-        lane_info = policy.get(patron.external_type)
-        if lane_info is None:
+        lane_identifier = policy.get(patron.external_type)
```
aslagle: This seems redundant given Lane.root_for_patron_type. Should this look for a Lane that's the root for the patron's type so we don't need the policy?

leonardr: Yes, exactly. This is the payoff for adding root_for_patron_type to the database -- we don't need to put the policy into JSON anymore.

api/opds.py
```diff
@@ -241,17 +241,20 @@ def lane_url(self, lane):
...
 
-        if not lane.sublanes and not lane.parent:
+        if lane and isinstance(lane, Lane) and lane.sublanes:
+            url = self.groups_url(lane)
+        elif lane and (
+            isinstance(lane, Lane)
+            or isinstance(lane, WorkBasedLane)
+            or isinstance(lane, ContributorLane)
+            or isinstance(lane, SeriesLane)
+            ):
```
leonardr: This is a big list that we might forget to add to in the future. Can you create a superclass of all these custom worklists, or invert the logic and create a special class for the top-level WorkList?

scripts.py
```diff
@@ -1018,3 +1017,11 @@ class OPDSForDistributorsReaperScript(OPDSImportScript):
...
     MONITOR_CLASS = OPDSForDistributorsReaperMonitor
     PROTOCOL = OPDSForDistributorsImporter.NAME
+
+class LaneResetScript(LibraryInputScript):
+    """Reset a library's lanes based on language configuration or estimates
+    of the library's current collection."""
+
+    def process_library(self, library):
+        create_default_lanes(self._db, library)
+        self._db.commit()
```
leonardr: This is dangerous enough that it should take a special argument and act on a 5-second delay, like `adobe_account_id_reset`.

aslagle: And maybe also have a dry run option so you can see what the new lanes will be.

api/opds.py
aslagle: Any suggestions on a name for the superclass?

leonardr: Maybe DynamicLane? They form OPDS lanes, but they're not ORM objects like Lane.

leonardr:
This looks good to me!

PR #759, opened 2017-11-02
Docker submodule update.

courte:
This submodule update incorporates the changes to scripts between circulation collections and the metadata wrangler.


leonardr:
I believe I merged this update as part of a branch that updated core as well, but there might be another docker commit I'm missing.

courte:
You did!

PR #763, opened 2017-11-06
Ask RBdigital for loans of specific lengths, depending on media type, using the 'days' parameter.

leonardr:
This branch takes advantage of an undocumented parameter in the RBdigital API that lets the caller specify the length of a loan in days.

Two new per-library configuration options are created that allow an administrator to specify the length of all ebook and audiobook loans initiated by the circulation manager.

The 'default loan period' parameter is suppressed for RBdigital collections, since we don't need to guess at the loan period -- we know it precisely because we specified it.

### DIFF
--- a/api/oneclick.py
+++ b/api/oneclick.py
@@ -59,12 +59,36 @@ from core.util.http import (
 class OneClickAPI(BaseOneClickAPI, BaseCirculationAPI):
 
     NAME = ExternalIntegration.RB_DIGITAL
+
+    # With this API we don't need to guess the default loan period -- we
+    # know which loan period we will ask for in which situations.
+    BASE_SETTINGS = [x for x in BaseCirculationAPI.SETTINGS
+                     if x['key'] != BaseCirculationAPI.DEFAULT_LOAN_PERIOD]
+
     SETTINGS = [
         { "key": ExternalIntegration.PASSWORD, "label": _("Basic Token") },
         { "key": Collection.EXTERNAL_ACCOUNT_ID_KEY, "label": _("Library ID") },
         { "key": ExternalIntegration.URL, "label": _("URL"), "default": BaseOneClickAPI.PRODUCTION_BASE_URL },
-    ] + BaseCirculationAPI.SETTINGS
+    ] + BASE_SETTINGS
     
+    AUDIOBOOK_LOAN_DURATION = 'audio_loan_duration'
+    EBOOK_LOAN_DURATION = 'ebook_loan_duration'
+
+    LIBRARY_SETTINGS = [
+        { "key" : AUDIOBOOK_LOAN_DURATION, 
+          "label": _("Audiobook Loan Duration (in Days)"),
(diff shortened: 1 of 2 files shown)


aslagle:
For ODL, I used the 'default loan period' parameter for the loan period we specify to the distributor. I think we should make that consistent.

leonardr:
The status quo is certainly confusing. I'm not sure if there's a way to make this consistent though. Let's talk it out and see.

All loans against a Bibliotheca or Axis 360 collection have the same duration. This duration is negotiated between the library and the distributor. The circulation manager has no access to this information. But when we create a LoanInfo object for a newly created loan, we need to make a _guess_ as to when the loan is _likely_ to expire. This is what we use the 'default loan duration' setting for. Once the patron hits their /loans feed this number isn't relevant because we can find out when the loan _actually_ will expire. But on the _initial_ return we use the default loan duration setting to say when the expiration date _probably_ is.

A patron who borrows an Overdrive book can choose their own default loan duration by logging in to the Overdrive website and changing their settings. For print books, the loan duration might be 7, 14, or 21 days. For audiobooks it might be 3, 5, or 7 days (or something -- the exact numbers don't matter). In this case the default loan duration says when the loan expiration time _probably_ is, _for this patron_, _assuming it's an ebook and not an audiobook_. Although most patrons don't change their default loan duration, IMO this makes the idea of a 'default loan duration' conceptually useless for Overdrive collections.

For ODL and RBdigital collections, the circulation manager itself can control the duration of the loan. In this case there is no need to have a 'default' duration because we never have to guess. Instead we need a setting that controls what the duration should _be_.

After talking with Datalogics as they set up an RBdigital library I think this setting should be set on a per-library per-collection basis, rather than purely on a per-collection basis. (But per-collection would also work, I don't think most people would mind.)

So I think we have two different settings here, which have different (but confusingly similar) names and purposes. I would like to change the ODL circulation API to look more like this branch. Let me know whether you agree with this.

Ideally I would like to get rid of DEFAULT_LOAN_PERIOD and DEFAULT_RESERVATION_PERIOD altogether, but I don't know what the effect would be on the mobile client if we stopped providing an expiration date in the OPDS entry we serve immediately after checkout.

I haven't investigated DEFAULT_RESERVATION_PERIOD yet but I think it might be possible to do away with that one. I'll take another look later.

aslagle:
I agree about changing the ODL API to match this branch.

I think DEFAULT_LOAN_PERIOD and DEFAULT_RESERVATION_PERIOD could be useful in the future if we ever want to attempt to improve our hold time guesses.

leonardr:
It might be smarter to empirically determine the average loan duration for a given medium + distributor by running some code on checkin/loan reap.

I don't know if we can empirically determine average reservation time (except for Bibliotheca) since most distributors don't treat it as a separate state, but it might be less confusing overall to just assume it's always 3 days. As far as I know, it's not a time period that's subject to negotiation, and it's hard to even explain that state to people.

PR #764, opened 2017-11-06
Make loan duration a per-library per-collection setting and make sure it's always corectly described.

leonardr:
This branch makes sure that all of our CollectionAPI subclasses (except the ones that offer unlimited-length loans) have one (ebooks) or two (ebooks+audiobooks) settings relating to loan durations.

For Overdrive, Axis, and Bibliotheca, these settings are informative only. They reflect the _default_ or _assumed average_ lengths of loans but the circulation manager has no power over the lengths of loans created through these services.

For ODL and RBdigital, these settings are active. When a loan is created, the duration of the loan is taken from this setting.

All this code existed before, but ODL and RBdigital were handled inconsistently, the settings were per-collection instead of per-library per-collection, and the descriptions of the settings didn't make it clear when these settings were informative and when they were active.

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -872,7 +872,7 @@ class BaseCirculationAPI(object):
         "key" : Collection.EBOOK_LOAN_DURATION_KEY, 
         "label": _("Ebook Loan Duration (in Days)"),
         "default": Collection.STANDARD_DEFAULT_LOAN_PERIOD,
-        "description": _("When a patron uses SimplyE to borrow an audio book from this collection, SimplyE will ask for a loan that lasts this number of days. This must be equal to or less than the maximum loan duration negotiated with the distributor.")
+        "description": _("When a patron uses SimplyE to borrow an ebook from this collection, SimplyE will ask for a loan that lasts this number of days. This must be equal to or less than the maximum loan duration negotiated with the distributor.")
     }
 
     # Add to LIBRARY_SETTINGS if your circulation API is for a
@@ -882,7 +882,7 @@ class BaseCirculationAPI(object):
         "key" : Collection.AUDIOBOOK_LOAN_DURATION_KEY,
         "label": _("Audiobook Loan Duration (in Days)"),
         "default": Collection.STANDARD_DEFAULT_LOAN_PERIOD,
-        "description": _("When a patron uses SimplyE to borrow an audio book from this collection, SimplyE will ask for a loan that lasts this number of days. This must be equal to or less than the maximum loan duration negotiated with the distributor.")
+        "description": _("When a patron uses SimplyE to borrow an audiobook from this collection, SimplyE will ask for a loan that lasts this number of days. This must be equal to or less than the maximum loan duration negotiated with the distributor.")
     }
 
     # Add to LIBRARY_SETTINGS if your circulation API is for a
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -865,13 +865,46 @@ class CirculationAPI(object):
 class BaseCirculationAPI(object):
     """Encapsulates logic common to all circulation APIs."""
 
-    DEFAULT_LOAN_PERIOD = "default_loan_period"
-    DEFAULT_RESERVATION_PERIOD = "default_reservation_period"
+    # Add to LIBRARY_SETTINGS if your circulation API is for a
+    # distributor which includes ebooks and allows clients to specify
+    # their own loan lengths.
+    EBOOK_LOAN_DURATION_SETTING = {
+        "key" : Collection.EBOOK_LOAN_DURATION_KEY, 
+        "label": _("Ebook Loan Duration (in Days)"),
+        "default": Collection.STANDARD_DEFAULT_LOAN_PERIOD,
+        "description": _("When a patron uses SimplyE to borrow an audio book from this collection, SimplyE will ask for a loan that lasts this number of days. This must be equal to or less than the maximum loan duration negotiated with the distributor.")
+    }
 
-    SETTINGS = [
-        { "key": DEFAULT_LOAN_PERIOD, "label": _("Default Loan Period (in Days)"), "optional": True, "type": "number" },
-        { "key": DEFAULT_RESERVATION_PERIOD, "label": _("Default Reservation Period (in Days)"), "optional": True, "type": "number" },
-    ]
+    # Add to LIBRARY_SETTINGS if your circulation API is for a
+    # distributor which includes audiobooks and allows clients to
+    # specify their own loan lengths.
+    AUDIOBOOK_LOAN_DURATION_SETTING = { 
+        "key" : Collection.AUDIOBOOK_LOAN_DURATION_KEY,
+        "label": _("Audiobook Loan Duration (in Days)"),
+        "default": Collection.STANDARD_DEFAULT_LOAN_PERIOD,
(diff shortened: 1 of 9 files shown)


api/circulation.py
```diff
@@ -865,13 +865,46 @@ def sync_bookshelf(self, patron, pin):
...
-        { "key": DEFAULT_RESERVATION_PERIOD, "label": _("Default Reservation Period (in Days)"), "optional": True, "type": "number" },
-    ]
+    # Add to LIBRARY_SETTINGS if your circulation API is for a
+    # distributor which includes ebooks and allows clients to specify
+    # their own loan lengths.
+    EBOOK_LOAN_DURATION_SETTING = {
+        "key" : Collection.EBOOK_LOAN_DURATION_KEY, 
+        "label": _("Ebook Loan Duration (in Days)"),
+        "default": Collection.STANDARD_DEFAULT_LOAN_PERIOD,
+        "description": _("When a patron uses SimplyE to borrow an audio book from this collection, SimplyE will ask for a loan that lasts this number of days. This must be equal to or less than the maximum loan duration negotiated with the distributor.")
```
aslagle: This description should say "ebook" not "audio"

api/odl.py
```diff
@@ -100,17 +100,22 @@ class ODLWithConsolidatedCopiesAPI(BaseCirculationAPI):
...
             "key": Collection.DATA_SOURCE_NAME_SETTING,
             "label": _("Data source name"),
         }
     ]
 
+    my_ebook_loan_duration_setting = dict(
+        BaseCirculationAPI.EBOOK_LOAN_DURATION_SETTING
+    )
+    my_ebook_loan_duration_setting.update(
+        description = _("When a patron uses SimplyE to borrow an ebook from this ODL server, SimplyE will ask for a loan that lasts this number of days. This must be equal to or less than the maximum loan duration negotiated with the distributor.")
```
aslagle: I don't think it's necessary to replace this description.

leonardr:
Assuming the tests pass I'm happy with these changes. I changed core quite a bit and I removed assertions from the controller tests to reflect the fact that these configuration settings are no longer directly associated with the collection.

api/odl.py
```diff
@@ -158,7 +156,7 @@ def __init__(self, _db, collection):
         self.username = collection.external_integration.username
         self.password = collection.external_integration.password
         self.consolidated_loan_url = collection.external_integration.setting(self.CONSOLIDATED_LOAN_URL_KEY).value
-        self.default_loan_period = collection.external_integration.setting(BaseCirculationAPI.DEFAULT_LOAN_PERIOD).int_value or 21
+        self.default_loan_period = collection.external_integration.setting(Collection.EBOOK_LOAN_DURATION_KEY).int_value or Collection.STANDARD_DEFAULT_LOAN_PERIOD
```
aslagle: doesn't this need to change to be per-library?

api/oneclick.py
```diff
@@ -71,24 +71,29 @@ class OneClickAPI(BaseOneClickAPI, BaseCirculationAPI):
...
+    )
+    my_audiobook_setting.update(
+        description = _("When a patron uses SimplyE to borrow an RBdigital audio book, SimplyE will ask for a loan that lasts this number of days. This must be equal to or less than the maximum loan duration negotiated with RBdigital."),
+        default=DEFAULT_LOAN_DURATION,
+    )
+    my_ebook_setting = dict(
+        BaseCirculationAPI.EBOOK_LOAN_DURATION_SETTING
+    )
+    my_ebook_setting.update(
+        description = _("When a patron uses SimplyE to borrow an RBdigital ebook, SimplyE will ask for a loan that lasts this number of days. This must be equal to or less than the maximum loan duration negotiated with RBdigital."),
```
aslagle: I'm not sure these descriptions need to be replaced either - they're the same as the default except with 'RBDigital' in them.

api/oneclick.py
```diff
@@ -103,11 +108,11 @@ def __init__(self, *args, **kwargs):
 
         integration = self.collection.external_integration
         self.audiobook_duration_days = (
-            integration.setting(self.AUDIOBOOK_LOAN_DURATION).int_value 
+            integration.setting(Collection.AUDIOBOOK_LOAN_DURATION_KEY).int_value 
             or self.DEFAULT_LOAN_DURATION
         )
         self.ebook_duration_days = (
-            integration.setting(self.EBOOK_LOAN_DURATION).int_value 
+            integration.setting(Collection.EBOOK_LOAN_DURATION_KEY).int_value
```
aslagle: these also don't use libraries.

tests/admin/test_controller.py
```diff
@@ -1969,14 +1957,6 @@ def test_collections_post_edit(self):
         eq_("website_id", setting.key)
         eq_("1234", setting.value)
 
-        setting = collection.external_integration.setting("default_loan_period")
-        eq_("default_loan_period", setting.key)
-        eq_("14", setting.value)
```
aslagle: This could be changed to check that the library + integration setting was set.

leonardr: Can you tell me what the setup code would look like? I tried passing in a value with the library short name and nothing happened (which might either indicate a bug or that I did it wrong).

aslagle: I think you have it right, but this is testing with an Overdrive collection and Overdrive doesn't have that setting.

aslagle: Actually, I looked and the admin interface code for collections doesn't handle library settings yet - this is the first time we've had them for collections.

aslagle: The JS part should be fine as is, but SettingsController.collections will need to be updated.

leonardr:
We need to add some code to the admin interface to handle setting library+collection-specific configuration information items. My brain can't handle this but Amy has agreed to tackle it tomorrow.

Here's the test I was working on when I discovered that the existing controller wasn't setting the configuration settings:

```
    def test_collections_post_edit_library_specific_configuration(self):
        # The collection exists.
        collection = self._collection(
            name="Collection 1",
            protocol=ExternalIntegration.RB_DIGITAL
        )

        l1, ignore = create(
            self._db, Library, name="Library 1", short_name="L1",
        )

        with self.app.test_request_context("/", method="POST"):
            flask.request.form = MultiDict([
                ("id", collection.id),
                ("name", "Collection 1"),
                ("protocol", ExternalIntegration.RB_DIGITAL),
                ("external_account_id", "1234"),
                ("username", "user2"),
                ("password", "password"),
                ("url", "http://rb/"),
                ("libraries", json.dumps([
                    {
                        "short_name": "L1", 
                        "ebook_loan_duration": "14",
                        "audio_loan_duration": "12"
                    }
                ])
                ),
            ])
            response = self.manager.admin_settings_controller.collections()
            eq_(response.status_code, 200)

        # Additional settings were set on the collection+library.
        set_trace()

        # Remove the connection between collection and library.
        with self.app.test_request_context("/", method="POST"):
            flask.request.form = MultiDict([
                ("id", collection.id),
                ("name", "Collection 1"),
                ("protocol", ExternalIntegration.OVERDRIVE),
                ("external_account_id", "1234"),
                ("username", "user2"),
                ("password", "password"),
                ("website_id", "1234"),
                ("ils_name", "the_ils"),
                ("libraries", json.dumps([])),
            ])
            response = self.manager.admin_settings_controller.collections()
            eq_(response.status_code, 200)

        eq_(collection.id, int(response.response[0]))

        # The settings associated with the collection+library were removed
        # when the connection between collection and library was deleted.

```

PR #771, opened 2017-11-09
Bibliotheca audiobooks have distinct delivery mechanism

leonardr:
This branch implements fulfillment for Bibliotheca audiobooks. Audiobooks are fulfilled through a distinct endpoint. The equivalent of the ACSM file is a Findaway JSON document which is served as `application/json`; I've created a more specific media type to use when serving this document to the client.

I've added two sample files (the Bibliotheca bibliographic metadata for an audiobook and a sample JSON document) for reference. They're not used in the tests.


api/bibliotheca.py
```diff
@@ -203,6 +214,12 @@ def get_fulfillment_file(self, patron_id, bibliotheca_id):
                    item_id=bibliotheca_id, patron_id=patron_id)
         body = self.TEMPLATE % args 
         return self.request('GetItemACSM', body, method="PUT")
+
+    def get_audio_fulfillment_file(self, patron_id, bibliotheca_id):
+        args = dict(request_type='AudioFulfillmentRequest', 
+                    item_id=bibliotheca_id, patron_id=patron_id)
+        body = self.TEMPLATE % args 
+        return self.request('GetItemAudioFulfillment', body, method="POST")
```
leonardr: This method takes POST even though GetItemACSM takes PUT.

PR #774, opened 2017-11-14
Fix metadata updates monitor

leonardr:
This branch changes the `MetadataWranglerCollectionUpdateMonitor` to reflect how the metadata wrangler's update feed actually works and to improve its robustness.

The monitor follows 'next' links to get more than one page of results. (Previously it was only processing one page.) Every time it processes a page of results, it forcibly updates the `Timestamp` associated with the monitor, setting it to the earliest timestamp encountered on that page. There shouldn't be any problems with 'next' loops, but if there are, the monitor will detect them and stop running. The monitor also stops running if it finds a page with a 'next' link but no entries -- this will let us remove a fairly expensive calculation on the server side if we want to.

We use the earliest timestamp rather than the latest because the metadata wrangler sends feeds composed of items from two different queries, and it's possible that books from one query might dramatically outnumber books from the other query. If this happens, using the latest timestamp found in a feed might jump over all the books from the big query.

### DIFF
--- a/api/coverage.py
+++ b/api/coverage.py
@@ -240,15 +240,21 @@ class MetadataWranglerCollectionManager(MetadataWranglerCoverageProvider):
         for message in self.process_feed_response(response, id_mapping):
             try:
                 identifier, _new = Identifier.parse_urn(self._db, message.urn)
-                mapped_batch.remove(identifier)
+                if identifier in mapped_batch:
+                    mapped_batch.remove(identifier)
             except ValueError as e:
                 # For some reason this URN can't be parsed. This
                 # shouldn't happen.
                 continue
 
             if message.status_code in success_codes:
-                result = id_mapping[identifier]
-                results.append(result)
+                if identifier in id_mapping:
+                    result = id_mapping[identifier]
+                    results.append(result)
+                else:
+                    # The server sent information about an identifier
+                    # we didn't ask for. Do nothing.
+                    pass
             elif message.status_code == 400:
                 # The URN couldn't be recognized. (This shouldn't happen,
                 # since if we can parse it here, we can parse it on MW, too.)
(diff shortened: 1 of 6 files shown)


api/coverage.py
```diff
@@ -240,15 +240,21 @@ def _process_batch(self, client_method, success_codes, batch):
...
             if message.status_code in success_codes:
-                result = id_mapping[identifier]
-                results.append(result)
+                if identifier in id_mapping:
+                    result = id_mapping[identifier]
+                    results.append(result)
+                else:
+                    # The server sent information about an identifier
+                    # we didn't ask for. Do nothing.
+                    pass
```
leonardr: I know why these changes are necessary but not why they only recently caused tests to fail -- I don't think any of the changes made in core recently would have triggered these edge cases.

tests/test_coverage.py
```diff
@@ -376,13 +376,14 @@ def opds_feed_identifiers(self):
...
-        # Mapped identifier.
+        # An identifier mapped to an identifier represented in the OPDS
+        # response.
         source = DataSource.lookup(self._db, DataSource.AXIS_360)
         mapped_id = self._identifier(
             identifier_type=Identifier.AXIS_360_ID, foreign_id=u'0015187876'
         )
         equivalent_id = self._identifier(
-            identifier_type=Identifier.ISBN, foreign_id=self._isbn
+            identifier_type=Identifier.ISBN, foreign_id='9781936460236'
```
leonardr: Again, not sure why this suddenly became a problem.

api/coverage.py
courte: This is so strange!

leonardr:
The two-queries thing is annoying, but I'm okay with waiting to figure out a solution, so long as the update script makes some amount of progress every time it runs.

PR #775, opened 2017-11-15
Immediately remove the 'reap' coverage records for relicensed works, …

leonardr:
…instead of adding all the relicensed works in with a union() call.

This branch removes the union() call for finding works that were reaped from a collection, but then relicensed. Instead, the 'reap' CoverageRecords are immediately removed and the database session committed. This was going to happen soon enough anyway, and doing it immediately lets us avoid treating relicensed works specially.

The use of union() was making `items_that_need_coverage` treat _all_ relicensed works as needing coverage, even ones that already had coverage records. These works were showing up over and over again, and being processed over and over again, blocking other titles from being published.

### DIFF
--- a/tests/test_coverage.py
+++ b/tests/test_coverage.py
@@ -268,7 +268,7 @@ class TestMetadataWranglerCoverageProvider(DatabaseTest):
         )
         cr = self._coverage_record(
             pool.identifier, self.provider.data_source,
-            operation=self.provider.OPERATION
+            operation=self.provider.OPERATION, collection=self.collection
         )
 
         # We have a coverage record already, so this book doesn't show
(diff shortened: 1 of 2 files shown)


tests/test_coverage.py
```diff
@@ -268,7 +268,7 @@ def test_items_that_need_coverage_respects_cutoff(self):
         )
         cr = self._coverage_record(
             pool.identifier, self.provider.data_source,
-            operation=self.provider.OPERATION
+            operation=self.provider.OPERATION, collection=self.collection
```
leonardr: Here I'm bringing the circulation tests up to the standards expected by the new revision of core by adding collections to our sample CoverageRecords.

PR #777, opened 2017-11-16
Real short client token configuration script

leonardr:
We have a script called `short_client_token_library_configuration` but it's intended for use by NYPL for registering the short client token configurations for _other_ libraries. This branch renames that script and provides a new `short_client_token_library_configuration` script that's intended for use by a library that's just getting things set up. NYPL runs the old script and the other library runs the new script with similar values, and we're all set up. This will keep us going until SimplyE starts using the library registry.

### DIFF
--- a/api/adobe_vendor_id.py
+++ b/api/adobe_vendor_id.py
@@ -723,6 +723,7 @@ class AuthdataUtility(object):
         secret = ConfigurationSetting.for_library_and_externalintegration(
             _db, ExternalIntegration.PASSWORD, library, integration
         ).value
+        set_trace()
 
         other_libraries = None
         adobe_integration = ExternalIntegration.lookup(
@@ -1009,7 +1010,7 @@ class AuthdataUtility(object):
         return patron_identifier_credential, delegated_identifier
 
 
-class ShortClientTokenLibraryConfigurationScript(Script):
+class VendorIDLibraryConfigurationScript(Script):
 
     @classmethod
     def arg_parser(cls):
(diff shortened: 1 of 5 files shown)


api/adobe_vendor_id.py
```diff
@@ -1040,7 +1041,7 @@ def do_run(self, _db=None, cmd_args=None, output=sys.stdout):
         if not adobe_integration:
             output.write(
                 "Could not find an Adobe Vendor ID integration for default library %s.\n" %
-                library.short_name
+                default_library.short_name
```
leonardr: Unrelated problem I found while testing.

api/adobe_vendor_id.py
```diff
@@ -1009,7 +1010,7 @@ def create_function():
         return patron_identifier_credential, delegated_identifier
 
 
-class ShortClientTokenLibraryConfigurationScript(Script):
+class VendorIDLibraryConfigurationScript(Script):
```
leonardr: I renamed this because "Vendor ID" is the term we use for NYPL's library configuration (which includes Short Client Token secrets for many other libraries) and "Short Client Token" is the term we use when helping other libraries get their configuration set up.

api/adobe_vendor_id.py
```diff
@@ -1092,3 +1093,93 @@ def explain(self, output, libraries, website):
...
+class ShortClientTokenLibraryConfigurationScript(Script):
+
+    @classmethod
+    def arg_parser(cls):
+        parser = argparse.ArgumentParser()
+        parser.add_argument(
+            '--website-url', 
+            help="The URL to this library's patron-facing website (not their circulation manager), e.g. \"https://nypl.org/\". This is used to uniquely identify a library.",
+            required=True,
+        )
```
leonardr: I use the patron-facing URL, rather than some other method for identifying which library you're editing, because patron-facing URL is what I ask for when setting up a SCT integration for a library.

PR #779, opened 2017-11-20
Replace import / sync / reap metadata wrangler coverage cycle with import /reap cycl

leonardr:
This branch changes the way the circulation manager communicates with the metadata wrangler. The motivation is the case where a book is new to your collection, but the book's metadata is old. Without this branch, you'll never get metadata about that book unless you look up that specific identifier. That will never happen because the lookup controller has been disabled for performance reasons.

We don't want servers asking about every book in the catalog over and over again. But we do need them to ask about every book in the catalog _once_.

The `MetadataWranglerCollectionRegistrar` uses the re-enabled `lookup` endpoint to register a collection's items with the metadata wrangler. If the MW happens to have bibliographic information about these items, it's provided immediately -- otherwise it will be provided later, once it becomes available, through the `MetadataWranglerCollectionUpdateMonitor`.

This used to be handled by two different CoverageProviders: `MetadataWranglerCollectionSync` made sure your collection was mirrored on the metadata wrangler, and `MetadataWranglerCoverageProvider` got the metadata. Now `MetadataWranglerCoverageProvider` is purely a superclass and `MetadataWranglerCollectionSync` is gone.

The `MetadataWranglerCollectionReaper` makes sure titles that drop out of a collection are _removed_ from the corresponding collection on the metadata wrangler. This hasn't changed much.


api/coverage.py
```diff
@@ -55,25 +101,10 @@ def __init__(self, collection, lookup_client, **kwargs):
         super(OPDSImportCoverageProvider, self).__init__(collection, **kwargs)
         self.lookup_client = lookup_client
 
-    @classmethod
```
leonardr: Maybe this needs to be there and we don't have a test for it.

api/coverage.py
```diff
@@ -95,9 +126,15 @@ def process_batch(self, batch):
...
-        for failure in error_messages_by_id.values():
-            results.append(failure)
+        # Anything left over is either a CoverageFailure, or an
+        # Identifier that used to be a CoverageFailure, indicating
+        # that a simplified:message that a normal OPDSImporter would
+        # consider a 'failure' should actually be considered a
+        # success.
+        for failure_or_identifier in sorted(error_messages_by_id.values()):
+            if isinstance(failure_or_identifier, CoverageFailure):
+                failure_or_identifier.collection = self.collection
```
leonardr: This should be collection_or_not

api/coverage.py
```diff
@@ -143,46 +186,53 @@ def create_identifier_mapping(self, batch):
...
-    OPERATION = CoverageRecord.IMPORT_OPERATION
     DATA_SOURCE_NAME = DataSource.METADATA_WRANGLER
-    INPUT_IDENTIFIER_TYPES = [
-        Identifier.OVERDRIVE_ID,
-        Identifier.BIBLIOTHECA_ID,
-        Identifier.AXIS_360_ID,
-        Identifier.ONECLICK_ID,
-        Identifier.URI,
-    ]
-
```
leonardr: This needs to come back.

api/enki.py
```diff
@@ -461,10 +461,6 @@ def _simple_http_get(self, url, headers, *args, **kwargs):
 
 class EnkiBibliographicCoverageProvider(BibliographicCoverageProvider):
     """Fill in bibliographic metadata for Enki records.
-
-    Currently this is only used by BibliographicRefreshScript. It's
-    not normally necessary because the Enki API combines
-    bibliographic and availability data.
```
leonardr: Bring this back

tests/test_coverage.py
```diff
@@ -257,6 +400,203 @@ def test_create_identifier_mapping(self):
...
+        )
+        [result] = self.provider.process_batch([valid_id])
+        eq_(True, isinstance(result, CoverageFailure))
+        eq_(valid_id, result.obj)
+        eq_('418: Mad Hatter', result.exception)
+
+        # The OPDS importer didn't know which Collection to associate
+        # with this CoverageFailure, but the CoverageProvider does,
+        # and it set .collection appropriately.
+        eq_(self.provider.collection, result.collection)
```
leonardr: This might not be true anymore -- core might take care of it.

tests/test_coverage.py
```diff
@@ -257,6 +400,203 @@ def test_create_identifier_mapping(self):
...
+        cr = self._coverage_record(
+            pool.identifier, self.provider.data_source,
+            operation=self.provider.OPERATION, 
+            collection=self.provider.collection
+        )
+        eq_([], qu.all())
+
+    def test_identifier_reaped_from_one_collection_covered_in_another(self):
+        """A LicensePool that's not actually available doesn't need coverage.
+        """
```
leonardr: Docstring is wrong

tests/test_coverage.py
leonardr: Turns out this is still true.

api/coverage.py
```diff
@@ -55,25 +71,10 @@ def __init__(self, collection, lookup_client, **kwargs):
         super(OPDSImportCoverageProvider, self).__init__(collection, **kwargs)
         self.lookup_client = lookup_client
 
-    @classmethod
-    def all(cls, _db, **kwargs):
```
aslagle: Why did you remove this?

leonardr: That was only used by `ContentServerBibliographicCoverageProvider`, which is also removed in this branch. The idea was to make sure that we only used the `lookup` protocol with the content server for collections that ultimately derived from the content server. But we weren't using that -- we get our canonical metadata direct from the metadata wrangler.

PR #787, opened 2017-11-24
Fulfill an RBdigital book by converting its proprietary manifest format to a Readium Web Publication.

leonardr:
The generated RWP should contain all the information necessary to download chapters of an audiobook. When the links expire you can get a new RWP by fulfilling the book again.

### DIFF
--- a/api/oneclick.py
+++ b/api/oneclick.py
@@ -41,6 +41,7 @@ from core.model import (
     DeliveryMechanism,
     Edition,
     ExternalIntegration,
+    Hyperlink,
     Identifier, 
     Library,
     LicensePool,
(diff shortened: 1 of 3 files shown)


api/oneclick.py
```diff
@@ -1032,7 +1035,113 @@ def run_once(self, start, cutoff):
...
+        """Import an RBdigital spine item as a Web Publication Manifest
+        spine item.
+        """
+        href = file_data.get('downloadUrl')
+        duration = file_data.get('minutes') * 60
+        title = file_data.get('display')
+
+        id = file_data.get('id')
+        size = file_data.get('size')
+        filename = file_data.get('filename')
```
aslagle: Some of these aren't used and others are extracted again below.

PR #801, opened 2017-12-06
Use precalculated lane size in admin interface instead of running the count every time

leonardr:
This should dramatically improve the user-visible performance of the lane editor.

This branch also removes some script code that was moved into core by https://github.com/NYPL-Simplified/server_core/pull/744.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -917,8 +917,7 @@ class LanesController(CirculationManagerController):
                 return [{ "id": lane.id,
                           "display_name": lane.display_name,
                           "visible": lane.visible,
-                          # TODO: getting the counts is very slow.
-                          "count": fast_query_count(lane.works(self._db).limit(None)),
+                          "count": lane.size,
                           "sublanes": lanes_for_parent(lane),
                           "custom_list_ids": [list.id for list in lane.customlists],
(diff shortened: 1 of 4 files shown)


api/admin/controller.py
```diff
@@ -983,6 +982,7 @@ def lanes_for_parent(parent):
             for list in lane.customlists:
                 if list.id not in custom_list_ids:
                     lane.customlists.remove(list)
+            lane.update_size(self._db)
```
leonardr: This ensures that the count is updated whenever the lane is edited, without having to wait for the materialized view script to run again.

scripts.py
```diff
@@ -296,44 +297,6 @@ def open(self):
         return StringIO(representation.content)
 
 
-class LaneSweeperScript(LibraryInputScript):
```
leonardr: The part of this code that really needs to be in circulation (because it deals with setting up a fake app server) has been moved to `CacheRepresentationPerLane`.

api/admin/controller.py
aslagle: One other place you could do it is when a custom list is edited, if the list is associated with any lanes.

leonardr: Good idea. This is ready for another look.

PR #802, opened 2017-12-11
Added a changelog that covers work done since 2.1.0.

leonardr:
Now that we have an audience we should start keeping a changelog. I'll make sure this is up to date before doing a release but it shouldn't be hard to remember to keep it updated as we make changes to the circulation branch (even if the branch is just bringing in a change made in core).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,30 @@
+# Unreleased (2.1.1)
+
+## Circulation changes
+
+* The simple authentication provider can authenticate multiple test
+  identifiers, so long as all of them have the same password.
+
+## Core changes
+
+* Performance improvement: Lane size is calculated ahead of time and
+  stored in the database.
+
+* Connector to the Odilo catalog/circulation API.
+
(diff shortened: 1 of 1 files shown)


leonardr:
Should they go there _instead_ of here?

courte:
I think so. That's an easier place to edit, since it's not tied with version control, and you can point someone at the URL to see the diff. I'm thinking about possible drawbacks, but I don't think it's important that the release notes are version controlled.

The only issue with putting it in releases, is that you can only edit / update them when you create the tag. So it all happens at the end, and the notes are written by whoever is in charge of releasing--as opposed to the developer who made the change. In a file like this, devs can input changes as they go, pre-release. We can keep track of what all is going into 2.1.1 as we go and make clearer decisions to cut a release.

Could we keep notes for the next prerelease in CHANGELOG, and transfer the latest release notes to releases once they're tagged and out? That gives us the best of both worlds.

leonardr:
That sounds good. I've made a 2.1.0 release and removed those change records from CHANGELOG.md.

PR #803, opened 2017-12-14
Convert Findaway license documents into semistandard Web Publication Manifests

leonardr:
This branch relieves clients of the need to understand everything about Findaway license documents. We still serve documents with a custom media type, but the documents are basically Web Publication Manifest documents with some extensions and some missing fields. Bibliographic information (not present in the Findaway documents) is filled in from our own data.

A client can parse this document with its Web Publication Manifest parser and take note of a few extension fields, rather than needing to parse a slightly different document and then fill in missing data from its local copy of the book's bibliographic information.

### DIFF
--- a/api/bibliotheca.py
+++ b/api/bibliotheca.py
@@ -347,8 +347,7 @@ class BibliothecaAPI(BaseBibliothecaAPI, BaseCirculationAPI):
         # of the LicensePool and its Work.
         manifest.update_bibliographic_metadata(license_pool)
 
-        # Add Findaway-specific information as extra metadata. TODO:
-        # Make this more linked-data-y.
+        # Add Findaway-specific information as extra metadata.
         for findaway_extension in [
                 'accountId', 'checkoutId', 'fulfillmentId', 'licenseId',
                 'sessionKey'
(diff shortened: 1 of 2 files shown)


api/bibliotheca.py
```diff
@@ -195,23 +197,32 @@ def fulfill(self, patron, password, pool, internal_delivery):
         )
         if drm_scheme == DeliveryMechanism.FINDAWAY_DRM:
             fulfill_method = self.get_audio_fulfillment_file
-            # The provided media type is application/json, which
-            # is too vague for us.
-            force_content_type = DeliveryMechanism.FINDAWAY_DRM
+            content_transformation = self.findaway_license_to_webpub_manifest
```
daryanypl: I see, so this is where you start the webpub manifest off with the information from the opds feed?

api/bibliotheca.py
```diff
@@ -303,6 +314,94 @@ def apply_circulation_information_to_licensepool(self, circ, pool, analytics=Non
...
+            # milliseconds. This assumption makes our example
+            # audiobook take about 7.9 hours, and no other reasonable
+            # assumption is in the right order of magnitude. But this
+            # needs to be explicitly verified.
+            duration = part.get('duration', 0) / 1000.0
+
+            kwargs = {}
+
+            sequence = part.get('sequence')
+            if sequence is not None:
```
courte: Why are we skipping over missing keys here (sequence, part) vs the Findaway-specific metadata, where missing keys are set to None? I'm not familiar with what a part vs a sequence is, but could this cause problems with parsing in the mobile client?

api/bibliotheca.py
leonardr: Not the OPDS feed per se -- the same underlying information from the database that is also used to populate the OPDS feed.

api/bibliotheca.py
leonardr: I also don't know what these items mean, but it's always good to avoid a situation where the parser expects a key to be there and it's not there.

courte: 👍

PR #808, opened 2018-01-03
Modified the circulation branch to be compatible with the single-query lane feed API

leonardr:
This branch bring the circulation branch up to date with the slightly different API used in my working code to generate lane feeds.

* There is no more passing around a work_model -- all lane queries go against the new materialized view.
* Lane.inherit_parent_restrictions is True by default.
* I created a custom Facets subclass used to generate the featured subset of a SeriesLane.
* SeriesLane no longer does an aliased join against Edition. We're always using the new materialized view, which contains the information we need to filter and sort.


api/lanes.py
```diff
@@ -859,20 +859,49 @@ def fetch_recommendations(self, _db):
...
             _db, self.recommendations, base_query=qu
         )
         return super(RecommendationLane, self).apply_filters(
-            _db, qu, work_model, facets, pagination, featured=featured)
+            _db, qu, facets, pagination, featured=featured
+        )
+
+
+class SeriesFacets(Facets):
+    """A custom Facets object for ordering a lane based on series."""
```
leonardr: `Facets.apply` has logic for setting DISTINCT and ORDER BY on a query, logic that was duplicated in SeriesLane and that broke when the core logic around ORDER BY changed. Making a custom `Facets` subclass lets us create custom joins on the query that are used only when we're applying the ORDER BY.

api/lanes.py
```diff
@@ -906,26 +930,34 @@ def url_arguments(self):
...
 
     def featured_works(self, _db):
-        qu = self.works(_db)
-
         # Aliasing Edition here allows this query to function
         # regardless of work_model and existing joins.
-        work_edition = aliased(Edition)
-        qu = qu.join(work_edition).order_by(work_edition.series_position, work_edition.title)
-        target_size = self.get_library(_db).featured_lane_size
-        qu = qu.limit(target_size)
```
leonardr: The motivation behind `FeaturedQueryFacets` is getting rid of this code, which is already present in `Facets`.

tests/test_lanes.py
```diff
@@ -528,6 +531,7 @@ def test_works_query(self):
         # Works in the series are returned as expected.
         w1 = self._work(with_license_pool=True)
         w1.presentation_edition.series = series_name
+        self._db.commit()
```
leonardr: We need to commit to make sure the series name shows up in the materialized view. Previously this wasn't an issue because we didn't use the materialized view to filter the series name; we did an aliased join against Edition and filtered that.

api/lanes.py
```diff
@@ -906,26 +917,26 @@ def url_arguments(self):
...
 
     def featured_works(self, _db):
-        qu = self.works(_db)
-
-        # Aliasing Edition here allows this query to function
-        # regardless of work_model and existing joins.
-        work_edition = aliased(Edition)
-        qu = qu.join(work_edition).order_by(work_edition.series_position, work_edition.title)
-        target_size = self.get_library(_db).featured_lane_size
-        qu = qu.limit(target_size)
```
leonardr: The motivation behind FeaturedQueryFacets is getting rid of this code, which is already present, more or less, in Facets. This code broke when the rules about DISTINCT changed.

PR #809, opened 2018-01-05
Added basic support for holds in an ODL collection.

aslagle:
With ODL, a distributor doesn't manage the holds queue and the circulation manager has full control. This branch adds support for managing holds queues within a single circ manager.

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -343,69 +343,70 @@ class CirculationAPI(object):
 
         new_loan = False
 
-        # TODO: If the book isn't available and the patron is putting it on hold,
-        # we shouldn't check the loan limit.
         loan_limit = patron.library.setting(Configuration.LOAN_LIMIT).int_value
         non_open_access_loans_with_end_date = [loan for loan in patron.loans if loan.license_pool.open_access == False and loan.end]
-        if loan_limit and len(non_open_access_loans_with_end_date) >= loan_limit:
-            raise PatronLoanLimitReached()
+        at_loan_limit = (loan_limit and len(non_open_access_loans_with_end_date) >= loan_limit)
 
-        try:
-            loan_info = api.checkout(
-                patron, pin, licensepool, internal_format
-            )
-
-            # We asked the API to create a loan and it gave us a
-            # LoanInfo object, rather than raising an exception like
-            # AlreadyCheckedOut.
-            #
-            # For record-keeping purposes we're going to treat this as
-            # a newly transacted loan, although it's possible that the
-            # API does something unusual like return LoanInfo instead
-            # of raising AlreadyCheckedOut.
-            new_loan = True
-        except AlreadyCheckedOut:
-            # This is good, but we didn't get the real loan info.
-            # Just fake it.
-            identifier = licensepool.identifier            
-            loan_info = LoanInfo(
-                licensepool.collection,
-                licensepool.data_source,
-                identifier.type,
-                identifier.identifier,
-                start_date=None,
-                end_date=now + datetime.timedelta(hours=1)
-            )
-            if existing_loan:
-                loan_info.external_identifier=existing_loan.external_identifier
-        except AlreadyOnHold:
-            # We're trying to check out a book that we already have on hold.
-            hold_info = HoldInfo(
-                licensepool.collection, licensepool.data_source,
-                licensepool.identifier.type, licensepool.identifier.identifier,
-                None, None, None
-            )
-        except NoAvailableCopies:
-            if existing_loan:
-                # The patron tried to renew a loan but there are
-                # people waiting in line for them to return the book,
-                # so renewals are not allowed.
-                raise CannotRenew(
-                    _("You cannot renew a loan if other patrons have the work on hold.")
+        # If we're at the loan limit, skip trying to check out the book and just try
+        # to place a hold. Otherwise, try to check out the book even if we think it's
+        # not available.
+        if not at_loan_limit:
+            try:
+                loan_info = api.checkout(
+                    patron, pin, licensepool, internal_format
                 )
-            else:
-                # That's fine, we'll just (try to) place a hold.
+
+                # We asked the API to create a loan and it gave us a
+                # LoanInfo object, rather than raising an exception like
+                # AlreadyCheckedOut.
                 #
+                # For record-keeping purposes we're going to treat this as
+                # a newly transacted loan, although it's possible that the
+                # API does something unusual like return LoanInfo instead
+                # of raising AlreadyCheckedOut.
+                new_loan = True
+            except AlreadyCheckedOut:
+                # This is good, but we didn't get the real loan info.
+                # Just fake it.
+                identifier = licensepool.identifier            
+                loan_info = LoanInfo(
+                    licensepool.collection,
+                    licensepool.data_source,
+                    identifier.type,
+                    identifier.identifier,
+                    start_date=None,
+                    end_date=now + datetime.timedelta(hours=1)
+                )
+                if existing_loan:
+                    loan_info.external_identifier=existing_loan.external_identifier
+            except AlreadyOnHold:
+                # We're trying to check out a book that we already have on hold.
+                hold_info = HoldInfo(
+                    licensepool.collection, licensepool.data_source,
+                    licensepool.identifier.type, licensepool.identifier.identifier,
+                    None, None, None
+                )
+            except NoAvailableCopies:
+                if existing_loan:
+                    # The patron tried to renew a loan but there are
+                    # people waiting in line for them to return the book,
+                    # so renewals are not allowed.
+                    raise CannotRenew(
+                        _("You cannot renew a loan if other patrons have the work on hold.")
+                    )
+                else:
+                    # That's fine, we'll just (try to) place a hold.
+                    #
+                    # Since the patron incorrectly believed there were
+                    # copies available, update availability information
+                    # immediately.
+                    api.update_availability(licensepool)
+            except NoLicenses, e:
                 # Since the patron incorrectly believed there were
-                # copies available, update availability information
+                # licenses available, update availability information
                 # immediately.
                 api.update_availability(licensepool)
-        except NoLicenses, e:
-            # Since the patron incorrectly believed there were
-            # licenses available, update availability information
-            # immediately.
-            api.update_availability(licensepool)
-            raise e
+                raise e
 
         if loan_info:
             # We successfuly secured a loan.  Now create it in our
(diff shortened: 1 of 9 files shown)


api/odl.py
```diff
@@ -304,21 +344,33 @@ def checkout(self, patron, pin, licensepool, internal_format):
...
             raise CannotLoan()
+        start = datetime.datetime.utcnow()
         expires = doc.get("potential_rights", {}).get("end")
         if expires:
             expires = datetime.datetime.strptime(expires, self.TIME_FORMAT)
+
+        # We need to set the start and end dates on our local loan since
+        # the code that calls this only sets them when a new loan is created.
+        loan.start = start
+        loan.end = expires
```
aslagle: This was a bug in my earlier code.

aslagle:
This approach duplicates some information and I'm concerned that there could be bugs if parts of it get out of sync. It seems like it would be easy to miss updating something.

api/odl.py
```diff
@@ -289,6 +324,11 @@ def checkout(self, patron, pin, licensepool, internal_format):
         if loan.count() > 0:
             raise AlreadyCheckedOut()
 
+        # TODO: Should this look at all loans instead of only loans from the ODL distributor?
```
leonardr: Yes, the discussion in https://github.com/NYPL-Simplified/Simplified/issues/67 makes me think this should be a per-library setting.

api/odl.py
```diff
@@ -359,17 +411,183 @@ def fulfill(self, patron, pin, licensepool, internal_format):
...
+            if len(current_loans) > copy_index:
+                next_cycle_start = current_loans[copy_index].end
+            else:
+                reservation = current_reservations[copy_index - len(current_loans)]
+                if reservation.end:
+                    next_cycle_start = reservation.end + datetime.timedelta(days=default_loan_period)
+                else:
+                    # TODO: If there's no reservation period and someone has a reserved hold, 
+                    # the maximum wait time is infinite.
+                    # Maybe reservation period should be required so this won't happen?
```
leonardr: I think it's fine to make reservation period required and set it to 3 days by default.

api/odl.py
```diff
@@ -359,17 +411,183 @@ def fulfill(self, patron, pin, licensepool, internal_format):
...
+        # Update the next hold in the queue when a license is reserved.
+        _db = Session.object_session(licensepool)
+
+        next_hold = _db.query(Hold).filter(
+            Hold.license_pool_id==licensepool.id
+        ).filter(
+            Hold.position > 0
+        ).order_by(
+            Hold.start
+        ).limit(1).all()
```
leonardr: If you use `one` here you won't have to treat `next_hold` as a list.

api/odl.py
```diff
@@ -359,17 +411,183 @@ def fulfill(self, patron, pin, licensepool, internal_format):
...
+            licensepool.licenses_available += 1
+
     def place_hold(self, patron, pin, licensepool, notification_email_address):
         """Holds aren't supported yet, so attempting to place one will
         raise an exception.
         """
-        raise CannotHold()
+        _db = Session.object_session(patron)
+
+        # TODO: Should this look at all holds instead of only holds from the ODL distributor?
```
leonardr: I think it should, and that the code should be in `CirculationAPI.borrow`.

api/odl.py
```diff
@@ -359,17 +411,183 @@ def fulfill(self, patron, pin, licensepool, internal_format):
...
+            licensepool.identifier.type,
+            licensepool.identifier.identifier,
+            start_date=hold.start,
+            end_date=hold.end,
+            hold_position=hold.position,
+        )
 
     def release_hold(self, patron, pin, licensepool):
         """Holds aren't supported yet, so attempting to release one will
         raise an exception.
```
leonardr: This docstring can be changed.

api/odl.py
```diff
@@ -570,6 +831,32 @@ def process_one_page(self, response):
...
+        ).filter(
+            LicensePool.collection_id==self.api.collection_id
+        ).filter(
+            Hold.end<datetime.datetime.utcnow()
+        )
+
+        for hold in expired_holds:
+            pool = hold.license_pool
+            self._db.delete(hold)
+            self.api.update_hold_queue(pool)
```
leonardr: You can make this more efficient by keeping track of the LicensePools that need to be updated, and then calling `update_hold_queue` on each pool in a separate loop. Since holds are concentrated in a few very popular books, it's pretty likely that the same LicensePool will show up more than once here.

tests/test_odl.py
```diff
@@ -134,6 +136,44 @@ def test_checkin_success(self):
         eq_(7, self.pool.licenses_available)
         eq_(0, self._db.query(Loan).count())
 
+    def test_checkin_success_with_holds_queue(self):
```
leonardr: I found this test and the next one difficult to get into; some narrative comments would help.

leonardr:
I share your concern about things getting out of sync. Some ideas:

* You may not need to keep LicensePool.licenses_reserved or patrons_in_hold_queue updated. We use licenses_available and licenses_owned in various queries, so they need to be semi-accurate, but I think patrons_in_hold_queue is only used informationally, and I don't know if licenses_reserved is used at all. You could decide that you don't keep those fields uploaded, the way that the RBdigital integration has decided that licenses_available is either a 1 or a 0.

* It should be fairly easy to write a multithreaded script that simulates a large number of lifecycle operations on a LicensePool and then compares the actual queues and `LicensePool` fields to expected values. Also a script that does a lot of weird stuff and makes sure that you never get an unexpected scenario like -1 patrons in the hold queue, without trying to keep track of exactly what the outcome 'should' be. These wouldn't work as unit tests, but they would help you shake out any race conditions and edge cases.

api/odl.py
```diff
@@ -304,21 +344,33 @@ def checkout(self, patron, pin, licensepool, internal_format):
             raise CannotLoan()
 
         # We have successfully borrowed this book.
-        licensepool.licenses_available -= 1
+        if not hold:
+            licensepool.licenses_available -= 1
+        else:
+            licensepool.licenses_reserved -= 1
+            licensepool.patrons_in_hold_queue -= 1
+            _db.delete(hold)
```
leonardr: This would be a natural place to put a simple check that prevents these numbers from going below zero.

leonardr:
My main concern with merging this into master is the limit on loans and holds, which I would rather have as a per-library setting. If that's the general consensus, it seems like that can go into a separate branch.

Apart from that, I'm okay whether this branch becomes a feature branch (if you want to smoke out edge cases) or gets merged into master.

aslagle:
Thanks @leonardr, I addressed a lot of your comments. I moved the loan and hold limits to a per-library setting on this branch, but there's a separate commit (929f641) if we want to revert or cherry-pick that part to another branch.

I'm using LicensePool.licenses_reserved internally because it's easier to keep that updated than Hold.position. The way I have it now, Hold.position gets out of date and it's fixed using LicensePool.licenses_reserved whenever it's needed. If I stopped using that I'd have to update every Hold on the pool whenever something changed. I did add a bunch of checks to prevent negative numbers and I like your idea about the multithreaded script. I'll try that and then if everything seems okay I'd like to merge.

aslagle:
I changed this so the derivative info gets recalculated and replaced rather than incremented/decremented, so if it was somehow wrong it would be fixed.

aslagle:
One side effect of this is that you can no longer use the same ODL collection on multiple circ managers that don't allow holds, since it requires the circ manager to know about all the loans. I think that's okay for now, I can revisit it later.

Also, I noticed the circulation API is deleting my end date estimates and replacing them with None.

api/circulation.py
```diff
@@ -342,6 +342,15 @@ def borrow(self, patron, pin, licensepool, delivery_mechanism,
             )
 
         new_loan = False
+
+        # TODO: If the book isn't available and the patron is putting it on hold,
+        # we shouldn't check the loan limit. Or should there be one limit for loans
+        # and holds?
```
leonardr: Separate loan and hold limits is the way to go, I think. I agree this is a TODO but it's kind of a marginal case where you're at your loan limit but not at your hold limit, and you try to borrow a book but end up putting it on hold.

aslagle: I agree. But this code also runs when you know a book's not available and try to put it on hold - we try a loan first even if we expect it won't work.

api/circulation.py
```diff
@@ -342,6 +342,15 @@ def borrow(self, patron, pin, licensepool, delivery_mechanism,
...
+        loan_limit = patron.library.setting(Configuration.LOAN_LIMIT).int_value
+        non_open_access_loans = [loan for loan in patron.loans if loan.license_pool.open_access == False]
+        if loan_limit and len(non_open_access_loans) >= loan_limit:
+            raise PatronLoanLimitReached()
```
leonardr: We want to count loans towards the limit only if your loan is interfering with someone else's ability to read the book. So open access books shouldn't count, but neither should books obtained through OPDS For Distributors. I don't have a good idea for determining whether a book's loan should 'count' apart from checking how the loan was set up.

api/config.py
```diff
@@ -37,6 +37,15 @@ class Configuration(CoreConfiguration):
     # of fines a patron can have before losing lending privileges.
     MAX_OUTSTANDING_FINES = u"max_outstanding_fines"
 
+    # The name of the per-library settings that set the maximum amounts
+    # of books a patron can have on loan or on hold at once.
+    # (Note: depending on distributor settings, a patron may be able
+    # to exceed the limits by checking out books directly from a distributor's
+    # app. They may also get a limit exceeded error before they reach these
+    # limits if a distributor has a smaller limit.)
```
leonardr: I think this explanatory text should go into the `description`s of these fields.

api/circulation.py
aslagle: That's a good point. How about checking if the loan has an end date?

leonardr: Yeah, that seems pretty reliable.

api/odl.py
```diff
@@ -102,7 +105,13 @@ class ODLWithConsolidatedCopiesAPI(BaseCirculationAPI):
...
             "key": Collection.DATA_SOURCE_NAME_SETTING,
             "label": _("Data source name"),
-        }
+        },
+        {
+            "key": Collection.DEFAULT_RESERVATION_PERIOD_KEY,
+            "label": _("Default Reservation Period (in Days)"),
+            "type": "number",
+            "default": Collection.STANDARD_DEFAULT_RESERVATION_PERIOD,
+        },
```
leonardr: This needs a `description` to help admins distinguish between a 'reservation period' and the period where you have something on hold and you're waiting to get to the front of the queue.

api/circulation.py
leonardr: You're right, that's a different case. If you try to put a book on hold and you're over your loan limit, it should try to put the book on hold, without trying to borrow it first. If the hold fails because the book is actually available, that's the time to raise PatronLoanLimitReached.

leonardr:
Apart from the issue with https://github.com/NYPL-Simplified/circulation/pull/809#discussion_r160208125I think this is ready to go.

PR #810, opened 2018-01-05
Improve feed generation performance

leonardr:
In addition to an important submodule update, this branch finds alternate ways to determine certain pieces of information that don't require additional database access.

### DIFF
--- a/api/opds.py
+++ b/api/opds.py
@@ -191,7 +191,6 @@ class CirculationManagerAnnotator(Annotator):
             # the loan/hold.
             return loan.license_pool
         else:
-
             # There is no active loan. Use the default logic for
             # determining the active license pool.
             return super(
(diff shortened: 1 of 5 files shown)


api/opds.py
```diff
@@ -350,16 +349,11 @@ def related_books_available(cls, work, library):
         """:return: bool asserting whether related books might exist for
         a particular Work
         """
-        if isinstance(work, Work):
-            edition = work.presentation_edition
-        else:
-            # This is a MaterializedWork*, so we're gonna grab the
-            # edition where we can.
-            edition = work.license_pool.presentation_edition
+        contributions = work.sort_author and work.sort_author != Edition.UNKNOWN_AUTHOR
```
leonardr: We don't need to look at the contributions (which would cause a database hit) here, we just need to see whether, when we calculated the materialized work's sort_author, we ended up with a non-default value.

scripts.py
```diff
@@ -474,16 +474,16 @@ def parse_args(self, cmd_args=None):
...
-            languages = None
-            lane_name = None
+            # Presumably this is the top-level WorkList.
+            lane_id = None
 
         library = lane.get_library(self._db)
         url = self.app.manager.cdn_url_for(
-            "feed", languages=languages, lane_name=lane_name, library_short_name=library.short_name
+            "feed", lane_identifier=lane_id,
+            library_short_name=library.short_name
```
leonardr: The signature for the `feed` controller method changed recently.

PR #812, opened 2018-01-09
Lanes only contain books

leonardr:
This branch changes the circulation manager to use the standard way of creating a top-level WorkList for a library (defined in https://github.com/NYPL-Simplified/server_core/pull/765).

It also changes the lanes created by default so that they only contain ebooks. (The core branch contains a migration script that puts this in place for existing lanes.)

### DIFF
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -51,25 +51,18 @@ def load_lanes(_db, library):
     Otherwise, a WorkList containing the visible top-level lanes is
     returned.
     """
+    top_level = WorkList.top_level_for_library(_db, library)
+    to_expunge = None
 
-    # Load all Lane objects from the database.
-    lanes = _db.query(Lane).filter(Lane.library==library).order_by(Lane.priority)
+    # It's likely this WorkList will be used across sessions, so
+    # expunge any data model objects from the database.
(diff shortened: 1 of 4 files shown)


api/lanes.py
```diff
@@ -235,6 +228,13 @@ def create_lanes_for_large_collection(_db, library, languages, priority=0):
     YA = [Classifier.AUDIENCE_YOUNG_ADULT]
     CHILDREN = [Classifier.AUDIENCE_CHILDREN]
 
+    common_args = dict(
+        languages=languages,
+        media=[Edition.BOOK_MEDIUM]
```
aslagle: might be good to change this to `Edition.FULFILLABLE_MEDIA` now that it exists

leonardr: I thought about that, and it might end up that way, but it's more likely we'll end up putting audiobooks in separate lanes, which would have a different `media` setting.

aslagle: That's true, it would be better to have audiobooks in different lanes.

leonardr: We're doing a survey to see how people want to encounter different types of books. Certainly movies would be in a separate section.

PR #813, opened 2018-01-12
Limit newly created lanes to BOOK_MEDIUM

leonardr:
This branch brings in recent submodule updates, and also makes sure that when you create a lane through the admin interface it only contains ebooks.


bin/opds_entry_coverage
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python
```
leonardr: See https://github.com/NYPL-Simplified/circulation-docker/pull/73 for this.

aslagle: Maybe you could make this a symlink to the script in core/bin and then you wouldn't have to duplicate it.

leonardr: I don't think a symlink would work because the import also needs to change.

aslagle: I'd have to test it but I think the core script would import relative to itself regardless of where you ran it from.

leonardr: I'll try it out.

leonardr: No, it doesn't work. :(

```
simplified/circulation$ bin/opds_entry_coverage
  File "bin/opds_entry_coverage", line 8, in <module>
    from coverage import OPDSEntryWorkCoverageProvider
ImportError: No module named coverage
```

PR #816, opened 2018-01-15
Made outstanding metadata editable in the admin interface.

aslagle:
This brings in https://github.com/NYPL-Simplified/circulation-web/pull/135 and https://github.com/NYPL-Simplified/server_core/pull/769.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -378,7 +378,7 @@ class SignInController(AdminController):
 
 class WorkController(CirculationManagerController):
 
-    STAFF_WEIGHT = 1
+    STAFF_WEIGHT = 1000
 
     def details(self, identifier_type, identifier):
         """Return an OPDS entry with detailed information for admins.
@@ -625,7 +625,7 @@ class WorkController(CirculationManagerController):
                     _("The rating must be a number between %(low)s and %(high)s.",
                       low=scale[0], high=scale[1]))
             if (new_rating - scale[0]) / (scale[1] - scale[0]) != work.quality:
-                primary_identifier.add_measurement(staff_data_source, Measurement.RATING, new_rating)
+                primary_identifier.add_measurement(staff_data_source, Measurement.RATING, new_rating, weight=WorkController.STAFF_WEIGHT)
                 changed = True
                 changed_rating = True
 
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -610,6 +610,7 @@ class WorkController(CirculationManagerController):
         # the calculated quality rather than the staff rating, which will be
         # confusing. It might also be useful to make it more clear how this
         # relates to the quality threshold in the library settings.
+        changed_rating = False
         new_rating = flask.request.form.get("rating")
         if new_rating != None and new_rating != '':
             try:
@@ -626,7 +627,9 @@ class WorkController(CirculationManagerController):
             if (new_rating - scale[0]) / (scale[1] - scale[0]) != work.quality:
                 primary_identifier.add_measurement(staff_data_source, Measurement.RATING, new_rating)
                 changed = True
+                changed_rating = True
 
+        changed_summary = False
         new_summary = flask.request.form.get("summary") or ""
         if new_summary != work.summary_text:
             old_summary = None
@@ -644,6 +647,7 @@ class WorkController(CirculationManagerController):
                 self._db.delete(old_summary)
 
             changed = True
+            changed_summary = True
 
         if changed:
             # Even if the presentation doesn't visibly change, we want
@@ -654,8 +658,8 @@ class WorkController(CirculationManagerController):
                 classify=True,
                 regenerate_opds_entries=True,
                 update_search_index=True,
-                calculate_quality=True,
-                choose_summary=True,
+                calculate_quality=changed_rating,
+                choose_summary=changed_summary,
             )
             work.calculate_presentation(policy=policy)
 
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -33,6 +33,7 @@ from core.model import (
     Collection,
     Complaint,
     ConfigurationSetting,
+    Contributor,
     CustomList,
     DataSource,
     Edition,
@@ -44,6 +45,7 @@ from core.model import (
     Library,
     LicensePool,
     Loan,
+    Measurement,
     Patron,
     PresentationCalculationPolicy,
     Representation,
@@ -58,7 +60,10 @@ from core.util.problem_detail import (
 )
 from core.util.http import HTTP
 from problem_details import *
-from core.util import fast_query_count
+from core.util import (
+    fast_query_count,
+    LanguageCodes,
+)
 
 from api.config import (
     Configuration, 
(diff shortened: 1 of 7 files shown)


api/admin/controller.py
```diff
@@ -947,7 +1110,11 @@ def lanes_for_parent(parent):
             parent_id = flask.request.form.get("parent_id")
             display_name = flask.request.form.get("display_name")
             custom_list_ids = json.loads(flask.request.form.get("custom_list_ids", "[]"))
-            inherit_parent_restrictions = flask.request.form.get("inherit_parent_restrictions", False)
+            inherit_parent_restrictions = flask.request.form.get("inherit_parent_restrictions")
+            if inherit_parent_restrictions == "true":
+                inherit_parent_restrictions = True
+            else:
+                inherit_parent_restrictions = False
```
aslagle: This is an unrelated bug I found.

api/admin/opds.py
```diff
@@ -1,15 +1,17 @@
...
 from core.model import (
     BaseMaterializedWork,
     LicensePool,
     Session,
 )
 from core.opds import AcquisitionFeed
+from core.util.opds_writer import AtomFeed
 
-class AdminAnnotator(CirculationManagerAnnotator):
+class AdminAnnotator(VerboseAnnotator, CirculationManagerAnnotator):
```
aslagle: I found a bug in the complaints and hidden books feeds due to this - they're missing acquisition links.

api/admin/controller.py
```diff
@@ -411,9 +416,61 @@ def complaints(self, identifier_type, identifier):
...
     def edit(self, identifier_type, identifier):
         """Edit a work's metadata."""
 
+        # TODO: It would be nice to use the metadata layer for this, but
+        # this code handles empty values differently than other metadata
+        # sources. When a staff member deletes a value, that indicates
+        # they think it should be empty. This needs to be indicated in the
+        # db so that it can overrule other data sources that set a value,
+        # unlike other sources which set empty fields to None.
+
```
leonardr: I thought about this for a while and the only solution I can see is to define a special string that can be put into the database and means "suppress any other value".

api/admin/controller.py
```diff
@@ -462,6 +555,74 @@ def edit(self, identifier_type, identifier):
...
 
+        new_medium = flask.request.form.get("medium")
+        if new_medium:
+            if new_medium not in Edition.medium_to_additional_type.keys():
+                self._db.rollback()
+                return UNKNOWN_MEDIUM.detailed(
+                    _("Medium %(medium)s is not one of the known media.",
+                      medium=new_medium))
+            staff_edition.medium = new_medium
+            changed = True
```
leonardr: If you change a title's medium, it seems like there's currently a risk that you'll remove it from the interface with no way to see it or change it back.

api/admin/controller.py
```diff
@@ -462,6 +555,74 @@ def edit(self, identifier_type, identifier):
...
+        if new_issued != staff_edition.issued:
+            staff_edition.issued = new_issued
+            changed = True
+
+        # TODO: This lets library staff add a 1-5 rating, which is used in the
+        # quality calculation. However, this doesn't work well if there are any
+        # other measurements that contribute to the quality. The form will show
+        # the calculated quality rather than the staff rating, which will be
+        # confusing. It might also be useful to make it more clear how this
+        # relates to the quality threshold in the library settings.
```
leonardr: We could change the quality calculation so that if a rating is present from a certain source, no other ratings are considered.

api/admin/controller.py
```diff
@@ -462,6 +555,74 @@ def edit(self, identifier_type, identifier):
...
+                self._db.rollback()
+                return INVALID_RATING
+            scale = Measurement.RATING_SCALES[DataSource.LIBRARY_STAFF]
+            if new_rating < scale[0] or new_rating > scale[1]:
+                self._db.rollback()
+                return INVALID_RATING.detailed(
+                    _("The rating must be a number between %(low)s and %(high)s.",
+                      low=scale[0], high=scale[1]))
+            if (new_rating - scale[0]) / (scale[1] - scale[0]) != work.quality:
+                primary_identifier.add_measurement(staff_data_source, Measurement.RATING, new_rating)
```
leonardr: You can also set `weight` really high here to swamp the effect of other ratings.

api/admin/controller.py
```diff
@@ -489,9 +650,11 @@ def edit(self, identifier_type, identifier):
                 classify=True,
                 regenerate_opds_entries=True,
                 update_search_index=True,
-                choose_summary=True
+                calculate_quality=True,
```
leonardr: I don't know how much time it will save to change this, but it occurs to me that you could pass in calculate_quality=True or choose_summary=True only if those fields changed in the current request. Those methods are potentially expensive, especially choose_summary, which performs textual analysis.

api/admin/opds.py
```diff
@@ -19,6 +21,12 @@ def annotate_work_entry(self, work, active_license_pool, edition, identifier, fe
 
         super(AdminAnnotator, self).annotate_work_entry(work, active_license_pool, edition, identifier, feed, entry)
 
+        if edition.imprint:
+            # TODO: Is there a better tag to use for this, and should it be in core?
```
leonardr: I think https://bib.schema.org/publisherImprint is the one to use.

tests/admin/test_controller.py
```diff
@@ -227,6 +227,86 @@ def test_details(self):
...
+
+    def test_edit_invalid_date_format(self):
+        response = self._make_test_edit_request(
+            [('issued', self._str)])
+        eq_(400, response.status_code)
+        eq_(INVALID_DATE_FORMAT.uri, response.uri)
+
+    def test_edit_invalid_rating_not_number(self):
+        response = self._make_test_edit_request(
+            [('rating', self._str)])
```
leonardr: self._str returns a string version of a number; you might be getting this status code because the 'number' is too large. Recommend you change this to an obvious non-number string like "abc".

api/admin/opds.py
leonardr: I don't follow -- CirculationManagerAnnotator has the acquisition links but making it also inherit VerboseAnnotator takes them away? Is VerboseAnnotator missing a super() call?

aslagle: Correct, that was the problem. But I need to investigate more, I think there may be something else wrong with those feeds as well.

api/admin/controller.py
aslagle: Do you think we should do that? How many other sources do we get ratings from?

We could also do something like we do with classifications where we show all the sources and weights.

api/admin/controller.py
leonardr: I think the first special string doesn't have to be a string; it can be an object like `NoValue` in `PatronData`. That might make things clearer.

The second "special string" could also be an object that indicates a special out-of-band value (a string or whatever it appropriate) should be stored in the database.

api/admin/controller.py
leonardr: The main other rating source is the distributor itself, e.g. Overdrive.

I like the idea of showing all the sources and weights, but weighting library staff opinion very heavily.

aslagle: I agree but I'm not sure how to present this information, so for now I think I'd like to expose the staff rating in the admin OPDS annotator and use that to populate the form rather than quality, and also give the staff rating a high weight.

leonardr: Sounds good.

api/admin/controller.py
```diff
@@ -373,7 +378,7 @@ def error_response(self, problem_detail):
 
 class WorkController(CirculationManagerController):
 
-    STAFF_WEIGHT = 1
+    STAFF_WEIGHT = 1000
```
aslagle: I changed this for ratings and it is also used for classifications, but changing it for those shouldn't matter. We originally were handling staff classifications by giving them a much higher weight than other classifications, but then we changed it so that if there are any staff classifications we use them and ignore anything else, so their weight doesn't matter except relative to other staff classifications. Maybe we'll end up doing the same thing with ratings eventually.

api/admin/controller.py
```diff
@@ -411,9 +416,61 @@ def complaints(self, identifier_type, identifier):
         
         return response
 
+    def roles(self):
+        """Return a mapping from MARC codes to contributor roles."""
+        # TODO: The admin interface only allows a subset of the roles
+        # listed in model.py since it uses the OPDS representation of
+        # the data, and some of the roles map to the same MARC code.
```
aslagle: @leonardr do you have any thoughts on this?

leonardr: In most cases this is fine because some other role means more or less the same thing (that's why they have the same MARC role). Just don't offer one or the other.

Use CONTRIBUTOR_ROLE for 'ctb' instead of COLLABORATOR_ROLE.
Use AFTERWORD_ROLE for 'aft' instead of COLOPHON_ROLE.
Use ASSOCIATED_ROLE for 'asn' instead of UNKNOWN_ROLE.
Use PRODUCER_ROLE for 'pro' instead of EXECUTIVE_PRODUCER_ROLE.

PRIMARY_AUTHOR_ROLE is the tricky one. We distinguish the primary author from other authors by putting them at the front of the list of contributors, rather than using a different MARC code. Whether changing the primary author means rearranging them in the user interface or selecting 'Primary Author' from a dropdown, it should be possible to convey the user's intention by reordering the contributor list in the OPDS entry sent back to the server.

aslagle: Okay, I could probably manage that by using PRIMARY_AUTHOR_ROLE for the first author that was submitted in the form. I think the UI will keep them in the order they were in the OPDS entry.

aslagle:
@leonardr I think this and https://github.com/NYPL-Simplified/circulation-web/pull/135 and https://github.com/NYPL-Simplified/server_core/pull/769 are all ready to merge now, could you look over my latest changes?

api/admin/controller.py
leonardr: I don't think it's a big deal right now, but that sounds like there's no way to change who the primary author is.

aslagle: The form will show them with the primary author first, and you can remove the authors and re-add them in the order you want. It's a little annoying since you can't just move them, but it works.

PR #828, opened 2018-01-22
Support authentication with no password

leonardr:
This branch changes the three authentication providers that use Basic Auth (simple, SIP2 and Millenium Patron) to handle the use case where a library authenticates patrons using only a barcode, with no password.

I added a [new value](https://github.com/NYPL-Simplified/Simplified/wiki/Authentication-For-OPDS-Extensions#keyboard) for the `keyboard` extension to Authentication For OPDS, called `No input`. This signals to clients that passwords should not be gathered or provided. On the server side, it controls what CM-side validation is done on passwords (basically, they have to be empty) and how the interaction with the authentication provider goes.

When a SIP2 integration is set to not use passwords, it omits the AD field from the SIP2 message even if a password was provided.

When a Millenium Patron integration is set to not use passwords, it does a patron lookup of the patron's identifier, rather than hitting the `pintest` endpoint, and returns success if the lookup succeeds.

When a simple authenticator is set not to use passwords, it just doesn't check the incoming password. If the barcode matches, authentication is successful.

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -1154,51 +1154,51 @@ class BasicAuthenticationProvider(AuthenticationProvider):
     
     # Identifiers can be presumed invalid if they don't match
     # this regular expression.
-    IDENTIFIER_REGULAR_EXPRESSION = 'identifier_regular_expression'
+    IDENTIFIER_REGULAR_EXPRESSION = u'identifier_regular_expression'
 
     # Passwords can be presumed invalid if they don't match this regular
     # expression.
-    PASSWORD_REGULAR_EXPRESSION = 'password_regular_expression'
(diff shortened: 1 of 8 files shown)


api/authenticator.py
```diff
@@ -1458,16 +1467,21 @@ def server_side_validation(self, username, password):
...
                 self.identifier_re.match(username) is not None
             )
-        if self.password_re:
-            valid = valid and password is not None and (
-                self.password_re.match(password) is not None
-            )
+
+        if not self.collects_password:
+            # The only legal password is an empty one.
+            valid = valid and password in (None, '')
```
leonardr: This makes sure submitted passwords aren't sent into the actual AuthenticationProviders, but the providers themselves also check so that if a submitted password does get in somehow, it's not sent over to the ILS.

PR #830, opened 2018-01-22
After creating a library, create some lanes for it.

leonardr:
This branch ensures that a new library immediately has a set of lanes created for it. These are probably not the lanes the library wants in the long term but it stops a common 'where are my books?' scenario during setup.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1544,6 +1544,11 @@ class SettingsController(CirculationManagerController):
                     ))
 
         if is_new:
+            # Now that the configuration settings are in place, create
+            # a default set of lanes.
+            create_default_lanes(self._db, library)
+
+        if is_new:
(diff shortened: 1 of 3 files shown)


aslagle:
Will this only work if the admin configures languages when the create the library? It won't have any books yet so it won't be able to get languages from the collection.

leonardr:
If the admin doesn't configure any languages then the library gets the default config (a large English collection) and that config is used.

aslagle:
I thought it didn't have defaults for language configuration and just looked at the library's collections if there were no settings. How does it get the default config?

leonardr:
I guess it's not correct to say that 'large english collection' is the default configuration, but that's [what `classify_holdings` returns](https://github.com/NYPL-Simplified/circulation/blob/master/api/config.py#L336)

PR #833, opened 2018-01-24
Update SIP read message conditions to allow for a SIP server configur…

alsrlw:
…ed to send CRLF line endings and correctly handle a condition where a message has a non-CR or LF line ending.

If the original message in data has neither a CR or LF last character, it will fall through the IF conditions and repeat the loop. The second time through the loop, there is likely not to be a message in the socket. But data will still have data from the first pass, so the 'if not data' condition will fail. This results in an uncaught infinite loop. (Without adding the condition for the LF line ending, the infinite loop occurs for any SIP server sending either CRLF or LF line endings. Koha sends CRLF by default, and Evergreen allows for CRLF.)

### DIFF
--- a/api/sip/client.py
+++ b/api/sip/client.py
@@ -707,9 +707,9 @@ class SIPClient(Constants):
         while not done:
             tmp = self.socket.recv(4096)
             data = data + tmp
-            if not data:
+            if not tmp:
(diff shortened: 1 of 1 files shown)


leonardr:
This is a great catch. Thanks for doing the legwork necessary to find these problems.

PR #836, opened 2018-01-25
Added admin controller for managing a book's list membership.

aslagle:
This brings in https://github.com/NYPL-Simplified/circulation-web/pull/136 and adds a controller method to support it.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -977,6 +977,8 @@ class WorkController(CirculationManagerController):
             else:
                 lists = []
 
+            affected_lanes = set()
+
             # Remove entries for lists that were not in the submitted form.
             submitted_ids = [l.get("id") for l in lists if l.get("id")]
             for entry in work.custom_list_entries:
@@ -984,7 +986,7 @@ class WorkController(CirculationManagerController):
                     list = entry.customlist
                     list.remove_entry(work)
                     for lane in Lane.affected_by_customlist(list):
-                        lane.update_size(self._db)
+                        affected_lanes.add(lane)
 
             # Add entries for any new lists.
             for list_info in lists:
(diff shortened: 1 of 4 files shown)


api/admin/controller.py
```diff
@@ -955,6 +955,58 @@ def _count_complaints_for_work(self, work):
...
+                lists = []
+
+            # Remove entries for lists that were not in the submitted form.
+            submitted_ids = [l.get("id") for l in lists if l.get("id")]
+            for entry in work.custom_list_entries:
+                if entry.list_id not in submitted_ids:
+                    list = entry.customlist
+                    list.remove_entry(work)
+                    for lane in Lane.affected_by_customlist(list):
+                        lane.update_size(self._db)
```
leonardr: There's a chance this could cause a big performance hit. I think it would be better to keep a set of affected lanes and update their size outside of this loop.

api/admin/controller.py
```diff
@@ -955,6 +955,58 @@ def _count_complaints_for_work(self, work):
...
+                    if not list:
+                        self._db.rollback()
+                        return MISSING_CUSTOM_LIST.detailed(_("Could not find list \"%(list_name)s\"", list_name=name))
+                else:
+                    list, is_new = create(self._db, CustomList, name=name, data_source=staff_data_source, library=library)
+                    list.created = datetime.now()
+                entry, was_new = list.add_entry(work)
+                if was_new:
+                    for lane in Lane.affected_by_customlist(list):
+                        lane.update_size(self._db)
```
leonardr: Same here -- keep adding to the set and update sizes later.

leonardr:
:+1: apart from the possibly redundant lane size updates. 

When you merge this, could you also do a submodule update on core to bring in my recent `opds:position` branch?

PR #838, opened 2018-01-25
Add ability to filter Millenium API patrons

jonathangreen:
This allows a per-library check when using the Millenium API for patron authentication to test if a field in the patron record matches a string value or a regex. This allows shared consortial ILS that use the patron record fields to distinguish libraries to work with the circ manager.


api/millenium_patron.py
```diff
@@ -99,6 +106,31 @@ class MilleniumPatronAPI(BasicAuthenticationProvider, XMLParser):
           "default": PIN_AUTHENTICATION_MODE
         }
     ] + BasicAuthenticationProvider.SETTINGS
+
+    LIBRARY_SETTINGS = [
+        { "key": PATRON_FIELD_RESTRICTION_FIELD,
+          "label": _("Patron record restriction field"),
+          "type": "select",
+          "description": _("Field on the patron record used for <i>patron record restriction</i> below."),
```
aslagle: My initial thought was that this setting should be optional since not everyone will use it, but it would be more work to validate that it gets set whenever the patron field restriction setting does, so I think it's okay to set it even when it's not being used.

I would maybe be more explicit in the description that it won't be used unless you set the patron record restriction setting.

api/millenium_patron.py
```diff
@@ -99,6 +106,31 @@ class MilleniumPatronAPI(BasicAuthenticationProvider, XMLParser):
...
+          "options": [
+              { "key": AGENCY_FIELD, "label": _(AGENCY_FIELD) },
```
aslagle: The `_` is from a localization library and part of the localization process is searching the code and pulling out strings that need to be translated, but I don't think it will work if you pass in a variable like this. Since this is a field in Sierra I don't think it should be translated anyway, so I would take out the `_`s here.

api/millenium_patron.py
```diff
@@ -99,6 +106,31 @@ class MilleniumPatronAPI(BasicAuthenticationProvider, XMLParser):
...
+        },
+        { "key": PATRON_FIELD_RESTRICTION_STRING,
+          "label": _("Patron record restriction"),
+          "description": _("<p>When multiple libraries share an ILS, a person may be able to " +
+                           "authenticate with the ILS but not be considered a patron of " +
+                           "<i>this</i> library. This setting contains the rule for determining " +
+                           "whether an identifier is valid for this specific library.</p>" +
+                           "This compares the <i>patron record restriction field</i> on the patron " +
+                           "record to this text. Match is treated as a regular expression if it starts " +
+                           "with ^, otherwise a string match is used."),
```
aslagle: I would also mention that this is mutually exclusive with the patron identifier prefix setting. I'm not sure what would happen if you put in both but I'm guessing it wouldn't work.

api/millenium_patron.py
```diff
@@ -136,24 +168,47 @@ def __init__(self, library, integration, analytics=None):
...
+        # Check that we match the patron record restriction
+        patrondata, match = self.patron_record_restriction_matches(username)
+        if not match:
+            return False
+
         if not self.collects_password:
             # We don't even look at the password. If the patron exists, they
             # are authenticated.
-            patrondata = self._remote_patron_lookup(username)
+            patrondata = patrondata if patrondata else self._remote_patron_lookup(username)
```
aslagle: Won't patrondata always be defined here? If for some reason it wasn't, it probably wouldn't help to do the remote patron lookup again.

api/millenium_patron.py
jonathangreen: I don't think you would ever want to, but they do work fine together. The patron identifier prefix is tested first [here](https://github.com/NYPL-Simplified/circulation/blob/master/api/authenticator.py#L1455-L1485), then if it passes then this check will be done.

aslagle: Cool, that's fine then. You might still want to mention that you probably don't need that setting if you use this one, but the description's long already. Up to you :)

api/millenium_patron.py
```diff
@@ -206,15 +261,54 @@ def remote_patron_lookup(self, patron_or_patrondata):
 
     def _remote_patron_lookup(self, identifier):
         """Look up patron information for the given identifier."""
-        path = "%(barcode)s/dump" % dict(barcode=identifier)
-        url = self.root + path
-        response = self.request(url)
-        return self.patron_dump_to_patrondata(identifier, response.content)
-        
+        raw_record = self._remote_patron_lookup_raw(identifier)
+        return self.patron_dump_to_patrondata(identifier, raw_record)
```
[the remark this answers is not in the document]
jonathangreen: I was thinking about doing this as well, then passing around `patrondata`, but since the `patrondata` object is shared by all the different authentication schemes, I wasn't sure if it was a good idea to add millennium specific data to it. I was considering adding something to it like `ils_data` where we could store ILS specific data on the patron record. 

What do you think? I can change this up to store the data on the patron record if it makes more sense.

leonardr:
It seems to me that PATRON_FIELD_RESTRICTION_STRING does the same thing as the standard PATRON_IDENTIFIER_RESTRICTION setting, and that this branch would work the same way it just used PATRON_FIELD_RESTRICTION_STRING.

The new thing here is that some value other than the patron identifier might need to be checked against the "patron identifier restriction". That's a good thing to support but it's not specific to Millenium Patron. I know there are SIP2 servers that store the relevant value in the "permanent location" (AQ) field; we probably also get some who store it in the "institution ID" (AO) field.

It should be possible to move `PATRON_FIELD_RESTRICTION_FIELD` into `AuthenticationProvider`. Every subclass can interpret the value however they want when creating the `PatronData` (I agree with Amy's comment re: adding this to `PatronData`).

api/millenium_patron.py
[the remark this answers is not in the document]
aslagle: That's a good point. I think the `ils_data` idea would work, or you could make a `MilleniumPatronData` subclass of `PatronData` that has the one extra thing.

jonathangreen:
@leonardr trying to wrap my head around the changes you want: 
- Move `PATRON_FIELD_RESTRICTION_FIELD` to `AuthenticationProvider`
- Remove `PATRON_FIELD_RESTRICTION_STRING` and use `PATRON_IDENTIFIER_RESTRICTION`
- Move restriction data to `PatronData`
- Create new function in `AuthenticationProvider` for evaluating the restriction that can be overridden in subclasses? 

Does that sound like I'm on the right track before I'm making changes?

leonardr:
Yes, except I don't think you should plan for the evaluation function to be overridden. The code that differs between subclasses is the code that reads the appropriate value from the ILS patron record and sticks it in the `PatronData`. That code will be in the `remote_patron_lookup` implementation.

The code that compares the value in `PatronData` to PATRON_IDENTIFIER_RESTRICTION will always work the same way, and it can stay in `patron_identifier_restriction_matches`, but 1) you probably want to rename that method, and 2) you'll need to remove the existing calls to that method (e.g. at the beginning of `server_side_validation`) and put in a new call that does the validation after the `PatronData` has been obtained.

aslagle:
The options for the PATRON_FIELD_RESTRICTION_FIELD setting will also be different between subclasses.

leonardr:
I would be okay if in the first pass we made PATRON_FIELD_RESTRICTION_FIELD a text entry. I don't know what fields are likely to be used for this purpose and I don't want to have to do a release every time we add one. Once we think we have it nailed down I'd be more comfortable with making it a drop-down selection.

jonathangreen:
The evaluation of `patron_identifier_restriction_matches` is actually already different already depending what you want to do. Thats why I didn't go down the path we are talking about when I initially wrote this patch. 

Currently if you are enforcing a barcode restriction, you do a prefix match when the value entered in `PATRON_IDENTIFIER_RESTRICTION`. However when doing a match against a field on a record, you don't necessarily want a prefix match, more likely you want an exact match.

jonathangreen:
I'm still seeing `PATRON_FIELD_RESTRICTION_FIELD` as a dropdown, so you can select the existing behaviour there. Something like: 
- None
- Indentifer Prefix
- AGENCY[p86]
- ect

To @aslagle point though, I'm not sure how to let subclasses add to these options, since they are defined in `LIBRARY_SETTINGS`.

leonardr:
That's a good point. We can further overload PATRON_IDENTIFIER_RESTRICTION so that sometimes it means a prefix and sometimes an exact match and sometimes a regex. But it might be better to add a PATRON_IDENTIFIER_RESTRICTION type that lets the admin select how the restriction should be applied against the value.

That would let us handle a weird case like "AQ field must be one of these two values" (which I believe I've actually seen in Connecticut).

leonardr:
Subclasses can define their own LIBRARY_SETTINGS, so you can have each subclass define a setting with the same name but different options. It would be a little icky and I have other reasons for not wanting this to be a drop-down just yet, but I think you can do it without much duplicated code.

jonathangreen:
@leonardr how does this look?

api/authenticator.py
```diff
@@ -936,49 +971,90 @@ def __init__(self, library, integration, analytics=None):
...
+            self.patron_restriction_type = self.PATRON_RESTRICTION_TYPE_NONE
+
         restriction = ConfigurationSetting.for_library_and_externalintegration(
-            _db, self.PATRON_IDENTIFIER_RESTRICTION, library, integration
+            _db, self.PATRON_RESTRICTION_STRING, library, integration
         ).value
-        if restriction and restriction.startswith("^"):
-            # Interpret the value a regular expression
+
+        if self.patron_restriction_type == self.PATRON_RESTRICTION_TYPE_REGEX:
```
leonardr: Putting this into a helper method will make it easy to test all these possibilities without a lot of setup code.

api/authenticator.py
```diff
@@ -936,49 +971,90 @@ def __init__(self, library, integration, analytics=None):
...
+                return True
+        elif match_type == cls.PATRON_RESTRICTION_TYPE_PREFIX:
+            if field.startswith(restriction):
+                return True
+        elif match_type == cls.PATRON_RESTRICTION_TYPE_STRING:
+            if field == restriction:
+                return True
+        elif match_type == cls.PATRON_RESTRICTION_TYPE_LIST:
+            for r in restriction:
+                if r == field:
```
leonardr: This can be reduced to `if field in restriction`

tests/test_authenticator.py
```diff
@@ -1373,21 +1423,46 @@ def test_patron_identifier_restriction(self):
...
+        # If its type is string make sure its a string
+        type_setting.value = MockProvider.PATRON_RESTRICTION_TYPE_STRING
+        string_setting.value = "abc"
         provider = MockProvider(library, integration)
-        eq_("^abcd", provider.patron_identifier_restriction.pattern)
+        eq_('abc', provider.patron_restriction_string)
 
-        # Otherwise, it's a regular string that is used as a prefix.
-        setting.value = "abcd"
+        # If its type is string make sure its a string
```
leonardr: This comment is wrong.

leonardr:
This is a solid branch overall. I'd like @aslagle to take a look at it as well. The main concern I have is that seeing it all written it out makes it obvious that 'patron restriction' is a bad name for this field. I chose this name originally so it's my fault. It kind of made sense 

I suggest that what we're looking at is actually a 'library identifier'. The library identifier might be a field value (such as AQ) or it might be part of some other field (such as the patron's barcode).

So we have the patron field that you look at to determine the patron's library identifier. We have the rule that is applied to the library identifier to see if the patron has chosen the correct library. And we have the meta-rule about how to apply the rule: as a string, prefix, regexp etc.

So, the code looks fine to me but I would like to see the fields renamed to LIBRARY_IDENTIFIER_FIELD, LIBRARY_IDENTIFIER_RESTRICTION, and LIBRARY_IDENTIFIER_RESTRICTION_TYPE.

I would also like to see a migration script that a) converts from "patron restriction" to "library identifier restriction" and b) sets "library identifier restriction type" to either "prefix" or "regex" depending on whether or not the identifier restriction started with a carat.

api/authenticator.py
```diff
@@ -936,49 +971,87 @@ def __init__(self, library, integration, analytics=None):
...
+        if self.library_identifier_restriction_type == self.LIBRARY_IDENTIFIER_RESTRICTION_TYPE_REGEX:
             try:
-                restriction = re.compile(restriction)
+                self.library_identifier_restriction = re.compile(restriction)
             except Exception, e:
-                self.log.error(
-                    "Could not interpret identifier restriction as a regular expression: %r", e
-                )
-        self.patron_identifier_restriction = restriction
+                self.log.error("Could not interpret patron record restriction as a regular expression: %r", e)
```
aslagle: I think this should maybe raise an Exception rather than logging the error - not sure if it would be worse to block actual library patrons or to let in patrons of other libraries when there's a typo in the regex.

And this isn't necessary for this branch, but: it might be worth adding a `type: regex` to settings so that the admin interface controller code could validate that the expression compiles before it's saved to the db.

aslagle:
This is looking pretty good, I like the renaming. The main concern I have now is that the setting will show up for all basic auth providers but only Millenium will extract the library identifier and put it in the patron data. I think SIP at least should do this as well. FirstBook and the simple authentication provider probably don't matter.

leonardr:
I figured adding support for this feature to SIP would come soon afterwards in a separate branch.

But, `LIBRARY_IDENTIFIER_RESTRICTION_TYPE` doesn't need to go into `AuthenticationProvider.LIBRARY_SETTINGS`. The setting can be defined as a class variable of `AuthenticationProvider` and then any specific authentication provider that wants to use it can include it (or a variation of it) in its `LIBRARY_SETTINGS`.

api/authenticator.py
jonathangreen: I added a check for the regex to the controller, so it won't save if the regex is invalid. I also added a test for it. Since there is this validation do you think we still need to throw an exception here?

jonathangreen:
@leonardr can you spell out your suggestion about removing `LIBRARY_IDENTIFIER_RESTRICTION_TYPE` from `AuthenticationProvider.LIBRARY_SETTINGS`? I don't understand what you want here?

api/authenticator.py
aslagle: Nice! With validation this should never happen. But if you did somehow end up with a bad regex in the db I think it would still be better to throw an exception than log it.

leonardr:
Sure. `AuthenticationProvider` has various subclasses for Millenium authentication, SIP2, etc. When you go into the admin interface for one of those subclasses, it looks at e.g. `MilleniumPatronAPI.LIBRARY_SETTINGS` to build the user interface.  Since the subclass doesn't define its own `LIBRARY_SETTINGS`, the superclass value is used.

Right now `AuthenticationProvider.LIBRARY_SETTINGS` contains instructions for building the three UI controls for the "library ID" settings and instructions on where to store the values in the database. Since none of the subclasses define `LIBRARY_SETTINGS`, they're all going to end up with three identical UI controls, even though some of them don't support this feature and some of them might want to support it slightly differently.

Compare this to the world of collection integrations. `BaseCirculationAPI` defines `LIBRARY_SETTINGS` as an empty list, so by default a collection integration has no settings. But `BaseCirculationAPI` also defines some commonly used settings like `DEFAULT_LOAN_DURATION_SETTING`. And then subclasses like `OverdriveAPI` define `LIBRARY_SETTINGS` to include a mix of common settings from `BaseCirculationAPI` and custom integration-specific settings.

```
    LIBRARY_SETTINGS = BaseCirculationAPI.LIBRARY_SETTINGS + [
        { "key": BaseOverdriveAPI.ILS_NAME_KEY, "label": _("ILS Name"),
          "default": BaseOverdriveAPI.ILS_NAME_DEFAULT,
          "description": _("..."),
        },
        BaseCirculationAPI.DEFAULT_LOAN_DURATION_SETTING
    ]
```

I'm saying treat these settings the same way `BaseCirculationAPI` treats default loan duration settings: as something that a specific integration can easily drop in to their `LIBRARY_SETTINGS`, but not in `LIBRARY_SETTINGS` by default.

jonathangreen:
Thank you, I understand now. 

Wouldn't removing these settings from `LIBRARY_SETTINGS` reduce functionality available, as right now all auth providers allow filtering based on prefix and regular expression on the barcode? I know we are using SIP with barcode prefix filtering at LYRASIS currently. 

Even though Millenium is the only one filling out the library identifier field on `patrondata`, all auth providers can use this functionality provided that the `library_identifier_field` is set to [barcode](https://github.com/NYPL-Simplified/circulation/pull/838/files#diff-44bd848f3bcb2624e25c57623ef74c4fR1039).

leonardr:
I see. I thought that the current `AuthenticationProvider.LIBRARY_SETTINGS` was promising functionality that wasn't yet available for all the ILSes. Looking at it again, I think the problem is a smaller one: the interface for `LIBRARY_IDENTIFIER_RESTRICTION_TYPE` is awkward for every `AuthenticationProvider` except `MilleniumPatronAPI`. You basically have the choice of typing in "barcode", leaving it blank, or typing in something else and having it do nothing.

So I suggest doing this:

`AuthenticationProvider` has two different constants containing different interfaces for `LIBRARY_IDENTIFIER_RESTRICTION_TYPE`:

* LIBRARY_IDENTIFIER_RESTRICTION_TYPE_TEXT_FIELD - the text field you currently have in 
`AuthenticationProvider.LIBRARY_SETTINGS`.
* LIBRARY_IDENTIFIER_RESTRICTION_TYPE_DEFAULT - a 'select' option where the two options are "barcode" and the empty string.

`AuthenticationProvider.LIBRARY_SETTINGS` contains a reference to `LIBRARY_IDENTIFIER_RESTRICTION_TYPE_DEFAULT`.

`MilleniumPatronAPI.LIBRARY_SETTINGS` does something like this:

```
LIBRARY_SETTINGS = AuthenticationProvider.LIBRARY_SETTINGS.replace(
 LIBRARY_IDENTIFIER_RESTRICTION_TYPE_DEFAULT, 
 LIBRARY_IDENTIFIER_RESTRICTION_TYPE_TEXT_FIELD
)
```

So `MilleniumPatronAPI` has the default configuration settings but with one of the fields changed to have a different user interface.

As we go forward, other `AuthenticationProvider`s can do the same thing, switching from the two-item selection box to a free text field or to a different selection box with protocol-specific values like SIP2's "AQ".

The process of building the user interface by combining fields of different types is awkward, but this should  get things working without reducing functionality or confusing the users of other authentication mechanisms.

jonathangreen:
@leonardr @aslagle does that look better? Users of any authentication provider other then `MilleniumPatronAPI ` can only choose `barcode` to apply authentication rules again.

api/authenticator.py
```diff
@@ -917,10 +917,13 @@ class AuthenticationProvider(OPDSAuthenticationFlow):
         },
         { "key": LIBRARY_IDENTIFIER_FIELD,
           "label": _("Library Identifier Field"),
+          "type": "select",
+          "options": [
+              {"key": LIBRARY_IDENTIFIER_RESTRICTION_BARCODE, "label": _("Barcode")},
+          ],
```
leonardr: Have you tried this out? I think you also need the empty option in `options`, so it would look more like this:

```
 "options": [
              {"key": "", "label": _("Do not gather a library identifier")},
              {"key": LIBRARY_IDENTIFIER_RESTRICTION_BARCODE, "label": _("Gather library identifier from barcode")},
          ],
```

jonathangreen: I have tried it out, and it seemed okay to me. Whats the concern here? 

The empty option is essentially the same as selecting [`LIBRARY_IDENTIFIER_RESTRICTION_TYPE_NONE `](https://github.com/NYPL-Simplified/circulation/pull/838/files#diff-44bd848f3bcb2624e25c57623ef74c4fR910), in that it doesn't apply a restriction. It seems strange to me to have 2 ways in the UI to turn off the enforcement.

leonardr: It is strange to have two ways to turn off the enforcement; I'd forgotten about the other way and was focused on the oddity of having a select drop-down with only a single option. So I agree that what you have will work.

But the strange thing is that we have two ways of turning off enforcement in the first place, not that we expose both ways in the UI. 

In fact, we have three ways of turning off enforcement, since you can also set the library identifier restriction to a value that matches everything.

So my concern is less about the code and more about the UI, which as it stands requires the following configuration setup for a standalone library:

* Gather library ID from the barcode. [option]
* Don't enforce any restriction on the library ID. [option]
* Don't match the library ID against any particular string. [empty text input]

But a standalone library doesn't have a "library ID" at all. It's not in the barcode; it's not anywhere. I don't want to have to explain to libraries that they should gather a nonexistent library ID from the barcode and do nothing with it. It would make more sense to an administrator of that library to be able to set up the following UI, even though it's technically redundant:

* Don't gather library ID at all. [option]
* Don't enforce any restriction on the library ID. [option]
* Don't match the library ID against any particular string. [empty text input]
 
Ideally the second two configuration options would only show up once you decided to gather the library ID from somewhere, but failing that I think it's best to be able to make it clear in the first configuration option that there is no library ID.

aslagle: I agree it would be nice to be able to configure fields that showed up conditionally based on your other selections, but we'd have to figure out how to represent that and it would also require more front-end work. It's probably not worth it until we have some more settings like that.

For now I think this PR is good as is but some of the description strings could be improved. The library identifier restriction field and library identifier restriction descriptions could say that they won't be used if the library identifier restriction type is no restriction.

leonardr: I agree that it's good enough to change the description strings.

jonathangreen: Here is my proposed verbiage, let me know if this will suffice and I will make the change. If not alternate suggestions for the text would be appreciated. 

----

`LIBRARY_IDENTIFIER_RESTRICTION_TYPE`
When multiple libraries share an ILS, a person may be able to authenticate with the ILS but not be considered a patron of *this* library. This setting contains the rule for determining whether an identifier is valid for this specific library. 

If this setting it set to 'No Restriction' then the values for *Library Identifier Field* and *Library Identifier Restriction* will not be used.

`LIBRARY_IDENTIFIER_FIELD`
This is the field on the patron record that the *Library Identifier Restriction Type* is applied to, different patron authentication methods provide different values here. This value is not used if *Library Identifier Restriction Type* is set to 'No restriction'.

`LIBRARY_IDENTIFIER_RESTRICTION`
This is the restriction applied to the *Library Identifier Field* using the method chosen in *Library Identifier Restriction Type*. This value is not used if *Library Identifier Restriction Type* is set to 'No restriction'.

aslagle: Sounds good to me.

leonardr: 👍

leonardr:
OK, I'm going to update this and merge it.

jonathangreen:
🍾 🎉 

Thank you both @leonardr and @aslagle for taking the time to review this.

PR #839, opened 2018-01-29
Made it possible to associate collections with a custom list.

aslagle:
This uses https://github.com/NYPL-Simplified/server_core/pull/788 and https://github.com/NYPL-Simplified/circulation-web/pull/138, and changes the custom lists controller for the admin interface to handle collections.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1002,7 +1002,7 @@ class WorkController(CirculationManagerController):
                 else:
                     list, is_new = create(self._db, CustomList, name=name, data_source=staff_data_source, library=library)
                     list.created = datetime.now()
-                entry, was_new = list.add_entry(work)
+                entry, was_new = list.add_entry(work, featured=True)
                 if was_new:
                     for lane in Lane.affected_by_customlist(list):
                         affected_lanes.add(lane)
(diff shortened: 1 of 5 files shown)


api/controller.py
```diff
@@ -408,6 +408,8 @@ def load_lane(self, lane_identifier):
         if lane_identifier is None:
             # Return the top-level lane.
             lane = self.manager.top_level_lanes[library_id]
+            if isinstance(lane, Lane):
+                lane = self._db.merge(lane)
```
aslagle: I was getting some errors from the getting the parentage of a Lane that was detached from the session.

api/admin/controller.py
```diff
@@ -1002,7 +1002,7 @@ def custom_lists(self, identifier_type, identifier):
                 else:
                     list, is_new = create(self._db, CustomList, name=name, data_source=staff_data_source, library=library)
                     list.created = datetime.now()
-                entry, was_new = list.add_entry(work)
+                entry, was_new = list.add_entry(work, featured=True)
```
leonardr: I feel like this should be controlled by a feature in the GUI rather than always being true.

leonardr: That said, I don't have a preference between having this always be true and having it always be false.

aslagle: I agree, but we don't have a design for that feature and until we have it implemented I prefer always true.

api/controller.py
leonardr: I thought I had already made this change, and I did start it, but for whatever reason didn't finish. I bring this up because my incomplete change also handles the case where `lane` is a `WorkList` but not a `Lane`. In this case I wrote that all of the _children_ of the `Lane` also need to be merged. 

Since `top_level_lanes` can return a `WorkList` (if a site has multiple top-level lanes) I think that step is important here.

api/admin/controller.py
leonardr: OK, it's cool.

api/controller.py
leonardr: Yeah, I double checked and I'm pretty sure we need this change. `load_lanes` expunges the children of a top-level `WorkList` so `load_lane` needs to merge the children back in.

PR #842, opened 2018-01-30
Default lane configuration for a library with no language configuration and no holdings

leonardr:
This branch ensures that a newly created library will always have some lanes, even if there is no basis on which to determine which languages are actually represented in the library's collection. This avoids an error on library creation and reduces the amount of work necessary to get a basic, English-language collection up and running.

### DIFF
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -63,6 +63,39 @@ def load_lanes(_db, library):
     map(_db.expunge, to_expunge)
     return top_level
 
+
+def _lane_configuration_from_collection_sizes(estimates):
+    """Sort a library's collections into 'large', 'small', and 'tiny'
+    subcollections based on language.
+
+    :param estimates: A Counter.
+
+    :return: A 3-tuple (large, small, tiny). 'large' will contain the
+    collection with the largest language, and any languages with a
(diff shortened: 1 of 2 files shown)


api/lanes.py
```diff
@@ -63,6 +63,39 @@ def load_lanes(_db, library):
     map(_db.expunge, to_expunge)
     return top_level
 
+
+def _lane_configuration_from_collection_sizes(estimates):
```
leonardr: I refactored this method to make it independently testable.

api/lanes.py
```diff
@@ -63,6 +63,39 @@ def load_lanes(_db, library):
...
+    :return: A 3-tuple (large, small, tiny). 'large' will contain the
+    collection with the largest language, and any languages with a
+    collection 10% that size. 'small' will contain any languages with
+    a collection 1% that size, and 'tiny' will contain all other
+    languages represented in `estimates`.
+    """
+    if not estimates:
+        # There are no holdings. Assume we have a large English
+        # collection and nothing else.
+        return [u'eng'], [], []
```
leonardr: This is the only new code.

PR #845, opened 2018-02-01
Separate route for custom list with entries

aslagle:
To improve performance of some pages in the admin interface when a list has a large number of entries, I made the /custom_lists endpoint return an entry count instead of all the entries of all the lists. I also changed the /custom_list/<id> endpoint to support GET and POST requests. GET will return the list metadata and its entries and collections, and POST will edit the list. I moved the code for editing or creating a list to a helper method that's used from both endpoints.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1061,22 +1061,13 @@ class CustomListsController(CirculationManagerController):
     def custom_lists(self):
         library = flask.request.library
 
-        data_source = DataSource.lookup(self._db, DataSource.LIBRARY_STAFF)
-
         if flask.request.method == "GET":
             custom_lists = []
             for list in library.custom_lists:
-                entries = []
-                for entry in list.entries:
-                    if entry.edition:
-                        entries.append(dict(pwid=entry.edition.permanent_work_id,
-                                            title=entry.edition.title,
-                                            authors=[author.display_name for author in entry.edition.author_contributors],
-                        ))
                 collections = []
                 for collection in list.collections:
                     collections.append(dict(id=collection.id, name=collection.name, protocol=collection.protocol))
-                custom_lists.append(dict(id=list.id, name=list.name, entries=entries, collections=collections))
+                custom_lists.append(dict(id=list.id, name=list.name, collections=collections, entry_count=len(list.entries)))
             return dict(custom_lists=custom_lists)
 
         if flask.request.method == "POST":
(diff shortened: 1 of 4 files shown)


api/admin/controller.py
```diff
@@ -1061,114 +1061,129 @@ class CustomListsController(CirculationManagerController):
...
-                    if entry.edition:
-                        entries.append(dict(pwid=entry.edition.permanent_work_id,
-                                            title=entry.edition.title,
-                                            authors=[author.display_name for author in entry.edition.author_contributors],
-                        ))
                 collections = []
                 for collection in list.collections:
                     collections.append(dict(id=collection.id, name=collection.name, protocol=collection.protocol))
-                custom_lists.append(dict(id=list.id, name=list.name, entries=entries, collections=collections))
+                custom_lists.append(dict(id=list.id, name=list.name, collections=collections, entry_count=len(list.entries)))
```
leonardr: Obviously the big change is the one you made, but we could get all of this stuff -- custom lists for a library with counts and collections for each list -- with a single database query if necessary.

api/admin/controller.py
```diff
@@ -1061,114 +1061,129 @@ class CustomListsController(CirculationManagerController):
...
-            new_pwids = [entry.get("pwid") for entry in entries]
-            for entry in old_entries:
-                if entry.edition.permanent_work_id not in new_pwids:
-                    list.remove_entry(entry.edition)
+        if entries:
+            entries = json.loads(entries)
+        else:
+            entries = []
+
+        old_entries = [x for x in list.entries if x.edition]
```
leonardr: Here you're still going to take a performance hit if there are thousands of items in the list. You're going to pull all the  CustomListEntry objects from the CustomList, and then look up an individual Work for every entry in the new list. That's not a reason to not land this branch but I don't think it's going to make list membership _changes_ any faster.

You can look at `Collection.catalog_identifiers` and `CoverageRecord.bulk_add` for examples of how we deal with making this sort of membership change, when (say) a hundred records out of millions total are changing.

api/admin/controller.py
aslagle: It seems fast enough already, but I'll come back to it later if it's a problem. Possibly if you have a large number of lists it would be important.

api/admin/controller.py
aslagle: That's a good point, and I think I'm going to need a follow-up branch to address performance on the custom list edit page anyway.

PR #851, opened 2018-02-07
Added crawlable acquisition feed for a list, sorted by recently updated.

aslagle:
This adds an endpoint for a crawlable feed for a list, and links to it from any lanes that are based on a single custom list.

Something similar could be done for collections and the top-level lane for a library, but I didn't do it here.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -223,6 +223,13 @@ def acquisition_groups(lane_identifier):
 def feed(lane_identifier):
     return app.manager.opds_feeds.feed(lane_identifier)
 
+@library_route('/crawlable/<list_name>')
+@has_library
+@allows_patron_web
+@returns_problem_detail
+def crawlable_feed(list_name):
+    return app.manager.opds_feeds.crawlable_feed(list_name)
+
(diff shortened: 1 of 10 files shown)


api/routes.py
```diff
@@ -223,6 +223,13 @@ def acquisition_groups(lane_identifier):
 def feed(lane_identifier):
     return app.manager.opds_feeds.feed(lane_identifier)
 
+@library_route('/crawlable/<list_name>')
```
leonardr: There is little I enjoy less than nitpicking URL designs, but since in the near future collections and libraries will also have crawlable feeds, this URL needs to contain an indication that the client is crawling a list rather than a collection of the same name.

I think either of these two setups would work. I prefer the first one because it opens up the possibility that you might get other sorts of OPDS feeds out of these things.

* `/<library>/crawlable`
* `/<library>/lists/<list_name>/crawlable`
* `/<library>/collections/<collection_name>/crawlable`

* `/<library>/crawlable`
* `/<library>/crawlable/lists/<list_name>`
* `/<library>/crawlable/collections/<collection_name>`

leonardr:
This looks good apart from the URL design issue.

PR #852, opened 2018-02-07
Upload collection from disk

leonardr:
This branch ports from the content server a script that creates a collection based on a local MARC file and directories full of EPUB documents and cover images.

The script was previously untested; it's now exhaustively tested. This might have been overkill, but it provides a platform for future work in making a circulation manager that can host content directly.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -1109,7 +1109,7 @@ class DirectoryImportScript(Script):
         :return: A 2-tuple (Collection, MirrorUploader)
         """
         collection, is_new = Collection.by_name_and_protocol(
-            self._db, collection_name, ExternalIntegration.DIRECTORY_IMPORT
+            self._db, collection_name, ExternalIntegration.MANUAL
         )
 
         if is_new:
--- a/scripts.py
+++ b/scripts.py
@@ -1199,7 +1199,8 @@ class DirectoryImportScript(Script):
         download, or None if no such book can be found.
         """
         ignore, book_media_type, book_content = self._locate_file(
-            identifier.identifier, ebook_directory, ['.epub', '.pdf'],
+            identifier.identifier, ebook_directory,
+            Representation.COMMON_EBOOK_EXTENSIONS,
             "ebook file",
         )
         if not book_content:
@@ -1242,7 +1243,7 @@ class DirectoryImportScript(Script):
        """
        cover_filename, cover_media_type, cover_content = self._locate_file(
            identifier.identifier, cover_directory, 
-           ['.jpg', '.jpeg', '.png', '.gif'], "cover image"
+           Representation.COMMON_IMAGE_EXTENSIONS, "cover image"
        )
 
        if not cover_content:
--- a/scripts.py
+++ b/scripts.py
@@ -1061,10 +1061,8 @@ class DirectoryImportScript(Script):
         if dry_run:
             mirror = None
 
-        replacement_policy = ReplacementPolicy(
-            rights=True, links=True, formats=True, contributions=True,
-            mirror=mirror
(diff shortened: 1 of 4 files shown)


scripts.py
```diff
@@ -995,6 +1004,340 @@ class OPDSForDistributorsReaperScript(OPDSImportScript):
...
+            '--dry-run',
+            help=u"Show what would be imported, but don't actually do the import.",
+            action='store_true',
+        )
+        return parser
+
+    def do_run(self, cmd_args=None):
+        parser = self.arg_parser(self._db)
+        parsed = parser.parse_args(cmd_args)
+        collection_name = parsed.data_source_name
```
aslagle: This should be collection name, not data source name

scripts.py
```diff
@@ -995,6 +1004,340 @@ class OPDSForDistributorsReaperScript(OPDSImportScript):
...
+
+        :param collection_name: Name of the Collection.
+        :param data_source_name: Associate this data source with
+            the Collection if it does not already have a data source.
+            A DataSource object will be created if necessary.
+
+        :return: A 2-tuple (Collection, MirrorUploader)
+        """
+        collection, is_new = Collection.by_name_and_protocol(
+            self._db, collection_name, ExternalIntegration.DIRECTORY_IMPORT
```
aslagle: I might go with something like "Manual import" as the protocol to leave open the possibility of adding individual books to this collection from the admin interface as well. I'm not sure we need to distinguish that it was a bulk directory import.

scripts.py
```diff
@@ -995,6 +1004,340 @@ class OPDSForDistributorsReaperScript(OPDSImportScript):
...
+        if dry_run:
+            self.log.warn(
+                "This is a dry run. No files will be uploaded and nothing will change in the database."
+            )
+        collection, mirror = self.load_collection(
+            collection_name, data_source_name
+        )
+
+        if dry_run:
+            mirror = None
```
aslagle: Is it a problem if it's not a dry run but there's no mirror?

tests/test_scripts.py
```diff
@@ -422,3 +441,489 @@ def test_set_secret(self):
...
+                return [metadata1, metadata2]
+
+            def work_from_metadata(self, *args):
+                self.work_from_metadata_calls.append(args)
+
+        # First, try a dry run.
+
+        # Make a change to a model object so we can track when the
+        # session is committed.
+        from core.analytics import Analytics
```
aslagle: Not sure why this import is here?

scripts.py
leonardr: Would "Manual intervention" or "Library staff" be appropriate? We already have constants for those.

aslagle: I think "Manual intervention" sounds more like a protocol than "Library staff".

scripts.py
leonardr: Currently, yes, but not in the long term. Right now if there's no mirror, the cover image and ebook will be stored in the database, and the cover image won't be scaled to a thumbnail. So that's not good.

But, there's no reason the circulation manager couldn't "fulfill" an open-access book by serving a copy of the book directly from the database, or stick a thumbnail image into an OPDS feed as a `data:` URL. In this case it would be possible to serve a collection of a couple books by local authors, without having to do a full S3 integration.

So I would like to take steps towards making everything work if there's no mirror. But right now it is a problem (because you can't see the books you imported) and I'm okay with raising an exception if you try to run this script with no S3 integration.

tests/test_scripts.py
leonardr: It's a non-interesting story.

PR #858, opened 2018-02-12
Added SIP2 socket timeout to fix condition where long periods between…

alsrlw:
… SIP2 communications caused download or return requests from mobile clients to fail; added exceptions in receiving socket data to handle errors and re-connect SIP2. Also, proposing to revert the change in conditional check of 'tmp' back to 'data' in favor of original code since the original did not contribute to either of our service issues.


api/sip/client.py
```diff
@@ -705,14 +707,22 @@ def read_message(self, max_size=1024*1024):
...
+                tmp = self.socket.recv(4096)
+            except socket.timeout:
+                raise socket.error
+            except socket.error, e:
+                raise IOError("Socket error occurred receiving SIP response: %s" % e)
+            except Exception, e:
+                raise IOError("Some other error occurred with SIP response: %s" % e.message)
             data = data + tmp
-            if not tmp:
+            if not data:
```
leonardr: I agree that it doesn't make a difference but I think checking `tmp` was the right move.

api/sip/client.py
```diff
@@ -705,14 +707,22 @@ def read_message(self, max_size=1024*1024):
...
                 raise IOError("No data read from socket.")
             if ord(data[-1]) == 13 or ord(data[-1]) == 10:
                 done = True
             if len(data) > max_size:
                 raise IOError("SIP2 response too large.")
+            # Do we need to test for a malformed response with no proper line termination?
```
leonardr: I don't know how we'd distinguish between a response with no line termination and an incomplete response that times out.

api/sip/client.py
alsrlw: With the benefit of experienced eyes, then, leaving it a tmp is good.

api/sip/client.py
alsrlw: I left the comment in from my working/testing copy, because it was part of my trying to figure out what was going on--and when using if not data: there is a possibility of an infinite loop. So maybe the best reason for leaving the condition at if not tmp as you suggested. tmp being empty would be caught, indicating the connection was closed remotely. Would a socket timeout _with data having content_ (nothing else received after receiving something) indicate a malformed/improperly terminated response? 

This may be more academic than it deserves treatment. But it was in my mind, so I left the comment.

api/sip/client.py
```diff
@@ -705,14 +707,22 @@ def read_message(self, max_size=1024*1024):
         data = ""
         tmp = ""
         while not done:
-            tmp = self.socket.recv(4096)
+            try:
+                tmp = self.socket.recv(4096)
+            except socket.timeout:
+                raise socket.error
```
leonardr: socket.error here is a class, not an exception object. Since socket.timeout is a subclass of socket.error, the simplest thing to do would be to take these two lines out and let the next two lines handle the problem. You could also raise a more specific IOError than is raised in the next two lines.

But, I expect you're reraising the exception as `IOError` so it can be dealt with by `except (IOError, socket.error), e:` on line 361. I don't think that's necessary either, since `socket.error` and `socket.timeout` (a subclass of `socket.error`) are already covered.

Looking at `circulation/tests/sip/test_client.py:TestNetworkError.retry_on_initial_timeout` I also see that we have a test that verifies we create a new socket connection when a timeout happens.

api/sip/client.py
```diff
@@ -705,14 +707,22 @@ def read_message(self, max_size=1024*1024):
...
+            except socket.error, e:
+                raise IOError("Socket error occurred receiving SIP response: %s" % e)
+            except Exception, e:
+                raise IOError("Some other error occurred with SIP response: %s" % e.message)
```
leonardr: So I feel like the only new exception handling code is here -- the code that takes any kind of exception, even if it's not an IOError, and turns it into an IOError. And that doesn't seem like it jibes with your problem.

I suspect your problem was the socket had _no_ timeout and that was causing a different type of error which wasn't caught here? I'm testing that locally now but I don't really get what happens when the socket errors are reraised as `IOError`.

leonardr: Looking at the code, I think the problem was that the lack of a timeout was causing the process to be killed rather than an exception to be raised. If a timeout is in place, the existing code should handle the resulting exception.

api/sip/client.py
leonardr: I'm not 100% on this, but... we're asking for four kilobytes of data. I don't think there's a maximum size of a SIP message (although we give up after receiving a megabyte) but I've never seen one longer than maybe one kilobyte.

`man recv` says "The receive calls normally return any data available, up to the requested amount,  rather  than  waiting  for receipt of the full amount requested."

If we ask for four kilobytes of data, and we get less than that, it's pretty likely that's the whole message. But I _think_ if we ask again we're going to read zero bytes rather than getting a timeout. At that point `tmp` will be empty (even though `data` is non-empty) and we'll proceed as normal.

api/sip/client.py
alsrlw: This is exactly the kind of help I needed. Thanks. It's what happens when a neophyte gets in the middle of a codebase :-). Yes, I was looking at getting the handler starting line 361 involved, since I was unsure how errors trickled up the code stack, and was mocking a block I saw in articles elsewhere. A couple suggested handling socket.timeout errors in an except block prior to general exception in order to differentiate. Perhaps that would better be handled, if desired, in the 361 block.

api/sip/client.py
alsrlw: Yes, agreed about the no timeout. My thinking is that the original problem had to do with a timeout on the remote end, where the socket was closed, but somehow not detected on our end. One would think that would have triggered the except (IOError, socket.error), but it didn't. Originally, without a timeout specified for the socket, the recv was blocking, and only the UWSGI timeout killed it by killing the worker. So the blocks I added were mainly designed to help me see what happened when I added a timeout. Without a thorough knowledge of socket programming/behavior, I need comments like these you've made to see where to go with this. Thanks.

leonardr:
Closing this very old PR rather than rebasing it to `main`.

PR #861, opened 2018-02-14
Set storage configuration for collection integrations

leonardr:
This branch adds a setting that only applies to collection-type integrations. The setting only shows up when there are one or more site-wide storage integrations configures, and it controls which (if any) of those integrations should be used as the collection's mirror integration.

We added this setting to all collections, not only OPDS import collections, because it's reasonable to want to make your own mirror of cover images from commercial sources rather than relying on the metadata wrangler.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1675,30 +1675,6 @@ class SettingsController(CirculationManagerController):
                 protocol["sitewide"] = sitewide
 
             settings = getattr(api, "SETTINGS", [])
-            integrations = self._db.query(ExternalIntegration).filter(
-                ExternalIntegration.goal==ExternalIntegration.STORAGE_GOAL
-            ).order_by(
-                ExternalIntegration.name
-            ).all()
-            if integrations:
-                mirror_integration_setting = {
-                    "key": "mirror_integration_id",
-                    "label": _("Mirror"),
-                    "description": _("Any cover images or free books encountered while importing content from this collection can be mirrored to a server you control."),
-                    "type": "select",
-                    "options" : [
-                        dict(
-                            key=None,
-                            label=_("None - Do not mirror cover images or free books")
-                        )
-                    ]
-                }
-                for integration in integrations:
-                    mirror_integration_setting['options'].append(
-                        dict(key=integration.id, label=integration.name)
-                    )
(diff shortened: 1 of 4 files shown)


api/admin/controller.py
```diff
@@ -1675,15 +1677,15 @@ def _get_integration_protocols(self, provider_apis, protocol_name_attr="__module
                 protocol["sitewide"] = sitewide
 
             settings = getattr(api, "SETTINGS", [])
-            protocol["settings"] = settings
+            protocol["settings"] = list(settings)
```
leonardr: This was necessary because the list may be modified later by the insertion of the 'mirror integration' setting. We can't modify the class SETTINGS object because the list of storage integrations might change later.

api/admin/controller.py
```diff
@@ -1834,6 +1836,13 @@ def collections(self):
                         ]
         protocols = self._get_integration_protocols(provider_apis, protocol_name_attr="NAME")
 
+        # If there are storage integrations, add a mirror integration
+        # setting to every protocol's 'settings' block.
+        mirror_integration_setting = self._mirror_integration_setting()
+        if mirror_integration_setting:
+            for protocol in protocols:
```
aslagle: Shouldn't this be more limited? It doesn't make sense for most of the protocols.

api/admin/controller.py
```diff
@@ -1934,6 +1945,20 @@ def collections(self):
...
+                if value == self.NO_MIRROR_INTEGRATION:
+                    integration_id = None
+                else:
+                    integration = get_one(
+                        self._db, ExternalIntegration, id=value
+                    )
+                    if not integration:
+                        return MISSING_SERVICE
+                    if integration.goal != ExternalIntegration.STORAGE_GOAL:
+                        return INTEGRATION_GOAL_CONFLICT
```
aslagle: I usually rollback before returning a problem detail so that none of the changes get saved when there's an error.

api/admin/controller.py
leonardr: I think it makes sense for every _collection_ protocol because of the cover image mirror thing.

aslagle: :+1:

Issue #864, opened 2018-02-16
Possible to create redundant ExternalIntegrations, crashing the admin interface


leonardr:
Create a "Local Analytics" integration, then create a second one. The admin interface (and possibly the circ manager in general) will no longer work.

PR #865, opened 2018-02-16
Allow an integration class to say that only one integration of its type should be created

leonardr:
This branch addresses https://github.com/NYPL-Simplified/circulation/issues/864 by allowing the class that provides an integration to set `CARDINALITY = 1` to say that there should only ever be one integration of that type on a given site.

The motivating case is the 'Local analytics' integration, but there are probably others, so I chose the tactic of refactoring rather than adding a special case for 'Local analytics'.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1658,7 +1658,8 @@ class SettingsController(CirculationManagerController):
             self._db.delete(library)
             return Response(unicode(_("Deleted")), 200)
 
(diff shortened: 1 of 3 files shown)


aslagle:
This looks good. I'd like to note that ideally the client would also handle cardinality - it could hide a protocol from the dropdown if its cardinality is 1 and there's already a service with that protocol.

aslagle:
The only other integrations I think this would apply to at the moment are the metadata services. Search and admin authentication are already sitewide.

PR #868, opened 2018-02-20
Logging configurations

EdwinGuzman:
This PR adds routes to GET, POSt, and DELTE logging settings, and functions to get sitewide settings data.


aslagle:
I'd suggest associating the logging settings with a sitewide integration with goal 'logging' and protocol 'loggly', so it would be easier to support other logging providers.

leonardr:
We have two groups of logging settings. One controls stuff like "which messages get logged?" and "what do the messages look like?". The other controls the Loggly credentials. Every logging mechanism, whether it's Loggly, syslog, or something else, is affected by the first group of settings.

Currently the first group of settings is controlled by a sitewide integration with goal="logging" and protocol="Internal logging". I've hacked in this integration into the databases hosted by NYPL, but I don't like that as a long-term solution. The second group of settings is controlled by a sitewide integration with goal="logging" and protocol="loggly", as you suggest.

The goal of this branch and https://github.com/NYPL-Simplified/server_core/pull/808/files is to treat the first group of settings as a special group of sitewide settings, rather than the settings associated with a sitewide integration. This is just because "which messages should be logged?" is a setting that affects _every_ logging mechanism regardless of protocol.

leonardr:
I tried to use this for real and then realized it was just controller code, so consider this next bit as feedback as advice for the UI branch.

The existing UI for the site-wide setting manager doesn't display a setting's `description`, should it have one. Some of the logging settings are complicated enough to need `description`s, and unlike most of the site-wide settings people are actually going to want to change them. (The other site-wide settings should also get `description`s, but that doesn't need to happen now.)

It's also possible that the existing UI for site-wide settings doesn't support control types other than "text input", but I haven't tested that.

leonardr:
Overall this looks good but the test code needs to change slightly to guarantee complete coverage.

Previously, you had a method called `sitewide_settings` which handled a bunch of stuff. Test coverage was provided by these five tests:

* test_sitewide_settings_get
* test_sitewide_settings_post_errors
* test_sitewide_settings_post_create
* test_sitewide_settings_post_edit
* test_sitewide_setting_delete

Now you have a method called `_sitewide_settings_controller` which handles all that stuff. So those five tests should become tests of `_sitewide_settings_controller`. It shouldn't be much more than renaming the test methods and having them call `_sitewide_settings_controller` instead of `sitewide_settings`.

At that point, `_sitewide_settings_controller` is tested but `sitewide_settings` and `logging_settings` are not. Fortunately, testing these methods is really easy because all they do is pass one object or another into `_sitewide_settings_controller`.

You don't need to test each of those methods five different ways. You can create a `Mock(SettingsController)` that defines `_sitewide_settings_controller` to record the arguments it was called with, and then test that `sitewide_settings` and `logging_settings` both call the method with the argument you'd expect.

As an example, check out `test_sitewide_service_management`, which creates a `Mock(SettingsController)` and verifies that `storage_services` and `storage_service` call the mock methods with the right arguments.

aslagle:
From a UI perspective I think it would be simpler to keep all the settings on the integration, even if most people will set them to the same thing for every logging integration when they have more than one.

leonardr:
I see what you're saying. That will work for LOG_FORMAT, LOG_APP_NAME, and LOG_MESSAGE_TEMPLATE, which are used to configure the Formatters, but it won't work for LOG_LEVEL and DATABASE_LOG_LEVEL, which are used to configure the Loggers.

What if we made LOG_LEVEL and DATABASE_LOG_LEVEL into normal sitewide configuration options, and then created "syslog logging" and "loggly logging" integrations for the other stuff?

aslagle:
Yeah, that sounds good.

api/admin/controller.py
```diff
@@ -2279,38 +2305,29 @@ def patron_auth_service(self, service_id):
...
         if flask.request.method == "DELETE":
             setting = ConfigurationSetting.sitewide(self._db, key)
             setting.value = None
             return Response(unicode(_("Deleted")), 200)
 
+    def logging_services(self):
+        detail = _("You tried to create a new logging service, but a logging service is already configured.")
+        return self._manage_sitewide_service(
+            ExternalIntegration.LOGGING_GOAL,
+            [Loggly] + [SysLogger],
```
aslagle: This is minor but I think `[Loggly, SysLogger]` would look nicer.

api/admin/controller.py
```diff
@@ -2676,6 +2693,19 @@ def storage_service(self, service_id):
...
+            'logging_services', detail
+        )
+
+    def logging_service(self, service_id):
+        return self._delete_integration(
+            service_id, ExternalIntegration.LOGGING_GOAL
+        )
```
aslagle: You seem to have two copies of these methods.

EdwinGuzman:
@aslagle updated based on feedback.

PR #871, opened 2018-02-22
Crawlable library and collection lane

leonardr:
This branch adds controllers that let you get a crawlable feed covering an entire library, or an entire collection (within the context of one of the libraries that uses that collection).

It also renames CrawlableCustomListFacets to CrawlableFacets, since there's now more than one type of crawlable feed and they all use the same Facets subclass.

The query that builds the feed is much slower than I expected, but that can be improved later.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -628,24 +628,45 @@ class OPDSFeedController(CirculationManagerController):
         )
         return feed_response(feed.content)
 
-    def crawlable_feed(self, list_name):
-        """Build or retrieve a crawlable, paginated acquisition feed, sorted
-        by update date.
+    def crawlable_library_feed(self):
+        """Build or retrieve a crawlable acquisition feed for the
+        request library.
+        """
+        library = flask.request.library
+        url = self.cdn_url_for(
+            "crawlable_library_feed",
+            library_short_name=library_short_name,
+        )
+        title = library.name
+        lane = CrawlableCollectionBasedLane(library)
+        return self._crawlable_feed(lane)
+
+    def crawlable_library_feed(self):
+        """Build or retrieve a crawlable acquisition feed for the
+        requested collection.
+        """
+        # TODO: This is a little tough because there's no active library.
+        # needs to be possible to get some facets without one.
+
+    def crawlable_list_feed(self, list_name):
+        """Build or retrieve a crawlable, paginated acquisition feed for the
+        named CustomList, sorted by update date.
         """
         library = flask.request.library
         list = CustomList.find(self._db, list_name, library=library)
         if not list:
             return NO_SUCH_LIST
         library_short_name = library.short_name
+        title = list.name
         url = self.cdn_url_for(
-            "crawlable_feed", list_name=list.name,
+            "crawlable_list_feed", list_name=list.name,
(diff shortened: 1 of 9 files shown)


api/lanes.py
```diff
@@ -1051,13 +1057,58 @@ def order_facet_to_database_field(cls, order_facet):
...
 class CrawlableCustomListBasedLane(DynamicLane):
     """A lane that consists of all works in a single CustomList."""
 
     uses_customlists = True
 
     def initialize(self, library, list):
-        super(CrawlableCustomListBasedLane, self).initialize(library, list.name)
+        super(CrawlableCustomListBasedLane, self).initialize(
+            library, "Crawlable feed: %s" % list.name
+        )
```
leonardr: Making sure every crawlable feed has "Crawlable" in its identifier guarantees that it won't clash with other types of feeds when it's cached using the `CacheFeed` code.

api/controller.py
```diff
@@ -628,26 +629,61 @@ def feed(self, lane_identifier):
...
+        lane = CrawlableCollectionBasedLane(library)
+        return self._crawlable_feed(library, title, url, lane)
+
+    def crawlable_collection_feed(self, collection_name):
+        """Build or retrieve a crawlable acquisition feed for the
+        requested collection.
+        """
+        # We use a library only for purposes of creating a Facets object.
+        # The requested collection does not have to be associated with
+        # the default library.
```
aslagle: This comment doesn't really apply anymore.

api/lanes.py
```diff
@@ -1051,13 +1057,62 @@ def order_facet_to_database_field(cls, order_facet):
         else:
             return Facets.order_facet_to_database_field(order_facet)
 
+
+class CrawlableCollectionBasedLane(DynamicLane):
+
+    MAX_CACHE_AGE = 60 * 60 * 12 # 12 hours
```
aslagle: Should this be configurable? Is it only used when there's no configuration setting for the paged OPDS feed cache time?

api/lanes.py
```diff
@@ -1051,13 +1057,62 @@ def order_facet_to_database_field(cls, order_facet):
...
+        self.initialize(library, "Crawlable feed: %s" % identifier)
+        if collections:
+            # initialize() set the collection IDs to all collections
+            # associated with the library. We may want to restrict that
+            # further.
+            self.collection_ids = [x.id for x in collections]
+
+    def get_library(self, _db):
+        """Find the Library object associated with this WorkList."""
+        return Library.by_id(_db, self.library_id)
```
aslagle: Why do you need to override this method?

api/lanes.py
leonardr: After taking a closer look at the CachedFeed code, I think the paged OPDS feed cache time is going to be used instead of MAX_CACHE_AGE, because it's going to be fetched as a 'page' type feed. Since my time is limited I'm going to take out this code and say that properly caching these feeds is a future piece of performance work, like speeding up the query.

leonardr: Maybe they should be separately configurable but the cache time for this feed should definitely be a lot longer than for a normal feed shown to patrons.

PR #875, opened 2018-03-06
Simplify the ORDER BY and DISTINCT field lists for crawlable feeds

leonardr:
Crawlable feeds are ordered by update date, and then by collection ID and work ID, period. There's no need to also order them by common `Facets` fields such as author and title, and with those extra fields in the query, Postgres doesn't know to use the `mv_works_for_lanes_by_recently_updated` index when sorting search results. This made it incredibly slow to generate crawlable feeds for collections or libraries.

This branch makes `CrawlableFacets` completely override `Facets.order_by` to simplify the query into something that Postgres can match against `mv_works_for_lanes_by_recently_updated`.

### DIFF
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -1054,8 +1054,10 @@ class CrawlableFacets(Facets):
         """Order the search results by last update time."""
         from core.model import MaterializedWorkWithGenre as work_model
         updated = func.greatest(work_model.availability_time, work_model.first_appearance, work_model.last_update_time)
+        collection_id = work_model.collection_id
         work_id = work_model.works_id
-        return [updated.desc(), work_id], [updated, work_id]
+        return ([updated.desc(), collection_id, work_id],
+                [updated, collection_id, work_id])
 
 
 class CrawlableCollectionBasedLane(DynamicLane):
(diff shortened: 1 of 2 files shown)


api/lanes.py
```diff
@@ -1050,12 +1050,12 @@ def default(cls, library):
...
-            from core.model import MaterializedWorkWithGenre as work_model
-            return func.greatest(work_model.last_update_time, work_model.first_appearance)
-        else:
-            return Facets.order_facet_to_database_field(order_facet)
+    def order_by(cls):
+        """Order the search results by last update time."""
+        from core.model import MaterializedWorkWithGenre as work_model
+        updated = func.greatest(work_model.availability_time, work_model.first_appearance, work_model.last_update_time)
+        work_id = work_model.works_id
+        return [updated.desc(), work_id], [updated, work_id]
```
aslagle: mv_works_for_lanes_by_recently_updated also has collection_id. Is that not needed here or added somewhere else?

leonardr: Yes, you're right, and making this change significantly speeds up the library-wide crawlable feed. I was thinking that we wanted to only show a work once if it was available through multiple collections, and maybe we do eventually, but this is better for now.

PR #879, opened 2018-03-09
Share ODL collections with libraries on other circ managers

aslagle:
This branch, along with https://github.com/NYPL-Simplified/circulation-web/pull/144 and https://github.com/NYPL-Simplified/server_core/pull/816, enables admins to share their ODL collections with libraries on other circ managers.


api/circulation.py
```diff
@@ -259,13 +259,15 @@ class HoldInfo(CirculationInfo):
...
-                 identifier, start_date, end_date, hold_position):
+                 identifier, start_date, end_date, hold_position,
+                 external_identifier=None):
         super(HoldInfo, self).__init__(
             collection, data_source_name, identifier_type, identifier
         )
         self.start_date = start_date
         self.end_date = end_date
         self.hold_position = hold_position
+        self.external_identifier = external_identifier
```
aslagle: Holds can now have an external_identifier. (Loans already could from my previous ODL work.) When a patron borrows or places a hold on a book from another circ manager, a URL for getting the status of the loan or hold is stored in external_identifier.

api/controller.py
```diff
@@ -187,6 +188,8 @@ def load_settings(self):
         new_top_level_lanes = {}
         # Create a CirculationAPI for each library.
         new_circulation_apis = {}
+        # Create a SharedCollectionAPI for each library.
+        new_shared_collection_apis = {}
```
aslagle: SharedCollectionAPI is meant to be similar to CirculationAPI. I'm not sure it's necessary, but I thought it might be useful if we ever want inter-library loan for other collection types.

api/lanes.py
```diff
@@ -1071,8 +1074,12 @@ def __init__(self, library, collections=None):
             all Collections associated with `library` will be used.
         """
         self.library_id = library.id
+        self.collection_feed = False
         if collections:
             identifier = " / ".join(sorted([x.name for x in collections]))
+            if len(collections) == 1:
+                self.collection_feed = True
+                self.collection_name = collections[0].name
```
aslagle: This fixes the next links in the crawlable feeds. There's no route for a crawlable feed for multiple collections but not all the collections in a library, so this won't work for that case.

api/odl.py
```diff
@@ -531,13 +571,21 @@ def update_hold_queue(self, licensepool):
...
+            new_licenses_reserved = len(holds)
+            new_patrons_in_hold_queue = len(holds)
+        licensepool.update_availability(
+            licensepool.licenses_owned,
+            new_licenses_available,
+            new_licenses_reserved,
+            new_patrons_in_hold_queue,
+            analytics=self.analytics,
+            as_of=datetime.datetime.utcnow(),
+        )
```
aslagle: I thought I would have to add a `last_availability_update_time` to the crawlable feed order by but it turns out `LicensePool.update_availability` already changes `Work.last_update_time`.

api/opds.py
```diff
@@ -253,7 +253,7 @@ def lane_url(self, lane):
     def annotate_work_entry(self, work, active_license_pool, edition, identifier, feed, entry):
         updated = None
         if isinstance(self.lane, CrawlableCustomListBasedLane) and isinstance(work, BaseMaterializedWork):
-            updated = max(work.last_update_time, work.first_appearance)
+            updated = max(work.last_update_time, work.first_appearance, work.availability_time)
```
aslagle: We missed this in an earlier branch.

api/admin/controller.py
```diff
@@ -1692,6 +1692,10 @@ def _get_integration_protocols(cls, provider_apis, protocol_name_attr="__module_
             if cardinality != None:
                 protocol['cardinality'] = cardinality
 
+            supports_registration = getattr(api, "SUPPORTS_REGISTRATION", None)
+            if supports_registration != None:
+                protocol['supports_registration'] = supports_registration
+
```
leonardr: Is this currently used or is solely it forward-looking?

aslagle: It's used in https://github.com/NYPL-Simplified/circulation-web/pull/144, to determine whether to display the library registration UI for a collection or service.

api/admin/controller.py
```diff
@@ -2886,3 +2872,98 @@ def _decrypt_shared_secret(self, encryptor, shared_secret):
...
+        """
+        response = do_get(catalog_url)
+        if isinstance(response, ProblemDetail):
+            return response
+        type = response.headers.get("Content-Type")
+        if type and type.startswith('application/opds+json'):
+            # This is an OPDS 2 catalog.
+            catalog = json.loads(response.content)
+            links = catalog.get("links", [])
+            vendor_id = catalog.get("metadata", {}).get("adobe_vendor_id")
```
leonardr: It seems odd to do this in general rather than only when registering with the library registry, but I don't think it will be a problem.

aslagle: Yeah, I agree.

api/admin/controller.py
```diff
@@ -2886,3 +2872,98 @@ def _decrypt_shared_secret(self, encryptor, shared_secret):
...
+                return INTEGRATION_ERROR.detailed(
+                    _("Remote service returned: \"%(problem)s\"", problem=problem.get("detail")))
+            else:
+                return INTEGRATION_ERROR.detailed(
+                    _("Remote service returned: \"%(problem)s\"", problem=response.content))
+            
+        catalog = json.loads(response.content)
+
+        # Since we generated a public key, the catalog should have the short name
+        # and shared secret for Short Client Tokens.
```
leonardr: Change this comment to something more general like "for future authenticated communication with the remote, e.g. through Short Client Tokens or authenticated API requests".

api/controller.py
```diff
@@ -1478,3 +1499,248 @@ def notify(self, loan_id):
...
+            # If we have a link to the content on a remote server, web clients may not
+            # be able to access it if the remote server does not support CORS requests.
+            # We need to fetch the content and return it instead of redirecting to it.
+            try:
+                response = do_get(fulfillment.content_link)
+                status_code = response.status_code
+                headers = dict(response.headers)
+                content = response.content
+            except RemoteIntegrationException, e:
+                return e.as_problem_detail_document(debug=False)
```
leonardr: This is fine for now, but since we have no control over the size of the thing at the other end of the link, I would eventually like to get to the point where the client can signal whether or not it needs us to do this.

aslagle: Sure, and the same applies to regular fulfills.

api/odl.py
leonardr: That does seem a little weird in isolation, but okay.

api/odl.py
```diff
@@ -890,3 +956,509 @@ def _url_for(self, *args, **kwargs):
...
+    def internal_format(self, delivery_mechanism):
+        """Each consolidated copy is only available in one format, so we don't need
+        a mapping to internal formats.
+        """
+        return delivery_mechanism
+
+    def collection(self, _db):
+        return get_one(_db, Collection, id=self.collection_id)
+
+    def _get(self, url, headers=None, patron=None, allowed_response_codes=None):
```
leonardr: There is a decent amount of code in this method and taking `HTTP.get_with_timeout` as an argument would make it easy to test.

leonardr: Or even better, making it a constructor argument so all the other methods can use it. But I haven't looked at your tests yet so don't know how you handle it now.

api/odl.py
```diff
@@ -890,3 +956,509 @@ def _url_for(self, *args, **kwargs):
         del kwargs["_external"]
         return "http://%s?%s" % ("/".join(args), "&".join(["%s=%s" % (key, val) for key, val in kwargs.items()]))
 
+
+class SharedODLAPI(BaseCirculationAPI):
```
leonardr: A general note for later: a lot of the parsing could be handled by a specialized OPDS parser. This was true before, when we were just parsing metadata out of the OPDS, but now that we're doing sophisticated matches against fulfillment mechanisms to find the right URL it seems a little more important.

api/odl.py
```diff
@@ -890,3 +956,509 @@ def _url_for(self, *args, **kwargs):
...
+        holds = _db.query(Hold).join(Hold.license_pool).filter(
+            LicensePool.collection_id==self.collection_id
+        ).filter(
+            Hold.patron==patron
+        )
+        
+        activity = []
+        for loan in loans:
+            info_url = loan.external_identifier
+            response = self._get(info_url, allowed_response_codes=["2xx", "3xx", "400"])
```
leonardr: Not much we can do at this point but this is a downside to 'every loan has a different patron identifier' that I hadn't previously noticed.

aslagle: For holds, it is likely that the hold position will change due to loans expiring or being returned on other circ managers. For loans, though, this would only be necessary if the loan was revoked through the DRM system without using the circ manager's revoke link, or was revoked by an admin on the ODL server.

api/odl.py
```diff
@@ -890,3 +956,509 @@ def _url_for(self, *args, **kwargs):
...
+            if response.status_code == 400:
+                # 400 is returned when the loan has been deleted. Leave this loan out of the result.
```
leonardr: 400 seems weird--I would expect a 404 or 410. What's going on here?

aslagle: It's the NO_ACTIVE_LOAN problem detail.

aslagle: Maybe it doesn't make sense to reuse that when we have a link to a specific loan.

api/odl.py
```diff
@@ -890,3 +956,509 @@ def _url_for(self, *args, **kwargs):
...
+                )
+            )
+        return activity
+
+
+class SharedODLImporter(OPDSImporter):
+    NAME = SharedODLAPI.NAME
+
+    @classmethod
+    def _detail_for_elementtree_entry(cls, parser, entry_tag, feed_url=None):
```
leonardr: This is the sort of thing I was thinking when I wrote above about making `OPDSImporter` handle other uses of OPDS. I think you could reuse this code in the `SharedODLAPI` instead of parsing out links and delivery mechanisms.

api/shared_collection.py
```diff
@@ -0,0 +1,231 @@
...
+            "key": EXTERNAL_LIBRARY_URLS,
+            "label": _("URLs for libraries on other circulation managers that use this collection"),
+            "type": "list",
+            "optional": True,
+        },
+        {
+            "key": Collection.EBOOK_LOAN_DURATION_KEY,
+            "label": _("Ebook Loan Duration for libraries on other circulation managers (in Days)"),
+            "default": Collection.STANDARD_DEFAULT_LOAN_PERIOD,
+            "description": _("When a patron uses SimplyE to borrow an ebook from this collection, SimplyE will ask for a loan that lasts this number of days. This must be equal to or less than the maximum loan duration negotiated with the distributor."),
```
leonardr: I would rephrase this as "When a patron of some other library... their circulation manager will ask for a loan..."

For the bit this is based on it's simpler to use "SimplyE" as shorthand for the circulation manager but here we should be more explicit.

tests/admin/test_controller.py
```diff
@@ -4718,6 +4720,137 @@ def test_library_registrations_post_success(self):
...
+            self.responses.append(MockRequestsResponse(200, content=feed, headers=headers))
+
+            response = self.manager.admin_settings_controller.collection_library_registrations(do_get=self.do_request, do_post=self.do_request)
+            eq_(REMOTE_INTEGRATION_FAILED.uri, response.uri)
+            eq_("The service at collection url did not provide a register link.", response.detail)
+            eq_([collection.external_account_id], self.requests[1:])
+            eq_("failure", ConfigurationSetting.for_library_and_externalintegration(
+                    self._db, "library-registration-status", library, collection.external_integration).value)
+
+    def test_collection_library_registrations_post_success(self):
```
leonardr: You've already written these tests and there's no need to change them, but when refactoring admin controller code (the way you did with `_register_library`) I've found it useful to write a suite of tests for the helper method, and then much simpler tests for the methods that call the helper method, which just verify that it called that method with the right arguments.

api/odl.py
```diff
@@ -268,8 +300,7 @@ def checkin(self, patron, pin, licensepool):
             raise NotCheckedOut()
         elif status == self.ACTIVE_STATUS:
             # This loan has already been fulfilled, so it needs to be returned through the DRM system.
-            # Do nothing.
-            return
+            raise CannotReturn(_("This loan has already been fulfilled, so it must be returned through the DRM system."))
```
leonardr: Is this always the case or does it depend on the DeliveryMechanism of the LicensePoolDeliveryMechanism? That is, if the book was delivered without DRM it should be returnable through the return link.

tests/test_odl.py
```diff
@@ -195,7 +201,10 @@ def test_checkin_already_fulfilled(self):
         })
 
         self.api.queue_response(200, content=lsd)
-        self.api.checkin(self.patron, "pin", self.pool)
+        assert_raises(
+            CannotReturn, self.api.checkin,
+            self.patron, "pin", self.pool,
+        )
```
leonardr: If you agree with what I say in https://github.com/NYPL-Simplified/circulation/pull/879/files#r173530170 then I imagine both cases would be tested here.

api/odl.py
aslagle: I don't think a book would ever have "active" status if it didn't have DRM - the DRM system notifies the distributor when the book is fulfilled and that makes it active. Without DRM the distributor wouldn't know and it would probably stay "ready".

aslagle: I could be wrong though - I'll have to check on this.

api/odl.py
leonardr: Yeah, NO_ACTIVE_LOAN is more like "you tried to do something to a book that required an active loan as a prerequisite, but you don't have an active loan", this is more like "you asked for information about a specific loan that doesn't exist".

leonardr:
I've mentioned some minor changes above; after fixing those I'm OK if you either land it or use this as the basis for a feature branch. Either way there's a next step that's pretty important. This is a public record of the stuff we discussed earlier on Slack.

You're expanding the circulation manager so that it can act as two different types of OPDS servers. For a long time we've had the OPDS server for a library, where users (humans) are authenticated through their library cards. Now we've got the OPDS server for a collection, where users (other circ managers) are authenticated through a shared secret created through a registration process.

In the first OPDS server there is no access to books except through a library. In the second OPDS server the libraries are not directly involved at all. The need to squeeze the second OPDS server into the structure of the first OPDS server is the underlying cause behind a lot of the stuff in here that I think is awkward -- primarily the hardcoded URLs and the need to have a Library object in places where no real library is involved.

I started this a little while ago when I made the crawlable collection feed work in the context of a library which it doesn't actually need. I didn't like this but I didn't see how to change it. Going through this branch I think I've identified the point of leverage: `CirculationManagerAnnotator`.

That's the class that implements all of the logic of the first OPDS server. For purposes of this discussion, that mainly means it controls what the `borrow` links look like. But it also adds a ton of links like 'report', 'recommendations', 'related', 'help', and the annotation links. These links are irrelevant to the second OPDS server, which is not intended for human use.

`CirculationManager` should be refactored and  the second OPDS server should have its own annotator based on the refactored superclass. This new annotator would serve `borrow` links that point to the `shared_collection_borrow` route, where the other annotator serves `borrow` links that point to `borrow`. (This is how you get rid of the hard-coded URLs).

Basically the `OPDSAnnotator` subclass controls which special features are available from a specific OPDS server. We've seen this before, but we haven't seen two different 'OPDS servers' within a single Python application server.

Once you have that done I think everything else will fall into place. You shouldn't need to pass in a Library to the new `OPDSAnnotator`, which means you shouldn't need to look up the current library in the controller, which means you should be able to use `@route` instead of `@library_route` in the routes and we can have `/collections/{collection}/crawlable` instead of `/SOMELIBRARY/collections/{collection}/crawlable`.

Finally, there is no `place_hold` route, so I don't think you need a `shared_collection_place_hold` route or the code behind it -- `shared_collection_borrow` should do the job of both. The client knows that someone wants to read a book, but they don't necessarily have up-to-date availability information, so it's the same HTTP request and code path whether the desire to read a book becomes an active loan or a hold.

api/odl.py
aslagle: I can't reuse much since the logic is different for borrow links and generic acquisition links, but I can move more of the code to this class.

leonardr: This is good for now but I think we could eventually move to a reusable OPDS parser that turns the XML (or JSON) into an abstract data structure.

api/odl.py
aslagle: I think maybe the best thing to do here would be to always try the return link for "active" or "ready" license if there is one, but raise an exception if the status doesn't change.

aslagle:
This and all the related branches are ready for another review.

api/controller.py
```diff
@@ -1571,14 +1568,20 @@ def borrow(self, collection_name, identifier_type, identifier, hold_id):
...
-                return pools
-            pools = [p for p in pools if p.collection == collection]
+            pools = self._db.query(LicensePool).join(LicensePool.collection).join(
+                LicensePool.identifier).filter(
+                    Identifier.type==identifier_type
+                ).filter(
+                    Identifier.identifier==identifier
+                ).filter(
+                    Collection.id==collection.id
+                ).all()
```
leonardr: This is a pretty minor thing, but is it necessary to join against Collection rather than filtering against collection_id.

api/odl.py
```diff
@@ -1085,15 +1085,32 @@ def checkout(self, patron, pin, licensepool, internal_format):
...
+                licensepool.identifier.type,
+                licensepool.identifier.identifier,
+                start,
+                end,
+                external_identifier=external_identifier
+            )
+        else:
+            # We tried to borrow this book but it wasn't available,
+            # so we got a hold.
+            position = entry.get("opds_holds", {}).get("position")
```
leonardr: Is this the only other option? Could there be a server that just doesn't support holds? Or in that case would the remote raise an exception before execution go to this point?

aslagle: I haven't provided a way to disable holds for these, but it might be a good idea to explicitly check for 'reserved' status just in case.

api/opds.py
```diff
@@ -938,8 +921,163 @@ def add_authentication_document_link(self, feed_obj):
...
+            url = self.url_for(
+                'shared_collection_loan_info',
+                collection_name=self.collection.name,
+                loan_id=active_loan.id,
+                _external=True)
+            kw = dict(href=url, rel='self')
+            info_link_tag = OPDSFeed.makeelement("link", **kw)
+            info_links.append(info_link_tag)
+
+        if active_hold and active_hold.integration_client:
```
leonardr: Why check that integration_client is set for holds but not loans?

api/opds.py
```diff
@@ -938,8 +921,163 @@ def add_authentication_document_link(self, feed_obj):
...
+            # If we can borrow this book, add this delivery mechanism
+            # to the borrow link as an <opds:indirectAcquisition>.
+            if format_types:
+                indirect_acquisition = AcquisitionFeed.indirect_acquisition(
+                    format_types
+                )
+                indirect_acquisitions.append(indirect_acquisition)
+
+        if not indirect_acquisitions:
+            raise UnfulfillableWork()
```
leonardr: I agree that this is the right behavior but I was surprised enough that I had to look it up. Could you add an explanatory comment that we want to cancel the creation of the OPDS entry altogether since there's no way to actually get the book?

leonardr: I still think the comment would be good but you didn't add this code, just moved it during the refactor. I probably wrote it originally.

api/shared_collection.py
```diff
@@ -27,15 +26,12 @@ class SharedCollectionAPI(object):
     collection types.
     """
 
-    def __init__(self, _db, library, api_map=None):
+    def __init__(self, _db, api_map=None):
         """Constructor.
 
         :param _db: A database session (probably a scoped session, which is
             why we can't derive it from `library`).
```
leonardr: This comment is now moot.

api/shared_collection.py
```diff
@@ -45,16 +41,15 @@ def __init__(self, _db, library, api_map=None):
            calls, we only instantiate one CirculationAPI per library,
            and keep them around as long as possible.
```
leonardr: This is copy-and-pasted. I believe the right thing to say is that we only instantiate one SharedCollectionAPI and try to keep it around as long as possible.

api/opds.py
aslagle: Left over from when there was one annotator that might have a patron or a client, I'll delete it.

Issue #884, opened 2018-03-13
Recommendation feed is unbearably slow.


leonardr:
The recommendation feed  gets a bunch of identifiers and then builds a feed out of works equivalent to those identifiers. It uses the fn_recursive_equivalents function on Identifier.id to do that.

The problem is that when you join that against mv_works_for_lanes, the restriction on Identifier.id isn't used to eliminate rows from mv_works_for_lanes based on its identifier_id. This causes a full table scan of the materialized view, which is bad enough to bring down the site if a lot of people are using it.

We've run into something similar for work id and license pool id, and the solution was to add a clause to apply the restriction twice, once on the materialized view and once on the table that was the source for that data.

It's also possible the answer is to use the restricted Identifier.id in the join against the materialized view, rather than to apply the same restriction twice.

PR #885, opened 2018-03-13
Speed up recommendation lanes

leonardr:
This branch is the circulation side of the fix to https://github.com/NYPL-Simplified/circulation/issues/884. In the query to build the recommendation feed, the identifier IDs of the recommendations are used to restrict `mv_works_for_lanes.identifier_id`, immediately cutting down the number of materialized view rows that must be scanned through. Previously, this query restricted `Identifier.id`, and then we very slowly scanned the materialized view to match its rows against the matching identifiers.


api/lanes.py
```diff
@@ -890,10 +890,11 @@ def fetch_recommendations(self, _db):
     def apply_filters(self, _db, qu, facets, pagination, featured=False):
         if not self.recommendations:
             return None
-
+        from core.model import MaterializedWorkWithGenre as wm
         qu = qu.join(LicensePool.identifier)
         qu = Work.from_identifiers(
-            _db, self.recommendations, base_query=qu
+            _db, self.recommendations, base_query=qu,
+            identifier_id_field=mw.identifier_id
```
courte: The variable name is off here. It should be `wm`. Or the one above should be `mw`.

Issue #886, opened 2018-03-13
Axis 360 reaper broke and we didn't notice


leonardr:
The Axis 360 reaper broke when we moved to a multi-collection architecture. We didn't notice because the reaper rarely makes a difference and we don't have test coverage of the reaper code.

PR #887, opened 2018-03-13
Fix and test Axis360API.update_licensepools_for_identifiers

leonardr:
This branch fixes https://github.com/NYPL-Simplified/circulation/issues/886 and adds test coverage so the problem won't recur.


api/axis.py
```diff
@@ -199,45 +199,61 @@ def update_licensepools_for_identifiers(self, identifiers):
         The book's LicensePool will be updated with current
         circulation information.
         """
-        identifier_strings = self.create_identifier_strings(identifiers)
-        response = self.availability(title_ids=identifier_strings)
         collection = self.collection
```
courte: It doesn't look like this `collection` is needed anymore, either.

api/axis.py
```diff
@@ -199,45 +199,61 @@ def update_licensepools_for_identifiers(self, identifiers):
...
+        response = self.availability(title_ids=identifier_strings)
+        parser = BibliographicParser(self.collection)
+        return parser.process_all(response.content)
+
+    def _reap(self, identifier):
+        """Update our local circulation information to reflect the fact that
+        the identified book has been removed from the remote
+        collection.
+        """
+        collection = self.collection
```
courte: I think this is a redundant variable assignment left from earlier, but the variable is then used later in the method.

leonardr: Yeah, I'll leave this one in because `collection` is a property that hits the database.

PR #891, opened 2018-03-19
Fixed admin session on first request after an app restart.

aslagle:
This fixes a bug where if an admin made the first request after the circ manager started, it would give them a null session and think they were logged out. This mainly affects developers working on the admin interface, but could also happen occasionally in a real app.


aslagle:
The session object contains information extracted from the current request's cookies, so it can't be shared across requests. Flask handles calling open_session for every request, but on the first request it happens before `before_first_request`.

leonardr:
OK, got it.

PR #898, opened 2018-03-21
Bring the loan and hold reapers into the generic reaper system

leonardr:
This branch removes `LoanAndHoldReaper`, replacing it with two `ReaperMonitor`s that fit into the general reaper system. The `loan_reaper` script is replaced with the `database_reaper` script.

The reapers do not affect collections where the circulation manager is the source of truth w/r/t loans and holds. Right now that means collections based on ODL or OPDS For Distributors.

### DIFF
--- a/api/monitor.py
+++ b/api/monitor.py
@@ -7,17 +7,23 @@ from lxml import etree
 from nose.tools import set_trace
 from StringIO import StringIO
 
-from sqlalchemy import or_
+from sqlalchemy import (
+    and_,
+    or_,
+)
 
 from core.monitor import (
     CollectionMonitor,
     EditionSweepMonitor,
+    ReaperMonitor,
 )
 from core.model import (
     DataSource,
     Edition,
+    Hold,
     Identifier,
     LicensePool,
+    Loan,
 )
 from core.opds_import import (
     MetadataWranglerOPDSLookup,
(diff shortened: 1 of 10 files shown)


api/opds_for_distributors.py
```diff
@@ -143,13 +143,23 @@ def checkin(self, patron, pin, licensepool):
             pass
 
     def checkout(self, patron, pin, licensepool, internal_format):
+        now = datetime.datetime.utcnow()
```
leonardr: I also started using utcnow() here instead of now().

aslagle:
I didn't notice these scripts before and wrote a separate Hold reaper for ODL. But now that I'm looking I think this might be a problem. 365 days doesn't seem all that long to have a popular book on hold, and for ODL you can't just sync it from somewhere else if it's accidentally deleted.

leonardr:
This is because in some cases the Loan/Hold object is the source of truth? Can we explicitly mark the source-of-truth items so they don't get reaped?

aslagle:
Maybe we could filter out things that have an external identifier?

api/monitor.py
```diff
@@ -228,3 +234,49 @@ def get_identifiers(self, url=None):
...
+    def where_clause(self):
+        """Find loans/holds that have either expired, or that were created a
+        long time ago and have no definite end date.
+        """
+        cls = self.MODEL_CLASS
+        end_field = getattr(cls, 'end')
+        start_field = getattr(cls, 'start')
+
+        now = datetime.datetime.utcnow()
+        expired = end_field < now
```
aslagle: I think for a Hold the end date is an estimate until it's available to check out, so an end date slightly in the past doesn't necessarily mean it's expired.

leonardr:
We could use another subquery to make sure license_pool_id is not in the list of license pools whose collection_id is one of the 'source of truth' collections. That might have bad performance characteristics though. The subqueries were the only way I could find to do all the deletes in one SQL statement rather than having to iterate in batches, but it might be faster in the worst case to iterate in batches.

aslagle:
It could be added to the existing subquery. It shouldn't be much worse than filtering on open_access, should it?

leonardr:
It depends on the collection. In a library where loans are being given out for open-access books (as opposed to people just downloading the books), it's likely that open-access books are mixed with non-open access, and there will be relatively few open-access loans. In a library that hosts a source-of-truth collection it seems likely that a significant number of the loans will be for that collection. But since the subquery is a query against Loan, not against LicensePool, it should be manageable.

Issue #792, opened 2017-11-30
Top-level search should search `Accept` languages only


leonardr:
We have a low-level but persistent problem where someone searches for a book and ends up borrowing the book they want but in the wrong language. The title of a book does not always reflect the language of the text. See https://nypl.overdrive.com/media/647767 for an example.

This is only a problem when someone searches from the top level of a collection, because once they go into a lane they're locked into a specific language. Perhaps when handling top-level searches we should only search the languages mentioned in the `Accept` header.

PR #907, opened 2018-03-27
Audiobook admin search

### LINKED ISSUE #895: Allow searching for specific media type independently of the lane's media type
The admin interface uses the standard `/search` endpoint to search a library's top-level lane. Currently this only searches ebooks. A different issue will cover changing the top-level lane to include all types of media.

The search controller needs to be changed further to allow the caller to pass in a specific type of media it wants to search. This goes beyond what OpenSearch allows for, but it's not a big deal and doesn't open up the full "advanced search" can of worms, because the URLs will only be used in the admin interface and in direct links where the context will be clear to the user.

The client should identify the media type it wants to search for using the schema.org URIs we've adopted: http://bib.schema.org/Audiobook and http://schema.org/EBook. If no media type is specified, all media available from the selected lane will be searched.

EdwinGuzman:
Fixes #895 and #792 
* Uses `accept_types` to parse `Accept-Language` header and use it as a param when searching.
* Reads query for media type and passes it when searching.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -383,9 +383,16 @@ class CirculationManager(object):
             library = lane.get_library(self._db)
         else:
             library = flask.request.library
+
+        # Some features are only available if a patron authentication
+        # mechanism is set up for this library.
+        authenticator = self.auth.library_authenticators.get(library.short_name)
+        auth_supported = authenticator.supports_patron_authentication
         return LibraryAnnotator(
             self.circulation_apis[library.id], lane, library,
-            top_level_title='All Books', *args, **kwargs
+            top_level_title='All Books',
+            library_supports_patron_authentication=auth_supported,
+            *args, **kwargs
         )
 
     @property
--- a/api/controller.py
+++ b/api/controller.py
@@ -718,6 +718,7 @@ class OPDSFeedController(CirculationManagerController):
         if isinstance(lane, ProblemDetail):
             return lane
         query = flask.request.args.get('q')
+        media = flask.request.args.get('media')
         library_short_name = flask.request.library.short_name
         this_url = self.url_for(
             'lane_search', lane_identifier=lane_identifier,
@@ -732,13 +733,17 @@ class OPDSFeedController(CirculationManagerController):
             return pagination
 
         # Run a search.
+        if media:
+            media_url = "&media=" + urllib.quote(media.encode("utf8"))
+        else:
+            media_url = ''
         this_url += "?q=" + urllib.quote(query.encode("utf8"))
         annotator = self.manager.annotator(lane)
         info = OpenSearchDocument.search_info(lane)
         opds_feed = AcquisitionFeed.search(
             _db=self._db, title=info['name'],
             url=this_url, lane=lane, search_engine=self.manager.external_search,
-            query=query, annotator=annotator, pagination=pagination,
+            query=query, media=media, annotator=annotator, pagination=pagination,
         )
         return feed_response(opds_feed)
 
--- a/api/controller.py
+++ b/api/controller.py
@@ -74,6 +74,7 @@ from core.model import (
 from core.opds import (
     AcquisitionFeed,
 )
+from core.util import LanguageCodes
 from core.util.opds_writer import (
      OPDSFeed,
 )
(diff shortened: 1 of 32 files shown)


api/controller.py
```diff
@@ -727,18 +729,30 @@ def search(self, lane_identifier):
             # Send the search form
             return OpenSearchDocument.for_lane(lane, this_url)
 
+        language_header = flask.request.headers.get("Accept-Language")
+        if language_header:
+            languages = parse_header(language_header)
+            languages = map(lambda x: str(x), languages)
```
leonardr: `map(str, languages)` should do the same job.

api/controller.py
```diff
@@ -727,18 +729,30 @@ def search(self, lane_identifier):
...
         pagination = load_pagination_from_request(default_size=Pagination.DEFAULT_SEARCH_SIZE)
         if isinstance(pagination, ProblemDetail):
             return pagination
 
         # Run a search.
-        this_url += "?q=" + urllib.quote(query.encode("utf8"))
+        if media:
+            media_url = "&media=" + urllib.quote(media.encode("utf8"))
+        else:
+            media_url = ''
```
leonardr: I'm on the fence about whether, having loaded the initial set of languages from the request, that language set should be propagated to future pages through the search URL. Right now if someone loads page 1 of search results, then switches their language settings and follows the 'next' link, they will get the second page search results from their new language settings, rather than the second page of the search they originally ran.

In real life this difference doesn't matter, so I wouldn't put too much work into fixing this, but in general I think it's important to make the 'next' link explicitly propagate all the values that went into the original query, even values that didn't originally come from the query string.

aslagle:
I'd just like to point out this piece of localization code that might be able to tie into this somehow: https://github.com/NYPL-Simplified/circulation/blob/master/api/routes.py#L70

api/controller.py
```diff
@@ -732,7 +733,9 @@ def search(self, lane_identifier):
         language_header = flask.request.headers.get("Accept-Language")
         if language_header:
             languages = parse_header(language_header)
-            languages = map(lambda x: str(x), languages)
+            languages = map(str, languages)
+            languages = map(LanguageCodes.iso_639_2_for_locale, languages)
+            languages = [l for l in languages if l]
```
EdwinGuzman: Is there a better way to do the previous two lines? The `Accept-Language` header after mapping it to `str` was `['en-US', 'en; q=0.90']`, but running that through `iso_639_2_for_locale` turned the second value into `None` which raised a parsing error in a later method.

api/controller.py
EdwinGuzman: Updated.

EdwinGuzman:
@aslagle It looks like I could use that function to get the languages and pass that value as a param to `app.manager.opds_feeds.search(lane_identifier)` in `routes.py`, instead of parsing the Accept-Language header in the controller. Does that seem right?

aslagle:
@EdwinGuzman I wanted you to be aware of that function since it also uses languages from the header, but it sets the language to the best match among languages we have translations for, which is currently only English. For search you don't need to limit it based on whether we can localize to that language. I'm not sure if there's any code you could share at this point.

api/controller.py
leonardr: parse_header returns a list of AcceptableType objects. You can access the language part as `AcceptableType.mime_type`. That's not a great name because parse_header was designed to parse the `Accept` header, but the format is the same, with the language code standing in for the MIME type.

EdwinGuzman:
@aslagle got it, thanks.
@leonardr added a test in `TestFeedController.test_search`.

api/controller.py
```diff
@@ -725,6 +727,7 @@ def search(self, lane_identifier):
         if isinstance(lane, ProblemDetail):
             return lane
         query = flask.request.args.get('q')
+        media = flask.request.args.get('media')
```
leonardr: This is where we need to convert between the schema.org URIs (e.g. "http://bib.schema.org/Audiobook") and the enumerated type values that `search` can handle (e.g. "audio"). The lookup table is `Edition.additional_type_to_medium`. If the incoming value for `media` isn't found in that lookup table, you should send an INVALID_INPUT problem detail document.

Sorry, I didn't remember this until I saw you pass in "audiobooks" in the test. That value should be rejected with a problem detail, and passing in "http://bib.schema.org/Audiobook" should convert to `Edition.AUDIO_MEDIUM` and pass that into your mock `search` method.

PR #908, opened 2018-03-27
Replace "Other Languages" lane with a larger "World Languages" lane

leonardr:
This branch changes the default lane setup with respect to language collections other than the "large" sized ones. Previously, "small" sized collections got one top-level lane apiece, and all the "tiny" sized collections were kept in a lane called "Other Languages". Now all small and tiny collections are kept in a lane called "World Languages".

Depending on the library, this can dramatically reduce the size of the top-level OPDS feed and the number of images that need to be downloaded, most of which will not be viewed.

This could be improved further by better organizing the languages within the "World Languages" lane. It's tough to figure out a general way to do this, though, because there are two competing desires: you want to put the most commonly-accessed collections first but you also want to group the languages in some useful way.

At one NYPL branch library languages are grouped by region. The shelves look like East Asia | Western Europe (including Spanish) | Eastern Europe | India. Within a shelf, languages are ordered alphabetically by the first three letters of their English name.

### DIFF
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -107,23 +107,24 @@ def create_default_lanes(_db, library):
 
     The database will have the following top-level lanes for
     each large-collection:
-    'Adult Fiction', 'Adult Nonfiction', 'Young Adult Fiction',
-    'Young Adult Nonfiction', and 'Children'.
+    'Adult Fiction', 'Adult Nonfiction', 'Young Adult',
+    and 'Children'.
     Each lane contains additional sublanes.
     If an NYT integration is configured, there will also be a
     'Best Sellers' top-level lane.
 
-    The database will also have a top-level lane named after each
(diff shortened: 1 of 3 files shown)


api/lanes.py
```diff
@@ -107,23 +107,24 @@ def create_default_lanes(_db, library):
 
     The database will have the following top-level lanes for
     each large-collection:
-    'Adult Fiction', 'Adult Nonfiction', 'Young Adult Fiction',
-    'Young Adult Nonfiction', and 'Children'.
+    'Adult Fiction', 'Adult Nonfiction', 'Young Adult',
+    and 'Children'.
```
aslagle: I think the old version of this comment is correct, don't see any changes related to this.

leonardr: You're right, this is a proposed change I decided not to make (yet?)

PR #918, opened 2018-03-30
Mitigate issue where licensepools have no works attached to them

jonathangreen:
## Issue

I've been troubleshooting an issue with @leonardr where some books from Overdrive were not coming up in search results in the Circulation manager. Some sleuthing showed that many of the `licensepools` in this circulation manger were missing `works` (something like 11,870 of them). Looking at my other circulation managers, none had as large a problem, but still they all had a few `licensepools` without `works`. 

## Solution

I was not able to reproduce the problem where a license pool ended up without a work associated. Assuming its a transient problem perhaps caused by issues with the Overdrive API, this PR tries to mitigate the problem by making sure a `work` is created in `update_licensepool` if there isn't already one. 

With this patch applied when the Overdrive Reaper ran it fixed all of the `licensepools` missing `works` and all of the books became available in the circulation manger.

### DIFF
--- a/tests/test_overdrive.py
+++ b/tests/test_overdrive.py
@@ -1,6 +1,6 @@
 # encoding: utf-8
 from nose.tools import (
-    set_trace, eq_,
+    set_trace, eq_, ok_,
     assert_raises,
 )
 import pkgutil
@@ -567,6 +567,8 @@ class TestOverdriveAPI(OverdriveAPITest):
 
         self.api.queue_response(200, content=availability)
         self.api.queue_response(200, content=bibliographic)
+        self.api.queue_response(200, content=availability)
+        self.api.queue_response(200, content=bibliographic)
 
         # Now we're ready. When we call update_licensepool, the
         # OverdriveAPI will retrieve the availability information,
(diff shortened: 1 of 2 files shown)


tests/test_overdrive.py
```diff
@@ -565,6 +565,8 @@ def test_update_licensepool_provides_bibliographic_coverage(self):
         availability['id'] = identifier.identifier
         bibliographic['id'] = identifier.identifier
 
+        self.api.queue_response(200, content=availability)
+        self.api.queue_response(200, content=bibliographic)
```
leonardr: This test would read better if you queued these responses just before calling `update_licensepool` on line 608.

leonardr:
Just make a minor change to the test and I'll merge this.

leonardr:
Oops, did not mean to close this.

leonardr:
For the record, "transient problem with the Overdrive API" is my best guess, but after looking at the code I couldn't see a scenario where the BibliographicCoverageProvider completed successfully but there was no work, which is what would have had to happen. The other possibility is that a work was created but, for some reason, deleted later.

jonathangreen:
@leonardr updated with code review feedback

leonardr:
Thanks for the branch!

PR #920, opened 2018-04-01
Admin auth methods are no longer mutually exclusive.

aslagle:
This branch changes the admin authentication code so that both passwords and Google OAuth can be used on the same circ manager, and admins logging in with google can be managed individually instead of by domain.

Visiting an admin route will show a sign in page with a link to sign in with google, a form for logging in with a password, or both, depending on configuration.

It's no longer required to whitelist domains for logging in with google. If you don't want to whitelist an entire domain, you can add individual admins instead, and they will be able to choose whether to log in with google or their password if their email is a google account. Any admins that already existed will still be able to log in with google even if they don't have a password.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -156,30 +156,43 @@ class AdminController(object):
         self.cdn_url_for = self.manager.cdn_url_for
 
     @property
-    def auth(self):
+    def auth_providers(self):
+        auth_providers = []
         auth_service = ExternalIntegration.admin_authentication(self._db)
         if auth_service and auth_service.protocol == ExternalIntegration.GOOGLE_OAUTH:
-            return GoogleOAuthAdminAuthenticationProvider(
+            auth_providers.append(GoogleOAuthAdminAuthenticationProvider(
                 auth_service,
                 self.url_for('google_auth_callback'),
                 test_mode=self.manager.testing,
-            )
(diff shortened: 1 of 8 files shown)


api/admin/controller.py
```diff
@@ -156,30 +156,43 @@ def __init__(self, manager):
         self.cdn_url_for = self.manager.cdn_url_for
 
     @property
-    def auth(self):
+    def auth_providers(self):
```
leonardr: Can you call this `admin_auth_providers` to distinguish it from patron auth?

api/admin/controller.py
```diff
@@ -291,87 +305,70 @@ class SignInController(AdminController):
...
+        if not auth:
             return ADMIN_AUTH_MECHANISM_NOT_CONFIGURED
 
-        admin_details, redirect_url = self.auth.callback(flask.request.args)
+        admin_details, redirect_url = auth.callback(flask.request.args)
         if isinstance(admin_details, ProblemDetail):
             return self.error_response(admin_details)
 
-        if not self.staff_email(admin_details['email']):
+        if not auth.staff_email(self._db, admin_details['email']):
```
leonardr: Since you're moving `staff_email` into the Authenticator class you might have an opportunity to streamline this method by having `callback` also check `staff_email` and return INVALID_ADMIN_CREDENTIALS if there's a problem.

PR #923, opened 2018-04-04
Handle the (unlikely) case where a service exists in the database but there is no configuration for its protocol.

leonardr:
This branch fixes a problem I saw in production due to handmade database changes. If a service exists in the database, but the code contains no definition for that service, _get_integration_info will raise an exception and the corresponding tab won't load in the admin interface.

This happened in production because I temporarily disabled an integration by renaming its protocol to "x (disabled)".


leonardr:
It turns out https://github.com/NYPL-Simplified/server_core/pull/841 was also causing this problem.

PR #927, opened 2018-04-09
Exposing an entry's identifier to use as a link for the UI.

EdwinGuzman:
This is to get a working url for each item in the entry list for the list editor. That url is updated in the UI to make it link to each item's detail page.


EdwinGuzman:
Thanks.

PR #929, opened 2018-04-11
Media entrypoints

leonardr:
This branch changes the OPDS generation for `groups`, `feed`, and `search` so that any requested facets are passed in.

* The currently selected entry point is propagated to related feeds, like the subsequent pages of a list.
* If the currently selected WorkList has multiple entry points, a facet group is created that links to the same page with different entry points.

It suspect there are still some links that need to be changed to propagate the currently selected entry point, but my hope is that doing this through the facets rather than through a completely new mechanism will limit the number of problems of this sort.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -753,8 +753,6 @@ class OPDSFeedController(CirculationManagerController):
         else:
             languages = None
 
-        # Create a function that, when called, generates a URL to the
-        # search controller.
         facets = load_facets_from_request(
             worklist=lane, base_class=SearchFacets
         )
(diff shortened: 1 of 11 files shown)


api/controller.py
```diff
@@ -759,26 +786,18 @@ def search(self, lane_identifier):
...
         # Run a search.
-        if media:
-            media_url = "&media=" + urllib.quote(media.encode("utf8"))
-        else:
-            media_url = ""
-        if languages:
-            languages_url = "&" + urllib.urlencode(dict(language=languages), doseq=True)
-        else:
-            languages_url = ""
-
```
leonardr: This whole thing is replaced by what `url_for` does with keyword arguments.

leonardr:
That's fine, I'm in no hurry to merge this.

PR #932, opened 2018-04-12
Removing code that grabbed media query in favor of entry points.

### LINKED ISSUE #931: When building lists, facet search based on available entry points, not media
Currently when you're building a custom list in the admin interface, you have the choice to search for titles in different media. The idea was that the end-user interface will eventually _also_ feature a choice of media.

Since then, the end-user design has changed to use EntryPoints, which may or may not be directly associated with media. The entry points we have now, "Books" and "Audio", do happen to be directly associated with media, but future entry points might be more like "Kids" or "Research", and include media of different types.

Someone building a custom list is likely to have a specific EntryPoint in mind, so the possibilities available should be the library's enabled EntryPoints rather than a list of media. You can get the currently enabled EntryPoints for a library with `Library.entrypoints`, use that to generate the HTML form, and have the value submitted under the name `FacetConstants.ENTRY_POINT_FACET_GROUP_NAME` (which is "entrypoint").

In `OPDSFeedController.search()`, then, there is no longer any need to gather the `media` argument from the Flask request or pass it in to `AcquisitionFeed.search`. The EntryPoint is pulled out of the query string in `load_facets_from_request` and passed in as a `SearchFacets` object.

This change would be a branch based on https://github.com/NYPL-Simplified/circulation/pull/929.

EdwinGuzman:
Fixes #931.

### DIFF
--- a/api/admin/package.json
+++ b/api/admin/package.json
@@ -9,6 +9,6 @@
   "author": "NYPL",
   "license": "Apache-2.0",
   "dependencies": {
-    "simplified-circulation-web": "0.0.63"
+    "simplified-circulation-web": "0.0.64"
   }
 }
--- a/api/controller.py
+++ b/api/controller.py
@@ -741,7 +741,6 @@ class OPDSFeedController(CirculationManagerController):
         if isinstance(lane, ProblemDetail):
             return lane
         query = flask.request.args.get('q')
-        media = flask.request.args.get('media')
         library_short_name = flask.request.library.short_name
 
         language_header = flask.request.headers.get("Accept-Language")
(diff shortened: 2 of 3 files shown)


aslagle:
This is probably from Leonard's previous branch, but while testing this I noticed that "Books" shows as checked on the library settings page when there are no entry points, or when it is the only entry point. If you uncheck it and save, it just stays checked. So there are two configurations that look identical on the library settings page but produce a different UI on the list page.

leonardr:
Yeah, that is my branch's fault. If you uncheck all the tabs you go back to the default value which is only books.

PR #935, opened 2018-04-19
Admin roles

aslagle:
This branch checks an admin's roles in admin interface controllers and allows some admins to configure admin roles. I'm not done with the front-end yet but I'm ready for some preliminary feedback on this part.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -28,6 +28,7 @@ from core.model import (
     get_one,
     get_one_or_create,
     Admin,
+    AdminRole,
     CirculationEvent,
     Classification,
     Collection,
@@ -2240,9 +2241,14 @@ class SettingsController(CirculationManagerController):
     def individual_admins(self):
         if flask.request.method == 'GET':
             admins = []
-            admins_with_password = Admin.with_password(self._db)
-            if admins_with_password.count() != 0:
-                admins=[dict(email=admin.email) for admin in admins_with_password]
+            for admin in self._db.query(Admin):
+                roles = []
+                for role in admin.roles:
+                    if role.library:
+                        roles.append(dict(role=role.role, library=role.library.short_name))
+                    else:
+                        roles.append(dict(role=role.role))
+                admins.append(dict(email=admin.email, roles=roles))
 
             return dict(
                 individualAdmins=admins,
@@ -2250,17 +2256,45 @@ class SettingsController(CirculationManagerController):
 
         email = flask.request.form.get("email")
         password = flask.request.form.get("password")
+        roles = flask.request.form.get("roles")
 
-        if not email or not password:
+        if not email:
             return INCOMPLETE_CONFIGURATION
 
+        if roles:
+            roles = json.loads(roles)
+        else:
+            roles = []
+
         admin, is_new = get_one_or_create(self._db, Admin, email=email)
-        admin.password = password
+        if password:
+            admin.password = password
         try:
             self._db.flush()
         except ProgrammingError as e:
             self._db.rollback()
             return MISSING_PGCRYPTO_EXTENSION
+
+        old_roles = admin.roles
+        for role in roles:
+            if role.get("role") not in AdminRole.ROLES:
+                self._db.rollback()
+                return UNKNOWN_ROLE
+            library = None
+            library_short_name = role.get("library")
+            if library_short_name:
+                library = Library.lookup(self._db, library_short_name)
+                if not library:
+                    self._db.rollback()
+                    return LIBRARY_NOT_FOUND.detailed(_("Library \"%(short_name)s\" does not exist.", short_name=library_short_name))
+            admin.add_role(role.get("role"), library)
+        new_roles = set((role.get("role"), role.get("library")) for role in roles)
+        for role in old_roles:
+            library = None
+            if role.library:
+                library = role.library.short_name
+            if not (role.role, library) in new_roles:
+                admin.remove_role(role.role, role.library)
 
         if is_new:
             return Response(unicode(admin.email), 201)
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -63,6 +63,7 @@ from core.util.problem_detail import (
 from core.mirror import MirrorUploader
 from core.util.http import HTTP
 from problem_details import *
+from exceptions import *
 from core.util import (
     fast_query_count,
     LanguageCodes,
@@ -256,27 +257,22 @@ class AdminCirculationManagerController(CirculationManagerController):
     def require_system_admin(self):
         admin = flask.request.admin
         if not admin or not admin.is_system_admin():
-            return ADMIN_NOT_AUTHORIZED
-        return True
+            raise AdminNotAuthorized()
 
     def require_sitewide_library_manager(self):
         admin = flask.request.admin
         if not admin or not admin.is_sitewide_library_manager():
-            return ADMIN_NOT_AUTHORIZED
-        return True
+            raise AdminNotAuthorized()
 
     def require_library_manager(self, library):
         admin = flask.request.admin
         if not admin or not admin.is_library_manager(library):
-            return ADMIN_NOT_AUTHORIZED
-        return True
+            raise AdminNotAuthorized()
 
     def require_librarian(self, library):
         admin = flask.request.admin
         if not admin or not admin.is_librarian(library):
-            return ADMIN_NOT_AUTHORIZED
-        return True
-
+            raise AdminNotAuthorized()
 
 class ViewController(AdminController):
     def __call__(self, collection, book, path=None):
@@ -413,9 +409,7 @@ class WorkController(AdminCirculationManagerController):
 
         This includes relevant links for editing the book.
         """
-        authorized = self.require_librarian(flask.request.library)
-        if isinstance(authorized, ProblemDetail):
-            return authorized
+        self.require_librarian(flask.request.library)
 
         work = self.load_work(flask.request.library, identifier_type, identifier)
         if isinstance(work, ProblemDetail):
@@ -430,9 +424,7 @@ class WorkController(AdminCirculationManagerController):
 
     def complaints(self, identifier_type, identifier):
         """Return detailed complaint information for admins."""
-        authorized = self.require_librarian(flask.request.library)
-        if isinstance(authorized, ProblemDetail):
-            return authorized
+        self.require_librarian(flask.request.library)
 
         work = self.load_work(flask.request.library, identifier_type, identifier)
         if isinstance(work, ProblemDetail):
@@ -496,9 +488,7 @@ class WorkController(AdminCirculationManagerController):
 
     def edit(self, identifier_type, identifier):
         """Edit a work's metadata."""
-        authorized = self.require_librarian(flask.request.library)
-        if isinstance(authorized, ProblemDetail):
-            return authorized
+        self.require_librarian(flask.request.library)
 
         # TODO: It would be nice to use the metadata layer for this, but
         # this code handles empty values differently than other metadata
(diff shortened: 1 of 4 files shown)


api/admin/controller.py
```diff
@@ -247,6 +251,28 @@ def generate_csrf_token(self):
         """Generate a random CSRF token."""
         return base64.b64encode(os.urandom(24))
 
+class AdminCirculationManagerController(CirculationManagerController):
+    """Parent class that provides methods for verifying an admin's roles."""
+
+    def require_system_admin(self):
+        admin = flask.request.admin
+        if not admin or not admin.is_system_admin():
+            raise AdminNotAuthorized()
```
aslagle: It might be better to use decorators for some of this, but some controllers have more complex logic around who can do things. I like having it at the controller-level rather than the route because it's easier to test.

api/admin/controller.py
```diff
@@ -383,6 +409,7 @@ def details(self, identifier_type, identifier):
 
         This includes relevant links for editing the book.
         """
+        self.require_librarian(flask.request.library)
```
aslagle: I'd also like to move in the direction of raising more exceptions that can be converted to problem details rather than returning problem details, I think it makes the code cleaner.

api/admin/controller.py
```diff
@@ -2240,42 +2330,104 @@ def admin_auth_service(self, protocol):
...
+                if not library:
+                    self._db.rollback()
+                    return LIBRARY_NOT_FOUND.detailed(_("Library \"%(short_name)s\" does not exist.", short_name=library_short_name))
+            if library:
+                self.require_library_manager(library)
+            elif role.get("role") == AdminRole.SYSTEM_ADMIN:
+                self.require_system_admin()
+            else:
+                self.require_sitewide_library_manager()
+            admin.add_role(role.get("role"), library)
```
aslagle: I have a lot of logic on the client to adjust all the roles as things are added or removed. For example, if you are a librarian of a few libraries and you become a sitewide librarian, it removes the individual librarian roles. I wasn't sure if it was necessary to handle that on the server as well. If everyone's using the client to set roles it won't be necessary.

tests/admin/test_controller.py
```diff
@@ -2626,46 +2921,67 @@ def test_collections_post_errors(self):
         )
 
         with self.app.test_request_context("/", method="POST"):
+            self._db.begin_nested()
```
aslagle: I was thinking I could use savepoints to reduce the amount of duplicated setup code in these tests. However, it's not working the way I expected and things aren't getting rolled back at the end of the test. I will probably revert all of the changes related to this and repeatedly create new admins instead, but I'd like to know if you have other ideas.

api/admin/controller.py
leonardr: What if a ProblemDetail _was_ an exception that could be raised instead of returned?

api/admin/controller.py
leonardr: Could you decorate the controller methods rather than the routes? That would the non-complicated cases clear while keeping everything testable.

api/admin/controller.py
```diff
@@ -1912,10 +1995,12 @@ def collections(self):
                 )
                 if c.protocol in [p.get("name") for p in protocols]:
                     [protocol] = [p for p in protocols if p.get("name") == c.protocol]
-                    libraries = [
-                            self._get_integration_library_info(
-                                c.external_integration, library, protocol)
-                            for library in c.libraries]
+                    libraries = []
+                    for library in c.libraries:
+                        if not flask.request.admin or not flask.request.admin.is_librarian(library):
```
leonardr: When I was reviewing the core branch, I didn't mind too much that there was no caching of AdminRole objects, because I figured permissions were only being checked once per request. But this code gives me pause. If we have thirty libraries that share five collections (which I think is reasonable) one call to `collections will make 5*30*2=300 calls to `is_librarian`, and each `is_librarian` call might make 3 SQL queries.

Can you change it so looking up an `Admin` object preloads all their `AdminRoles`? Making `AdminRole` a `HasFullTableCache` would also basically solve this problem.

api/admin/controller.py
leonardr: I think it's not necessary to handle that on the server since it's not inconsistent for you to have a few libraries + a sitewide role, just redundant.

tests/admin/test_controller.py
```diff
@@ -141,7 +148,8 @@ def test_redirect_to_default_library(self):
             response = self.manager.admin_view_controller(None, None)
             eq_(302, response.status_code)
             location = response.headers.get("Location")
-            assert "admin/web/collection/%s" % self._default_library.short_name in location
+            library = Library.default(self._db)
+            assert "admin/web/collection/%s" % library.short_name in location
```
leonardr: Nothing wrong with this code but I don't see why the change is necessary.

api/admin/controller.py
aslagle: That could work, I might try that but it maybe it would be better to do it in a separate pr.

api/admin/controller.py
aslagle: It can be inconsistent if you go the other direction, like if you are a sitewide librarian and removed from one library. On the client removing you from a library when you have a sitewide role will remove the sitewide role and create individual roles for every other library.

aslagle: It doesn't leave an inconsistent state in the database, but it could be inconsistent with users' expectations.

api/admin/controller.py
leonardr: It would also be easier to test, because in most cases just need to verify that the decorator was applied, you wouldn't actually have to set up a fake request.

api/admin/controller.py
leonardr: Yeah, that's a big change.

api/admin/controller.py
```diff
@@ -2240,42 +2330,102 @@ def admin_auth_service(self, protocol):
...
+            for admin in self._db.query(Admin):
+                roles = []
+                for role in admin.roles:
+                    if role.library:
+                        if not flask.request.admin or not flask.request.admin.is_librarian(role.library):
+                            continue
+                        roles.append(dict(role=role.role, library=role.library.short_name))
+                    else:
+                        roles.append(dict(role=role.role))
+                admins.append(dict(email=admin.email, roles=roles))
```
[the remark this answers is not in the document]
aslagle: I think that's correct, because we'd like a library manager to be able to add them to their library even if they don't have any other roles that are visible to that manager yet.

api/admin/controller.py
leonardr: I'm OK with that translation happening entirely on the client side. If a request somehow goes in to remove a sitewide admin from one specific library, I would expect that to be a no-op.

api/admin/controller.py
```diff
@@ -2240,42 +2330,102 @@ def admin_auth_service(self, protocol):
...
 
             return dict(
                 individualAdmins=admins,
             )
 
         email = flask.request.form.get("email")
         password = flask.request.form.get("password")
+        roles = flask.request.form.get("roles")
 
-        if not email or not password:
```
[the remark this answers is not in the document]
aslagle: The admin interface currently has one page where you can change an admin's password and also edit their roles. I plan to move the password change out of that page, but until I do I want other admins to be able to edit someone's roles without submitting a password.

tests/admin/test_controller.py
```diff
@@ -2124,14 +2361,17 @@ def setup(self):
...
+
     def do_request(self, url, *args, **kwargs):
         """Mock HTTP get/post method to replace HTTP.get_with_timeout or post_with_timeout."""
         self.requests.append(url)
         response = self.responses.pop()
         return HTTP.process_debuggable_response(response)
 
     def test_get_integration_protocols(self):
-        """Test the _get_integration_protocols helper method."""
+        # Test the _get_integration_protocols helper method.
```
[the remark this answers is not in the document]
leonardr: This annoys me as well, but a docstring just seems better from a maintanability perspective. I'm not gonna nitpick in the context of this review but we should be able to figure out a different way to run the tests or something to get the display we want.

tests/admin/test_controller.py
aslagle: This is related to the stuff I changed to use nested transactions - it started committing things to the database and broke self._default_library. Once I figure out how to handle that I can change this back.

tests/admin/test_controller.py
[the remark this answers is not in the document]
aslagle: The actual reason I did this was that it made it easier to comment out large chunks of the file while debugging, and I decided to leave it because I liked it better anyway. I'm fine with changing it back for now since it's unrelated to this pr.

api/admin/controller.py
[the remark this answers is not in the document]
leonardr: Got it.

aslagle:
The thing I'm least confident about in this PR is the way is having an Admin model object and using it in the context as well as in a component state. I definitely could be making better use of Redux in this whole app.

api/admin/controller.py
```diff
@@ -203,13 +203,17 @@ def authenticated_admin_from_request(self, email=None, type=None):
...
             self._db, Admin, email=admin_details['email']
         )
         admin.update_credentials(
             self._db,
             credential=admin_details.get('credentials'),
         )
+        if is_new and admin_details.get("roles"):
+            for role in admin_details.get("roles"):
+                library = Library.lookup(self._db, role.get("library"))
+                admin.add_role(role.get("role"), library)
```
leonardr: Either this code or `add_role` should verify that the role being added is one of the actually existing roles.

aslagle: What should it do in that case? Log a warning and ignore the role?

leonardr: Yeah, that seems fine. It won't happen in real life so no need to make a ProblemDetail out of it.

api/admin/routes.py
```diff
@@ -48,20 +45,6 @@ def setup_admin(_db=None):
...
-    # the flask session object, but we can create a new
-    # session object in order to extract the admin's email
-    # and store it for later.
-    if not flask.session and flask.request:
-        temp_session = app.open_session(flask.request)
-        email = temp_session.get("admin_email")
-        type = temp_session.get("auth_type")
-        if email and type:
-            app.admin_email = email
-            app.admin_auth_type = type
```
aslagle: This turned out to cause problems when you can log out, it sometimes kept using these variables on subsequent requests. I don't have another fix for it, so now you may get a 401 on the first request after the app reloads again.

aslagle:
@leonardr could you take one more look at the latest changes to this?

api/admin/controller.py
```diff
@@ -212,8 +212,11 @@ def authenticated_admin(self, admin_details):
         )
         if is_new and admin_details.get("roles"):
             for role in admin_details.get("roles"):
-                library = Library.lookup(self._db, role.get("library"))
-                admin.add_role(role.get("role"), library)
+                if role.get("role") in AdminRole.ROLES:
+                    library = Library.lookup(self._db, role.get("library"))
+                    admin.add_role(role.get("role"), library)
```
leonardr: If the lookup fails, `library` will end up `None` and `add_role` will create the role globally, which I don't think we want.

api/admin/controller.py
```diff
@@ -179,33 +181,42 @@ def admin_auth_provider(self, type):
...
         )
         admin.update_credentials(
             self._db,
             credential=admin_details.get('credentials'),
         )
+        if is_new and admin_details.get("roles"):
+            for role in admin_details.get("roles"):
+                if role.get("role") in AdminRole.ROLES:
+                    library = Library.lookup(self._db, role.get("library"))
+                    admin.add_role(role.get("role"), library)
```
leonardr: My previous comment is outdated -- this code must have moved. This is the bit I'm talking about. Elsewhere in the code you can see the transaction rolled back if the library lookup fails.

PR #939, opened 2018-04-26
Add custom index views, e.g. a COPPA age gate

leonardr:
This branch introduces the concept of a custom index view. If a custom index view is associated with a library, `IndexController` will invoke it for that library instead of the normal code (which redirects to /groups/ or to the appropriate root lane for the patron type).

This branch adds one custom index view, which serves an OPDS navigation feed based on the one at https://instantclassics.librarysimplified.org/index.xml. The OPDS feed guides patrons to one of two lanes

In the future, the code that redirects to different root lanes based on the authenticated patron type could become a second custom index view. The code is similar, as is the use case -- both features are basically hard-to-configure one-offs for particular libraries.


api/custom_index.py
```diff
@@ -0,0 +1,215 @@
...
+        """
+        del self.BY_PROTOCOL[view_class.PROTOCOL]
+
+    @classmethod
+    def for_library(cls, library):
+        """Find the appropriate CustomIndexView for the given library."""
+        _db = Session.object_session(library)
+        integration = ExternalIntegration.one_for_library_and_goal(
+            _db, library, cls.GOAL
+        )
```
aslagle: This is not really an "external" integration since the custom view code is in the circ manager, but I guess the same is true for something like the simple authentication provider.

api/custom_index.py
```diff
@@ -0,0 +1,215 @@
...
+    @classmethod
+    def gate_tag(cls, restriction, met_url, not_met_url):
+        """Create a simplified:gate tag explaining the boolean option
+        the client is faced with.
+        """
+        # The original Instant Classics OPDS feed incorrectly omitted
+        # the simplified: namespace on 'restriction-met' and
+        # 'restriction-not-met', and the SimplyE client depends on the
+        # namespace not being present. For now, we include both a
+        # namespaced and a non-namespaced version.
```
aslagle: I see the namespace on https://instantclassics.librarysimplified.org/index.xml. Is it currently broken in the app?

api/custom_index.py
```diff
@@ -0,0 +1,215 @@
...
+            raise CannotLoadConfiguration(
+                "Unregistered custom index protocol: %s" % name
+            )
+        view_class = cls.BY_PROTOCOL[protocol]
+        return view_class(library, integration)
+
+    def __init__(self, library, integration):
+        raise NotImplementedError()
+
+    def __call__(self, library, annotator):
```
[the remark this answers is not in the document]
leonardr: I had kwargs in an earlier version but it turned out not to be necessary so I cut it rather than making sure it worked. I agree that some subclasses will not need an annotator. I guess when we move in the kwargs we can make 'annotator' one of the kwargs.

PR #944, opened 2018-04-27
Add error log on 401 response from OverDrive

datalogics-cgreen:
OverDrive returns a 401 code, but it is unclear what is exactly returned from OverDrive. There is evidently a problem with the OAuth Bearer Token. This PR aims to help diagnose such problems.

Ideally, this should not reveal information that is too sensitive, which may include patron information.

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -131,6 +131,7 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI):
         if response.status_code == 401:
             if exception_on_401:
                 # This is our second try. Give up.
+                self.log.error("Exception on %s: %s", response.status_code, response.content)
                 raise Exception("Something's wrong with the patron OAuth Bearer Token!")
             else:
                 # Refresh the token and try again.


leonardr:
`log.error` is the maximum volume something can be logged at, so it should always emit something.

leonardr:
I don't understand what's going on here. The line you wrote should be logged with `"name": "Overdrive API"`. If the exception gets logged, your item should also be logged. 

It is possible that your code isn't the code that's running? If you change the error message to incorporate the `response.content`, is it displayed and what does it look like?

datalogics-cgreen:
The above output is from `scripts`. My change went into `webapp`. I confirmed that the container I ran was created from the image I built containing my change. Perhaps the problem is that my change is in `circulation` and not `server_core`?

However, when I ran the monitor from the `webapp` container, I still didn't get any additional output.

leonardr:
Your Overdrive credential can be used with what Overdrive calls "Client Authentication." (https://developer.overdrive.com/apis/client-auth). This is for basic stuff like looking at the collection and seeing how many copies of what are available.

The problem you're having happens when your circulation manager tries to act on behalf of a specific patron. This requires what Overdrive calls "Patron Authentication". (https://developer.overdrive.com/apis/patron-auth) 

Scripts like `metadata_wrangler_collection_registrar` and `overdrive_monitor_full` only use Client Authentication, because they never act on behalf of a specific patron. The `patron_request` method, which is where your logging lives, will only be called from the app server, when someone tries to do something like borrow a book.

This also points to a likely explanation for your problem: I bet this Overdrive API key is authorized for Client Authentication, but not Patron Authentication. This has happened before, and it's something you'd need to work out with Overdrive. Seeing the error message from Overdrive should let us confirm or disprove this.

leonardr:
Closing this very old PR rather than rebasing it to `main`.

PR #951, opened 2018-05-01
Disabled strict slashes for admin routes.

### LINKED ISSUE #946: Possible problem with admin routes without slashes when running from docker
When I run the app from docker, going to /admin/ works but going to /admin gives me this:
```{"host": "05af3e533f95", "name": "root", "level": "ERROR", "timestamp": "2018-04-30T02:35:16.704907", "app": "simplified", "traceback": "Traceback (most recent call last):\n  File \"/var/www/circulation/env/local/lib/python2.7/site-packages/flask/app.py\", line 1813, in full_dispatch_request\n    rv = self.dispatch_request()\n  File \"/var/www/circulation/env/local/lib/python2.7/site-packages/flask/app.py\", line 1791, in dispatch_request\n    self.raise_routing_exception(req)\n  File \"/var/www/circulation/env/local/lib/python2.7/site-packages/flask/app.py\", line 1774, in raise_routing_exception\n    raise request.routing_exception\nRequestRedirect: 301 Moved Permanently: None", "message": "Exception in web app: 301 Moved Permanently: None", "filename": "app_server.py"}```

aslagle:
This fixes #946 which turned out to be a real problem.

At first I thought removing the route without the slash would make flask redirect to the route with the slash, but it didn't fix the error.

### DIFF
--- a/api/admin/routes.py
+++ b/api/admin/routes.py
@@ -579,8 +579,8 @@ def admin_sign_in_again():
         return redirect(app.manager.url_for('admin_sign_in', redirect=redirect_url))
     return flask.render_template_string(sign_in_again_template)
 
-@app.route('/admin/web')
-@app.route('/admin/web/')
+@app.route('/admin/web', strict_slashes=False)
+@app.route('/admin/web/', strict_slashes=False)
 @app.route('/admin/web/collection/<path:collection>/book/<path:book>')
 @app.route('/admin/web/collection/<path:collection>')
 @app.route('/admin/web/book/<path:book>')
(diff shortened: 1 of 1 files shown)


leonardr:
Can you get the same behavior by removing the route _without_ the slash and keeping `strict_slashes=False`?

aslagle:
No - the route with the slash would work, but without the slash would be a 404.

aslagle:
Actually, I will test this.

aslagle:
You're right, that seems to work fine. I also found one more route that needed strict_slashes turned off and a few that I had didn't actually need it.

Issue #945, opened 2018-04-27
Under some conditions, 'borrow' should be possible even if library has no authentication mechanism


leonardr:
Currently, if a library has no authentication mechanism, the CirculationManagerAnnotator never displays 'borrow' links for books. In most cases this is the right move, because you can't create a loan without authenticating.

But it's possible for a library with no authentication mechanism to grant anonymous access to a collection for which the library is the source of truth. A specific example is The SimplyE Collection, which will provide anonymous access to the short titles distributed through Plympton. As long as the agreement between the library and the distributor alows for this, it's perfectly fine for the library to give out bearer tokens without authenticating patrons or storing information about loans.

From an OPDS perspective these titles would effectively be open-access titles (in the sense of "direct download with no hassle", not in the sense of "CC-licensed").

PR #953, opened 2018-05-01
In some circumstances, a title can be fulfilled when there is no underlying loan

leonardr:
This is the first half of my work on https://github.com/NYPL-Simplified/circulation/issues/945. The goal is to make the `fulfill` route work even when there is no underlying loan, in cases where this is both technically possible and allowed by library policy.

The main conceptual change here is in `LibraryAuthenticator`. I introduce the concept of whether an authentication mechanism "identifies individuals". An authentication mechanism identifies individuals if it is capable of recognizing the same patron across visits to the site. A library identifies individuals if all of its authentication mechanisms identify individuals.

Currently, all authentication mechanisms identify individuals. This means all libraries identify individuals except the ones with no authentication mechanisms at all. In the future we will probably be supporting authentication mechanisms such as IP and geographic gates, which control access but do not identify individuals across visits.

If a library does not identify individuals, then it can't create any loans, which means it's okay to fulfill titles when there is no underlying loan. The other piece of the puzzle is whether this is technically possible. Currently this is technically possible with OPDS For Distributors. (It's also possible for open-access collections, but those are already handled separately, not through the `fulfill` route.)

In addition to the unit tests, I've verified that this works with a real OPDS For Distributors collection on a library that has no patron authentication.

The other half of this work is making sure patrons of these libraries see links to the `fulfill` route when appropriate.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -1013,11 +1013,10 @@ class LoanController(CirculationManagerController):
         """Fulfill a book that has already been checked out,
         or which can be fulfilled with no active loan.
 
-        If successful, this will serve the patron a downloadable copy of
-        the book, or a DRM license file which can be used to get the
-        book). Alternatively, for a streaming delivery mechanism it may
-        serve an OPDS entry with a link to a third-party web page that
-        streams the content.
+        If successful, this will serve the patron a downloadable copy
+        of the book, a key (such as a DRM license file or bearer
+        token) which can be used to get the book, or an OPDS entry
+        containing a link to the book.
         """
         do_get = do_get or Representation.simple_http_get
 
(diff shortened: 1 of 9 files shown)


api/controller.py
```diff
@@ -1031,20 +1036,25 @@ def fulfill(self, license_pool_id, mechanism_id=None, do_get=None):
...
             if isinstance(mechanism, ProblemDetail):
                 return mechanism
 
+        if (not loan or not loan_license_pool) and not (
+                self.can_fulfill_without_loan(
+                    library, patron, requested_license_pool, mechanism)
+        ):
+            return NO_ACTIVE_LOAN.detailed(
+                _("You have no active loan for this title.")
+            )
```
aslagle: When you can't fulfill without a loan but no patron is authenticated, is this the problem detail? It might be better to have a 401 in that case.

api/controller.py
```diff
@@ -1010,17 +1012,20 @@ def best_lendable_pool(self, library, patron, identifier_type, identifier,
...
+        of the book, a key (such as a DRM license file or bearer
+        token) which can be used to get the book, or an OPDS entry
+        containing a link to the book.
         """
         do_get = do_get or Representation.simple_http_get
 
-        patron = flask.request.patron
+        # Unlike most controller methods, this one does not require
+        # that the patron be authenticated.
+        patron = getattr(flask.request, 'patron', None)
```
aslagle: Since you removed `@requires_auth` from the route, does `flask.request.patron` ever get set?

leonardr:
My latest commit should address both of your concerns.

PR #954, opened 2018-05-02
Show direct fulfillment links when appropriate

leonardr:
This is the other half of https://github.com/NYPL-Simplified/circulation/issues/945. 

This first half, https://github.com/NYPL-Simplified/circulation/pull/953, makes the `fulfill` controller able to fulfill books without an underlying loan in cases where library policy allows for this and it's technically possible. (Basically, OPDS For Distributors on libraries that don't authenticate their patrons.)

This half adds links to the `fulfill` controller that would not otherwise be present. Without this branch, there are no links to that controller in the absence of an underlying loan.

The links use the link relation `http://opds-spec.org/acquisition/open-access`. This doesn't have any implications about the rights under which the books are distributed; the definition of that link relation just says "the complete representation of the Resource can be retrieved without any requirement, which includes payment and registration."

I'm going back and forth on using `http://opds-spec.org/acquisition/` instead, to eliminate any ambiguity, but we're pretty sure SimplyE can't handle `http://opds-spec.org/acquisition/` without modification.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -401,11 +401,10 @@ class CirculationManager(object):
         # Some features are only available if a patron authentication
         # mechanism is set up for this library.
         authenticator = self.auth.library_authenticators.get(library.short_name)
-        auth_supported = authenticator.supports_patron_authentication
         return LibraryAnnotator(
             self.circulation_apis[library.id], lane, library,
             top_level_title='All Books',
-            library_supports_patron_authentication=auth_supported,
+            library_identifies_patrons=authenticator.identifies_individuals,
             *args, **kwargs
         )
 
--- a/api/controller.py
+++ b/api/controller.py
@@ -69,6 +69,7 @@ from core.model import (
     LicensePool,
     Loan,
     LicensePoolDeliveryMechanism,
+    Patron,
     PatronProfileStorage,
     Representation,
     Session,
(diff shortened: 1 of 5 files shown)


api/controller.py
```diff
@@ -401,11 +401,10 @@ def annotator(self, lane, *args, **kwargs):
         # Some features are only available if a patron authentication
         # mechanism is set up for this library.
         authenticator = self.auth.library_authenticators.get(library.short_name)
-        auth_supported = authenticator.supports_patron_authentication
         return LibraryAnnotator(
             self.circulation_apis[library.id], lane, library,
             top_level_title='All Books',
-            library_supports_patron_authentication=auth_supported,
+            library_identifies_patrons=authenticator.identifies_individuals,
```
leonardr: This is what we really wanted all along. Bookshelf, annotations, and DRM discovery tags are all things that only work if a library identifies individual patrons. If a library "supports authentication" with a geo-gate or something, those features still won't work.

aslagle:
I'm concerned about using the open-access rel because our code makes assumptions about the rights, even if the spec does not, and we don't include rights information in our OPDS feeds. Maybe you could add the rights info as part of this PR.

leonardr:
That's a good solution.

PR #957, opened 2018-05-04
Novelist snapshot

### LINKED ISSUE #878: Synchronize collections with Novelist Collection Data API
The Novelist Collection Data API gives libraries a way to tell Novelist what's in their collection without having to set up a separate ILS integration. But it only works if you use it! A library with a Novelist integration should seek coverage of its entire collection from a CoverageProvider that covers a batch of Identifiers by telling the Novelist Collection Data API about them.

EdwinGuzman:
Fixes #878 

I think I may be missing something to fully complete the task. I'm returning the response from NoveList but not doing anything with it.

### DIFF
--- a/api/novelist.py
+++ b/api/novelist.py
@@ -60,7 +60,8 @@ class NoveListAPI(object):
     SETTINGS = [
         { "key": ExternalIntegration.USERNAME, "label": _("Profile") },
         { "key": ExternalIntegration.PASSWORD, "label": _("Password") },
-        { "key": AUTHORIZED_IDENTIFIER, "label": _("Authorized Identifier") },
+        { "key": AUTHORIZED_IDENTIFIER, "label": _("Authorized Identifier"),
+          "description": "The authorization key used as a Header for using the NoveList collections API endpoint." },
     ]
 
     # Different libraries may have different NoveList integrations
--- a/api/novelist.py
+++ b/api/novelist.py
@@ -38,15 +38,6 @@ from sqlalchemy.sql import (
     join,
     and_,
     or_,
-    literal_column,
-)
-from sqlalchemy import (
-    event,
-    exists,
-    func,
-    MetaData,
-    Table,
-    text,
 )
 from sqlalchemy.orm import aliased
 from core.util.http import HTTP
@@ -468,18 +459,16 @@ class NoveListAPI(object):
         LEFT_OUTER_JOIN = True
         i1 = aliased(Identifier)
         i2 = aliased(Identifier)
-        e = aliased(Equivalency)
-        lp = aliased(LicensePool)
 
         isbnQuery = select(
             [i1.identifier, i1.type, i2.identifier]
         ).select_from(
-            join(lp, i1, i1.id==lp.identifier_id)
-            .join(e, i1.id==e.input_id, LEFT_OUTER_JOIN)
-            .join(i2, e.output_id==i2.id, LEFT_OUTER_JOIN)
+            join(LicensePool, i1, i1.id==LicensePool.identifier_id)
+            .join(Equivalency, i1.id==Equivalency.input_id, LEFT_OUTER_JOIN)
+            .join(i2, Equivalency.output_id==i2.id, LEFT_OUTER_JOIN)
         ).where(
             and_(
-                lp.collection_id.in_(collectionList),
+                LicensePool.collection_id.in_(collectionList),
                 or_(i1.type=="ISBN", i2.type=="ISBN")
             )
         ).alias('lp_isbns')
@@ -488,9 +477,9 @@ class NoveListAPI(object):
 
         isbns = []
         for res in result:
-            if (res[1] == "ISBN"):
+            if (res[1] == Identifier.ISBN):
                 isbns.append(res[0])
-            else:
+            elif res[2] is not None:
                 isbns.append(res[2])
 
         return isbns
--- a/api/novelist.py
+++ b/api/novelist.py
@@ -46,6 +46,10 @@ class NoveListAPI(object):
 
     PROTOCOL = ExternalIntegration.NOVELIST
     NAME = _("Novelist API")
+
+    # Hardcoded authentication key used as a Header for calling the NoveList
+    # Collections API. It identifies the client, and lets NoveList know that
+    # SimplyE is making the requests.
     AUTHORIZED_IDENTIFIER = u"62521fa1-bdbb-4939-84aa-aee2a52c8d59"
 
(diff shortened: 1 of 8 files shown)


api/novelist.py
```diff
@@ -80,27 +103,30 @@ def values(cls, library):
...
 
     @classmethod
     def is_configured(cls, library):
         if (cls.IS_CONFIGURED is None or
             library.id != cls._configuration_library_id
         ):
-            profile, password = cls.values(library)
-            cls.IS_CONFIGURED = bool(profile and password)
+            profile, password, authorized_identifier = cls.values(library)
+            cls.IS_CONFIGURED = bool(profile and password and authorized_identifier)
```
leonardr: Is NoveList always going to provide an authorized identifier? I kind of thought you might only have profile and password, with some other technique being used to populate the recommendation dataset.

api/novelist.py
```diff
@@ -433,6 +459,58 @@ def get_recommendations(self, metadata, recommendations_info):
...
+
+        isbnQuery = select(
+            [i2.identifier]
+        ).select_from(
+            join(lp, i1, i1.id==lp.id)
+            .join(e, i1.id==e.input_id)
+            .join(i2, e.output_id==i2.id)
+        ).where(
+            and_(lp.collection_id.in_(collectionList), i2.type=="ISBN")
+        ).alias('lp_isbns')
```
leonardr: This will miss ISBNs that are the primary identifier for their LicensePool, rather associated with some proprietary identifier that is associated with an ISBN through some Equivalency. Right now Plympton is the only distributor that uses ISBNs as its primary identifier, but I'd prefer to get this working now and not have to come back to it later.

There are two ways of doing this. You can do two different queries, one of which doesn't have the join against `Equivalency`, or you can do an `outerjoin` against `Equivalency` instead of a regular join, and match on `or_(i1.type=="ISBN", i2.type=="ISBN")`.

Either way, you would test this by creating a LicensePool directly associated with an ISBN and making sure this method picked it up.

api/novelist.py
EdwinGuzman: I wasn't sure about this but you can't call the API endpoint provided with just the profile and password set since the authorized identifier header is needed. Is it safe to say that not every library will call the API endpoint periodically and instead use the manual option? If that's the case then this check is not needed here.

api/novelist.py
```diff
@@ -433,6 +459,58 @@ def get_recommendations(self, metadata, recommendations_info):
...
+        isbns = []
+        for r in result:
+            isbns.append(r[0])
+
+        if isbns:
+            response = self.put(
+                self.COLLECTION_DATA_API,
+                { "AuthorizedIdentifier": self.authorized_identifier },
+                data=self.make_novelist_data_object(isbns)
+            )
```
leonardr: It'll be easier to test this if the code to pull the ISBNs from the database is a different method than the code to make the PUT request.

bin/novelist_update
```diff
@@ -0,0 +1,11 @@
+#!/usr/bin/env python
+"""Show a collection or the full list of collections."""
```
leonardr: This description is left over from a copy-and-paste.

api/novelist.py
leonardr: This code is also missing a test.

EdwinGuzman:
@leonardr updated.

api/novelist.py
```diff
@@ -28,17 +28,40 @@
...
+    LicensePool,
+    Collection,
 )
 from core.util import TitleProcessor
+from sqlalchemy.sql import (
+    select,
+    join,
+    and_,
+    or_,
+    literal_column,
```
leonardr: literal_column import is not necessary.

api/novelist.py
```diff
@@ -433,6 +460,73 @@ def get_recommendations(self, metadata, recommendations_info):
...
+    def get_isbns_from_query(self, library):
+        collectionList = []
+        for c in library.collections:
+            collectionList.append(c.id)
+
+        LEFT_OUTER_JOIN = True
+        i1 = aliased(Identifier)
+        i2 = aliased(Identifier)
+        e = aliased(Equivalency)
+        lp = aliased(LicensePool)
```
leonardr: This is a minor point but you don't need to alias Equivalency or LicensePool, since you're only using those tables once.

api/novelist.py
leonardr: Changing over to two different integrations would also resolve this question.

api/novelist.py
```diff
@@ -433,6 +460,73 @@ def get_recommendations(self, metadata, recommendations_info):
...
+                lp.collection_id.in_(collectionList),
+                or_(i1.type=="ISBN", i2.type=="ISBN")
+            )
+        ).alias('lp_isbns')
+
+        result = self._db.execute(isbnQuery)
+
+        isbns = []
+        for res in result:
+            if (res[1] == "ISBN"):
```
leonardr: Use Identifier.ISBN here.

api/novelist.py
```diff
@@ -433,6 +460,73 @@ def get_recommendations(self, metadata, recommendations_info):
...
+                isbns.append(res[0])
+            else:
```
leonardr: This is a minor point but I think this ought to be `elif res[2] is not null` to handle the case where an ISBN is marked as equivalent to another ISBN.

api/novelist.py
```diff
@@ -433,6 +460,73 @@ def get_recommendations(self, metadata, recommendations_info):
...
+        return isbns
+
+    def put_isbns_novelist(self, library):
+        isbns = self.get_isbns_from_query(library)
+
+        content = None
+        if isbns:
+            response = self.put(
+                self.COLLECTION_DATA_API,
+                { "AuthorizedIdentifier": self.authorized_identifier },
```
leonardr: Given that the credentials we got from NoveList in the most recent email include both a profile and a password, It's confusing to me why only the 'password' is necessary here.

EdwinGuzman: I'm using the `authorized identifier` as a header, but I'm using the `profile` and `password` in the data object request that gets sent in line 507.

EdwinGuzman:
The new credentials returned the same "failed" response for the collections API. I've asked about that and am waiting to hear back. But, the new credentials did work for the NoveList Select API.

EdwinGuzman:
Updated - removed the `authorized identifier` as an integration setting and hardcoded it.

leonardr:
Looks good.

jce1028:
nice work guys. Thank you.

PR #958, opened 2018-05-04
Fix problems that showed up when I used the COPPA gate for real

leonardr:


api/custom_index.py
```diff
@@ -10,6 +10,7 @@
 from sqlalchemy.orm.session import Session
 
 from config import CannotLoadConfiguration
+from core.app_server import cdn_url_for
```
leonardr: This missing import didn't cause tests to fail because the tests pass in a mock `url_for`.

api/custom_index.py
```diff
@@ -151,7 +152,7 @@ def _navigation_feed(self, library, annotator, url_for=None):
         feed = OPDSFeed(title=library.name, url=base_url)
         opds = feed.feed
         yes_url = url_for(
-            'feed',
+            'acquisition_groups',
```
leonardr: Using `acquisition_groups` is more reliable than `feed` because we want this to work whether a lane has groups or not. If a lane doesn't have groups, the fallback position of `acquisition_groups` is to generate a regular feed. The `feed` route doesn't go in the other direction.

PR #961, opened 2018-05-09
Add an explanation to each language field that only ISO language codes are accepted.

leonardr:
I've tested this manually and verified that the descriptions show up properly.


aslagle:
This looks good, but I'll just note that it would be even better to make the inputs have autocomplete with full language names. I think I did that in another part of the admin interface already.

leonardr:
You did, in the metadata editor for books. I filed https://jira.nypl.org/browse/SIMPLY-272 to add that feature to this part of the admin interface.

PR #962, opened 2018-05-11
Include a signed JWT in the sitewide registration document.

leonardr:
This is the circulation side of https://github.com/NYPL-Simplified/metadata_wrangler/pull/167. Together these two branches fix https://jira.nypl.org/browse/SIMPLY-277.

In addition to the unit tests I've tested this with a local circulation manager registering with a local metadata wrangler. I did a registration, then deleted the shared secret on the circulation manager side and did the registration again, creating a new shared secret.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -16,6 +16,7 @@ from flask import (
     redirect,
 )
 from flask_babel import lazy_gettext as _
+import jwt
 from sqlalchemy.exc import ProgrammingError
 from PIL import Image
 from StringIO import StringIO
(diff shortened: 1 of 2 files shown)


api/admin/controller.py
```diff
@@ -2782,17 +2783,20 @@ def sitewide_registration(self, integration, do_get=HTTP.debuggable_get,
 
         # If the integration has an existing shared_secret, use it to access the
         # server and update it.
+        #
+        # NOTE: This is no longer technically necessary since we prove
+        # ownership with a signed JWT.
```
leonardr: We can take this out now or later.

api/admin/controller.py
```diff
@@ -2813,6 +2817,22 @@ def sitewide_registration(self, integration, do_get=HTTP.debuggable_get,
         shared_secret = encryptor.decrypt(base64.b64decode(shared_secret))
         integration.password = unicode(shared_secret)
 
+    def sitewide_registration_document(self, private_key):
```
aslagle: This is called sitewide registration even though it's only used by the metadata wrangler. If we did add another component that circ managers needed to register with, I'm not sure we'd want to use this.

leonardr: Assuming we get to control the registration protocol, I would like to reuse this document. The trust boundaries would be the same as with the metadata wrangler. We might have to add extra things to the JWT or to the resulting document but that would be a pretty easy change. Either way, we can handle this when it comes up.

PR #967, opened 2018-05-24
Cached OPDS entries need to be completed with Identifier-specific information

leonardr:
This is the metadata portion of https://jira.nypl.org/browse/SIMPLY-273. OPDS entries from the cache can't be served as-is, because they only contain bibliographic information about the Work. They need to be completed with information about the Identifier that was requested.

### DIFF
--- a/api/admin/opds.py
+++ b/api/admin/opds.py
@@ -23,7 +23,7 @@ class AdminAnnotator(LibraryAnnotator):
     def annotate_work_entry(self, work, active_license_pool, edition, identifier, feed, entry):
 
         super(AdminAnnotator, self).annotate_work_entry(work, active_license_pool, edition, identifier, feed, entry)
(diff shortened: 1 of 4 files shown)


api/admin/opds.py
```diff
@@ -23,7 +23,7 @@ def __init__(self, circulation, library, test_mode=False):
     def annotate_work_entry(self, work, active_license_pool, edition, identifier, feed, entry):
 
         super(AdminAnnotator, self).annotate_work_entry(work, active_license_pool, edition, identifier, feed, entry)
-        VerboseAnnotator.annotate_work_entry(work, active_license_pool, edition, identifier, feed, entry)
+        VerboseAnnotator.add_ratings(work, entry)
```
leonardr: This is still a little inelegant, but it avoids doing the work of `annotate_work_entry` twice by only doing the part of it that is specific to VerboseAnnotator. There may be a way to make multiple inheritance do this job but I've never trusted multiple inheritance in Python.

PR #968, opened 2018-05-29
Admins can change covers

aslagle:
This branch adds controller methods to enable admins to choose a new open-access image to use as a book cover. There's one route to generate a preview of the new cover and one to actually save it.


api/admin/controller.py
```diff
@@ -530,6 +538,14 @@ def media(self):
         """Return the supported media types for a work and their schema.org values."""
         return Edition.additional_type_to_medium
 
+    def rights_status(self):
+        """Return the supported rights status values with their names and whether
+        they are open access."""
+        return {uri: dict(name=name,
+                          open_access=(uri in RightsStatus.OPEN_ACCESS),
+                          allows_derivatives=(uri in RightsStatus.ALLOWS_DERIVATIVES))
+                for uri, name in RightsStatus.NAMES.iteritems()}
```
aslagle: Only rights statuses that allow derivatives are selectable for a new cover image.

api/admin/controller.py
```diff
@@ -1030,6 +1046,228 @@ def edit_classifications(self, identifier_type, identifier):
...
+        )
+
+        metadata = Metadata(data_source, links=[cover_data])
+        metadata.apply(work.presentation_edition,
+                       collection,
+                       replace=replacement_policy)
+
+        # metadata.apply only updates the edition, so we also need
+        # to update the work.
+        work.calculate_presentation(policy=presentation_policy)
```
aslagle: There may be some extra work happening in metadata.apply since I reused the policy.

api/admin/opds.py
```diff
@@ -66,7 +67,22 @@ def annotate_work_entry(self, work, active_license_pool, edition, identifier, fe
...
+        )
+        if storage_integrations.count() > 0:
+            feed.add_link_to_entry(
+                entry,
+                rel="http://librarysimplified.org/terms/rel/change_cover",
+                href=self.url_for(
+                    "work_change_book_cover",
+                    identifier_type=identifier.type,
+                    identifier=identifier.identifier, _external=True)
+            )
```
aslagle: The presence of this link is used to determine whether to show the "Cover" tab. If there's no storage integration, it won't show.

api/admin/controller.py
leonardr: This could be enforced on the data model level but I think this is the right level; we should leave room for someone modifying an image that they own the rights to.

api/admin/controller.py
```diff
@@ -1030,6 +1046,228 @@ def edit_classifications(self, identifier_type, identifier):
 
         return Response("", 200)
 
+    def _validate_cover_image(self, image):
+        image_width, image_height = image.size
+        if image_width < 600 or image_height < 900:
```
leonardr: Please make these into constants.

api/admin/controller.py
```diff
@@ -1030,6 +1046,228 @@ def edit_classifications(self, identifier_type, identifier):
...
+           return INVALID_COVER_IMAGE.detailed(_("Cover image must be at least 600px in width and 900px in height."))
+        return True
+
+    def _process_cover_image(self, work, image, title_position):
+        title = work.presentation_edition.title
+        author = work.presentation_edition.author
+        if author == Edition.UNKNOWN_AUTHOR:
+            author = ""
+
+        if title_position in ["top", "center", "bottom"]:
```
leonardr: This should also be constants, especially because it's not exhaustive -- 'none' is also possible but I didn't learn this until later.

api/admin/controller.py
leonardr: This is likely to be a real case for works published through the library -- the author wants a photo they own to be used as the cover. Maybe we don't need to handle that yet, but "permission granted" is a reason why a RightsStatus that wouldn't ordinarily belong here should go here anyway.

api/admin/opds.py
leonardr: This should only happen if there's a storage integration _for this work_, i.e. either a sitewide integration or an integration for one of its collections.

Refactoring https://github.com/NYPL-Simplified/circulation/pull/968/files#diff-d3633d1a0ad08a784fffc12ecdba4065R1185 might help this.

Not a big deal -- we could even simplify it by saying only a sitewide integration counts.

tests/admin/test_controller.py
```diff
@@ -1081,6 +1112,236 @@ def test_classifications(self):
...
+
+        # Without a title position, the image won't be changed.
+        self.manager.admin_work_controller._process_cover_image(work, processed, "none")
+
+        image_histogram = original.histogram()
+        expected_histogram = processed.histogram()
+
+        root_mean_square = math.sqrt(reduce(operator.add,
+                                            map(lambda a,b: (a-b)**2, image_histogram, expected_histogram))/len(image_histogram))
+        assert root_mean_square < 10
```
leonardr: This is validating that the processed image is not changed that much from the original?

leonardr: Yeah, you're comparing the actual to expected with some fuzz accounting for slight variations in how PIL decides to handle the image.

api/admin/controller.py
aslagle: Yes, good idea. I was also wondering if there's anything else I should enforce here like aspect ratio, but since these are uploaded by admins they'll see if the preview looks bad.

tests/admin/test_controller.py
aslagle: Correct.

api/admin/controller.py
aslagle: Yeah, I'd like to handle that case but I'm not sure how to make it clear in the UI.

tests/admin/test_controller.py
```diff
@@ -1081,6 +1112,236 @@ def test_classifications(self):
...
+
+        class TestFileUpload(StringIO):
+            headers = { "Content-Type": "image/png" }
+        base_path = os.path.split(__file__)[0]
+        resource_path = os.path.join(base_path, "..", "files", "images")
+        path = os.path.join(resource_path, "blue.jpg")
+        original = Image.open(path)
+        buffer = StringIO()
+        original.save(buffer, format="PNG")
+        image_data = buffer.getvalue()
```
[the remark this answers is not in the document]
leonardr: Actually you can mock _process_cover_image here, as you do elsewhere, so long as an Image object is returned. I was thinking of simplifying further (and speeding up the tests) by having the mock return something that's not an image.

api/admin/controller.py
leonardr: Maybe just let 'In copyright' be an option. The rights_explanation will explain why it's okay to make a derivative work. That's the reason you'd need a rights_explanation, right? When it's not obvious from the RightsStatus.

tests/admin/test_controller.py
```diff
@@ -1081,6 +1112,236 @@ def test_classifications(self):
...
+
+            resource = link.resource
+            assert identifier.urn in resource.url
+            eq_(staff_data_source, resource.data_source)
+            eq_(RightsStatus.CC_BY, resource.rights_status.uri)
+            eq_("explanation", resource.rights_explanation)
+
+            derivation = self._db.query(ResourceDerivation).filter(ResourceDerivation.derivative_id==resource.id).one()
+            original_resource = derivation.original
+            assert resource != original_resource
```
leonardr: Can you also verify that this resource's thumbnail is different from the old resource's thumbnail?

api/admin/controller.py
aslagle: I was thinking it was more to justify why the work is open access in case it's challenged later.

leonardr: OK, the rights_explanation on the original resource would have the reason why the image is open access, and the rights_explanation on the derivative would have the reason why the derivative could be made from a non-open-access original.

Again, it's not necessary to implement this in this branch, I just want to make sure we don't exclude the possibility.

aslagle: Yeah, that would work fine in the data model but it would make the UI complicated. For now I'm just using the same rights explanation for both.

leonardr: I think the UI wouldn't need to change. You would have the standard free licenses and 'other', plus a box to write an explanation. If you choose one of the standard licenses then your explanation is associated with the original resource, and the derivative explanation is 'the original resource was licensed for this'. Otherwise the explanation is associated with the derivative resource as 'why did we do this'.

aslagle: If you choose 'other', is the rights status unknown?

aslagle: I added 'in copyright' and 'other' as options, and made it save the rights explanations differently depending on the URI.

PR #969, opened 2018-05-29
Allow Enki requests to recover from a timeout

datalogics-dans:
We've been seeing some timeout errors when the Enki import script
runs. This aims to address that by running with a 20 second timeout
on the first attempt, but if a timeout occurs it'll try again with
a 60 timeout before finally giving up.


leonardr:
You may have been conflating timeouts with other types of HTTP errors, where trying the same request again can be a good strategy.

Do you have any way of empirically determining a good timeout time before making a PR? If you need to wait for a product release to try this out, you could be spending a long time with a nonworking Enki integration.

leonardr:
This won't resolve your docker problems, but if you're changing the original timeout period it's not also necessary to catch the timeout exception and retry. You can just change the number.

PR #974, opened 2018-06-03
Changed list controllers to use identifiers instead of pwids.

aslagle:
@EdwinGuzman I gave you some bad advice when you were working on the bugs related to adding works to lists. I originally used pwids because they were convenient to get from the OPDS feed and I thought they were unique enough, but that keeps causing problems. The latest problem is that sometimes the data source in the OPDS entry and license pool doesn't match the data source on the edition, so the query doesn't find anything. @leonardr suggested using identifiers instead, which should be unique enough and makes more sense than having to keep adding more filters and joins to the query.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1440,38 +1440,22 @@ class CustomListsController(AdminCirculationManagerController):
         old_entries = [x for x in list.entries if x.edition]
         membership_change = False
         for entry in entries:
-            pwid = entry.get("pwid")
-            medium = entry.get("medium")
-            language = entry.get("language")
-            data_source = entry.get("data_source")
+            urn = entry.get("identifier_urn")
 
-            data_source = DataSource.lookup(self._db, data_source)
-            if data_source:
-                data_source_id = data_source.id
-            else:
-                data_source_id = None
-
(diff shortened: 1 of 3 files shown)


api/admin/controller.py
```diff
@@ -1440,48 +1440,32 @@ def _create_or_update_list(self, library, name, entries, collections, id=None):
...
             ).join(
                 Edition, Edition.id==Work.presentation_edition_id
+            ).join(
+                LicensePool, LicensePool.work_id==Work.id
+            ).join(
+                Collection, LicensePool.collection_id==Collection.id
             ).filter(
-                Edition.permanent_work_id==pwid
+                Edition.primary_identifier_id==identifier.id
+            ).filter(
```
leonardr: I'm not convinced it's necessary to join against Edition at all -- can you filter against LicensePool.identifier_id instead? That will be simpler and I think will be more resistant to edge cases (since a Work might consolidate LicensePools with a number of different identifiers).

EdwinGuzman:
Cool, so using identifiers means it's unique enough that we don't need language or medium to query by anymore?

aslagle:
Right, books with different languages or mediums would have their own identifiers.

PR #975, opened 2018-06-03
Convert uploaded cover images to RGB mode before drawing on them.

aslagle:
Michelle Bickert found a bug where images with mode "L" (black and white with 8-bit pixels) caused an exception in PIL when drawing a rectangle:

```
self._process_cover_image(work, image, title_position)\n  File \"/Users/aslagle/dev/circulation/api/admin/controller.py\", line 1114, in _process_cover_image\n    fill=(255,255,255))\n  File \"/Users/aslagle/dev/circulation/env/lib/python2.7/site-packages/PIL/ImageDraw.py\", line 191, in rectangle\n    ink, fill = self._getink(outline, fill)\n  File \"/Users/aslagle/dev/circulation/env/lib/python2.7/site-packages/PIL/ImageDraw.py\", line 118, in _getink\n    fill = self.draw.draw_ink(fill, self.mode)\nTypeError: function takes exactly 1 argument (3 given)", "message": "Exception in web app: function takes exactly 1 argument (3 given)"
```
I don't really understand this but converting the image to 'RGB' mode first avoids it.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -1067,6 +1067,10 @@ class WorkController(AdminCirculationManagerController):
             author = ""
 
         if title_position in self.TITLE_POSITIONS:
+            # Convert image to 'RGB' mode if it's not already, so drawing on it works.
+            if image.mode != 'RGB':
+                image = image.convert("RGB")
+
             draw = ImageDraw.Draw(image)
             image_width, image_height = image.size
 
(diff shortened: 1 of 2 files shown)


api/admin/controller.py
```diff
@@ -1213,6 +1216,10 @@ def change_book_cover(self, identifier_type, identifier, mirror=None):
             if not original_href:
                 original_href = Hyperlink.generic_uri(data_source, work.presentation_edition.primary_identifier, Hyperlink.IMAGE, content=original_content)
                 
+            # Convert image to 'RGB' mode if it's not already, so drawing on it works.
+            if image.mode != 'RGB':
+                image = image.convert("RGB")
+
```
leonardr: Can this go into _process_cover_image so you don't need the code twice?

PR #977, opened 2018-06-05
Add fields novelist request

EdwinGuzman:
@leonardr you mentioned joining with the Contributions and Contributors table but the author was in the Editions table. When I did join the Contributions and Contributors tables, I got mixed book titles and authors (but I might have been doing something wrong). I was also getting multiple ISBNS entries because there were different contributor types, like "Author" and "Primary Author", but if joining those two tables is the better approach, would I have to limit the query to just one of those types?
_I think_ this covers both cases for the identifiers.


leonardr:
Yeah, I meant to join against Edition and then against Contribution through Contributor. I forgot that we store sort_author as a string in Edition.

For purposes of getting this working I think it's okay to just get Edition.sort_author and split it on the semicolon to get a list. (Edition.sort_author is the output of Edition.calculate_author, which creates output like `King, Stephen ; Angelou, Maya`). We'll see if NoveList needs something more.

To get the narrator of an audiobook we will have to join against Contribution through Contributor, since the fact that someone had a 'narrator' role isn't stored in Edition -- it's stored in Contribution.

api/novelist.py
```diff
@@ -461,54 +462,74 @@ def get_isbns_from_query(self, library):
         i2 = aliased(Identifier)
 
         isbnQuery = select(
-            [i1.identifier, i1.type, i2.identifier]
+            [i1.identifier, i1.type, i2.identifier,
+            Edition.title, Edition.medium, Edition.author],
```
leonardr: I'm pretty sure NoveList will want `sort_author`.

tests/test_novelist.py
```diff
@@ -357,33 +357,47 @@ def test_choose_best_metadata(self):
...
+        items = self.novelist.get_items_from_query(self._default_library)
+        eq_(items, [])
 
         edition = self._edition(identifier_type=Identifier.ISBN)
         pool = self._licensepool(edition, collection=self._default_collection)
 
-        isbns = self.novelist.get_isbns_from_query(self._default_library)
+        items = self.novelist.get_items_from_query(self._default_library)
+        item = dict(
+            Author=edition.author,
```
leonardr: Again, sort_author here I think.

EdwinGuzman:
Thanks. I joined the tables to get the Narrator if it exists. I'm not sure if Novelist is expecting a list of authors from `sort_author` so I just extracted the first one. I'll add more comments from my latest commit.

api/novelist.py
```diff
@@ -461,54 +464,83 @@ def get_isbns_from_query(self, library):
         i2 = aliased(Identifier)
 
         isbnQuery = select(
-            [i1.identifier, i1.type, i2.identifier]
+            [i1.identifier, i1.type, i2.identifier,
+            Edition.title, Edition.medium, Edition.sort_author, Edition.author,
```
EdwinGuzman: I got both `sort_author` and `author` because in the test `Edition.sort_author` was empty. So just covering the case where it could be empty from the table. Does that sound right?

api/novelist.py
```diff
@@ -461,54 +464,83 @@ def get_isbns_from_query(self, library):
...
+                or_(Edition.primary_identifier_id==i1.id, Edition.primary_identifier_id==i2.id)
+            )
+            .join(Contribution, Edition.id==Contribution.edition_id)
+            .join(Contributor, Contribution.contributor_id==Contributor.id)
         ).where(
             and_(
                 LicensePool.collection_id.in_(collectionList),
-                or_(i1.type=="ISBN", i2.type=="ISBN")
+                or_(i1.type=="ISBN", i2.type=="ISBN"),
+                or_(Contribution.role=="Primary Author", Contribution.role=="Narrator"),
```
EdwinGuzman: I added `"Primary Author"` check here because there were multiple rows of the same ISBN but the only difference was this role value which could've been "Author", "Primary Author", "Unknown", "Illustrator", etc.

api/novelist.py
leonardr: If you're getting the contributions (which is better anyway) then you don't need Edition.sort_author and Edition.author at all.

api/novelist.py
leonardr: You also need to pick up "Author", for the case where a book has multiple authors or no primary author is specified. So instead of this `or` clause, use a clause defined in the following way:

```
roles = list(Contributor.AUTHOR_ROLES)
roles.append(Contributor.NARRATOR_ROLE)
clause = Contribution.role.in_(roles)
```

api/novelist.py
```diff
@@ -498,13 +504,18 @@ def create_item_object(self, object):
         elif object[2] is not None:
             isbn = object[2]
 
+        title = object[3]
         mediaType = Edition.medium_to_additional_type[object[4]]
+        author = object[5].split(';')[0] if object[5] else object[6]
```
leonardr: Then, instead of this, pull an author name from object[8] `if role in Edition.AUTHOR_ROLES` and add it to a list. Use that list as the value of `Author`.

leonardr: If we have a Contributor but we're missing the sort name then I would just ignore it -- it won't do NoveList any good.

EdwinGuzman: Novelist's documentation says the Author is a string value, but it's a good question to ask to see if they can accept a list, too.

api/novelist.py
```diff
@@ -549,7 +554,7 @@ def create_item_object(self, object, currentIdentifier, existingItem):
             # If we encounter a new ISBN, we take whatever author value is
             # initially given to us.
             title = object[3]
-            mediaType = Edition.medium_to_book_format_type_values[object[4]]
+            mediaType = self.medium_to_book_format_type_values[object[4]]
```
leonardr: This will work, but just to avoid crashes, do self.medium_to_book_format_type_values.get(object[4], ""). That will handle anything unusual in the database.

leonardr:
Good job! I think this is ready.

PR #978, opened 2018-06-06
Restored MARC parsing code that was lost when moving directory import from content server.

aslagle:
In #852 the `load_metadata` method was mocked for testing so we didn't notice that it was not actually implemented :|

I brought it back, but I don't actually have a MARC file anymore so I can't test this.

For my next PR, I will be expanding this script to support ONIX as well as MARC.


leonardr:
Here's a MARC file you can test with.

[marc.zip](https://github.com/NYPL-Simplified/circulation/files/2077487/marc.zip)

aslagle:
I just noticed there's one in the content server repo too.

PR #980, opened 2018-06-07
Added ONIX option to directory import script.

aslagle:
I also made a couple minor changes to the script so the missing mirror doesn't cause exceptions during a dry run.

### DIFF
--- a/api/onix.py
+++ b/api/onix.py
@@ -170,58 +170,41 @@ class ONIXExtractor(object):
         root = tree.getroot()
 
         for record in root.findall('product'):
-            title_tag = parser._xpath(record, 'descriptivedetail/titledetail/titleelement/b203')
-            title = None
-            if title_tag:
-                title = title_tag[0].text
-            else:
-                title_prefix_tag = parser._xpath(record, 'descriptivedetail/titledetail/titleelement/b030')
-                title_without_prefix_tag = parser._xpath(record, 'descriptivedetail/titledetail/titleelement/b031')
-                if title_prefix_tag and title_without_prefix_tag:
-                    title = title_prefix_tag[0].text + ' ' + title_without_prefix_tag[0].text
+            title = parser.text_of_optional_subtag(record, 'descriptivedetail/titledetail/titleelement/b203')
+            if not title:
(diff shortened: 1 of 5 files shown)


api/onix.py
```diff
@@ -0,0 +1,281 @@
...
+        'F01': Contributor.PHOTOGRAPHER_ROLE, # 'Filmed/photographed by'
+        'F02': Contributor.EDITOR_ROLE, # 'Editor (film or video)'
+        'F99': Contributor.UNKNOWN_ROLE, # Other type of recording
+        'Z01': Contributor.UNKNOWN_ROLE, # 'Assisted by'
+        'Z02': Contributor.UNKNOWN_ROLE, # 'Honored/dedicated to'
+        'Z99': Contributor.UNKNOWN_ROLE, # Other creative responsibility
+    }
+    
+    @classmethod
+    def parse(cls, file, data_source_name):
```
leonardr: What about making this a subclass of core/xmlparser.py:XMLParser? I think many chunks of this code could each be simplified by using _xpath and _xpath1. It would also handle namespaces, though judging from your sample, ONIX files aren't super concerned with namespaces.

api/onix.py
```diff
@@ -0,0 +1,281 @@
...
+                                                        family_name=family_name,
+                                                        roles=[cls.CONTRIBUTOR_TYPES[type]],
+                                                        biography=bio))
+
+            collateral_tags = parser._xpath(record, 'collateraldetail/textcontent')
+            links = []
+            for tag in collateral_tags:
+                type = tag.find('x426').text
+                # TODO: '03' is the summary in the example I'm testing, but that
+                # might not be generally true.
```
aslagle: I find the ONIX documentation difficult to use but if it's from "List 42" then "03" would be "Body matter", which is "Text components such as Part, Chapter, Section etc which appear as part of the main body of text content in a product", so that's not necessarily a summary.

api/onix.py
aslagle: Yeah, I don't think I could do much with namespaces but I could probably use some of the parser methods more effectively.

aslagle:
I cleaned this up a lot just by using XMLParser methods. @leonardr do you think there's more I could improve with a subclass?

leonardr:
No, subclasses are mainly useful for keeping track of different namespaces.

PR #984, opened 2018-06-19
Implement integration self-tests

leonardr:
This branch implements basic self-tests for five integrations:

* BasicAuthenticationProvider (any ILS system that uses HTTP Basic Auth, notably SIP2 and Millenium Patron)
* Overdrive
* Odilo
* OPDS For Distributors
* The NYT best-seller API

More will follow; these are the ones that were necessary for me to convince myself that the architecture is sound and can handle both simple and complex cases.

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -4,8 +4,8 @@ from config import (
     CannotLoadConfiguration,
     IntegrationException,
 )
-from core.external_integration import (
-    HasSelfTest,
+from core.selftest import (
+    HasSelfTests,
 )
 from core.opds import OPDSFeed
 from core.model import (
@@ -1217,7 +1217,7 @@ class AuthenticationProvider(OPDSAuthenticationFlow):
         raise NotImplementedError()
 
 
-class BasicAuthenticationProvider(AuthenticationProvider, HasSelfTest):
+class BasicAuthenticationProvider(AuthenticationProvider, HasSelfTests):
     """Verify a username/password, obtained through HTTP Basic Auth, with
     a remote source of truth.
     """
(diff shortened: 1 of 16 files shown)


api/authenticator.py
```diff
@@ -1444,10 +1448,47 @@ def testing_patron(self, _db):
 
         :return: A 2-tuple (Patron, password)
         """
-        if self.test_username is None or self.test_password is None:
-            return None, None
+        if self.test_username is None:
+            return self.test_username, self.test_password
```
leonardr: This lets us handle the case where the test user doesn't need a password.

api/nyt.py
```diff
@@ -76,20 +83,31 @@ def from_config(cls, _db, **kwargs):
...
         if not metadata_client:
-            metadata_client = MetadataWranglerOPDSLookup.from_config(self._db)
+            try:
+                metadata_client = MetadataWranglerOPDSLookup.from_config(
+                    self._db
+                )
+            except CannotLoadConfiguration, e:
+                self.log.error(
+                    "Metadata wrangler integration is not configured, proceeding without one."
+                )
```
leonardr: It doesn't look like this needs to be a fatal error -- worst case we'll use the default code to turn "Surname, Name" into "Name Surname".

api/odilo.py
```diff
@@ -116,6 +146,9 @@ def get_patron_credential(self, patron, pin):
         def refresh(credential):
             return self.get_patron_access_token(credential, patron, pin)
 
+        return self._credential_lookup(patron, refresh)
```
aslagle: Should this be `self._patron_credential_lookup`?

leonardr: The fact that this error didn't cause big problems rang an alarm bell for me. I investigated and it turns out that the Odilo API actually has no concept of a "patron token". The API doesn't issue patron tokens and when it acts on behalf of a patron it uses the global token. The 'patron token' stuff is just something that was copied from the Overdrive code this was based on.

I filed https://jira.nypl.org/browse/SIMPLY-676 to clean up the code, but the upshot for purposes of this branch is that the Odilo self-test doesn't need to test functionality that doesn't exist.

PR #985, opened 2018-06-19
Lanes.media is not currently used

leonardr:
This branch changes the contents of `lanes` to reflect the fact that we don't currently restrict lanes to specific media types. This is theoretically possible, and may be useful in the future, but for now we're solely using entrypoints for this, and the lanes themselves should not have any restrictions on media type.


EdwinGuzman:
There's just the broken test, but I can now see Audio filtered items.

PR #986, opened 2018-06-20
Implement external_integration implementations now that HasSelfTests requires it

leonardr:
This is the circulation counterpart to https://github.com/NYPL-Simplified/server_core/pull/898, and it addresses https://jira.nypl.org/browse/SIMPLY-677. 

All this does it implement external_integrations for the API classes that implement HasSelfTests. Going forward I'll implement external_integration as part of the HasSelfTests work.


api/axis.py
```diff
@@ -91,6 +95,43 @@ class Axis360API(BaseAxis360API, Authenticator, BaseCirculationAPI):
...
+            "Refreshing bearer token", self.refresh_bearer_token
+        )
+        yield result
+        if not result.success:
+            # If we can't get a bearer token, there's no point running
+            # the rest of the tests.
+            return
+
+        def _count_events():
+            now = datetime.utcnow()
```
EdwinGuzman: I think this one is missing `.datetime` in there?

leonardr: axis.py does `from datetime import datetime` instead of `import datetime`, so `datetime.utcnow()` is correct here, although it would be better to change the import to the way it's used elsewhere.

api/bibliotheca.py
```diff
@@ -87,6 +92,34 @@ class BibliothecaAPI(BaseBibliothecaAPI, BaseCirculationAPI):
         [v,k] for k, v in delivery_mechanism_to_internal_format.items()
     )
 
+    def external_integration(self, _db):
+        return self.collection.external_integration
+
+    def _run_self_tests(self, _db):
+        def _count_events():
+            now = datetime.datetime.utcnow()
+            five_minutes_ago = now - datetime.timedelta(minutes=5)
```
EdwinGuzman: Is there a difference between `timedelta(minutes=5)` (line 113) and `datetime.timedelta(minutes=5)`?

leonardr: Only because this file imports the same class differently.

PR #988, opened 2018-06-21
Implement self-test for RBDigital

leonardr:
This branch implements a self-test for RBDigital integrations which counts the number of ebooks and audiobooks in the collection, and counts the loans and holds for the default patron (creating an account for them if necessary).

While testing this against a real RBDigital integration, I discovered that the RBDigital API has changed its rules for creating remote patrons and the documents they send when a patron has created. This broke our production code, but not our tests, which were mocking the old behavior. I've fixed the problem I ran into but there may be more undocumented changes.


leonardr:
This addresses https://jira.nypl.org/browse/SIMPLY-683.

PR #989, opened 2018-06-27
Enki self test

leonardr:
This branch implements a basic self-test that verifies an Enki collection is properly configured. It addresses https://jira.nypl.org/browse/SIMPLY-687.

While writing the self-test I discovered that the Enki client itself isn't working properly. Instead of getting a delta of circulation events and new titles added to the collection, the Enki collection monitor retrieves the entire list of books every time it runs. That's only supposed to happen on initial setup. This is most likely a quick-and-dirty initial implementation that we never revisited.

This branch does not address the larger client problem, but the API methods it implements can be used to solve the problem later.


api/enki.py
```diff
@@ -385,8 +477,9 @@ def patron_activity(self, patron, pin):
                 raise CirculationException()
         for loan in result['checkedOutItems']:
             yield self.parse_patron_loans(loan)
-        for hold in result['holds']:
-            yield self.parse_patron_holds(hold)
+        for type, holds in result['holds'].items():
+            for hold in holds:
+                yield self.parse_patron_holds(hold)
```
leonardr: The Enki client is also missing holds support, so this doesn't matter yet, but when there are patron holds the old code crashes.

tests/test_enki.py
```diff
@@ -46,14 +52,82 @@ def get_data(cls, filename):
         path = os.path.join(cls.resource_path, filename)
         return open(path).read()
 
-
-class TestEnkiAPI(DatabaseTest, BaseEnkiTest):
-
```
leonardr: Other test classes were subclassing TestEnkiAPI, which made the tests defined in TestEnkiAPI run multiple times.

PR #990, opened 2018-06-28
Improve usage of Enki API

leonardr:
This branch adds complete test coverage of our Enki API code and fixes a number of serious issues:

* There was no way of reading in incremental changes from Enki, even though the Enki API had ways of learning what happened since a certain time. The only way to find out about the collection was to iterate over the whole thing. Now it takes two different API calls (updated_titles for new books and recent_activity for checkouts) but we can do it incrementally after an initial full import.
* Important information such as subject classifications, descriptions, and thumbnail images was not being brought in.
* Inconsistent usage of the "recordId" field to create Enki Identifiers versus the "id". We now consistently use the "id".
* We were assuming every book was in English, even though there is a data field for a book's language.
* I removed the EnkiBibliographicCoverageProvider (which should never be needed in normal usage) rather than get it working and add test coverage for it.

Holds are still not supported.

This addresses https://jira.nypl.org/browse/SIMPLY-688.


api/enki.py
```diff
@@ -233,126 +222,83 @@ def recent_activity(self, minutes=5):
...
         args['method'] = "getAllTitles"
         args['id'] = "secontent"
         args['strt'] = strt
         args['qty'] = qty
         args['lib'] = self.library_id
         response = self.request(url, params=args)
-        return response
-
-    @classmethod
-    def create_identifier_strings(cls, identifiers):
```
leonardr: This method was never used and was left over from copy-and-paste.

api/enki.py
```diff
@@ -233,126 +222,83 @@ def recent_activity(self, minutes=5):
...
-            else:
-                value = i
-            identifier_strings.append(value)
-
-        return identifier_strings
+        for metadata in BibliographicParser().process_all(response.content):
+            yield metadata
 
     @classmethod
-    def parse_token(cls, token):
```
leonardr: This method was never used and was left over from copy-and-paste.

api/enki.py
```diff
@@ -190,39 +162,56 @@ def count_loans_and_holds(patron, pin):
...
 
     def request(self, url, method='get', extra_headers={}, data=None,
-                params=None, exception_on_401=False):
-        """Make an HTTP request, acquiring/refreshing a bearer token
-        if necessary.
-        """
+                params=None, retry_on_timeout=True, **kwargs):
+        """Make an HTTP request to the Enki API."""
         headers = dict(extra_headers)
-        if exception_on_401:
```
leonardr: This code was copied-and-pasted from Overdrive code and assumes that requests are being made using an authorization token, which isn't the case.

api/enki.py
```diff
@@ -233,126 +222,83 @@ def recent_activity(self, minutes=5):
...
-                params=params, timeout=90, **kwargs
-            )
-        except RequestTimedOut, e:
-            self.log.info("Request to %s timed out once. Trying a second time.", url)
-            return HTTP.request_with_timeout(
-                method, url, headers=headers, data=data,
-                params=params, timeout=90, **kwargs
-            )
-
-    def reaper_request(self, identifier):
```
leonardr: This is just a getItem request; the API class doesn't need to know what will be done with the results.

api/enki.py
```diff
@@ -421,11 +366,18 @@ def fulfill(self, patron, pin, licensepool, internal_format):
                 self.log.error("User validation against Enki server with %s / %s was unsuccessful."
                     % (patron.authorization_identifier, pin))
                 raise AuthorizationFailedException()
-        drm_type = self.get_enki_drm_type(book_id)
```
leonardr: There's no reason to make another HTTP request to get the DRM type -- it's always the type associated with the DeliveryMechanism.

api/enki.py
```diff
@@ -539,226 +479,253 @@ def queue_response(self, status_code, headers={}, content=None):
...
-                )
-                batch_results.append(result)
-        return batch_results
-
-    def process_item(self, identifier):
-        results = self.process_batch([identifier])
-        return results[0]
-
-    def handle_success(self, identifier):
-        return self.set_presentation_ready(identifier)
```
leonardr: The only real downside of getting rid of the BibliographicCoverageProvider is that the process of creating a work and making it presentation-ready had to be moved into the Monitor.

api/enki.py
```diff
@@ -539,226 +479,253 @@ def queue_response(self, status_code, headers={}, content=None):
...
-        sort_name = element["author"]
-        if not sort_name:
-            sort_name = Edition.UNKNOWN_AUTHOR
+
+        contributors = []
+        sort_name = element.get("author", None) or Edition.UNKNOWN_AUTHOR
         contributors.append(ContributorData(sort_name=sort_name))
-        primary_identifier = IdentifierData(EnkiAPI.ENKI_ID, element["id"])
-        image_url = element["large_image"]
-        thumbnail_url = element["large_image"]
```
leonardr: This error meant the thumbnail and large image were always the same.

api/enki.py
```diff
@@ -773,23 +740,21 @@ def process_book(self, bibliographic, availability):
...
         )
-        if new_edition:
-            bibliographic.apply(edition, self.collection, replace=policy)
+        bibliographic.apply(edition, self.collection, replace=policy)
+        if not license_pool.work:
+            work, is_new = license_pool.calculate_work(
+                even_if_no_author=True
+            )
+            if work:
+                work.set_presentation_ready()
```
leonardr: Here's the code I had to add because I got rid of the BibliographicCoverageProvider.

api/enki.py
leonardr: Update: I did some spot checks and I _think_ it's better to treat the small_image as the thumbnail. I'm not 100% sure but I don't trust 'large_image' to actually be pretty small.

api/enki.py
```diff
@@ -190,39 +162,56 @@ def count_loans_and_holds(patron, pin):
...
-            return response
 
+    def _request(self, method, url, headers, data, params, **kwargs):
+        """Actually make an HTTP request.
+
+        MockEnkiAPI overrides this method.
+        """
+        return HTTP.request_with_timeout(
+            method, url, headers=headers, data=data,
+            params=params, timeout=90, disallowed_response_codes=None,
```
aslagle: Doesn't matter much, but `timeout` and `disallowed_response_codes` are in the kwargs too.

api/enki.py
```diff
@@ -539,226 +479,253 @@ def queue_response(self, status_code, headers={}, content=None):
...
+        """Extract Metadata and CirculationData from a dictionary
+        of information from Enki.
+
+        :return: A Metadata with attached CirculationData.
+        """
+        # TODO: it's not clear what these are or whether we'd find them
+        # useful:
+        #  dateSaved
+        #  length
+        #  publishDate
```
aslagle: publishDate seems like it would be useful.

api/enki.py
leonardr: Thanks, it turns out this does matter a lot.

api/enki.py
leonardr: The problem is I've never seen a value specified for publishDate, and since "publication date" can mean so many different things I can't tell where it should go if it were to show up.

PR #992, opened 2018-07-02
A library may be registered with a registry in a 'testing' stage or in a 'production' stage.

leonardr:
This branch changes the controller that registers a library with a registry so that the admin can say whether they want the library to be registered for testing or for production.

I've updated the [https://github.com/NYPL-Simplified/Simplified/wiki/OPDS-Directory-Registration-Protocol](OPDS Directory Registration Protocol) to accommodate this addition as well as the recent 'contact email' addition. Both of the new fields are optional.

The registry side of this is not implemented yet. The most likely implementation is that the _library's_ desired stage ("testing" or "production") will be tracked separately from the _registry's_ desired stage ("registered", "approved", or "rejected"). A library will only show up in the production feed if it is in production/approved. A library in testing/registered, testing/approved, or production/registered will show up in the testing feed. A library in */rejected will not show up in any feeds at all.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -3385,7 +3385,7 @@ class SettingsController(AdminCirculationManagerController):
 
             integration_id = flask.request.form.get("integration_id")
             library_short_name = flask.request.form.get("library_short_name")
-            state = flask.request.form.get("registration_state")
+            stage = flask.request.form.get("registration_stage")
 
             integration = get_one(self._db, ExternalIntegration,
                                   goal=ExternalIntegration.DISCOVERY_GOAL,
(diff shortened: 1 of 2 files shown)


leonardr:
I'm not sure this is even a good idea -- maybe we should ditch this branch altogether and let the registry admin coordinate out-of-band with the library about when a library should go live.

leonardr:
I'm withdrawing my request for review until I get a better understanding of how we really want this to work.

leonardr:
I think long term we have to have a double-opt-in where both the library and the registry need to be ready to see their library in the main SimplyE app, so this branch is necessary and I would in fact like you to review it.

aslagle:
I don't have any problem with this code, but as far as the registration process/UI I think new registrations should always be for testing, and I wouldn't even have the option to go straight to production.

I'd probably also want the registry to have to approve the registration before the production option is available, but for early libraries it might be more important for a registry admin to be monitoring things rather than a library admin, so it might be better to have the registry admin be the final approval. Either way, we'll need to add something to check the library's status with the registry independent of registering.

PR #993, opened 2018-07-12
Entrypoints tab

EdwinGuzman:
Passing the facets down to the annotator so that feeds have the correct entrypoint value in their link.


api/opds.py
```diff
@@ -454,11 +456,8 @@ def groups_url(self, lane, facets=None):
             **kwargs
         )
 
-    def default_lane_url(self):
-        return self.groups_url(None)
-
-    def lane_url(self, lane):
```
EdwinGuzman: I think this was a duplicate `lane_url` method.

EdwinGuzman:
My only concern is that the default entrypoint of "All" might always clutter the url now.

leonardr:
I think that's a small price to pay. We can clean up the URLs later, let's get everything working first.

api/controller.py
```diff
@@ -653,7 +654,7 @@ def groups(self, lane_identifier):
             worklist=lane, base_class=FeaturedFacets,
             base_class_constructor_kwargs=facet_class_kwargs
         )
-        annotator = self.manager.annotator(lane)
+        annotator = self.manager.annotator(lane, facets)
```
aslagle: Are you sure this is the only place with entrypoints? What about search?

EdwinGuzman: Just added it to the annotator when making a search, thanks. There are more instances of `self = self.manager.annotator` but I don't think they are relevant?

PR #994, opened 2018-07-16
Auth document fixes

leonardr:
This branch fixes the parts of https://jira.nypl.org/browse/SIMPLY-1019 that are easy to fix and well-defined.

* Added some more color schemes to reflect the schemes actually supported by SimplyE.
* Added a configuration setting for the barcode format of an identifier.
* Stopped publishing the maximum password length as the maximum identifier length.

In addition to the automated tests, which test the behavior, I manually inspected the configuration screens to make sure the interface rendered correctly.


api/config.py
```diff
@@ -162,15 +162,25 @@ class Configuration(CoreConfiguration):
...
+                dict(key="amber", label=_("Amber")),
+                dict(key="black", label=_("Black")),
+                dict(key="blue", label=_("Blue")),
+                dict(key="bluegrey", label=_("Blue Grey")),
+                dict(key="brown", label=_("Brown")),
+                dict(key="cyan", label=_("Cyan")),
+                dict(key="darkorange", label=_("Dark Orange")),
+                dict(key="darkpurple", label=_("Dark Purple")),
+                dict(key="green", label=_("Green")),
+                dict(key="grey", label=_("Grey")),
```
aslagle: You changed the spelling of gray here and thus would probably need a migration for anyone who already picked it. Also, why not store these values as hex? Do we expect to change the exact values in the future?

aslagle: Hex would also make it possible to show a preview of the color in the admin interface.

leonardr: They don't correspond to exact values even now: Android uses colors that are more saturated than iOS uses. In the future, they won't necessarily correspond to a single color -- there could be a normal color and a highlight color for a given color scheme.

aslagle: Accounts.json currently has a hex value though, and it's the same across platforms, so they at least use the same base value to generate their color schemes.

leonardr:
On Slack we talked with Greg about this. The iOS Accounts.json uses a specific hex value, but that's going to go away. Android can't support arbitrary hex values because an Android app can only choose from a number of precompiled themes.

PR #995, opened 2018-07-17
Specific lane library root

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1020. It's now possible to set up the library root so it always redirects to a specific lane.

### DIFF
--- a/api/custom_index.py
+++ b/api/custom_index.py
@@ -64,7 +64,7 @@ class CustomIndexView(object):
         protocol = integration.protocol
         if not protocol in cls.BY_PROTOCOL:
             raise CannotLoadConfiguration(
-                "Unregistered custom index protocol: %s" % name
+                "Unregistered custom index protocol: %s" % protocol
             )
         view_class = cls.BY_PROTOCOL[protocol]
         return view_class(library, integration)
@@ -247,3 +247,4 @@ class COPPAGate(CustomIndexView):
         return tag
 
 CustomIndexView.register(COPPAGate)
+CustomIndexView.register(LaneRedirect)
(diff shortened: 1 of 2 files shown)


api/custom_index.py
```diff
@@ -64,7 +64,7 @@ def for_library(cls, library):
         protocol = integration.protocol
         if not protocol in cls.BY_PROTOCOL:
             raise CannotLoadConfiguration(
-                "Unregistered custom index protocol: %s" % name
+                "Unregistered custom index protocol: %s" % protocol
```
leonardr: This is an unrelated bug I found when getting this to work for real.

aslagle:
I don't think this actually solves the problem with Open Bookshelf, since it would also change the root lane in the admin interface. The underlying problem is that we need some lanes that are visible to admins but not patrons.

aslagle:
Maybe it would work if the admin interface was set up to always go to /groups instead of the custom index, but that seems like a hack.

leonardr:
That doesn't sound like a hack... I would say the admin interface should always show the top of the lane hierarchy, even if that's not the server root.

The SimplyE Collection's server root is an OPDS navigation feed, and for that one I would expect the admin interface to show the actual lanes rather than ask me if I'm over 13.

aslagle:
The way the lanes are set up in Open Bookshelf is a hack - the top level lane should actually be the Open Bookshelf lane, I just need somewhere to put the internal admin lanes and there isn't anywhere.

Also, eventually, I want the server root for Open Bookshelf to be a web page rather than an OPDS feed.

aslagle:
That would work better for Open Bookshelf.

leonardr:
Abandoning after a Slack conversation; https://github.com/NYPL-Simplified/circulation/pull/996 is my second attempt at solving this problem.

PR #996, opened 2018-07-17
Custom patron catalogs

leonardr:
This is a replacement for https://github.com/NYPL-Simplified/circulation/pull/995, which I abandoned after a Slack conversation.

Rather than trying to get "library root redirects to a specific lane" working, I've created a new concept called the "custom patron catalog". This code is based on CustomIndex, but it's in a separate file with its own test. This is because I intend this concept to replace the "custom index" concept. We can't replace it yet because iOS versions of SimplyE need the OPDS navigation feed generated by the custom index.

The custom patron catalog is a customizable bit of code that modifies a library's Authentication for OPDS document. Since this document is (will be) the point of entry for SimplyE and other patron-facing interfaces, this lets us customize the behavior of those interfaces without changing the behavior of other views which might be needed by other clients, such as the admin interface.

This branch includes two custom patron catalogs. The `CustomRootLane` lets a library modify the `start` link to tell patron-facing clients that the entry point for the library is the feed for a particular lane. The `COPPAGate` lets a library add a custom authentication mechanism that tells clients to go to one lane if they're under 13 and another lane if they're 13 or up.

### DIFF
--- a/tests/test_custom_patron_catalog.py
+++ b/tests/test_custom_patron_catalog.py
@@ -118,6 +118,7 @@ class TestCustomPatronCatalog(DatabaseTest):
             dict(rel="leave-me-alone", href="link2"),
             dict(rel="replace-me", href="link4", type="text/html"),
         ]
+        eq_(doc['links'], links)
 
 
 class TestCustomRootLane(DatabaseTest):
--- a/tests/test_custom_patron_catalog.py
+++ b/tests/test_custom_patron_catalog.py
@@ -1,0 +1,271 @@
+from nose.tools import (
+    assert_raises_regexp,
+    eq_,
+    set_trace,
+)
+from lxml import etree
+
+from flask import Response
+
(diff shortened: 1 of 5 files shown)


tests/test_custom_patron_catalog.py
```diff
@@ -0,0 +1,271 @@
...
+        CustomPatronCatalog.replace_link(
+            doc, "replace-me", href="link4", type="text/html"
+        )
+
+        # Both replace-me links have been removed, and a a new link
+        # with the same relation has been added.
+        links = [
+            dict(rel="leave-me-alone", href="link2"),
+            dict(rel="replace-me", href="link4", type="text/html"),
+        ]
```
aslagle: You forgot to check that the doc's new links equal this.

aslagle:
I'm not sure this needs to use an ExternalIntegration, but it's nice to keep the ConfigurationSettings together.

PR #998, opened 2018-07-20
Refactor registration code so it can be used outside of controllers

leonardr:
This branch refactors the code for carrying out the OPDS Directory Registration Protocol so that it can be used outside of controllers -- specifically in a command-line script I plan to write.

In a follow-up branch I'll remove a lot of the controller test code (which is very slow) and replace it with  tests of RemoteRegistry and Registration (which should be much faster). For now I just want to get your review of the design of the system, because I'm not familiar with how this protocol works when it comes to shared collections.

The design is that you need an ExternalIntegration corresponding to the remote service you're registering with. You can turn this ExternalIntegration into a RemoteRegistry object. A Registration object represents the connection between that ExternalIntegration and a Library. You can set settings on this relationship with `Registration.setting()`, or you can invoke the registration protocol with `Registration.push()`.

These are the only substantive changes:

1. Creating a Registration object automatically puts the registration into the "failure" state, which will only be changed once the registration is pushed. Previously a registration could be in a null state.
2. All Registration objects have a 'stage', which controls whether the library admin wants the registration to be in production or testing mode. Previously only library registry registrations had a 'stage'. This doesn't make much of a difference because a Registration for a shared collection will always be used in the 'production' stage.

The permission change to bin/novelist_update is unrelated; this is just a convenient branch to put it in.


api/admin/controller.py
```diff
@@ -3291,23 +3301,32 @@ def discovery_services(self):
...
+                        RemoteRegistry.DEFAULT_LIBRARY_REGISTRY_URL
+                    )
+
+            services = self._get_integration_info(goal, protocols)
             return dict(
                 discovery_services=services,
                 protocols=protocols,
             )
 
+        # Beyond this point the user wants to create a new discovery service.
```
aslagle: or edit one.

api/registry.py
```diff
@@ -0,0 +1,342 @@
...
+    # registration may have failed.
+    LIBRARY_REGISTRATION_STATUS = u"library-registration-status"
+    SUCCESS_STATUS = u"success"
+    FAILURE_STATUS = u"failure"
+
+    # A library may be registered in a 'testing' stage or a
+    # 'production' stage. This represents the _library's_ opinion
+    # about whether it's ready for production. The library won't
+    # actually show up in production feeds until the _registry_ also
+    # thinks it should.
```
aslagle: How will the registry's stage be handled?

api/admin/controller.py
```diff
@@ -3316,228 +3335,101 @@ def discovery_services(self):
...
             library = get_one(self._db, Library, short_name=library_short_name)
             if not library:
                 return NO_SUCH_LIBRARY
 
-            integration.libraries += [library]
-            status = ConfigurationSetting.for_library_and_externalintegration(
-                self._db, LIBRARY_REGISTRATION_STATUS, library, integration)
-            status.value = FAILURE
-            registered = self._register_library(integration.url, library, integration, stage=stage, do_get=do_get, do_post=do_post, key=key)
+            registration = Registration(registry, library)
```
aslagle: In the old code, we set the registration status back to failure when you attempted to re-register. That probably wasn't great, since you might still be registered even if the update failed. But we should have some way to track that the update failed.

api/registry.py
leonardr: Both the library and the registry stages will be stored on the registry side. Once we get it working I can have the registry send information about both stages as part of the library's informational OPDS feed, alongside more standard items such as the library's name. Then clients (whether they be SimplyE or the admin interface) will know the status of a given library.

api/admin/controller.py
leonardr: I changed the code to do this consistently inside registry.py -- it was working the way you said for shared collections but not for the library registry.

I agree that later on we should figure out a way to distinguish between "reregistration failed but you're still good" and "this went so badly that you're no longer registered".

PR #1000, opened 2018-07-23
Implement a script to register libraries with a registry in bulk

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1012.
The new script will register any number of libraries with a given registry, in either the 'production' or 'testing' stages.

### DIFF
--- a/api/custom_index.py
+++ b/api/custom_index.py
@@ -68,7 +68,7 @@ class CustomIndexView(object):
         protocol = integration.protocol
         if not protocol in cls.BY_PROTOCOL:
             raise CannotLoadConfiguration(
-                "Unregistered custom index protocol: %s" % name
+                "Unregistered custom index protocol: %s" % protocol
             )
         view_class = cls.BY_PROTOCOL[protocol]
         return view_class(library, integration)
--- a/api/registry.py
+++ b/api/registry.py
@@ -423,9 +423,9 @@ class LibraryRegistrationScript(LibraryInputScript):
             default=RemoteRegistry.DEFAULT_LIBRARY_REGISTRY_URL
         )
         parser.add_argument(
-            '--testing',
-            help="Tell the registry that you do not consider these libraries ready for production.",
-            action='store_true'
+            '--stage',
+            help="Register these libraries in the 'testing' stage or the 'production' stage.",
+            choices=(Registration.TESTING_STAGE, Registration.PRODUCTION_STAGE)
         )
         return parser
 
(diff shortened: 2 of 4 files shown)


api/custom_index.py
```diff
@@ -68,7 +68,7 @@ def for_library(cls, library):
         protocol = integration.protocol
         if not protocol in cls.BY_PROTOCOL:
             raise CannotLoadConfiguration(
-                "Unregistered custom index protocol: %s" % name
+                "Unregistered custom index protocol: %s" % protocol
```
leonardr: An unrelated error I found during real-world testing.

api/registry.py
```diff
@@ -384,3 +406,79 @@ def _process_registration_result(self, catalog, encryptor, desired_stage):
...
+        parser.add_argument(
+            '--registry-url',
+            help="Register libraries with the given registry.",
+            default=RemoteRegistry.DEFAULT_LIBRARY_REGISTRY_URL
+        )
+        parser.add_argument(
+            '--testing',
+            help="Tell the registry that you do not consider these libraries ready for production.",
+            action='store_true'
+        )
```
leonardr: The default is "production" because in general it'll be more dangerous to set a production library to "testing" than to set a library to "production" before you're ready. A library doesn't go into SimplyE production until both the library and the registry think it should. Whereas taking a library from "production" to "testing" can immediately remove it from view for most SimplyE users.

However I'm not sure we want it to be possible to put a library in "testing" once it goes into production, so this could change in the future.

api/registry.py
```diff
@@ -64,6 +68,24 @@ def for_protocol_and_goal(cls, _db, protocol, goal):
...
+    def for_protocol_goal_and_url(cls, _db, protocol, goal, url):
+        """Get a LibraryRegistry for the given protocol, goal, and
+        URL. Create the corresponding ExternalIntegration if necessary.
+        """
+        try:
+            integration = ExternalIntegration.with_setting_value(
+                _db, protocol, goal, ExternalIntegration.URL, url
+            ).one()
+        except NoResultFound:
+            integration = None
```
aslagle: should you also handle multiple results found?

api/registry.py
aslagle: Yeah, I would make it so the default is "testing" if the library isn't registered yet, but if it's in production it stays there.

aslagle:
Did you forget to add a script file for this?

api/registry.py
leonardr: one() will raise an exception if there are multiple results, which I think is appropriate.

api/registry.py
leonardr: Thanks, this is a really good idea, I just had to sit with this a while to figure out how to do it.

PR #1001, opened 2018-07-26
Added focus_area and service_area to library configuration.

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-510 in a minimal way by introducing text fields for `focus_area` and `service_area`, the two Authentication For OPDS (https://github.com/NYPL-Simplified/Simplified/wiki/Authentication-For-OPDS-Extensions) extensions that allow a public library to explain the geographic area it serves.

This implementation can easily handle what is by far the most common case: a library that serves a single city. It can also handle more complex cases, since the text field can contain any JSON-encoded value, but creating such a value will require extra skill or help.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -80,6 +80,10 @@ class Configuration(CoreConfiguration):
 
     # The library-wide logo setting.
     LOGO = "logo"
+
+    # Settings for geographic areas associated with the library.
+    LIBRARY_FOCUS_AREA = "focus_area"
+    LIBRARY_SERVICE_AREA = "service_area"
    
     # Names of the library-wide link settings.
     TERMS_OF_SERVICE = 'terms-of-service'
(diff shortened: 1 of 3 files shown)


api/config.py
```diff
@@ -140,12 +144,32 @@ class Configuration(CoreConfiguration):
...
+        {
+            "key": HELP_WEB,
+            "label": _("Patron support web site"),
+            "description": _("A URL for patrons to get help."),
+            "optional": True,
+        },
+        {
+            "key": HELP_URI,
+            "label": _("Patron support custom integration URI"),
+            "description": _("A custom help integration like Helpstack, e.g. 'helpstack:nypl.desk.com'."),
```
leonardr: I moved these fields above the "designated agent" field because otherwise the description of that field, which mentions the generic patron support address, is unclear.

leonardr:
I filed https://jira.nypl.org/browse/SIMPLY-1039 to cover the part of 510 that's not handled by this branch.

PR #1004, opened 2018-07-31
Update Novelist document format

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1045. There was a performance problem which turned out not to be an issue, but the NoveList API has updated their document format and we were still using the old format.

While updating the document format I discovered that we don't handle the Narrator role correctly. A book with a narrator was showing up with no author. Since Novelist considers 'narrator' an optional field, I've taken 'narrator' out altogether rather than put in the extra work to get it working -- we can always add it later and I just want to get this release out.

### DIFF
--- a/api/novelist.py
+++ b/api/novelist.py
@@ -477,6 +477,8 @@ class NoveListAPI(object):
         i1 = aliased(Identifier)
         i2 = aliased(Identifier)
         roles = list(Contributor.AUTHOR_ROLES)
+        # TODO: We should handle the Narrator role properly, by
+        # setting the 'narrator' field in the NoveList API document.
         # roles.append(Contributor.NARRATOR_ROLE)
 
         isbnQuery = select(
--- a/api/novelist.py
+++ b/api/novelist.py
@@ -87,7 +87,7 @@ class NoveListAPI(object):
 
     medium_to_book_format_type_values = {
         Edition.BOOK_MEDIUM : u"EBook",
-        Edition.AUDIO_MEDIUM : u"AudiobookFormat",
+        Edition.AUDIO_MEDIUM : u"Audiobook",
     }
 
     @classmethod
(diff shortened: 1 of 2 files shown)


api/novelist.py
```diff
@@ -572,24 +575,33 @@ def put_items_novelist(self, library):
...
                 content = json.loads(response.content)
+                logging.info(
+                    "Success from NoveList: %r", response.content
+                )
+            else:
+                logging.error("Data sent was: %r", data)
+                logging.error(
+                    "Error %s from NoveList: %r", response.status_code,
+                    response.content
+                )
```
leonardr: Adding this logging will help libraries communicate with NoveList.

leonardr:
API documentation is here: http://www.novelistcollectiondata.com/api/swagger/ui/index

api/novelist.py
```diff
@@ -477,7 +477,7 @@ def get_items_from_query(self, library):
         i1 = aliased(Identifier)
         i2 = aliased(Identifier)
         roles = list(Contributor.AUTHOR_ROLES)
-        roles.append(Contributor.NARRATOR_ROLE)
+        # roles.append(Contributor.NARRATOR_ROLE)
```
aslagle: Would probably be good to have a comment explaining why this was removed.

api/novelist.py
```diff
@@ -539,31 +539,34 @@ def create_item_object(self, object, currentIdentifier, existingItem):
...
-                Title=title,
-                MediaType=mediaType,
-                Author=author,
-                Role=role,
-                Narrator=narrator
+                isbn=isbn,
+                title=title,
+                mediaType=mediaType,
+                author=author,
+                role=role,
```
aslagle: I don't see role in the API documentation.

leonardr: It's removed from the final document, but we keep track of it so we know whether or not we've found the primary author.

I don't think we need to keep track of it _in the document_, but I didn't want to change that.

PR #1006, opened 2018-08-01
Patron information

jonathangreen:
Not sure if this is more widely useful, but I thought I'd submit a PR in case it is. 

Just a small informational script to grab and print the `patron_data` for a given patron.


bin/informational/patron_information
```diff
@@ -0,0 +1,72 @@
...
+from api.authenticator import LibraryAuthenticator, PatronData
+
+class PatronInformationScript(Script):
+    @classmethod
+    def arg_parser(cls):
+        parser = argparse.ArgumentParser()
+        parser.add_argument(
+            '--library-id',
+            help="ID of the library the patron belongs to.",
+            required=True,
```
leonardr: You can subclass `core.scripts.LibraryInputScript` to get code that takes a library name from the command line and turns it into a `Library` object.

See `circulation/scripts.py:NovelistSnapshotScript` for an example. `parse_command_line` will put a list of libraries in `parsed_libraries` and you can either pick the first one or iterate over the list.

bin/informational/patron_information
```diff
@@ -0,0 +1,72 @@
...
+        else:
+            auth = LibraryAuthenticator.from_config(_db, library).basic_auth_provider
+            patron_data = auth.remote_authenticate(args.barcode, args.pin)
+            self.explain(patron_data)
+
+    def explain(self, patron_data):
+        if patron_data is None or patron_data is False:
+            print "Account does not exist."
+        else:
+            print "Permanent ID: %s" % self.print_data(patron_data.permanent_id)
```
leonardr: Generally we use `self.log.info()` instead of `print` statements. It doesn't matter much here but this makes sure the logging information goes to wherever logs are stored as well as standard output.

Now that I think about it, though, we probably _don't_ want these logs to go where logs are stored, because they contain patron information, so this is better.

jonathangreen:
@leonardr updated based on the code review!

leonardr:
Thanks for doing this work!

PR #1007, opened 2018-08-02
Use a distinct problem detail URI for the situation where a patron's credentials have been suspended.

leonardr:
A very small fix I noticed while looking at https://jira.nypl.org/browse/SIMPLY-485. The problem detail URI we send for a patron whose account has been suspended is the same as the one we send for a patron whose account has expired. This branch gives "suspended" its own unique URI.

### DIFF
--- a/api/problem_details.py
+++ b/api/problem_details.py
@@ -31,7 +31,7 @@ EXPIRED_CREDENTIALS = pd(
 )
 
 BLOCKED_CREDENTIALS = pd(
-      "http://librarysimplified.org/terms/problem/credentials-expired",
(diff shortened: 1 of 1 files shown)


api/problem_details.py
```diff
@@ -31,7 +31,7 @@
 )
 
 BLOCKED_CREDENTIALS = pd(
-      "http://librarysimplified.org/terms/problem/credentials-expired",
+      "http://librarysimplified.org/terms/problem/credentials-suspended",
       403,
       _("Expired credentials."),
```
aslagle: You should also change this title.

PR #1010, opened 2018-08-07
Remove the ServiceStatus object and all references to it.

leonardr:
This branch resolves https://jira.nypl.org/browse/SIMPLY-1066

With the introduction of a more robust self-test system, the old `service_status` controller is no longer necessary. Since this controller was never widely publicized, I think it's fine to just remove it.

The `service_status` controller does offer one feature that's not in the new self-tests: it can borrow a book, fulfill the book, and return it. This feature never worked very well because it required knowing about a book from a certain vendor that was always guaranteed to be available. We can always add this feature to the new self-test system if we figure out a reliable way to make it work.


leonardr:
You could run it as a script or you could visit the /service_status URL to see it as HTML.

PR #1011, opened 2018-08-07
Fixed a couple small problems with registration.

aslagle:
I found these when I tested the library registry locally.

### DIFF
--- a/api/registry.py
+++ b/api/registry.py
@@ -334,12 +334,12 @@ class Registration(object):
 
         :return: Either a ProblemDetail or a requests-like Response object.
         """
-        # Allow 401 so we can provide a more useful error message.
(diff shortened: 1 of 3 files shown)


api/registry.py
```diff
@@ -334,12 +334,12 @@ def _send_registration_request(cls, register_url, payload, do_post):
 
         :return: Either a ProblemDetail or a requests-like Response object.
         """
-        # Allow 401 so we can provide a more useful error message.
+        # Allow 400 and 401 so we can provide a more useful error message.
```
aslagle: 400 is returned for something like a missing help email. That error should be passed on to the admin interface.

leonardr:
Needs a test fix that should be trivial.

PR #1013, opened 2018-08-08
Annotate a library's authentication document with absolute URLs

leonardr:
`url_for` doesn't generate absolute URLs, so this leads to relative URLs in authentication documents, which some clients have problems with. See https://circulation.librarysimplified.org/CLASSICS/authentication_document for a live example.

Using `cdn_url_for` ensures we generate absolute URLs, and we also go right to the CDN if appropriate.


aslagle:
Why do this instead of passing in `_external=True` when calling `url_for`? Maybe it doesn't matter but it seems like it would be better for the auth document to have the non-CDN version of the start URL.

PR #1014, opened 2018-08-08
Merge duplicative odilo files

adriana-alter:
Partially resolves https://jira.nypl.org/browse/SIMPLY-676

Removed odilo files from core and core/tests, and merged their functionality into the circulation odilo files.


aslagle:
Is there going to be another PR for the core piece of this? It would probably be easier to review them together.

aslagle:
It looks like this is also included in #1016 and I already reviewed it there. Probably should have reviewed and merged this first, but I guess let's just close this and use the other one?

aslagle:
#1017 would need to be pulled into the other branch.

PR #1015, opened 2018-08-08
Remove trailing whitespace

leonardr:
This will clean up the code generally and put a stop to future branches that include irrelevant whitespace changes.

I made this change with the following script:

`find ./ -type f -name "*.py" -exec sed -ir 's/ \{1,\}$//g' {} \;`

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -96,7 +96,7 @@ class PatronData(object):
     # Patron was asked to return an item so someone else could borrow it,
     # but didn't return the item.
     RECALL_OVERDUE = 'recall overdue'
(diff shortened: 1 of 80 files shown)


api/authenticator.py
```diff
@@ -28,6 +28,7 @@
     AuthenticationForOPDSDocument,
     OPDSAuthenticationFlow,
 )
+from core.app_server import cdn_url_for
```
aslagle: this isn't trailing whitespace :)

leonardr:
I redid this branch based on a fresh copy of master.

PR #1016, opened 2018-08-08
Merge odilo classes

adriana-alter:
Removed unnecessary Odilo monitor subclasses for getting recent vs. full collections, since their parent class (OdiloCirculationMonitor) already has that functionality.


api/odilo.py
```diff
@@ -65,12 +368,45 @@ class OdiloAPI(BaseOdiloAPI, BaseCirculationAPI, HasSelfTests):
     }
 
     def __init__(self, _db, collection):
-        super(OdiloAPI, self).__init__(_db, collection)
+        # super(OdiloAPI, self).__init__(_db, collection)
```
aslagle: why is this commented out?

api/odilo.py
```diff
@@ -139,14 +531,8 @@ def _make_absolute_url(self, url):
...
-        """Create an OAuth token for the given patron.
-
-        TODO: This code does nothing and should be removed.
-        """
-
-        def refresh(credential):
-            return self.get_patron_access_token(credential, patron, pin)
+    def refresh(credential):
+        return self.get_patron_access_token(credential, patron, pin)
         return self._patron_credential_lookup(patron, refresh)
```
aslagle: I'm not sure what's going on here, did you mean to delete these lines?

api/odilo.py
```diff
@@ -408,6 +826,69 @@ def release_hold(self, patron, pin, licensepool):
...
+            kwargs['allow_redirects'] = True
+
+        response = HTTP.get_with_timeout(url, headers=headers, **kwargs)
+        return response.status_code, response.headers, response.content
+
+    @staticmethod
+    def _do_post(url, payload, headers, **kwargs):
+        # More time please
+        if 'timeout' not in kwargs:
+            kwargs['timeout'] = 600
```
aslagle: This was 60 in core ...

aslagle: Same for _do_get

api/odilo.py
```diff
@@ -490,44 +971,150 @@ def get_url(self, limit, modification_date, offset):
...
-    """Monitor every single book in the Odilo collection.
+        # Modify the record of the request to include the patron information.
+        original_data = self.requests[-1]
 
-    This tells us about books added to the Odilo collection that
-    are not found in our collection.
+        # The last item in the record of the request is keyword arguments.
+        # Stick this information in there to minimize confusion.
+        original_data[-1]['_patron'] = patron
+        original_data[-1]['_pin'] = patron
```
aslagle: Should this be set to `pin` instead of `patron`?

bin/odilo_monitor_full
```diff
@@ -1,13 +0,0 @@
-#!/usr/bin/env python
```
aslagle: Why did you remove this script, is it not needed? Also, just a heads up that it will need to be removed from the crontab in the docker repo too.

api/odilo.py
adriana-alter: I meant to delete it--just fixed that!  (Because I merged the two separate Odilo classes, this no longer has a parent class (other than Object) to be calling super on.)

api/odilo.py
adriana-alter: Yes; see https://github.com/NYPL-Simplified/circulation/pull/1017

api/odilo.py
adriana-alter: It was already set to "patron" in the existing code, but yes, I think it should be set to "pin" instead; I just changed it.

aslagle:
This looks fine but the change from #1017 needs to be merged to this branch.

PR #1018, opened 2018-08-09
Set content type on OpenSearchDocument and added test

adriana-alter:


adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-481

PR #1020, opened 2018-08-09
Self test controller

EdwinGuzman:
This is route and controller pieces for adding previous self test results to the collections endpoint and to getting (previous self tests) and posting (running new self tests) for individual collections.

This is part of [circulation-web #164](https://github.com/NYPL-Simplified/circulation-web/pull/164).


api/admin/controller.py
```diff
@@ -2323,6 +2325,109 @@ def _sitewide_settings_controller(self, configuration_object):
...
+                              description=_("Books will be manually added to the circulation manager, not imported automatically through a protocol."),
+                              settings=[],
+                              ))
+
+        return protocols
+
+    def _get_prior_test_results(self, collection, newCollection):
+        """This helper function returns previous self test results for a given
+        collection if it has a protocol. The `collection` is the collection found
+        in the database and the `newCollection` is the collection being constructed.
```
aslagle: I don't understand this comment. Why is a new collection being constructed?

api/admin/controller.py
```diff
@@ -2323,6 +2325,109 @@ def _sitewide_settings_controller(self, configuration_object):
...
+            if collection["protocolClass"]:
+                if (collection["protocol"] == OPDSImportMonitor.PROTOCOL):
+                    value, results = collection["protocolClass"].run_self_tests(self._db, collection["protocolClass"], self._db, collectionFound, OPDSImporter)
+                else:
+                    value, results = collection["protocolClass"].run_self_tests(self._db, collection["protocolClass"], self._db, collectionFound)
+
+                if (value):
+                    return Response("Successfully ran new self tests", 200)
+                else:
+                    return Response("Failed to run self tests", 200)
```
aslagle: Why is this a 200 and not a problem detail?

api/admin/controller.py
```diff
@@ -2323,6 +2325,109 @@ def _sitewide_settings_controller(self, configuration_object):
...
+
+            return Response("No protocol found in the collection", 200)
```
aslagle: This also seems like it should be a problem detail.

api/admin/controller.py
```diff
@@ -2323,6 +2325,109 @@ def _sitewide_settings_controller(self, configuration_object):
...
+                         FeedbooksOPDSImporter,
+                        ]
+        protocols = self._get_protocols(provider_apis)
+
+        if not identifier:
+            return Response("No Identifier", 200)
+
+        collection = dict()
+        collectionFound = None
+        for col in self._db.query(Collection).order_by(Collection.name).all():
```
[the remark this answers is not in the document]
EdwinGuzman: Ah, thanks. That's much better.

api/admin/controller.py
```diff
@@ -2323,6 +2325,109 @@ def _sitewide_settings_controller(self, configuration_object):
...
+                collection["self_test_results"] = self._get_prior_test_results(col, collection)
+
+        if flask.request.method == 'GET':
+            if "protocolClass" in collection:
+                del collection["protocolClass"]
+            return dict(collection=collection)
+
+        if flask.request.method == "POST":
+            if collection["protocolClass"]:
+                if (collection["protocol"] == OPDSImportMonitor.PROTOCOL):
```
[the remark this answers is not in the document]
EdwinGuzman: I think there's this and possibly Feedbooks that are special cases. The `OPDSImporter` class doesn't inherit from `HasSelfTests` so that's why I'm making this case here.

api/admin/controller.py
EdwinGuzman: Wasn't too sure of the proper way to do that and forgot to update it when making the PR. I'll update it.

api/admin/controller.py
EdwinGuzman: I'm not too happy with how I worked this out. I'll try to see if I can do it differently, but passing just the protocol class is a start.

EdwinGuzman:
Updated based on feedback except the provider apis array.

api/admin/controller.py
```diff
@@ -2323,6 +2325,109 @@ def _sitewide_settings_controller(self, configuration_object):
...
+                         OverdriveAPI,
+                         OdiloAPI,
+                         BibliothecaAPI,
+                         Axis360API,
+                         OneClickAPI,
+                         EnkiAPI,
+                         ODLWithConsolidatedCopiesAPI,
+                         SharedODLAPI,
+                         FeedbooksOPDSImporter,
+                        ]
```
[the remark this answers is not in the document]
leonardr: I agree with this but I think it can go in a separate branch that maybe also addresses the OPDSImporter inconsistency.

api/admin/controller.py
```diff
@@ -2323,6 +2325,107 @@ def _sitewide_settings_controller(self, configuration_object):
...
+        protocolClass = None
+        for col in self._db.query(Collection).filter(Collection.id==int(identifier)):
+            collectionFound = col
+            collection = dict(
+                id=col.id,
+                name=col.name,
+                protocol=col.protocol,
+                parent_id=col.parent_id,
+                settings=dict(external_account_id=col.external_account_id),
+            )
```
leonardr: You can build the `collection` object inside `if flask.request.method == 'GET'`. I think that will save you from having to mock `self._get_prior_test_results` in some of the tests. Not a big deal though.

api/admin/controller.py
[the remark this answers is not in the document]
leonardr: Actually, looking this over I think it would be good to just consolidate these lists in this branch.

api/admin/controller.py
EdwinGuzman: Updated. That was helpful to not have to mock some functions.

api/admin/controller.py
```diff
@@ -2410,26 +2425,93 @@ def _sitewide_settings_controller(self, configuration_object):
...
-                         OdiloAPI,
-                         BibliothecaAPI,
-                         Axis360API,
-                         OneClickAPI,
-                         EnkiAPI,
-                         ODLWithConsolidatedCopiesAPI,
-                         SharedODLAPI,
-                         FeedbooksOPDSImporter,
-                        ]
+    def _get_protocols(self, provider_apis):
```
aslagle: I would rename this `_get_collection_protocols` to avoid confusion with `_get_integration_protocols`.

api/admin/controller.py
```diff
@@ -2410,26 +2425,93 @@ def _sitewide_settings_controller(self, configuration_object):
...
+            else:
+                self_test_results = protocol.prior_test_results(self._db, protocol, self._db, collection)
+
+        return self_test_results
+
+    def collection_self_tests(self, identifier):
+        protocols = self._get_protocols(self.PROVIDER_APIS)
+
+        if not identifier:
+            return Response("No Identifier", 200)
```
aslagle: This looks like another thing that should be a problem detail.

api/admin/controller.py
```diff
@@ -2410,26 +2425,93 @@ def _sitewide_settings_controller(self, configuration_object):
...
+
+            if protocolClass:
+                if (collectionProtocol == OPDSImportMonitor.PROTOCOL):
+                    protocolClass = OPDSImportMonitor
+                    value, results = protocolClass.run_self_tests(self._db, protocolClass, self._db, collection, OPDSImporter)
+                else:
+                    value, results = protocolClass.run_self_tests(self._db, protocolClass, self._db, collection)
+
+                if (value):
+                    return Response("Successfully ran new self tests", 200)
```
aslagle: We usually wrap messages like this in `_(..)` in case we ever decide to translate them.

aslagle: Doesn't need to be done now, but if we wanted to optimize this more in the future I think you could actually return the new test results here so the client only has to make one request.

EdwinGuzman: I'm guessing only the string is wrapped and not the whole `Response`?

aslagle: right

api/admin/controller.py
```diff
@@ -2410,26 +2425,93 @@ def _sitewide_settings_controller(self, configuration_object):
...
+        if flask.request.method == "POST":
+            collection = dict()
+            collectionProtocol = None
+            protocolClass = None
+            for col in self._db.query(Collection).filter(Collection.id==int(identifier)):
+                collection = col
+                collectionProtocol = col.protocol
+
+                if collectionProtocol in [p.get("name") for p in protocols]:
+                    [protocolClassFound] = [p for p in self.PROVIDER_APIS if p.NAME == col.protocol]
```
aslagle: This doesn't work if you have a manual collection. protocols includes manual but PROVIDER_APIS does not.

api/admin/controller.py
```diff
@@ -2471,6 +2555,10 @@ def collections(self):
                                 value = c.external_integration.setting(key).value
                             collection["settings"][key] = value
 
+                    [protocolClassFound] = [p for p in self.PROVIDER_APIS if p.NAME == c.protocol]
```
aslagle: This has the same problem with a manual collection.

api/admin/controller.py
```diff
@@ -2410,26 +2425,93 @@ def _sitewide_settings_controller(self, configuration_object):
...
+                if collectionProtocol in [p.get("name") for p in protocols]:
+                    [protocolClassFound] = [p for p in self.PROVIDER_APIS if p.NAME == col.protocol]
+                    protocolClass = protocolClassFound
+
+            if protocolClass:
+                if (collectionProtocol == OPDSImportMonitor.PROTOCOL):
+                    protocolClass = OPDSImportMonitor
+                    value, results = protocolClass.run_self_tests(self._db, protocolClass, self._db, collection, OPDSImporter)
+                else:
+                    value, results = protocolClass.run_self_tests(self._db, protocolClass, self._db, collection)
```
aslagle: This needs to check if the protocol class is a subclass of HasSelfTests and do something reasonable if not.

EdwinGuzman: Updated.

EdwinGuzman:
@aslagle updated the check for a collection's protocol so the manual import errors properly.

PR #1021, opened 2018-08-10
Annotate authentication document with loans and profile links; annotate patron profile with Short Client Token

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-1092 and https://jira.nypl.org/browse/SIMPLY-1093.

We've updated https://github.com/NYPL-Simplified/Simplified/wiki/User-Profile-Management-Protocol#profile-registry to reserve the `drm` key and define the semantics of the `drm:` namespace. We use the same keys here as we do in the loans feed -- this is just a faster way to get the same information.

### DIFF
--- a/api/opds.py
+++ b/api/opds.py
@@ -923,19 +923,6 @@ class LibraryAnnotator(CirculationManagerAnnotator):
         link_tag.attrib.update(dict(href=fulfill_url))
         return link_tag
 
-    # @classmethod
-    # def _adobe_patron_identifier(self, patron):
-    #     _db = Session.object_session(patron)
-    #     internal = DataSource.lookup(_db, DataSource.INTERNAL_PROCESSING)
-    #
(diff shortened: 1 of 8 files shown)


api/opds.py
```diff
@@ -976,27 +963,15 @@ def adobe_id_tags(self, patron_identifier):
...
-                )
             cached = []
             authdata = AuthdataUtility.from_config(self.library)
             if authdata:
-                # TODO: We would like to call encode() here, and have
-                # the client use a JWT as authdata, but we can't,
-                # because there's no way to use authdata to deactivate
-                # a device. So we've used this alternate technique
-                # that's much smaller than a JWT and can be smuggled
-                # into username/password.
```
leonardr: At this point we're giving up on the idea of ever going back to authdata JWTs.

PR #1022, opened 2018-08-13
Controllers for Adobe ID reset

leonardr:
This branch does the controller work for https://jira.nypl.org/browse/SIMPLY-342. It provides two controller endpoints used to get personal information about a patron and to reset all Adobe-related Credentials associated with them. The personal information is never stored in the database; it's intended to be shown to the library admin.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,11 +26,6 @@ before_install:
   - git submodule update --init --recursive
 
 install:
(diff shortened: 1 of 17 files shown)


EdwinGuzman:
Looks good. We'll probably need new admin routes for the front-end to hit but we can work on that later when we get to the front-end piece of this feature.

PR #1023, opened 2018-08-13
Add a link to create an 'open_book' analytics event

leonardr:
Currently SimplyE modifies the 'alternate' link to find the URL to the circulation manager's analytics event controller. This branch introduces an explicit link in a book's OPDS entry which triggers a particular analytics event.

This resolves https://jira.nypl.org/browse/SIMPLY-518; the client part of the work is https://jira.nypl.org/browse/SIMPLY-1096.

### DIFF
--- a/api/opds.py
+++ b/api/opds.py
@@ -20,6 +20,7 @@ from core.util.opds_writer import (
     OPDSFeed,
 )
 from core.model import (
+    CirculationEvent,
     ConfigurationSetting,
     Credential,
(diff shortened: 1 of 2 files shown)


leonardr:
It would be nice to use a Link Template here, but OPDS 1.x doesn't support Link Templates. Since there's only one analytics event initiated by the client, we decided to punt and maybe revisit this in OPDS 2.

PR #1024, opened 2018-08-14
Send shared secret with registration request, if available

leonardr:
This is the other part of https://jira.nypl.org/browse/SIMPLY-1031. A circulation manager running this code will transparently re-register itself at a new URL.

### DIFF
--- a/api/registry.py
+++ b/api/registry.py
@@ -234,10 +234,16 @@ class Registration(object):
         if isinstance(payload, ProblemDetail):
             return payload
 
+        headers = self._create_registration_headers()
+
+        if isinstance(payload, ProblemDetail):
+            return payload
+
         # Send the document.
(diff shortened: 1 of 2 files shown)


leonardr:
Amy pointed out that the library registry already includes code to extract the shared secret from the `Authorization` header, so we'll be changing this branch to send the shared secret in that header rather than in the form submission.

api/registry.py
```diff
@@ -231,13 +231,19 @@ def push(self, stage, url_for, catalog_url=None, do_get=HTTP.debuggable_get,
 
         # Build the document we'll be sending to the registration URL.
         payload = self._create_registration_payload(url_for, stage)
+        if isinstance(payload, ProblemDetail):
+            return payload
```
aslagle: You have this twice.

PR #1026, opened 2018-08-16
Change 'spine' to 'readingOrder'

adriana-alter:
Partially resolves https://jira.nypl.org/browse/SIMPLY-1105


io7m:
It looks like there are test failures related to the change:

```
AttributeError: 'AudiobookManifest' object has no attribute 'add_reading_order'
```

Is there some code in there still using the old name?

leonardr:
It kind of looks like a submodule issue. We'll look into it.

PR #1027, opened 2018-08-16
Merge duplicative bibliotheca files and their tests

adriana-alter:


api/bibliotheca.py
```diff
@@ -3,15 +3,25 @@
...
 import logging
+import base64
+import urlparse
+import time
+import hmac
+import hashlib
+import os
+import re
+import logging
+
```
leonardr: Some of these are imported twice.

tests/test_bibliotheca.py
```diff
@@ -67,12 +76,25 @@ def setup(self):
...
+    base_path = os.path.split(__file__)[0]
+    resource_path = os.path.join(base_path, "files", "bibliotheca")
+
     @classmethod
     def sample_data(self, filename):
         return sample_data(filename, 'bibliotheca')
+    @classmethod
+    def get_data(cls, filename):
+        path = os.path.join(cls.resource_path, filename)
+        return open(path).read()
```
leonardr: get_data and sample_data do the same thing -- you should be able to remove get_data and change all the references.

PR #1029, opened 2018-08-17
Add library check

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-1124

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -1647,7 +1647,10 @@ class BasicAuthenticationProvider(AuthenticationProvider, HasSelfTests):
         the patron belongs to the library associated with thie BasicAuthenticationProvider."""
 
         patron_info = self._remote_patron_lookup(patron_or_patrondata)
-        return self.enforce_library_identifier_restriction(patron_info.authorization_identifier, patron_info)
+        if patron_info:
+            return self.enforce_library_identifier_restriction(patron_info.authorization_identifier, patron_info)
+        else:
+            return patron_info
 
     @property
     def collects_password(self):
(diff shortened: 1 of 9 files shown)


api/authenticator.py
```diff
@@ -1237,6 +1237,12 @@ def _restriction_matches(cls, field, restriction, match_type):
 
     def enforce_library_identifier_restriction(self, identifier, patrondata):
         """Does the given patron match the configured library identifier restriction?"""
+        if isinstance(patrondata, Patron):
+            if self.library_id == patrondata.library_id:
+                return patrondata
+            else:
+                return False
+
```
leonardr: Since remote_patron_lookup can take a Patron object, this method now needs to be able to handle being given a Patron object.

api/authenticator.py
```diff
@@ -1362,12 +1369,17 @@ def remote_patron_lookup(self, patron_or_patrondata):
...
+
+
             return patron_or_patrondata
         raise ValueError(
             "Unexpected object %r passed into remote_patron_lookup." %
             patron_or_patrondata
         )
 
+    _remote_patron_lookup = remote_patron_lookup
```
aslagle: Without going through this code more carefully, this is confusing. I don't understand the purposes of these two methods - sometimes they are the same, sometimes one calls the other. Why do we need two?

scripts.py
```diff
@@ -654,27 +648,6 @@ def process_patron(self, patron):
             if self.delete:
                 self._db.delete(credential)
 
-
-class BibliographicCoverageProvidersScript(RunCoverageProvidersScript):
```
leonardr: This script wasn't used or tested, and it was importing classes that were moved, so now is a good time to get rid of the code rather than try and get it working again.

PR #1030, opened 2018-08-20
Merge OneClick files from core into circulation

adriana-alter:


api/oneclick.py
```diff
@@ -18,74 +22,98 @@
...
 
     SETTINGS = [
         { "key": ExternalIntegration.PASSWORD, "label": _("Basic Token") },
         { "key": Collection.EXTERNAL_ACCOUNT_ID_KEY, "label": _("Library ID") },
-        { "key": ExternalIntegration.URL, "label": _("URL"), "default": BaseOneClickAPI.PRODUCTION_BASE_URL },
+        { "key": ExternalIntegration.URL, "label": _("URL"), "default": PRODUCTION_BASE_URL },
     ] + BASE_SETTINGS
 
     # The loan duration must be specified when connecting a library to an
     # RBdigital account, but if it's not specified, try one week.
```
leonardr: This comment should be moved up and replace the comment on lines 99-100.

api/oneclick.py
```diff
@@ -101,11 +129,43 @@ class OneClickAPI(BaseOneClickAPI, BaseCirculationAPI, HasSelfTests):
     ]
 
     EXPIRATION_DATE_FORMAT = '%Y-%m-%d'
+    SERVER_NICKNAMES = {
+        "production" : PRODUCTION_BASE_URL,
+        "qa" : QA_BASE_URL,
+    }
+
```
leonardr: Move this up to where those constants are defined.

api/oneclick.py
```diff
@@ -1027,12 +1497,28 @@ def process_access_document(self, access_document):
         expires = datetime.datetime.utcnow() + datetime.timedelta(minutes=14)
         return content_type, content_link, expires
 
+class MockOneClickAPI(OneClickAPI):
 
-class MockOneClickAPI(BaseMockOneClickAPI, OneClickAPI):
+    @classmethod
+    def make_collection(self, _db):
```
leonardr: I don't think this needs to be its own method -- you can move this into the beginning of `mock_collection` and get rid of the `collection = self.make_collection(_db)` line.

api/rbdigital.py
```diff
@@ -1938,34 +1938,34 @@ def invoke(self):
...
-class OneClickCirculationMonitor(CollectionMonitor):
-    """Maintain LicensePools for OneClick titles.
+class RBDigitalCirculationMonitor(CollectionMonitor):
+    """Maintain LicensePools for RBDigital titles.
 
     Bibliographic data isn't inserted into new LicensePools until
     we hear from the metadata wrangler.
     """
-    SERVICE_NAME = "OneClick CirculationMonitor"
+    SERVICE_NAME = "RBDigital CirculationMonitor"
```
leonardr: We need a migration script for this change, so that existing installations don't lose track of when they last ran their monitor.

migration/20170224-2-copy-collection-configuration-into-database.py
```diff
@@ -100,12 +100,12 @@ def convert_axis(_db, library):
     collection.external_account_id = library_id
     collection.external_integration.url = url
 
-def convert_one_click(_db, library):
-    config = Configuration.integration('OneClick')
+def convert_rbdigital(_db, library):
+    config = Configuration.integration('RBDigital')
```
leonardr: Leave this bit alone; this is an old migration script that's looking in a JSON configuration document for OneClick credentials. There should not be any sites that need to run this anymore, but if there are that JSON document will still say "OneClick". The rest of the file is  fine though.

tests/files/rbdigital/response_catalog_all_sample.json
```diff
@@ -18,15 +18,15 @@
     "images": [
       {
         "name": "small",
-        "url": "http://images.oneclickdigital.com/1746552/1746552_image_71x108.jpg"
+        "url": "http://images.rbdigital.com/1746552/1746552_image_71x108.jpg"
```
leonardr: We don't use these URLs for real, but they need to stay the same -- they still work and I don't know what the new ones would be.

tests/files/rbdigital/response_catalog_delta.json
```diff
@@ -20,15 +20,15 @@
         "images": [
           {
             "name": "small",
-            "url": "http://images.oneclickdigital.com/78226/78226_image_71x108.jpg"
+            "url": "http://images.rbdigital.com/78226/78226_image_71x108.jpg"
```
leonardr: Same with these images, and anything that currently matches "images.rbdigital.com".

tests/files/rbdigital/response_patron_checkouts_200_list.json
```diff
@@ -8,55 +8,55 @@
         "id": 1646766,
         "display": "Introduction",
         "filename": "78226_001_IN.mp3",
-        "downloadUrl": "http://api.oneclickdigital.us/v1/media/9781456103859/parts/1646766/download-url?s3=78226&f=78226_001_IN",
+        "downloadUrl": "http://api.rbdigital.us/v1/media/9781456103859/parts/1646766/download-url?s3=78226&f=78226_001_IN",
```
leonardr: It's okay to change this, and we don't use these URLs, but to be accurate it should be changed to api.rbdigital.com

tests/files/rbdigital/response_patron_checkouts_200_list.json
```diff
@@ -71,39 +71,39 @@
...
+        "url": "http://images.rbdigital.com/78226/78226_image_512x512_iTunes.png"
       }
     ],
     "titleId": 0,
     "isbn": "9781456103859",
     "title": "The Greatest: Muhammad Ali",
     "publisher": "Recorded Books, Inc.",
     "expiration": "2016-11-19",
-    "output": "Due to Microsoft's decision to no longer support DRM encryption, protected OneClickdigital titles can no longer be downloaded as .zip files. You can still listen to protected titles in the OneClickdigital Media Manager, stream the titles from your library’s OneClickdigital website, or listen to the title with one of the OneClickdigital mobile apps on Apple, Android, or Kindle Fire devices.",
+    "output": "Due to Microsoft's decision to no longer support DRM encryption, protected rbdigital titles can no longer be downloaded as .zip files. You can still listen to protected titles in the rbdigital Media Manager, stream the titles from your library’s rbdigital website, or listen to the title with one of the rbdigital mobile apps on Apple, Android, or Kindle Fire devices.",
```
leonardr: Change this line back -- we don't know what the output is, but it's more likely what it was before the change.

migration/20170224-2-copy-collection-configuration-into-database.py
```diff
@@ -114,8 +114,8 @@ def convert_rbdigital(_db, library):
 
     collection, ignore = get_one_or_create(
         _db, Collection,
-        protocol=Collection.RBDIGITAL,
-        name="RBDigital"
+        protocol=Collection.ONECLICK,
+        name="OneClick"
```
leonardr: It turns out this whole thing is broken, because we changed the way Collection and ExternalIntegration relate to each other. Since everyone who was going to run this code has already run it, it's fine to leave the whole file as is.

api/rbdigital.py
leonardr: Note to posterity: the migration script was added to circulation, since only circulation managers use RBDigital.

leonardr:
There's no Docker change necessary because the scripts are already named `rbdigital_*`.

PR #1032, opened 2018-08-23
More works are presentation ready circ

adriana-alter:
Partially resolves https://jira.nypl.org/browse/SIMPLY-1055

### DIFF
--- a/migration/20180823-recalculate-work.py
+++ b/migration/20180823-recalculate-work.py
@@ -31,4 +31,4 @@ license_pools = _db.query(LicensePool)
 
 for license_pool in license_pools:
     license_pool.calculate_work()
-    _db.commit
+    _db.commit()
--- a/migration/20180823-recalculate-work.py
+++ b/migration/20180823-recalculate-work.py
@@ -20,13 +20,15 @@ works = _db.query(Work).filter(Work.fiction == None).order_by(Work.id)
 for work in works:
     work.set_presentation_ready_based_on_content()
 
+_db.commit()
+
+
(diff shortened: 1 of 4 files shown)


migration/20180823-recalculate-work.py
```diff
@@ -0,0 +1,32 @@
...
+_db = production_session()
+
+works = _db.query(Work).filter(Work.fiction == None).order_by(Work.id)
+for work in works:
+    work.set_presentation_ready_based_on_content()
+
+license_pools = _db.query(LicensePool)
+                    .join(LicensePool.presentation_edition)
+                    .filter(LicensePool.work == None)
+                    .filter(Edition.author == None)
```
leonardr: This should be `"[Unknown]"`, the string we use to indicate that we couldn't find any authors associated with an Edition.

leonardr:
Re: the migration, on NYPL's circ manager there are 25 works that are not presentation ready and have no fiction status. There are 529 LicensePools that have no Work and no author information.

migration/20180823-recalculate-work.py
```diff
@@ -0,0 +1,34 @@
...
+
+license_pools = _db.query(LicensePool)
+                    .join(LicensePool.presentation_edition)
+                    .filter(LicensePool.work == None)
+                    .filter(Edition.author == "[Unknown]")
+                    .order_by(LicensePool.id)
+
+for license_pool in license_pools:
+    license_pool.calculate_work()
+    _db.commit
```
aslagle: This line is missing its parentheses.

PR #1034, opened 2018-08-28
language links open in new tab

mgiraldo:
or else user loses all their work if they click on it


leonardr:
Hi, Mauricio. :penguin: The config.py change looks great but I'm not sure about the other change -- even if it's helpful to upgrade those packages (which is what I assume is going on) it shouldn't go into this branch.

mgiraldo:
see #1035

PR #1036, opened 2018-08-29
If HasSelfTests.prior_test_results raises an exception, act as though there were no test results.

leonardr:
This fixes a problem I discovered in production. If the Overdrive configuration is incorrect, it's not possible to even instantiate an OverdriveAPI object. This causes `prior_test_results` to raise an unhandled exception. The unhandled exception prevents the admin interface from loading, so there's no way to go in and fix the Overdrive configuration.

### DIFF
--- a/api/admin/controller.py
+++ b/api/admin/controller.py
@@ -2466,10 +2466,10 @@ class SettingsController(AdminCirculationManagerController):
                 # this whole process -- that might prevent an admin from
                 # making the configuration changes necessary to fix
                 # this problem.
-                logging.warn(
-                    "Exception getting self-test results for %r",
-                    collection, exc_info=e
-                )
+                message = _("Exception getting self-test results for collection %s: %s")
+                args = (collection.name, e.message)
+                logging.warn(message, *args, exc_info=e)
+                self_test_results = message % args
 
         return self_test_results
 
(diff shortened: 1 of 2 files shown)


EdwinGuzman:
Thanks for catching that. So the UI will still see what it expects, in this case "no self tests results for ...." or should maybe that message that it could not get self test results also be shown in the UI?

api/admin/controller.py
```diff
@@ -2451,11 +2451,25 @@ def _get_prior_test_results(self, collection, protocolClass):
...
+                )
+            except Exception, e:
+                # This is bad, but not so bad that we should short-circuit
+                # this whole process -- that might prevent an admin from
+                # making the configuration changes necessary to fix
+                # this problem.
+                message = _("Exception getting self-test results for collection %s: %s")
+                args = (collection.name, e.message)
+                logging.warn(message, *args, exc_info=e)
+                self_test_results = message % args
```
aslagle: sorry for the retroactive comment but the args should be passed into `_` - I don't think it will work this way.

EdwinGuzman: It does seem to work. I get this in the front-end `"Exception getting self-test results for collection manual test: Collection protocol is Manual intervention, but passed into EnkiAPI!"`.

aslagle: Interesting. Maybe the `%` operator is converting it to a string and leaving in the `%s`'s since it has those instead of named args.

PR #1044, opened 2018-09-11
Port search code to new system

leonardr:
This branch brings circulation up to date with the core changes of https://github.com/NYPL-Simplified/server_core/pull/942.

Basically, it changes the way we call the `search()` method to reflect the fact that information like "currently active languages" is now pulled in from `SearchFacets.from_request` and doesn't need to be passed in separately.


api/controller.py
```diff
@@ -758,29 +756,21 @@ def search(self, lane_identifier):
...
-            kwargs['language'] = languages
-        kwargs.update(dict(facets.items()))
 
         # Create a function that, when called, generates a URL to the
         # search controller.
+        #
+        # We'll call this one way if there is no query string in the
+        # request arguments, and another way if there is a query
+        # string.
+        make_url_kwargs = dict(facets.items())
```
leonardr: Previously these URLs were not propagating the SearchFacets items. I don't think this made a difference, though, since it would only have affected the `self` link.

tests/test_controller.py
```diff
@@ -2801,11 +2802,35 @@ def test_search(self):
...
+                    check = '%s=%s' % tuple(map(urllib.quote, (k,v)))
+                    assert check in link['href']
+
+            feed_links = feed['feed']['links']
+            for rel in ('next', 'previous', 'self'):
+                [link] = [link for link in feed_links if link.rel == rel]
+
+                assert_propagates_facets(None, link)
+                assert 'q=t' in link['href']
+
```
leonardr: I screwed up the code that this is testing and then noticed that it wasn't being tested, so I added some coverage.

PR #1046, opened 2018-09-18
Generate public/private key pairs as soon as possible, and store them in the database permanently

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1216 by removing the only point in the circulation manager where we do a database commit during a request (rather than at the very end).

Previously we created a public/private key pair, saved the public key to the database, sent an HTTP request to another server, and awaited a response. The other server would make a request to _us_, use the public key in some way, and respond to our original request. We would then use the private key to validate the response to our original request, and delete the public key from the database.

Now we create a public/private key pair for the site when the CirculationManager is initialized -- that is, on the very first request. We create a public/private key pair for a library when its LibraryAuthenticator is created -- again, on the first request.

Unless the very first request made to the circulation manager kicks off this registration process, this ensures that the public key is always available when the callback request is made.

These keys are intended to be transient, and we could just as easily create new ones every single time a CirculationManager or LibraryAuthenticator object is created -- they just need to be in the database _before_ that complicated request begins.

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -3,7 +3,12 @@ import re
 from nose.tools import set_trace
 import contextlib
 from copy import deepcopy
+
+from Crypto.PublicKey import RSA
+from Crypto.Cipher import PKCS1_OAEP
+
 from flask_babel import lazy_gettext as _
+
 from core.config import (
     Configuration as CoreConfiguration,
     CannotLoadConfiguration,
@@ -507,6 +512,19 @@ class Configuration(CoreConfiguration):
             private_setting.value = key.exportKey()
 
         return public_setting.value, private_setting.value
+
+    @classmethod
+    def cipher(cls, key):
+        """Create a Cipher for a public or private key.
+
+        This just wraps some hard-to-remember Crypto code.
+
+        :param key: A string containing the key.
+
+        :return: A Cipher object which will support either
+        encrypt() (public key) or decrypt() (private key).
+        """
+        return PKCS1_OAEP.new(RSA.import_key(key))
 
 
 @contextlib.contextmanager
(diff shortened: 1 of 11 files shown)


api/config.py
```diff
@@ -469,6 +479,53 @@ def configuration_contact_uri(cls, library):
...
+
+        TODO: This could go into ConfigurationSetting or core Configuration.
+
+        :param public_setting: A ConfigurationSetting for the public key.
+        :param private_setting: A ConfigurationSetting for the private key.
+
+        :return: A 2-tuple (public key, private key)
+        """
+        # Prevent bad code where public and private key are mixed up.
+        if 'public' not in public_setting.key.lower():
```
aslagle: Why not check that it equals the constant?

leonardr: I thought this would be useful for storing other public/private key pairs, but we don't need that now.

tests/test_config.py
```diff
@@ -13,6 +17,75 @@
...
+            self._db, "irrelevant"
+        )
+        irrelevant.value = "blue"
+
+        # If you pass in a pair of ConfigurationSettings, and at least
+        # one of them is missing its value, a new key pair is created.
+        public_key, private_key = Configuration.key_pair(public, private)
+        assert 'BEGIN PUBLIC KEY' in public_key
+        assert 'BEGIN RSA PRIVATE KEY' in private_key
+        eq_(public_key, public.value)
```
aslagle: You could check that this is different from “old value”

aslagle: I guess you did because “ BEGIN PUBLIC KEY” wasn’t there

PR #1047, opened 2018-09-18
Store key pairs in a single Configurationsetting object, and fix call to push()

leonardr:
A simple follow-up branch to https://github.com/NYPL-Simplified/circulation/pull/1046. I should have done an end-to-end test on a local circ manager.

Storing the public and private key in the same ConfigurationSetting will avoid the possibility of a race condition. I don't think that explains the problems I've been seeing, but it seems like a good safety measure.


leonardr:
Edwin and I went over this in person.

PR #1048, opened 2018-09-21
Fix the 2018083-recalculate-work script and rename it so it will run on the next migration.

leonardr:
The 2018083-recalculate-work migration script had a couple of Python syntax problems, so when it was used in a real migration it crashed. I got it to work against a real site and renamed the file so that it will run again on the next migration.

This isn't a crucial migration -- it just makes some books available that weren't available before -- so we can decide whether or not to do a new release just for this script.


migration/20180921-recalculate-work.py
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python
+import os
+import sys
+import logging
+from pdb import set_trace
+bin_dir = os.path.split(__file__)[0]
+package_dir = os.path.join(bin_dir, "..")
```
leonardr: The package_dir had an extra ".."  -- I think it was copied from a migration script in core.

migration/20180921-recalculate-work.py
```diff
@@ -0,0 +1,39 @@
...
+a = 0
+for work in works:
+    logging.info("Processing %s", work.title)
+    work.set_presentation_ready_based_on_content()
+    if not a % 10:
+        _db.commit()
+_db.commit()
+
+
+license_pools = _db.query(LicensePool).join(LicensePool.presentation_edition).filter(LicensePool.work == None).filter(Edition.title != None).filter(Edition.author == "[Unknown]").order_by(LicensePool.id)
```
leonardr: This was on separate lines with no continuation characters, so Python didn't parse it.

PR #1049, opened 2018-09-25
Use new model

leonardr:
This branch makes the circulation manager use the refactored model package. We made a lot of changes in model so that this branch could stay simple.


api/app.py
```diff
@@ -30,14 +30,13 @@
...
-    db_url = Configuration.database_url(testing)
+
+    db_url = Configuration.database_url()
     if autoinitialize:
         SessionManager.initialize(db_url)
     session_factory = SessionManager.sessionmaker(db_url)
     _db = flask_scoped_session(session_factory, app)
     app._db = _db
-    if autoinitialize:
-        SessionManager.initialize_data(_db)
```
leonardr: initialize() calls this, and calling it again was redundant.

api/app.py
```diff
@@ -30,14 +30,13 @@
 @app.before_first_request
 def initialize_database(autoinitialize=True):
     testing = 'TESTING' in os.environ
-    db_url = Configuration.database_url(testing)
+
+    db_url = Configuration.database_url()
```
leonardr: database_url checks `TESTING` so there's no need to pass it in.

PR #1050, opened 2018-09-28
Propagate facet links

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-1255 for the `audiobooks` branch. This fix will need to be ported to the `master` branch.

1255 describes a few different problems, but in every case the issue is that the active Facets object is not used in a call to `url_for` or equivalent.

Refactoring the feed-generation scripts may be overkill but it removes almost all of the code from the `do_generate` methods, making it easy to inspect `do_generate` and verify that the `Facets` object is used in every method call.


api/app.py
```diff
@@ -30,7 +30,7 @@
 @app.before_first_request
 def initialize_database(autoinitialize=True):
     testing = 'TESTING' in os.environ
-    db_url = Configuration.database_url(testing)
+    db_url = Configuration.database_url()
```
leonardr: This is an unrelated change caused by the Configuration refactoring, necessary to get the app server to start. I believe this change made it into `master` after the `audiobooks` branch was created.

api/opds.py
```diff
@@ -725,13 +725,25 @@ def annotate_feed(self, feed, lane):
...
+            facet_kwargs.update(dict(self.facets.items()))
+
         # Add a 'search' link if the lane is searchable.
         if lane and lane.search_target:
+            # TODO: On the top level lane, when the library supports
+            # multiple entry points and the currently selected entry
+            # point is the default, the entrypoint of the search form
+            # should be EverythingEntryPoint, even though the
+            # currently selected entry point is not
+            # EverythingEntryPoint.
```
leonardr: This TODO corresponds to https://jira.nypl.org/browse/SIMPLY-1262.

scripts.py
```diff
@@ -334,11 +336,29 @@ def arg_parser(cls, _db):
...
+
+        :param _db: A database connection.
+        :param cmd_args: A mock set of command-line arguments, to use instead
+           of looking at the actual command line.
+        :param testing: If this method creates a CirculationManager object,
+           this value will be passed in to its constructor as its value for
+           `testing`.
+        :param manager: A mock CirculationManager object, to use instead
+           of creating a new one (creating a CirculationManager object is
+           very time-consuming).
```
leonardr: `manager` is new here.

scripts.py
```diff
@@ -517,32 +590,50 @@ def do_generate(self, lane):
...
+        title = lane.display_name
+        library = lane.get_library(self._db)
+        annotator = self.app.manager.annotator(lane, facets=facets)
+        url = annotator.feed_url(lane, facets=facets, pagination=pagination)
+        feed_class = feed_class or AcquisitionFeed
+        return feed_class.page(
+            _db=self._db, title=title, url=url, lane=lane,
+            annotator=annotator, facets=facets, pagination=pagination,
+            force_refresh=True
+        )
```
leonardr: This script didn't previously have a problem with propagating its Facets.

scripts.py
```diff
@@ -553,41 +644,46 @@ def should_process_lane(self, lane):
...
         if not lane.children:
             return False
-        if self.max_depth and lane.depth > self.max_depth:
+        if self.max_depth is not None and lane.depth > self.max_depth:
             return False
         return True
 
-    def do_generate(self, lane):
-        feeds = []
-        annotator = self.app.manager.annotator(lane)
```
leonardr: This is where the problem was -- we were creating an annotator that didn't know which facets were selected.

scripts.py
```diff
@@ -412,18 +432,57 @@ def process_library(self, library):
...
+                )
+                a = time.time()
+                feed = self.do_generate(lane, facets, pagination)
+                b = time.time()
+                if feed:
+                    cached_feeds.append(feed)
+                    self.log.info(
+                        "Took %.2f sec to make %d bytes.", (b-a), len(feed)
+                    )
+        total_size = sum(len(x) for x in cached_feeds)
```
aslagle: you're not using this total size

scripts.py
```diff
@@ -517,32 +590,50 @@ def do_generate(self, lane):
         )
         chosen_collections = self.collections or [default_collection]
 
-        for order in chosen_orders:
-            if order not in allowed_orders:
-                logging.warn("Ignoring unsupported ordering %s" % order)
+        for entrypoint_name in chosen_entrypoints:
+            entrypoint = EntryPoint.BY_INTERNAL_NAME.get(entrypoint_name)
+            if not entrypoint:
+                logging.warn("Ignoring unknown entry point %s" % entrypoint_name)
```
aslagle: should you also check that a user-specified entry point is one of the library's enabled entry points?

PR #1051, opened 2018-10-01
Propagate facet links from master

adriana-alter:
This branch ports https://github.com/NYPL-Simplified/circulation/pull/1050 to the master branch.


scripts.py
```diff
@@ -202,15 +204,12 @@ def process_batch(self, batch):
 class MetadataCalculationScript(Script):
 
     """Force calculate_presentation() to be called on some set of Editions.
-
```
leonardr: There are a number of places where blank lines were removed when they shouldn't have been -- probably done automatically by Atom.

PR #1054, opened 2018-10-02
Handle default entry point differently for search controller

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-1262 as well as a related problem.

If the currently selected EntryPoint is selected because it's the default (as opposed to because the user deliberately chose it), then a search from that feed will no longer search the currently selected EntryPoint -- it will search the entire collection.

The related problem: if you hit the search controller directly, without specifying an EntryPoint, you want to search everything, not the library's default entrypoint.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -32,6 +32,7 @@ from core.app_server import (
     HeartbeatController,
     URNLookupController,
 )
+from core.entrypoint import EverythingEntryPoint
 from core.external_search import (
     ExternalSearchIndex,
     MockExternalSearchIndex,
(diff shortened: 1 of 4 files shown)


leonardr:
Amy had concerns with this context switch from "you're looking at ebooks" to "you're looking at everything". She pointed out that user research shows that patrons have a specific format in mind when they perform a search and will generally not accept a desired book in an undesirable format.

We don't do the context switch in hopes of getting patrons to switch formats. The goal is to handle the common case where someone loads up SimplyE and goes directly to search without explicitly choosing an entry point. They do a search first, and if they don't immediately see what they want, they refine their search -- possibly by choosing one of the other entry points, possibly by typing something else into the search box.

Rather than propagating the "ebooks" entry point from browse to search, this branch is propagating the fact that the user hasn't (as far as we can tell) made a decision yet. In general, library catalog searches work this way -- they mix different media types in a single result stream and allow the patron to refine the search afterwards.

PR #1058, opened 2018-10-05
Default entrypoint in scripts

leonardr:
This branch (hopefully) completes my work for https://jira.nypl.org/browse/SIMPLY-1262. After testing my initial 1262 work I found that `entrypoint_is_default` was set for `Facets` objects created from a request, but not for `Facets` objects created by scripts. Since the feed with the problematic search link is generated by a script, the code to generate the correct search link was never running.

When you generate a `Facets` object in the context of a script, the default entrypoint is the library's first enabled entrypoint. The default entrypoint is _not_ the first entrypoint generated by the script. Depending on how the script is run, the library's default entrypoint might not be used at all.


scripts.py
```diff
@@ -569,11 +569,14 @@ def facets(self, lane):
         allowed_orders = library.enabled_facets(Facets.ORDER_FACET_GROUP_NAME)
         chosen_orders = self.orders or [default_order]
 
-        allowed_entrypoints = [
+        allowed_entrypoint_names = [
```
leonardr: This is just renaming the variable to make it clear that this list contains names, not EntryPoint objects.

PR #1059, opened 2018-10-09
Brought in configuration that was previously in config files in the web catalog.

aslagle:
This branch adds library settings for background and foreground colors and header links for the web catalog, and also adds login button images for some authentication providers.


aslagle:
I had a couple builds get stuck on the scoped session test - but I don't think I did anything here that would affect it.

api/admin/controller.py
```diff
@@ -2153,6 +2155,27 @@ def libraries(self):
...
+                  contrast_check_url=contrast_check_url))
+
+        # Verify that header links and labels are the same length.
+        header_links = flask.request.form.getlist(Configuration.WEB_HEADER_LINKS)
+        header_labels = flask.request.form.getlist(Configuration.WEB_HEADER_LABELS)
+        if len(header_links) != len(header_labels):
+            self._db.rollback()
+            return INVALID_CONFIGURATION_OPTION.detailed(
+                _("There must be the same number of web header links and web header labels."))
+
```
leonardr: I think you could avoid the need to rollback by checking these things before the `create` call.

leonardr: Are the links and labels guaranteed to come out of the form in the same order?

api/authenticator.py
```diff
@@ -921,12 +921,31 @@ def create_authentication_document(self):
...
+        header_urls = ConfigurationSetting.for_library(
+            Configuration.WEB_HEADER_LINKS, library).json_value
+        if header_urls:
+            header_labels = ConfigurationSetting.for_library(
+                Configuration.WEB_HEADER_LABELS, library).json_value
+            header_links = []
+            for (url, label) in zip(header_urls, header_labels):
+                header_links.append(dict(href=url, title=label));
+            doc["web_header_links"] = header_links
+
```
leonardr: These things should be documented as [extensions](https://github.com/NYPL-Simplified/Simplified/wiki/Authentication-For-OPDS-Extensions) to Authentication For OPDS.

For `web_colors` I think it would be good enough to call it `web_color_scheme` to fit with `color_scheme`.

For `web_header_links` it's more difficult. I can think of a couple solutions:

* Leave `web_header_links` alone. Define it as being the same as `links` but for links specific to the 'web header' role. The problem with this is that `links` requires a `rel` for each link. So we'd either define `rel` as optional for `web_header_links` or make up a fake rel for it.

* Incorporate `web_header_links` into `links` somehow, with "web header" as the `role` (which would be different from a `rel` -- `rel` describes the relationship between the two pages, `role` describes the purpose of the link within the document.)

* Create a separate object called `web` and reuse the standard names `links` and `color_scheme` inside that object, possibly defining them differently.

leonardr: This doesn't absolutely need to be resolved in this branch but it would be best to reach an agreement before this is deployed on servers we don't control.

aslagle: I'd like to resolve this now, I was waiting for your feedback before I updated the wiki.

I kind of like the second solution. Which one do you prefer?

api/authenticator.py
```diff
@@ -1979,13 +2005,19 @@ def _authentication_flow_document(self, _db):
...
             labels=dict(login=unicode(localized_identifier_label),
                         password=unicode(localized_password_label)),
             inputs = dict(login=login_inputs,
                           password=password_inputs)
         )
+        if self.LOGIN_BUTTON_IMAGE:
+            # TODO: I'm not sure if logo is appropriate for this, since it's a button
+            # with the logo on it rather than a plain logo. Perhaps we should use plain
+            # logos instead.
+            flow_doc["links"] = [dict(rel="logo", href=url_for("static_image", filename=self.LOGIN_BUTTON_IMAGE, _external=True))]
```
leonardr: I do think plain logos would be better. It seems like we can turn them into buttons on the client side. Unless there are background color issues.

For the regular auth document `logo` I insert the actual logo as a `data:` URI. This reduces the load on the server since the client (the library registry for a circ manager, SimplyE for the library registry) almost always wants that logo. If you did this here the `static_image` controller would not be necessary.

api/authenticator.py
leonardr: I also prefer the second solution. The problem is coming up with a `rel` -- we'd need to either let the user specify an appropriate `rel` for each link (which is kind of silly because in a web context it doesn't matter), or make up a generic fake `rel`, or have 'web header link' be the rel, or leave it blank and serve invalid Authentication For OPDS documents.

leonardr: Letting 'web header link' be the `rel` is the simplest solution but it twists what a `rel` is for.

api/authenticator.py
aslagle: I agree about plain logos, but I'm not sure if I want to change that now. It was originally done as a button rather than a plain logo because Clever provides a button, and students who use Clever will likely be familiar with the button. The button images are the same ones used by the Open eBooks mobile apps, though the apps still have them hard-coded.

For these images, using the URL means they don't need to be downloaded until the user tries to log in, so the initial request for the auth document can be faster. The client probably wants the images most of the time, but doesn't need them immediately.

api/routes.py
```diff
@@ -512,3 +512,11 @@ def loadstorm_verify(code):
...
 def health_check():
     return Response("", 200)
+
+@app.route("/images/<filename>")
+def static_image(filename):
+    directory = os.path.join(os.path.abspath(os.path.dirname(__file__)), "..", "resources", "images")
+    cache_timeout = ConfigurationSetting.sitewide(
+        app._db, Configuration.STATIC_FILE_CACHE_TIME
+    ).int_value
+    return flask.send_from_directory(directory, filename, cache_timeout=cache_timeout)
```
leonardr: I looked up the docs for `send_from_directory` and `send_file` and didn't see an explicit indication that it handles hack attempts like `/images/../../../etc/passwd`. But it does say `send_from_directory` is "secure" so I assume that's what it's referring to.

leonardr: Based on the `send_from_directory` docs I think `/images/<path:filename>` would be a better route.

leonardr: Also there's no test for this controller in this branch. Is it feasible to add one?

api/authenticator.py
aslagle: What if we used "related" as the `rel` for these and kept "web header link" as the `role`?

api/authenticator.py
leonardr: Eventually SimplyE will be retrieving the circ manager's auth document, and in that case it will not need the logo, so maybe we should eventually switch over to having a controller to serve that logo.

api/routes.py
aslagle: This is a route and we don't have tests for any routes. Should I make a controller for it? The admin static files routes could use it as well.

api/authenticator.py
leonardr: I like that, and I think the role should be `navigation` based on https://www.w3.org/1999/xhtml/vocab.

api/routes.py
leonardr: You're right, I mentally put this code in a controller. I do think it's complex enough to deserve a controller.

api/authenticator.py
aslagle: That does make sense. We already get footer links from the feed (about, TOS, etc).

api/admin/controller.py
```diff
@@ -21,6 +21,8 @@
 from PIL import Image, ImageDraw, ImageFont
 import textwrap
 from StringIO import StringIO
+import wcag_contrast_ratio
```
EdwinGuzman: 🏆

api/admin/controller.py
EdwinGuzman: That could get confusing if a user has a lot of links to enter and they can't see them all in the small form input. Does the flask break up the list by commas or spaces?

api/authenticator.py
```diff
@@ -2254,6 +2286,11 @@ def _authentication_flow_document(self, _db):
             links=[dict(rel="authenticate",
                         href=self._internal_authenticate_url(_db))]
         )
+        if self.LOGIN_BUTTON_IMAGE:
+            # TODO: I'm not sure if logo is appropriate for this, since it's a button
+            # with the logo on it rather than a plain logo. Perhaps we should use plain
+            # logos instead.
+            flow_doc["links"] += [dict(rel="logo", href=url_for("static_image", filename=self.LOGIN_BUTTON_IMAGE, _external=True))]
```
EdwinGuzman: What's the difference between this one and the one above?

aslagle: They're the same but the basic and oauth auth providers have different implementations of this function so I put it in two places.

tests/test_controller.py
```diff
@@ -3839,3 +3839,35 @@ def test_scoped_session(self):
...
+        filename = "CleverLoginButton280.png"
+        with open(os.path.join(directory, filename)) as f:
+            expected_content = f.read()
+
+        with self.app.test_request_context("/"):
+            response = self.app.manager.static_files.image(filename)
+
+        eq_(200, response.status_code)
+        eq_(expected_content, response.response.file.read())
+
```
leonardr: Can you test a 404 as well? Or use a mock to verify that `send_from_directory` is being called, and then we trust that `send_from_directory` does the right thing.

api/authenticator.py
```diff
@@ -921,12 +921,31 @@ def create_authentication_document(self):
...
         if description:
             doc['color_scheme'] = description
 
+        # Add the library's web colors, if it has any.
+        background = ConfigurationSetting.for_library(
+            Configuration.WEB_BACKGROUND_COLOR, library).value
+        foreground = ConfigurationSetting.for_library(
+            Configuration.WEB_FOREGROUND_COLOR, library).value
+        if background or foreground:
+            doc["web_colors"] = dict(background=background, foreground=foreground)
```
leonardr: I would still like to see this called `web_color_scheme` to match `color_scheme`.

api/authenticator.py
leonardr: Yeah, that looks a lot better.

leonardr:
Other than my very minor comments this looks ready to go.

PR #1061, opened 2018-10-09
SIMPLY-1170 add enki pdf format check

datalogics-tsmith:
This pull request adds functionality to the Enki API parser to distinguish between PDF and EPUB books coming from the Enki API

Addresses [SIMPLY-1170](https://jira.nypl.org/browse/SIMPLY-1170)


api/enki.py
```diff
@@ -223,7 +223,7 @@ def recent_activity(self, since):
         for element in data['result']['recentactivity']:
             identifier = IdentifierData(Identifier.ENKI_ID, element['id'])
             yield parser.extract_circulation(
-                identifier, element['availability']
+                identifier, element['availability'], {} #The recent activity API does not include format info
```
leonardr: `availability` is supposed to be a dictionary, so if the value is missing then it's appropriate to substitute an empty dictionary (as on line 611, below). But `formattype` is supposed to be a scalar value, so if the value is missing then the best substitute is `None`.

api/enki.py
```diff
@@ -608,12 +608,12 @@ def extract_bibliographic(self, element):
             subjects=subjects,
         )
         circulationdata = self.extract_circulation(
-            primary_identifier, element.get('availability', {})
+            primary_identifier, element.get('availability', {}), element.get('formattype', {})
```
leonardr: Again, the best default value for `formattype` is None.

api/enki.py
```diff
@@ -626,12 +626,18 @@ def extract_circulation(self, primary_identifier, availability):
...
         formats = []
-        formats.append(
-            FormatData(
-                content_type=Representation.EPUB_MEDIA_TYPE,
-                drm_scheme=drm_type
+        if formattype != {}:
+            if formattype == 'PDF':
+                content_type = Representation.PDF_MEDIA_TYPE
+            elif formattype == 'EPUB':
+                content_type=Representation.EPUB_MEDIA_TYPE
```
leonardr: This is basically fine, apart from checking `if formattype is None` instead of `if formattype != {}`, but if there's ever a new, unrecognized `formattype` this is going to blow up because `content_type` won't get defined.

Something like this would be more reliable:
```
content_type = None
if formattype == 'PDF':
    content_type = Representation.PDF_MEDIA_TYPE
elif formattype == 'EPUB':
    content_type = Representation.EPUB_MEDIA_TYPE

if content_type:
  ...
else:
   self.log.error("Unrecognized formattype: %s", formattype)
```

That way if there's an unexpected formattype, the title won't get a FormatData at all, and an error will be logged.

tests/test_enki.py
```diff
@@ -505,6 +505,18 @@ def test_extract_bibliographic(self):
...
+        data = json.loads(self.get_data("pdf_document_entry.json"))
+        parser = BibliographicParser()
+        m = parser.extract_bibliographic(data['result'])
+        assert isinstance(m, Metadata)
+
+        # The book is available as an ACS-encrypted PDF.
+        circulation = m.circulation
+        assert isinstance(circulation, CirculationData)
+        [format] = circulation.formats
+        eq_(Representation.PDF_MEDIA_TYPE, format.content_type)
```
leonardr: The comment mentions this is an ACS-encrypted PDF, so check the `drm_scheme` while you're at it.

datalogics-tsmith:
@leonardr The requested changes are up

tests/test_enki.py
```diff
@@ -505,6 +505,19 @@ def test_extract_bibliographic(self):
...
+        parser = BibliographicParser()
+        m = parser.extract_bibliographic(data['result'])
+        assert isinstance(m, Metadata)
+
+        # The book is available as a non-ACS PDF.
+        circulation = m.circulation
+        assert isinstance(circulation, CirculationData)
+        [format] = circulation.formats
+        eq_(Representation.PDF_MEDIA_TYPE, format.content_type)
+        eq_(DeliveryMechanism.NO_DRM, format.drm_scheme)
```
leonardr: OK, that's not that I was expecting, but looking at pdf_document_entry.json it is in fact a DRM-free PDF. So line 515 is wrong. Just change that comment and you'll be good to go.

datalogics-tsmith: I did, It now says "The book is available as a non-ACS PDF" or should it be worded differently?

leonardr: No, that's great.

PR #1065, opened 2018-10-11
When only one entry point is enabled, use that entry point as the SearchFacets default

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-1293. The problem was that the EverythingEntryPoint was used as the SearchFacets default, even when there was only one enabled entry point and so the EverythingEntryPoint wasn't even an option. This made AcquisitionFeed.add_entrypoint_links() in core decide to explicitly add a link to the selected entry point (because the selected entry point differed from the default), but the problem wasn't there -- the default was wrong.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -757,6 +757,21 @@ class OPDSFeedController(CirculationManagerController):
         )
         return feed_response(feed)
 
+    def _load_search_facets(self, lane):
+        entrypoints = list(flask.request.library.entrypoints)
+        if len(entrypoints) > 1:
+            # There is more than one enabled EntryPoint.
+            # By default, search them all.
+            default_entrypoint = EverythingEntryPoint
+        else:
+            # There is only one enabled EntryPoint,
+            # and no need for a special default.
+            default_entrypoint = None
+        return load_facets_from_request(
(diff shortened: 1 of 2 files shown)


api/controller.py
```diff
@@ -757,18 +757,29 @@ def _crawlable_feed(self, library, title, url, lane, annotator=None):
         )
         return feed_response(feed)
 
+    def _load_search_facets(self, lane):
+        entrypoints = list(flask.request.library.entrypoints)
+        if len(entrypoints) > 1:
+            # There is more than one enabled EntryPoint.
+            # By default, search them all.
+            default_entrypoint = EverythingEntryPoint
```
aslagle: What if there are multiple enabled entry points, but EverythingEntryPoint isn't one of them?

leonardr: EverythingEntryPoint is still the right choice. When you do a search, and there is more than one entry point, and you don't specify an entry point, you're searching everything, even if EverythingEntryPoint isn't an option for the site as a whole.

Selecting EverythingEntryPoint from the admin interface makes "All" show up as an option for navigation. It's always there for search, as long as there are multiple entry points enabled for navigation.

PR #1066, opened 2018-10-11
Backported fix to SIMPLY-1293.

leonardr:
Backport of https://github.com/NYPL-Simplified/circulation/pull/1065 to the `audiobooks` branch.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -759,6 +759,21 @@ class OPDSFeedController(CirculationManagerController):
         )
         return feed_response(feed)
 
+    def _load_search_facets(self, lane):
+        entrypoints = list(flask.request.library.entrypoints)
+        if len(entrypoints) > 1:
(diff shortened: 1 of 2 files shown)


api/controller.py
```diff
@@ -759,6 +759,21 @@ def _crawlable_feed(self, library, title, url, lane, annotator=None):
...
+            default_entrypoint = EverythingEntryPoint
+        else:
+            # There is only one enabled EntryPoint,
+            # and no need for a special default.
+            default_entrypoint = None
+        return load_facets_from_request(
+            worklist=lane, base_class=SearchFacets,
+            default_entrypoint=default_entrypoint,
+        )
+
```
leonardr: The bit of test_controller.py that uses this method in `assert_propagates_facets` isn't in this branch branch, so this method is only used once here.

PR #1067, opened 2018-10-11
Make SIP Connections short lived

jonathangreen:
LYRASIS was seeing issues using SIP2 authentication with the Yavapai library network. There were a large number of SIP processes were being created in the ILS (SirsiDynix Symphony) which was causing ILS performance problems.

This patch updates the SIP client to:
  - Connect Socket
  - Send Login message - `93` (if necessary)
  - Send Patron Information message - `63`
  - Send End Patron Session - `35`
  - Close Socket

There is no attempt at long lived connections or socket reuse, this process is done every time. Based on conversations with Yavapai this appears to be how Overdrives SIP2 authentication works.

This PR is based on a slack conversation with @leonardr and @alsrlw.

### DIFF
--- a/api/sip/__init__.py
+++ b/api/sip/__init__.py
@@ -100,8 +100,7 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
                 client = SIPClient(
                     target_server=server, target_port=port,
                     login_user_id=login_user_id, login_password=login_password,
-                    location_code=location_code, separator=field_separator,
-                    connect=connect
+                    location_code=location_code, separator=field_separator
                 )
 
         except IOError, e:
--- a/api/sip/__init__.py
+++ b/api/sip/__init__.py
@@ -116,9 +116,10 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
             )
 
     def _remote_patron_lookup(self, patron_or_patrondata):
-        return self.patron_information(
+        info = self.patron_information(
             patron_or_patrondata.authorization_identifier, None
         )
+        return self.info_to_patrondata(info, False)
 
     def remote_authenticate(self, username, password):
         """Authenticate a patron with the SIP2 server.
@@ -135,7 +136,7 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
         return self.info_to_patrondata(info)
 
     @classmethod
-    def info_to_patrondata(cls, info):
+    def info_to_patrondata(cls, info, validate_password=True):
 
         """Convert the SIP-specific dictionary obtained from
         SIPClient.patron_information() to an abstract,
@@ -146,7 +147,7 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
             # library. Don't return any data.
             return None
 
-        if info.get('valid_patron_password') == 'N':
+        if info.get('valid_patron_password') == 'N' and validate_password:
             # The patron did not authenticate correctly. Don't
             # return any data.
             return None
--- a/api/sip/__init__.py
+++ b/api/sip/__init__.py
@@ -98,7 +98,7 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
             if self.client:
                 sip = self.client
             else:
-                sip = self.client(
+                sip = SIPClient(
                     target_server=self.server, target_port=self.port,
                     login_user_id=self.login_user_id, login_password=self.login_password,
                     location_code=self.location_code, separator=self.field_separator
@@ -106,7 +106,7 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
             sip.connect()
             sip.login()
             info = sip.patron_information(username, password)
-            sip.end_session()
+            sip.end_session(username, password)
             sip.disconnect()
             return info
 
--- a/api/sip/__init__.py
+++ b/api/sip/__init__.py
@@ -84,33 +84,39 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
         super(SIP2AuthenticationProvider, self).__init__(
             library, integration, analytics
         )
+
+        self.server = integration.url
+        self.port = integration.setting(self.PORT).int_value
+        self.login_user_id = integration.username
+        self.login_password = integration.password
+        self.location_code = integration.setting(self.LOCATION_CODE).value
+        self.field_separator = integration.setting(self.FIELD_SEPARATOR).value or '|'
+        self.client = client
+
+    def patron_information(self, username, password):
         try:
-            server = None
-            if client:
-                if callable(client):
-                    client = client()
+            if self.client:
+                sip = self.client
             else:
-                server = integration.url
-                port = integration.setting(self.PORT).int_value
-                login_user_id = integration.username
-                login_password = integration.password
-                location_code = integration.setting(self.LOCATION_CODE).value
-                field_separator = integration.setting(
-                    self.FIELD_SEPARATOR).value or '|'
-                client = SIPClient(
-                    target_server=server, target_port=port,
-                    login_user_id=login_user_id, login_password=login_password,
-                    location_code=location_code, separator=field_separator
+                sip = self.client(
+                    target_server=self.server, target_port=self.port,
+                    login_user_id=self.login_user_id, login_password=self.login_password,
+                    location_code=self.location_code, separator=self.field_separator
                 )
+            sip.connect()
+            sip.login()
+            info = sip.patron_information(username, password)
+            sip.end_session()
+            sip.disconnect()
+            return info
 
         except IOError, e:
             raise RemoteIntegrationException(
-                server or 'unknown server', e.message
+                self.server or 'unknown server', e.message
             )
-        self.client = client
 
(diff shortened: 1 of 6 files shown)


api/sip/client.py
```diff
@@ -252,127 +248,94 @@ def __init__(self, target_server, target_port, login_user_id=None,
             escaped = self.separator
         self.separator_re = re.compile(escaped + "([A-Z][A-Z])")
 
+        self.sequence_number = 0
```
leonardr: I don't think we need sequence numbers at all since a given connection is never sending more than one message. (Resent messages don't count because you're not supposed to increment the sequence number.)

leonardr: The sequence number needs to be associated with the session. Previously the session and the SIPClient class were one and the same. Now there's no separate "session" object. If a SIPClient is used by multiple threads, we'll have multiple sessions going on at once and sharing a sequence number.

Try creating a SIPSession object which manages the connection and the sequence number.

api/sip/client.py
```diff
@@ -252,127 +248,94 @@ def __init__(self, target_server, target_port, login_user_id=None,
...
     def patron_information(self, *args, **kwargs):
         """Get information about a patron, possibly also verifying their
         password.
         """
+        connection = self.connect()
+        self.login(connection)
+        information = self._patron_information(connection, *args, **kwargs)
+        self.end_session(connection, *args, **kwargs)
+        self.disconnect(connection)
+        return information
```
leonardr: I think you can get to a point where the code looks like this:

```
with self.connect() as session:
    information = self._patron_information(session, *args, **kwargs)
return information
```

`connect()` is a context manager that creates a `SIPSession` and yields it. When the `SIPSession` is created it makes a socket connection and logs in. Once execution returns to the context manager, it ends the session and disconnects the context manager.

api/sip/client.py
```diff
@@ -405,10 +367,54 @@ def login_response_parser(self, message):
...
+            named.patron_identifier.required,
+            named.screen_message,
+            named.print_line
+        )
+
     def patron_information_request(
             self, patron_identifier, patron_password="", institution_id="",
             terminal_password="",
-            language=None, summary=None, start_item="", end_item=""
+            language=None, summary=None
```
leonardr: What's the effect of this change?

api/sip/client.py
jonathangreen: How about if the connection just moves back into the class, so the session and object are one in the same? Does that address your concern?

api/sip/client.py
jonathangreen: The parameters weren't used in the function.

api/sip/client.py
```diff
@@ -609,7 +614,6 @@ def parse_response(self, data, expect_status_code, *fields):
                     if field.required and field in required_fields_not_seen:
                         required_fields_not_seen.remove(field)
             i += 2
-            field = fields_by_sip_code
```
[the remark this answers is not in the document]
jonathangreen: 👍🏻 yes that is correct

leonardr:
It would also work to create a new `SIPClient` every time you need to make a request -- I don't know what the overhead would be but probably not too bad.

api/sip/client.py
leonardr: :ok:

api/sip/client.py
jonathangreen: @leonardr I see the elegance in the solution you propose of creating a `SIPSession` object and I will look into making that change. 

I am wondering about your comment on threading though. Is this a theoretical concern in case this is used by multiple threads in the future or do you see a threading problem as the code is used right now? 

My understanding is that currently each thread gets its own `SIP2AuthenticationProvider` and each of those has its own `SIPClient` so I don't see where `SIPClient` can be called into from multiple thread currently. Just want to make sure I am understanding correctly.

leonardr: The multi-threading thing is theoretical for us, but at the moment this is the most functional Python SIP2 client on the Internet, so I'd like to keep it usable in someone else's multi-threaded scenario (or our future scenario).

As I said in another comment, it should also be fine to treat `SIPClient` as lasting the length of a SIP session. In that case you would establish a connection immediately and the sequence number could be associated with the `SIPClient` object.  `SIPClient`s would be disposable, `SIPAuthenticationProvider` would create them all the time instead of holding on to them, and you wouldn't use them across threads.

leonardr:
While you're changing this, would you mind also making this (hopefully) simple change?
https://jira.nypl.org/browse/SIMPLY-1097

jonathangreen:
@leonardr I believe I addressed your feedback by moving the socket back into the `SIPClient` class, and making sure there is a new `SIPClient` for each request.

I can take a look at [SIMPLY-1097](https://jira.nypl.org/browse/SIMPLY-1097), but I'd rather get this merged first before adding more to this PR.

api/sip/__init__.py
```diff
@@ -84,34 +84,39 @@ def __init__(self, library, integration, analytics=None, client=None, connect=Tr
...
+                    target_server=self.server, target_port=self.port,
+                    login_user_id=self.login_user_id, login_password=self.login_password,
+                    location_code=self.location_code, separator=self.field_separator
                 )
+            sip.connect()
+            sip.login()
+            info = sip.patron_information(username, password)
+            sip.end_session()
+            sip.disconnect()
+            return info
```
leonardr: I was thinking the connect/login/end_session/disconnect loop would happen _inside_ the call to patron_information, but this also works, since the only 'real' SIP method we ever call is patron_information. We can revisit this if we ever start calling another method.

api/sip/__init__.py
```diff
@@ -84,33 +84,39 @@ def __init__(self, library, integration, analytics=None, client=None, connect=Tr
...
+        self.client = client
+
+    def patron_information(self, username, password):
         try:
-            server = None
-            if client:
-                if callable(client):
-                    client = client()
+            if self.client:
+                sip = self.client
```
leonardr: The `if callable` code should still be around, so that we can test by passing in a mock class rather than an instance, but it can be moved to line 94.

jonathangreen: Do we actually need this though? All the tests work by passing in an instance?

api/sip/__init__.py
leonardr: I also momentarily thought that this meant your treatment of `sequence_number` was wrong, but I think you have it right. Every time you call connect(), `sequence_number` is reset. So as long as a SIPClient is only used by one thread at a time, everything will be fine.

That said, I think there's no longer any reason to get rid of `socket_lock`. Putting it back would make it possible for multiple threads to use a SIPClient as part of a single session. Experiment with that and see if you can get it to work, but I'm okay with merging this code as is because the circ manager uses one SIPClient per thread.

api/sip/__init__.py
leonardr: You're right, if the tests are passing we don't need it.

jonathangreen:
Not sure what happened to the test here, maybe some blip with dependancies? 
https://github.com/NYPL-Simplified/circulation/pull/1067/commits/136fd70b3fec93390348beeab21b3028760aad2c shouldn't have broken anything, and the tests still run for me locally. Looking at the error it doesn't seem to be in any of the SIP code. 

Maybe someone could restart travis for me? 

Test failure for reference: 
```
======================================================================
ERROR: Failure: ImportError (No module named ordered_dict)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nose/loader.py", line 418, in loadTestsFromName
    addr.filename, addr.module)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nose/importer.py", line 47, in importFromPath
    return self.importFromDir(dir_path, fqname)
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/nose/importer.py", line 94, in importFromDir
    mod = load_module(part_fqname, fh, filename, desc)
  File "/home/travis/build/NYPL-Simplified/circulation/tests/__init__.py", line 4, in <module>
    from core.testing import (
  File "/home/travis/build/NYPL-Simplified/circulation/core/testing.py", line 14, in <module>
    from config import Configuration
  File "/home/travis/build/NYPL-Simplified/circulation/core/config.py", line 21, in <module>
    from util.http import IntegrationException
  File "/home/travis/build/NYPL-Simplified/circulation/core/util/http.py", line 3, in <module>
    import requests
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/requests/__init__.py", line 97, in <module>
    from . import utils
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/requests/utils.py", line 26, in <module>
    from ._internal_utils import to_native_string
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/requests/_internal_utils.py", line 11, in <module>
    from .compat import is_py2, builtin_str, str
  File "/home/travis/virtualenv/python2.7.14/lib/python2.7/site-packages/requests/compat.py", line 47, in <module>
    from urllib3.packages.ordered_dict import OrderedDict
ImportError: No module named ordered_dict
Ran 1 test in 0.001s
```

jonathangreen:
Sounds good. I merged the code from #1068 into this branch, so everything is here.

PR #1068, opened 2018-10-11
_remote_patron_lookup returning dict

jonathangreen:
`_remote_patron_lookup` is suposed to return a `PatronData` object but it is returning a dictionary of values from SIPClient.

Call the function to tranform it into a proper `PatronData` object and additionally add some exception handling.

Prevents this error that I was seeing when patrons are logging in: 
```
{"host": "6436e280159d", "name": "root", "level": "ERROR", "timestamp": "2018-10-08T16:50:58.598228", "app": "simplified", "traceback": "Traceback (most recent call last):
  File \"/var/www/circulation/env/local/lib/python2.7/site-packages/flask/app.py\", line 1813, in full_dispatch_request
    rv = self.dispatch_request()
  File \"/var/www/circulation/env/local/lib/python2.7/site-packages/flask/app.py\", line 1799, in dispatch_request
    return self.view_functions[rule.endpoint](**req.view_args)
  File \"/var/www/circulation/api/routes.py\", line 147, in decorated
    return f(*args, **kwargs)
  File \"/var/www/circulation/api/routes.py\", line 116, in wrapped_function
    resp = make_response(f(*args, **kwargs))
  File \"/var/www/circulation/api/routes.py\", line 89, in decorated
    patron = app.manager.index_controller.authenticated_patron_from_request()
  File \"/var/www/circulation/api/base_controller.py\", line 42, in authenticated_patron_from_request
    patron = self.authenticated_patron(header)
  File \"/var/www/circulation/api/base_controller.py\", line 69, in authenticated_patron
    self._db, authorization_header
  File \"/var/www/circulation/api/authenticator.py\", line 470, in authenticated_patron
    return self.invoke_authenticator_method(\"authenticated_patron\", _db, header)
  File \"/var/www/circulation/api/authenticator.py\", line 467, in invoke_authenticator_method
    return getattr(self.library_authenticators[short_name], method_name)(*args, **kwargs)
  File \"/var/www/circulation/api/authenticator.py\", line 729, in authenticated_patron
    return self.basic_auth_provider.authenticated_patron(_db, header)
  File \"/var/www/circulation/api/authenticator.py\", line 1296, in authenticated_patron
    patron = self.authenticate(_db, header)
  File \"/var/www/circulation/api/authenticator.py\", line 1779, in authenticate
    patrondata = self.remote_patron_lookup(patrondata)
  File \"/var/www/circulation/api/authenticator.py\", line 1664, in remote_patron_lookup
    return self.enforce_library_identifier_restriction(patron_info.authorization_identifier, patron_info)
AttributeError: 'dict' object has no attribute 'authorization_identifier'", "message": "Exception in web app: 'dict' object has no attribute 'authorization_identifier'", "filename": "app_server.py"}
```

### DIFF
--- a/api/sip/client.py
+++ b/api/sip/client.py
@@ -249,7 +249,7 @@ class SIPClient(Constants):
         self.separator_re = re.compile(escaped + "([A-Z][A-Z])")
 
         self.sequence_number = 0
-
+        self.connection = None
         self.login_user_id = login_user_id
         self.login_password = login_password
         if login_user_id and login_password:
@@ -259,11 +259,11 @@ class SIPClient(Constants):
             # We're implicitly logged in.
             self.must_log_in = False
 
-    def login(self, connection):
+    def login(self):
         """Log in to the SIP server if required."""
         if self.must_log_in:
             response = self.make_request(
-                connection, self.login_message, self.login_response_parser,
+                self.login_message, self.login_response_parser,
                 self.login_user_id, self.login_password, self.location_code
             )
             if response['login_ok'] != '1':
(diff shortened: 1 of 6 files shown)


jonathangreen:
Tagging @leonardr for review.

jonathangreen:
So I've been digging into this one a little bit. It looks like the bug was introduced in this PR: 
https://github.com/NYPL-Simplified/circulation/pull/1029

From my reading of it, no SIP2 patron that has never logged in before is able to log in without the error shown above because we call `_remote_authenticate` here: 
https://github.com/NYPL-Simplified/circulation/blob/master/api/authenticator.py#L1779

leonardr:
That makes sense to me.

jonathangreen:
So in trying to sort out a solution to this problem I'm wondering why `_remote_patron_lookup` was defined for SIP at all. 

Before #1029 doing a `remote_patron_lookup` was a no-op for SIP since `remote_authenticate` returns a full `PatronData` object. That seems correct to me, but I think I'm missing something here.

leonardr:
Most of the time we get a PatronData object as a side effect of authentication. The exception is in the
admin interface for a given library shows a patron lookup form. That means we need to look up a patron in the ILS, and then verify that they are a patron of a particular library, without authenticating them.

jonathangreen:
@leonardr here is an updated fix for this issue. The diff for the relevant bit is here 
https://github.com/NYPL-Simplified/circulation/pull/1068/commits/2b6039d897b97f509eb23f5c03f336fd5812d69d. Its dependant on the work in #1067. 

I'm open to handling this however. I can push that commit to #1067 and roll it all into one, or we can keep this PR open, and just merge this one after #1067. 

Let me know what you'd like to do.

api/sip/client.py
```diff
@@ -252,127 +248,82 @@ def __init__(self, target_server, target_port, login_user_id=None,
...
-        with self.socket_lock:
-            sock = socket.socket(socket.AF_INET, socket.SOCK_STREAM)
-            try:
-                sock.connect((self.target_server, self.target_port))
-            except TypeError:
-                raise IOError(
-                    "Could not connect to %s:%s" % (
-                        self.target_server, self.target_port
-                    )
+        try:
```
leonardr: One last thing. In this method, check if there's an active connection already, and call `disconnect` if necessary. That way if there's any bad code left in the system it won't be able to create dangling connections.

jonathangreen: Good call. I think the garbage collector should handle this since connection goes out of scope here and a new connection is created. Good to look both ways and be explicit though. Added that in.

PR #1070, opened 2018-10-15
Search integration test

adriana-alter:


api/config.py
```diff
@@ -186,8 +186,8 @@ class Configuration(CoreConfiguration):
         },
         {
             "key": DEFAULT_NOTIFICATION_EMAIL_ADDRESS,
-            "label": _("Default email address to use when sending vendor hold notifications"),
-            "description": _('This should be an address controlled by the library which rejects or trashes all email sent to it. Vendor hold notifications contain sensitive patron information, but <a href="https://confluence.nypl.org/display/SIM/About+Hold+Notifications" target="_blank">cannot be forwarded to patrons</a> because they contain vendor-specific instructions.')
```
aslagle: This looks like an accidental revert?

aslagle:
It would be nice to have some info about the requirements for running these tests. I'm guessing they only work against NYPL's index so others won't be able to run them?

integration_tests/test_search.py
```diff
@@ -1,25 +1,2078 @@
...
+        """
+        self.evaluate_search(hits)
+        first = hits[0]
+        self.evaluate_first(first)
+        self.evaluate_hits(hits)
+
+    def evaluate_search(self, hits):
+        """Evaluate the search itself."""
+        # By default, a search passes the check if it returns anything
+        # as opposed to returning nothing.
```
aslagle: There may be cases where we want to return nothing, but maybe not with NYPL's collection since it's so large.

integration_tests/test_search.py
```diff
@@ -1,25 +1,2078 @@
...
+
+    Most of the results must also be by that author.
+    """
+
+    def __init__(self, author, accept_book_about_author=False, threshold=0.5):
+        super(SpecificAuthor, self).__init__(author=author)
+        self.accept_book_about_author = accept_book_about_author
+        # I changed the threshold to 0 because searches involving less prolific authors were
+        # failing otherwise
+        self.threshold = 0
```
aslagle: Why not change the default argument instead?

integration_tests/test_search.py
```diff
@@ -1,25 +1,2078 @@
...
+        self.search("origin", FirstMatch(title="Origin"))
+
+    def test_simple_title_match_goldfinch(self):
+        self.search("goldfinch", FirstMatch(title="Goldfinch"))
+
+    def test_simple_title_match_beach(self):
+        self.search("Manhattan beach", FirstMatch(title="Manhattan Beach"))
+
+    def test_simple_title_match_testing(self):
+        self.search("The testing", FirstMatch(title="The testing"))
```
aslagle: This doesn't look like a real search :)

integration_tests/test_search.py
```diff
@@ -1,25 +1,2078 @@
...
+        self.search("blind assassin", FirstMatch(title="Blind Assassin"))
+
+    def test_simple_title_match_dry(self):
+        self.search("the dry", FirstMatch(title="The Dry"))
+
+    def test_simple_title_match_origin(self):
+        self.search("origin", FirstMatch(title="Origin"))
+
+    def test_simple_title_match_goldfinch(self):
+        self.search("goldfinch", FirstMatch(title="Goldfinch"))
```
aslagle: This is probably "The Goldfinch", right?

integration_tests/test_search.py
```diff
@@ -1,25 +1,2078 @@
...
+                AtLeastOne(title=re.compile("cosmetics")),
+            ]
+        )
+
+    def test_nonexistent_title_tower(self):
+        # NOTE: there is no book with this title.  The most
+        # likely scenario is that the user meant "The Dark Tower."  This
+        # doesn't currently work on either version of ES.
+
+        self.search("The night tower", FirstMatch(title="The Dark Tower"))
```
aslagle: It seems significantly less important for something like this to work where there's no good result. I'd be careful about including things like this and the next test in the comparison between the versions.

integration_tests/test_search.py
```diff
@@ -1,25 +1,2078 @@
...
+        self.search(
+            "Sebald after nature",
+            Common(author=re.compile("(w. g. sebald|w.g. sebald)"))
+        )
+
+class TestGenreMatch(SearchTest):
+
+    def test_sf(self):
+        # NOTE: This doesn't work.  On ES1, the top result has "science fiction" in
+        # the title, but has no genre; on ES6, the top result has "science fiction" if __name__ == '__main__':
```
aslagle: some extraneous code in this comment

integration_tests/test_search.py
```diff
@@ -1,25 +1,2078 @@
...
+        )
+
+    def test_travel_florida(self):
+        # NOTE: this doesn't work on either version of ES, but it's closer on ES6.
+
+        # Keyword
+        self.search(
+            "Florida",
+            Common(subject=re.compile("(travel|guide|fodors)"), first_must_match=False)
+        )
```
aslagle: This could also be Lauren Groff's Florida

leonardr:
Summary of the test results:

* 110 tests work in ES1 and ES6
* 58 tests fail in ES1 and ES6
* 13 tests fail in ES1 and work in ES6
* 17 tests work in ES1 and fail in ES6

I'm probably overfitting the data, since we don't have that many failures to generalize from, but overall I think ES6 handles these searches better:

* Topic searches
* We don't have the exact book the patron is searching for, but we do have books on related subjects

And I think overall  ES1 handles these searches better:

* Partial title searches that also contain typos
* Searches that combine a partial title name with a partial author name
* Searching for a genre when there are books whose titles include the genre name
* Searching for a specific book in a series, e.g. "foundation book 1" (This shouldn't work at all--not sure what's happening).

I think these differences are caused by internal changes to Elasticsearch between versions 1 and 6, and not something we can address as part of the ES6 upgrade -- instead we should upgrade and start thinking of ways to improve search overall.

leonardr:
> It would be nice to have some info about the requirements for running these tests. I'm guessing they only work against NYPL's index so others won't be able to run them?

The general format is that we check whether the search results did a good job of fulfilling the hypothetical patron's intent. If someone is searching for a specific book, we check whether that book was returned. For an author search we check whether the results are dominated by that author's books (or books about them in some cases). For a topic search we check the genres or topics of the results without being concerned with specific titles.

So, I think we should get similar results running this test against the index of any public library with a similarly-sized collection. This is also why I'm not too concerned about a title search test that fails because NYPL doesn't have the book. (Though I agree it shouldn't be counted in the totals.)

I can run this test against Brooklyn's search index to check this, but NYPL's and Brooklyn's books are purchased by the same people, so I don't know if that's a good test.

integration_tests/test_search.py
leonardr: I added a set of tests that validate that you get no results if you type a bunch of junk into the search box.

integration_tests/test_search.py
```diff
@@ -1,25 +1,2078 @@
...
+        if result.target_age.lower == 18:
+            return False, how_old_is_the_kid, "18+"
+
+        expect_upper, expect_lower = max(how_old_is_the_kid), min(how_old_is_the_kid)
+        expect_set = set(range(expect_lower, expect_upper + 1))
+
+        # There are a handful of books for which the ages are flipped--i.e. the
+        # upper age is listed under result.target_age.lower, and vice versa;
+        # hence using max() and min() to extract the values rather
+        # than just using the keys to access them.
```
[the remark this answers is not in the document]
leonardr: You are right. I investigated and filed https://jira.nypl.org/browse/SIMPLY-1320.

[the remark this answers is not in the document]
leonardr: I'll remove this code. Rebuilding the search index will solve this problem.

integration_tests/test_search.py
leonardr: It's a real book. I'll add an author check to make it more clear.

integration_tests/test_search.py
leonardr: It is; we have it under both names.

integration_tests/test_search.py
leonardr: I think we can take out the "a hand to reach" test altogether. It's really unclear what the patron is searching for there. 

Here there is a good result, we're just not sophisticated enough to provide it. If you search Google for "the night tower" it shows you _The Dark Tower_ results, and I think that's probably what this person was searching for. This might be past the edge of where we can reasonably expect to get with our search engine, though.

integration_tests/test_search.py
leonardr: That's true. I changed this to a similar test that isn't likely to have a novel with the same name.

integration_tests/test_search.py
```diff
@@ -1,25 +1,2078 @@
...
+        self.search(
+            "Harry potter",
+            Common(series="Harry Potter", threshold=0.9)
+        )
+
+    def test_harry_potter_2(self):
+        # NOTE: this does bring up the target book as the first result, but, ideally,
+        # other books from the same series would also show up at some point.  (Not
+        # necessarily a problem; just would be nice to have, particularly since there
+        # are other searches that do work that way.)
```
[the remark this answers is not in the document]
leonardr: They do match the search terms, at least the 'harry potter and the' part. It's better to match "harry potter and the" than to match "chambers" or "of secrets", which some of the other results here do.

leonardr:
Once we move over the ES6 I have a vague idea of further enhancing this with an annotation called `@fails`  or something. This annotation would invert the normal meaning of the test -- we would expect it to fail based on what we know about the search engine, and the overall test would fail if that failure _didn't_ happen.

With that we can get to a state where test_search.py accurately reflects the behavior of the search index: that is, it runs with no failures. At that point it's really easy to test improvements.

If the improvement works,it will cause a failure (because a test we expected to fail now succeeds). If the improvement has negative side effects it will cause a failure (because a test that used to succeed now fails). Either way, we need to investigate and get things back into the equilibrium state.

leonardr:
I think this branch is ready for another look, and then to be merged. I made some major changes which don't affect the functionality much but will make it easier to do further work.

I broke up tests that ran searches in a loop. If we ran six searches and the third one failed then we were losing information, because the other three searches never ran, and it was also unclear exactly why the test had failed. Every test now runs one and only one search.

There are now 249 tests in this class, up from 198 the last time we did a comparison. Of these, 158 tests pass in ES1 (up from 127) and 155 pass in ES6 (up from 123).  So, there's a net gain of 51 tests. ES1 has a net gain of 31 passing tests and ES6 has a net gain of 32 passing tests.

I also organized the unit tests into classes by what type of search the patron is trying to do. I got  rid of classes like TestKidsSearches and TestDifficultSearches which were categories we imposed on the searches. Here are the classes:

* TestGibberish - if you put junk into the search box you shouldn't get any results
* TestTitleMatch
* TestUnownedTitle - title matches for titles not in NYPL's collection
* TestMisspelledTitleSearch
* TestPartialTitleSearch
* TestMixedTitleAuthorMatch
* TestSubtitleMatch
* TestAuthorMatch
* TestGenreMatch
* TestSubjectMatch
* TestSeriesMatch
* TestSeriesTitleMatch
* TestCharacterMatch -- looking for books with a certain fictional character
* TestAgeRangeRestriction -- adding an age range restricts the results to children's books

There are also three classes for title searches like "law of the mountain man" where the title contains some word that activates some other part of the search algorithm:

* TestTitleGenreConflict
* TestTitleAuthorConflict
* TestTitleAudienceConflict

On top of that there are a number of VariantSearchTest subclasses: TestTheHateUGive, TestTimothyZahn, TestTravel, etc. The tests in these classes used to be a loop inside a single test. These classes run a bunch of slight variant searches and evaluate the results with the same Evaluator object.

PR #1071, opened 2018-10-16
Moved migration script in from core.

leonardr:
Circulation portion of https://github.com/NYPL-Simplified/server_core/pull/974. Updated submodule and moved in a migration script. The only change I made to the migration script was changing the code that modifies sys.path so that 'from core.model import' referred to the right place.

### DIFF
--- a/api/rbdigital.py
+++ b/api/rbdigital.py
@@ -1608,9 +1608,22 @@ class RBDigitalRepresentationExtractor(object):
             Identifier.RB_DIGITAL_ID, rbdigital_id
         )
 
+        # medium is both bibliographic and format information.
+
+        # options are: "eBook", "eAudio"
+        rbdigital_medium = book.get('mediaType', None)
+        if rbdigital_medium and rbdigital_medium not in cls.rbdigital_medium_to_simplified_medium:
+            cls.log.error(
+                "Could not process medium %s for %s", rbdigital_medium, rbdigital_id)
+
+        medium = cls.rbdigital_medium_to_simplified_medium.get(
+            rbdigital_medium, Edition.BOOK_MEDIUM
+        )
+
         metadata = Metadata(
             data_source=DataSource.RB_DIGITAL,
             primary_identifier=primary_identifier,
+            medium=medium,
         )
 
         if include_bibliographic:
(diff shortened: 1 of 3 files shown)


leonardr:
Not sure what's going on here but it's something about the Travis environment, not anything to do with this code.

Here's a branch that doesn't change anything and its integration tests fail the same way: https://github.com/NYPL-Simplified/circulation/pull/1072

leonardr:
A new version of urllib3 was just released and I think one of its changes, probably https://github.com/urllib3/urllib3/pull/1409, caused this problem. I don't know whether this needs to be fixed in urllib3, requests, or if we need to do something, but for now, refusing to use the new release seems to fix the problem.

requirements.txt
```diff
@@ -5,6 +5,7 @@ elasticsearch-dsl<2.0.0
 pillow
 psycopg2
 requests==2.18.4
+urllib3!=1.24
```
aslagle: I'd do < 1.24 since there may be more releases with the same problem.

api/rbdigital.py
```diff
@@ -1608,9 +1608,22 @@ def isbn_info_to_metadata(cls, book, include_bibliographic=True, include_formats
...
+        # options are: "eBook", "eAudio"
+        rbdigital_medium = book.get('mediaType', None)
+        if rbdigital_medium and rbdigital_medium not in cls.rbdigital_medium_to_simplified_medium:
+            cls.log.error(
+                "Could not process medium %s for %s", rbdigital_medium, rbdigital_id)
+
+        medium = cls.rbdigital_medium_to_simplified_medium.get(
+            rbdigital_medium, Edition.BOOK_MEDIUM
+        )
+
```
leonardr: I moved this code out of the `if include_bibliographic` block so it would always run and we would never have a situation where it was unclear what medium a title was. The medium is bibliographic information, but the RBdigital code also uses it to determine format information.

We could probably get rid of the `include_bibliographic` thing altogether, but there's no rush.

leonardr: I checked the other license sources and didn't find any other cases where we relied on the default Metadata.medium behavior that was removed from core.

PR #1074, opened 2018-10-18
The medium of an Edition created from a NYT best-seller list must now be explicitly specified

leonardr:
This branch deals with fallout from https://github.com/NYPL-Simplified/server_core/pull/978. All of the license sources explicitly specify a medium when they create an Edition, but the NYT best-seller list doesn't. This branch uses the name of a list to determine whether the NYT is talking about an audiobook or some other kind of book. We use 'BOOK_MEDIUM' to cover both ebooks and print books -- I don't think the distinction matters at this point. In fact I don't think the distinction between ebooks and audiobooks matters at this point, but it was easy to get that one right.

### DIFF
--- a/api/nyt.py
+++ b/api/nyt.py
@@ -30,6 +30,7 @@ from core.model import (
     get_one_or_create,
     CustomList,
     DataSource,
+    Edition,
     ExternalIntegration,
     Identifier,
     Representation,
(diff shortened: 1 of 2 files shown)


leonardr:
Actually I think it does matter, because we match Works against the best-seller list based on their permanent work ID. This branch should make it so that audiobooks which are best-sellers show up on the best-seller list, even if they aren't best-sellers as print books.

leonardr:
I checked metadata and none of its data providers seem to have this problem. They don't explicitly specify a medium for the Editions they create, but unlike with the NYT code they don't care what the edition is and they don't try to make a PWID (which requires that .medium be set).

leonardr:
I'm still concerned about the intersection of audiobook + bestseller. If we have an audio edition of a book that was a print best-seller, it seems like the audiobook should show up in the "Best sellers" lane. Right now that won't happen because the works have different PWIDs. But now at least _something_ will show up in the "Best sellers" lane.

PR #1075, opened 2018-10-19
Move instructions to confluence

adriana-alter:
Partially resolves https://jira.nypl.org/browse/SIMPLY-1319


aslagle:
What's the evidence that this is more user-friendly? I think it's easier to see things right in the form so you don't have to open another site.

EdwinGuzman:
Hopefully the links to the confluence page don't change... I think you also have a lot of files that may not necessarily be part of this PR?

aslagle:
The other problem is that it's possible the instructions could change for different versions of the circ manager.. it's easier to see the right instructions in the circ manager than to keep multiple versions in confluence and have the admin figure out the right one.

leonardr:
The model for this branch is https://github.com/NYPL-Simplified/circulation/pull/1062. 

I think the case for a separate page is stronger here than in 1062. In 1062 there's a really confusing thing that has to be explained at some length, but once you understand it, the decision is simple -- you pick an email address and put it in the form field. There's no need to leave the browser window.

For the changes in this branch, once you understand the situation, you have to go to a third-party site, log in, and follow step-by-step instructions. So you're already opening a new window and switching back and forth between windows. The person setting up the integration on the Google side might not even be the same person who's configuring that integration in the circ manager.

Changes between versions are a strong argument for keeping the documentation with the code, but that seems like a bigger issue. We have lots of documentation on the Confluence page that needs updating between versions. We need to solve it regardless of whether these bits of documentation go to Confluence. Maybe we can tag the Confluence site the way we tag the code base, and refer to it with a tag?

If we do decide to keep this documentation with the code, I would still prefer to have the documentation appear in a pop-up rather than putting it next to the form field.

mgiraldo:
what if the help is collapsed and the user can toggle it on upon click on the link? i think the help is way more useful in the page itself instead of some third page (even worse if that requires a login).

leonardr:
Collapsed help is a good idea. 

Robert put the Google Analytics documentation on Confluence and in doing so he dramatically expanded it to include step-by-step setup instructions: https://confluence.nypl.org/display/SIM/Configure+a+Google+Analytics+Integration

Is that amount of text okay for collapsed help?

FWIW the Confluence site does not require a login to view documentation.

aslagle:
The step-by-step instructions are mostly for getting to the form, so you wouldn't need them if you're already in the form. The main thing you need is the list of analytics fields, since you have to copy them over one at a time and get the order right. You need that every time you set up analytics, even if you've done it before and are very familiar with the instructions, so it's a little different than the other cases.

mgiraldo:
it would also be useful to provide a `target="_blank"` to https://analytics.google.com/analytics/web/ to follow the instructions provided

adriana-alter:
Proposed solution, as per a conversation with Jack (our UX designer): collapsed help, but 1) with a label making it clear that this is outlining the next step in the workflow, and 2) containing a link to the Google Analytics page (like Mauricio had suggested).

My understanding is that `target="_blank"` has some associated security risks and should only be used if absolutely necessary; opinions?

aslagle:
@adriana-alter that sounds good but I'm not sure what you mean about the label.

`target="_blank"` just makes the link open in a new tab, I don't see why that would be a security risk.

adriana-alter:
@aslagle 

Re: the label, I just meant that, above the collapsible instructions (probably next to the toggle for them), there would be a thing in bold type saying something along the lines of "Set up Google Analytics," formatted similarly to the labels above the input boxes (to indicate to the user that this is the next step in the process).

Re: `target="_blank"`, there are a lot of articles out there.  I have no idea how legit they are (and they might not even be relevant to this particular use case anyway)--it's just something I've heard about enough times for it to set off warning bells in my head. Here's one: https://news.softpedia.com/news/adding-target-blank-to-your-links-opens-the-door-for-phishing-attacks-507851.shtml

leonardr:
Taking the `target="_blank"` thing at face value, it should be fine here, since we control which URL we put in the `href` and we only link to reputable sites like google.com and loc.gov (for the list of language codes). But it wouldn't hurt to add `rel="noopener" rel="noreferer"` to the links.

aslagle:
What do you think is the distinction between "description" and "instructions"? Do we need both?

adriana-alter:
We do need both--on the front end, "description" is the existing, general term for any text that labels a form field (and is what's being used as the header for the collapsible panel); "instructions" are a separate, optional thing that are used as the content for the collapsible panel, and that only exist in a few, specific cases.

aslagle:
Many of the things that are currently in "description" could probably be collapsible though? They're short, but there are a lot of them so they do make the overall forms longer.

nyplkim:
My concern with this approach is that if we only put the instructions in the form, then users might not understand that they need to gather some information ahead of time and may be frustrated when they start to fill out the form and discover there is a complicated set of steps around getting their Google analytics ID for example. The Google analytics step could require working with someone else in their organization and could take some time. It would help the user if there was  a document in Confluence that stepped them through all the information they needed to have ready in order to complete the form, and then for those who just dive in we can have instructions in the form. It makes sense to me to have it in both places.

leonardr:
I filed https://jira.nypl.org/browse/SIMPLY-1340 to cover the larger issue of how to manage our documentation across versions of the server software.

api/admin/google_oauth_admin_authentication_provider.py
```diff
@@ -18,15 +18,26 @@
...
-                    "Create a project, click 'Create Credentials' in the left sidebar, and select 'OAuth client ID'. " +
-                    "If you get a warning about the consent screen, click 'Configure consent screen' and enter your library name as the product name. Save the consent screen information.</p>" +
-                    "<p>Choose 'Web Application' as the application type.</p>" +
-                    "<p>Leave 'Authorized JavaScript origins' blank, but under 'Authorized redirect URIs', add the url of your circulation manager followed by '/admin/GoogleAuth/callback', e.g. 'http://mycircmanager.org/admin/GoogleAuth/callback'.</p>" +
-                    "<p>Click create, and you'll get a popup with your new client ID and secret. Copy these values and enter them in the form below.</p>"),
+    DESCRIPTION = _("How to Configure a Google OAuth Integration")
     DOMAINS = "domains"
 
+    INSTRUCTIONS = _("<p>Configuring a Google OAuth integration in the Circulation Manager " +
+                    "will allow users to sign into the Admin interface with their Google/GMail credentials.</p>" +
```
aslagle: I might say "admins" instead of "users" here just so it's clear this isn't intended to allow anyone to log in, although it can be configured to allow anyone from a certain domain.

api/google_analytics_provider.py
```diff
@@ -13,11 +13,22 @@
...
-    NAME = _("Google Analytics")
 
-    DESCRIPTION = _("<p>Send circulation-related analytics events to Google Analytics." +
-                    "<p>To use this integration, set up a property in Google Analytics. " +
-                    "On the administration page for the property, go to Custom Definitions > Custom Dimensions, " +
+    NAME = _("Google Analytics")
+    DESCRIPTION = _("How to Configure a Google Analytics Integration")
+    INSTRUCTIONS = _("<p>In order to track usage statistics, you can configure the Circulation Manager " +
+                    "to connect to Google Analytics.  Currently, Google Analytics is the only analytics engine " +
+                    "supported for this function.</p>" +
```
aslagle: I don't think this paragraph is necessary. Particularly the last sentence doesn't really need to be on this particular page.

api/google_analytics_provider.py
```diff
@@ -13,11 +13,22 @@
...
+                    "<p>There are two components required to use Google Analytics to capture usage from your SimplyE service: " +
+                    "<ol><li><a href='https://analytics.google.com/analytics/web/provision/?authuser=0#/provision/create' " +
```
aslagle: Does this link work if you already have a google analytics account?

api/google_analytics_provider.py
```diff
@@ -13,11 +13,22 @@
...
+                    "rel='noopener' rel='noreferer' target='_blank'>Create a Google Analytics account</a> and configure it " +
+                    "to accommodate elements of SimplyE usage</li>" +
```
aslagle: "SimplyE" should be changed to "Library Simplified" throughout this since many of the events don't come from the app.

api/google_analytics_provider.py
adriana-alter: I agree about the second sentence; taking it out.  But I think it's nice to have the first sentence as an introduction (particularly for people who haven't used an analytics integration before)...

api/google_analytics_provider.py
```diff
@@ -13,11 +13,22 @@
...
+                    "<li>Add your Google Analytics account tracking ID to an integration record for your library " +
+                    "in the Circulation Manager Admin interface</li></ol></p>" +
+                    "<p>Configure Google Analytics Elements for SimplyE:</p>" +
+                    "<p>To capture data from the SimplyE Circulation Manager in your Google Analytics account, " +
+                    "you must set up a property in Google Analytics for SimplyE.  In your Google Analytics " +
```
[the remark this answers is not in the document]
adriana-alter: True.  But I think it still makes sense to have the instruction here--it doesn't seem confusing to me.  (Anyway, I can't think of a better/clearer way to phrase it; you?)

PR #1076, opened 2018-10-22
Fulfill Axis 360 audiobooks as Findaway manifest documents

leonardr:
This branch adds the ability to fulfill audiobooks from Axis 360 by hitting the new "GetFulfillmentInfo" API endpoint. This code is based on documents provided to me by Baker & Taylor, rather than a working API, so it may require some adjustment later.

This includes two major pieces of refactoring:

* The `FindawayManifest` is extracted from Bibliotheca code. We want to create the same type of manifest for every vendor that uses Findaway.
* The `APIAwareFulfillmentInfo` is taken from the RBdigital `RBFulfillmentInfo` code. Our circulation API assumes that `FulfillmentInfo` objects are cheap to create, and that there's no cost to creating them even if the user has no intention of actually fulfilling the book. But for RBdigital titles and Axis 360 audiobooks, creating a `FulfillmentInfo` requires a separate API call, which can be expensive. So we create an  `APIAwareFulfillmentInfo` instead, giving it the information it needs to make an extra API request and get the real fulfillment info. That API call is only made if the user actually tries to fulfill the book.

The major piece of brand new code is the `FulfillmentInfoResponseParser`, which creates a `FindawayManifest` from the result of a call to the Axis 360 "GetFulfillmentInfo" API.


api/axis.py
```diff
@@ -963,7 +982,7 @@ def extract_bibliographic(self, element, ns):
             data_source=DataSource.AXIS_360,
             title=title,
             language=language,
-            medium=Edition.BOOK_MEDIUM,
+            medium=medium,
```
leonardr: I'm missing a test for this bit because I don't have a real availability document for an audiobook to use in the test.

api/axis.py
```diff
@@ -379,6 +379,7 @@ def patron_activity(self, patron, pin, identifier=None):
         availability = self.availability(
             patron_id=patron.authorization_identifier,
             title_ids=title_ids)
+        set_trace()
```
EdwinGuzman: There's a set_trace here left over.

PR #1077, opened 2018-10-23
Refactor settings code in controller

adriana-alter:


leonardr:
You've effectively created a new controller class. `ConfigurationProcessor` deals with Flask requests and responds to GET, POST, and DELETE. This is good work, but `ConfigurationProcessor` doesn't sound like the name of a controller. It sounds like a tool which might be used in a controller or somewhere else--a command-line script, for instance. You've also left code in admin/controller.py which could be removed if `ConfigurationProcessor` embraced its destiny as a controller class.

So, to start with, how about renaming `ConfigurationProcessor` to `SitewideConfigurationSettingController` and renaming `configuration_processor.py` to `configuration_setting_controller.py`? Then you should be able to make the following changes:

* Completely remove `_sitewide_settings_controller`, `sitewide_settings`, and `sitewide_setting` from `admin/controller.py`.
* Change `setup_admin_controller` in `admin/controller.py` to create a `SitewideConfigurationSettingController` object and save it as `manager.admin_sitewide_configuration_settings_controller`, next to all the other admin controllers.
* Change `admin/routes.py` to reference the new `manager.admin_sitewide_configuration_settings_controller` object instead of `manager.admin_settings_controller` whenever necessary. This ensures that incoming requests will be routed to the new controller.
* Move the tests of the removed functions out of `tests/admin/test_controller.py` and into a new file, `tests/admin/test_configuration_setting_controller`, and a new class, `TestSitewideConfigurationSettingController`.

At this point we'll have completely removed a chunk of controller code from the huge `admin/controller.py` file and removed its tests from the huge `tests/admin/test_controller.py` file. That will be enough work to merge this branch.

I do want a class that does what `ConfigurationProcessor` sounds like it does -- a class that we could use in a controller or in a command-line script -- but we can do that later. The first step would be to get all the configuration setting code and tests into two relatively small files. Then you can look at those files and see how to best split the "handle an incoming HTTP request" code from the "validate and commit a bunch of proposed changes to the ConfigurationSettings table" code.

api/admin/configuration_settings_controller.py
```diff
@@ -0,0 +1,57 @@
...
+from api.config import Configuration
+from flask import Response
+from problem_details import *
+import flask
+from flask_babel import lazy_gettext as _
+
+class SitewideConfigurationSettingsController(AdminCirculationManagerController):
+
+    def process_settings(self):
+         self.require_system_admin()
```
leonardr: @aslagle I don't need you to review this whole branch, but I'd like your take on whether require_system_admin() should be called here, or individually in `process_get` and `process_post`. Also, would it be better to eliminate `process_settings` altogether by setting up two different routes in `admin/routes.py`, one for GET and one for POST?

aslagle: It doesn't matter about `require_system_admin` but it's good that it's called before anything else happens. I think it would be good to break up the routes. Almost all the admin routes have the same decorators and many of them also call the same method for GET and POST, but I think that's because the earlier ones had substantial shared code for both cases, and this only has one line.

api/config.py
```diff
@@ -292,18 +292,21 @@ class Configuration(CoreConfiguration):
...
             "type": "list",
             "description": LANGUAGE_DESCRIPTION,
+            "optional": True,
         },
         {
             "key": TINY_COLLECTION_LANGUAGES,
             "label": _("Other languages in this library's collection"),
             "type": "list",
             "description": LANGUAGE_DESCRIPTION,
+            "optional": True,
```
leonardr: It looks like these changes were brought in from SIMPLY-1318 work. This code shouldn't be in this branch, but as long as I'm talking about it, I think SMALL_ and TINY_COLLECTION_LANGUAGES can be made optional, I think LARGE_COLLECTION_LANGUAGES is still required. Although we could make ['eng'] the default.

leonardr: Actually, looking at https://github.com/NYPL-Simplified/circulation/blob/master/api/lanes.py#L135, I think the right thing to do is to make all three of these optional, with no default. If no value is set, we'll estimate values based on the library's actual holdings at the time the lanes are created.

tests/admin/controller/test_controller.py
```diff
@@ -3123,288 +3130,260 @@ def test_create_integration(self):
...
+    #         assert_raises(AdminNotAuthorized,
+    #                       self.manager.admin_settings_controller.library,
+    #                       library.uuid)
+    #
+    #         self.admin.add_role(AdminRole.SYSTEM_ADMIN)
+    #         response = self.manager.admin_settings_controller.library(library.uuid)
+    #         eq_(response.status_code, 200)
+    #
+    #     library = get_one(self._db, Library, uuid=library.uuid)
+    #     eq_(None, library)
```
leonardr: It looks like all this commented-out code was moved to test_library, so it can be removed from here.

tests/admin/controller/test_controller.py
```diff
@@ -6463,11 +6339,7 @@ def test_collection_library_registrations_post(self):
         ])
 
         class Mock(Registration):
-            # We reproduce the signature, even though it's not
-            # necessary for what we're testing, so that if the push()
-            # signature changes this test will fail.
-            def push(self, stage, url_for, catalog_url=None, do_get=None,
-                     do_post=None):
+            def push(self, *args, **kwargs):
```
leonardr: The version of this code that you remove here is the newer, more correct version.

tests/admin/controller/test_controller.py
```diff
@@ -6494,15 +6366,12 @@ def push(self, *args, **kwargs):
...
-            eq_(collection.external_account_id, kwargs.pop('catalog_url'))
-
-            # No other weird keyword arguments were passed in.
-            eq_({}, kwargs)
+            eq_(HTTP.debuggable_post, kwargs['do_post'])
+            eq_(HTTP.debuggable_get, kwargs['do_get'])
 
+            # We would have generated a fresh public key just for this
+            # transaction.
+            eq_(None, kwargs['key'])
```
leonardr: Same here -- revert these changes.

PR #1079, opened 2018-10-24
Add nearly complete test coverage to routes.py and fix a bug

leonardr:
Up until now we haven't tested the code in `routes.py` because it was difficult. But we do occasionally have problems with routes.py that would have been caught if we had test coverage.

I've figured out a way to get decent coverage of routes.py. We can verify that a certain URL is routed to a certain controller method with certain arguments, and that a real app server defines the same method, without having to spin up an entire app server for every test class (slow) and call that (potentially slow) method on realistic data. We do this with a mock Flask application that contains mock controllers which contain mock methods which create Response objects when called.

This branch adds coverage to every route defined in routes.py. We verify that the route goes where it's supposed to, that the URL arguments are propagated to the controller method, that controller and its method would actually exist in a real site, and that any restrictions on which HTTP methods can be used are enforced.

We also verify that certain controllers require authentication -- that is, we can test that the `requires_auth` annotation is in use and that it works properly.

We don't have coverage of the `allows_patron_web`, `has_library`, `library_route`, `library_dir_route`, or `returns_problem_detail` annotators -- we don't explicitly test their functionality or that they are present on any particular route. There are also a couple of very minor missing pieces of test coverage which I've marked with TODO.

While writing the tests I discovered that the URNLookupController is broken and has been broken for a while. I moved most of the code out of routes.py into a controller class where it could be tested in a realistic environment, and fixed it.


leonardr:
We can use `app.url_map` to test that the rules point to the right  function names.

leonardr:
Thanks to the `url_map` I was able to rewrite the tests so that you can pass in a URL and verify that a controller class was called with the appropriate arguments. That gives us good coverage of everything in routes.py apart from the annotations.

aslagle:
I do think this is useful, but we should still be testing the routes end-to-end as well.

leonardr:
What do you want to get out of end-to-end tests? Maybe we can figure out a way to get the same benefits without the performance cost.

aslagle:
I think we've had a few cases where we mocked something for a test but then it didn't work when the thing wasn't mocked, I'd like to avoid that.

leonardr:
I see. In that case there's no substitute for a full end-to-end test. :(

However this made me realize there was a bit of these tests that needed tightening up. We don't check `.args` or `.kwargs` unless we're expecting something to happen, but that doesn't catch the case where something _is_ passed in to the mock and we don't want it to be (as we saw earlier today with the leftover `key` variable). I've changed the test harness so it always checks both `.args` and `.kwargs`, which should stop that particular problem from happening.

I'll add some more test coverage based on the new harness, possibly for everything in routes.py, and then ask you for a full review.

api/routes.py
```diff
@@ -500,15 +496,6 @@ def odl_notify(loan_id):
...
 
-@app.route('/loadstorm-<code>.html')
-@returns_problem_detail
-def loadstorm_verify(code):
-    c = Configuration.integration("Loadstorm", required=True)
-    if code == c['verification_code']:
-        return Response("", 200)
-    else:
-        return Response("", 404)
-
```
leonardr: This is another route that stopped working a long time ago, but no one uses this code, so I just removed it.

tests/test_routes.py
```diff
@@ -0,0 +1,633 @@
...
+class TestIndex(RouteTest):
+
+    CONTROLLER_NAME = "index_controller"
+
+    def test_index(self):
+        for url in '/', '':
+            self.assert_request_calls(url, self.controller)
+
+    def test_authentication_document(self):
+        url = '/authentication_document'
```
leonardr: Throughout these tests I used the actual URL templates for the routes whenever possible, so it would be easy to match the test with the original code. The only exception is when a template included `<path:identifier>`, which I converted to `an/identifier`. This lets us verify that the <path:> thing works.

PR #1083, opened 2018-10-26
exception_handler should leave werkzeug exceptions alone.

leonardr:
Our exception handler deals with exceptions by converting them to problem detail documents if possible, and by generating 500 errors if not. There's a third class of exception we weren't handling: HTTPExceptions generated by werkzeug which handle things like 405 errors and (crucially) redirects generated by the routing code.

Since those exceptions don't get turned into problem detail documents, we were turning them into 500 errors. So redirects were turning into 500 errors. Pretty bad. This branch changes the exception handler to 'handle' HTTPExceptions by returning them -- they'll be converted into the proper HTTP responses.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -10,6 +10,7 @@ from flask import (
     make_response,
 )
 from flask_cors.core import get_cors_options, set_cors_headers
+from werkzeug.exceptions import HTTPException
 
 from app import app, babel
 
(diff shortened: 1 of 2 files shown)


aslagle:
Does this fix https://jira.nypl.org/browse/SIMPLY-995?

leonardr:
Yes, it does. Thanks for finding an already filed bug.

PR #1084, opened 2018-10-26
Required form fields

EdwinGuzman:
Fixes [SIMPLY-1318](https://jira.nypl.org/browse/SIMPLY-1318) along with [server-core #979](https://github.com/NYPL-Simplified/server_core/pull/979).

Removed `"optional": True` properties and added `"required": True` to explicitly state if a field is required. This will also be part of a `circulation-web` branch to display the following below. Let me know if I missed adding or removing any fields, or if I messed up the required/optional property for any field.

screen shot 2018-10-26 at 5 02 38 pm


api/admin/google_oauth_admin_authentication_provider.py
```diff
@@ -40,8 +40,8 @@ class GoogleOAuthAdminAuthenticationProvider(AdminAuthenticationProvider):
 
     SETTINGS = [
         { "key": ExternalIntegration.URL, "label": _("Authentication URI"), "default": "https://accounts.google.com/o/oauth2/auth" },
```
aslagle: This is also required, although if you don't provide it the default will be used. I saw some of the other ones had a default but were still required though.

api/admin/problem_details.py
```diff
@@ -177,6 +177,13 @@
     detail=_("You must identify the collection by its name."),
 )
 
+MISSING_ANALYTICS_NAME = pd(
+    "http://librarysimplified.org/terms/problem/missing-analytics-name",
+    status_code=400,
+    title=_("Missing analytics name."),
+    detail=_("You must identify the analytics by its name."),
```
aslagle: This should probably be "analytics service"?

api/authenticator.py
```diff
@@ -1541,9 +1541,10 @@ class BasicAuthenticationProvider(AuthenticationProvider, HasSelfTests):
         { "key": TEST_IDENTIFIER,
           "label": _("Test Identifier"),
           "description": _("A valid identifier that can be used to test that patron authentication is working."),
+          "required": True,
         },
         { "key": TEST_PASSWORD, "label": _("Test Password"), "description": _("The password for the test identifier."),
-          "optional": True,
+          "required": True,
```
aslagle: Did you mean to change this one? I think it needs to be optional because some libraries don't have passwords, only barcodes.

api/admin/google_oauth_admin_authentication_provider.py
aslagle: We should decide what to do for fields with defaults and be consistent about it, there are some that are required and some that aren't.

api/overdrive.py
```diff
@@ -58,10 +58,10 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI, HasSelfTests):
...
     SETTINGS = [
-        { "key": Collection.EXTERNAL_ACCOUNT_ID_KEY, "label": _("Library ID") },
-        { "key": BaseOverdriveAPI.WEBSITE_ID, "label": _("Website ID") },
-        { "key": ExternalIntegration.USERNAME, "label": _("Client Key") },
-        { "key": ExternalIntegration.PASSWORD, "label": _("Client Secret") },
+        { "key": Collection.EXTERNAL_ACCOUNT_ID_KEY, "label": _("Library ID"), "required": True },
+        { "key": BaseOverdriveAPI.WEBSITE_ID, "label": _("Website ID"), "required": True },
+        { "key": ExternalIntegration.USERNAME, "label": _("Client Key"), "required": True },
+        { "key": ExternalIntegration.PASSWORD, "label": _("Client Secret"), "required": True },
     ] + BaseCirculationAPI.SETTINGS
```
aslagle: You missed one in CHILD_SETTINGS below this.

api/sip/__init__.py
```diff
@@ -22,8 +22,8 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
     FIELD_SEPARATOR = "field separator"
 
     SETTINGS = [
-        { "key": ExternalIntegration.URL, "label": _("Server") },
-        { "key": PORT, "label": _("Port") },
+        { "key": ExternalIntegration.URL, "label": _("Server"), "required": True },
+        { "key": PORT, "label": _("Port"), "required": True },
         { "key": ExternalIntegration.USERNAME, "label": _("Login User ID"), "optional": True },
         { "key": ExternalIntegration.PASSWORD, "label": _("Login Password"), "optional": True },
         { "key": LOCATION_CODE, "label": _("Location Code"), "optional": True },
```
aslagle: You missed removing a few optionals here.

aslagle:
There are also some optionals in simple_authentication_provider that could be removed.

api/admin/google_oauth_admin_authentication_provider.py
EdwinGuzman: Right, I don't know which ones should or should not be required but I will make this one required. I don't think select drop downs need a required attribute since a choice will _have_ to be made even if it's the default. But for this one it could be accidentally left blank.

api/authenticator.py
EdwinGuzman: Yes, I meant to remove and not change to `required`.

EdwinGuzman:
Where is the `simple_authentication_provider` found?

leonardr:
`circulation/api/simple_authentication.py`

api/admin/google_oauth_admin_authentication_provider.py
leonardr: I think if a required field has a default, the default should show up in the text box and the 'Required' label should be shown. This makes it clear that you can't just wipe out the value in the text box.

EdwinGuzman:
Updated. Thanks.

api/admin/google_oauth_admin_authentication_provider.py
```diff
@@ -39,9 +39,14 @@ class GoogleOAuthAdminAuthenticationProvider(AdminAuthenticationProvider):
...
+          "key": ExternalIntegration.URL,
+          "label": _("Authentication URI"),
+          "default": "https://accounts.google.com/o/oauth2/auth",
+          "required": True,
+        },
+        { "key": ExternalIntegration.USERNAME, "label": _("Client ID"), "required": True },
+        { "key": ExternalIntegration.PASSWORD, "label": _("Client Secret"), "required": True },
     ]
 
     LIBRARY_SETTINGS = [
```
aslagle: There's still an `optional` in these library settings.

api/axis.py
```diff
@@ -109,9 +109,9 @@ class Axis360API(Authenticator, BaseCirculationAPI, HasSelfTests):
     DATE_FORMAT = "%m-%d-%Y %H:%M:%S"
 
     SETTINGS = [
-        { "key": ExternalIntegration.USERNAME, "label": _("Username") },
-        { "key": ExternalIntegration.PASSWORD, "label": _("Password") },
-        { "key": Collection.EXTERNAL_ACCOUNT_ID_KEY, "label": _("Library ID") },
+        { "key": ExternalIntegration.USERNAME, "label": _("Username"), "required": True },
+        { "key": ExternalIntegration.PASSWORD, "label": _("Password"), "required": True },
+        { "key": Collection.EXTERNAL_ACCOUNT_ID_KEY, "label": _("Library ID"), "required": True },
         { "key": ExternalIntegration.URL, "label": _("Server"), "default": PRODUCTION_BASE_URL },
```
aslagle: This one should also be required, if we're saying things with defaults should still be required.

api/enki.py
```diff
@@ -70,7 +70,7 @@ class EnkiAPI(BaseCirculationAPI, HasSelfTests):
 
     DESCRIPTION = _("Integrate an Enki collection.")
     SETTINGS = [
-        { "key": Collection.EXTERNAL_ACCOUNT_ID_KEY, "label": _("Library ID") },
+        { "key": Collection.EXTERNAL_ACCOUNT_ID_KEY, "label": _("Library ID"), "required": True },
         { "key": ExternalIntegration.URL, "label": _("URL"), "default": PRODUCTION_BASE_URL },
```
aslagle: This one should also be required.

api/google_analytics_provider.py
```diff
@@ -46,7 +46,7 @@ class GoogleAnalyticsProvider(object):
     DEFAULT_URL = "http://www.google-analytics.com/collect"
 
     SETTINGS = [
-        { "key": ExternalIntegration.URL, "label": _("URL"), "default": DEFAULT_URL },
+        { "key": ExternalIntegration.URL, "label": _("URL"), "default": DEFAULT_URL, "required": True },
     ]
 
     LIBRARY_SETTINGS = [
```
aslagle: Tracking ID should be required.

api/rbdigital.py
```diff
@@ -113,8 +113,8 @@ class RBDigitalAPI(BaseCirculationAPI, HasSelfTests):
                      if x['key'] != BaseCirculationAPI.DEFAULT_LOAN_PERIOD]
 
     SETTINGS = [
-        { "key": ExternalIntegration.PASSWORD, "label": _("Basic Token") },
-        { "key": Collection.EXTERNAL_ACCOUNT_ID_KEY, "label": _("Library ID") },
+        { "key": ExternalIntegration.PASSWORD, "label": _("Basic Token"), "required": True },
+        { "key": Collection.EXTERNAL_ACCOUNT_ID_KEY, "label": _("Library ID"), "required": True },
         { "key": ExternalIntegration.URL, "label": _("URL"), "default": PRODUCTION_BASE_URL },
```
aslagle: This should be required.

api/sip/__init__.py
```diff
@@ -22,11 +22,11 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
...
-        { "key": ExternalIntegration.USERNAME, "label": _("Login User ID"), "optional": True },
-        { "key": ExternalIntegration.PASSWORD, "label": _("Login Password"), "optional": True },
-        { "key": LOCATION_CODE, "label": _("Location Code"), "optional": True },
+        { "key": ExternalIntegration.URL, "label": _("Server"), "required": True },
+        { "key": PORT, "label": _("Port"), "required": True },
+        { "key": ExternalIntegration.USERNAME, "label": _("Login User ID") },
+        { "key": ExternalIntegration.PASSWORD, "label": _("Login Password") },
+        { "key": LOCATION_CODE, "label": _("Location Code") },
         { "key": FIELD_SEPARATOR, "label": _("Field Separator"),
           "default": "|",
```
aslagle: This should be required

PR #1085, opened 2018-10-28
Added OPDS navigation feed route.

aslagle:


tests/test_routes.py
```diff
@@ -283,6 +283,16 @@ def test_feed(self):
             url, self.controller.feed, '<lane_identifier>'
         )
 
+    def test_navigation_feed(self):
+        # An incoming lane identifier is passed in to the feed()
+        # method.
```
leonardr: This looks like a copy-and-paste that wasn't changed.

PR #1086, opened 2018-10-29
Store a library's web client URL when a registry provides one.

aslagle:
This uses the registry change in https://github.com/NYPL-Simplified/library_registry/pull/79 to get the web client URL for a library when it's successfully registered.

This also changes the sitewide web URL setting to support more than one web-based client.

### DIFF
--- a/api/registry.py
+++ b/api/registry.py
@@ -6,6 +6,7 @@ import json
 import logging
 from sqlalchemy.orm.exc import NoResultFound
 from sqlalchemy.orm.session import Session
+import urlparse
 
 from core.model import (
     create,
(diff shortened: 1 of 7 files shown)


api/registry.py
```diff
@@ -125,6 +126,10 @@ class Registration(object):
     PRODUCTION_STAGE = "production"
     VALID_REGISTRATION_STAGES = [TESTING_STAGE, PRODUCTION_STAGE]
 
+    # A registry may provide access to a web client. If so, we'll store
+    # the URL so we can link to it.
```
leonardr: I was also confused by this until you explained that we'll "link to it" by generating MARC records. Mentioning that here would be really helpful.

aslagle:
They are in fact separate settings, it's just that the sitewide setting contains all the library-specific domains as well as any sitewide ones. I agree that this is confusing and I can change it.

aslagle:
Yeah - that was one of the advantages of having it in the sitewide setting, you could also see every domain there. Now there's nowhere to see the ones from registries.

aslagle:
Yeah, the registry TOS would be a good place to describe this. I could do a follow up branch later to display something on the config page as well, if we come up with an idea of what that should look like.

PR #1089, opened 2018-10-29
Refactor settings code in controller

adriana-alter:


tests/admin/controller/test_individual_admins.py
```diff
@@ -0,0 +1,415 @@
...
+        test_changing_password(sitewide_librarian, librarian1)
+        test_changing_password(manager2, librarian1)
+        test_changing_password(librarian2, librarian1)
+        # Librarian 2
+        test_changing_password(system, librarian2, allowed=True)
+        test_changing_password(sitewide_manager, librarian2, allowed=True)
+        test_changing_password(manager2, librarian2, allowed=True)
+        test_changing_password(sitewide_librarian, librarian2)
+        test_changing_password(manager1, librarian2)
+        test_changing_password(librarian1, librarian2)
```
leonardr: This wasn't in the code before, but it's now obvious that we don't test that a librarian can change their own password, and we should add that test.

api/admin/controller/individual_admin_settings.py
```diff
@@ -0,0 +1,190 @@
...
+        return dict(
+            individualAdmins=admins,
+        )
+
+    def process_post(self):
+        # For readability: the person who is submitting the form is referred to as "user"
+        # rather than as something that could be confused with "admin" (the admin
+        # which the user is submitting the form in order to create/edit.)
+
+        user = flask.request.admin
```
leonardr: This is a good change.

api/admin/controller/individual_admin_settings.py
```diff
@@ -0,0 +1,190 @@
...
+        roles_error = self.handle_roles(user, admin, roles, settingUp)
+        if roles_error:
+            return roles_error
+
+        password = flask.request.form.get("password")
+        self.handle_password(password, admin, is_new, user)
+
+        return self.response(admin, is_new)
+
+    def check_permissions(self, user, admin, settingUp):
```
leonardr: You don't use the `user` argument.

api/admin/controller/individual_admin_settings.py
```diff
@@ -0,0 +1,190 @@
...
+                else:
+                    # An admin who isn't a librarian for the library won't be able to see
+                    # its roles, so might make requests that change other roles without
+                    # including this library's roles. Leave the non-visible roles alone.
+                    continue
+
+    def handle_password(self, password, admin, is_new, user):
+        if password:
+            # If the admin we're editing has a sitewide manager role, we've already verified
+            # the current admin's role above. Otherwise, an admin can only change that
```
leonardr: "above" now means "in check_permissions".

PR #1092, opened 2018-10-30
Collections controller

adriana-alter:


api/admin/controller/collection_settings.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def set_up_protocols(self):
+        self.protocols = self._get_collection_protocols(self.PROVIDER_APIS)
+
+        # If there are storage integrations, add a mirror integration
+        # setting to every protocol's 'settings' block.
+        mirror_integration_setting = self._mirror_integration_setting()
+        if mirror_integration_setting:
+            for protocol in self.protocols:
+                protocol['settings'].append(mirror_integration_setting)
```
leonardr: This is a little trickier than other protocol setup steps. We could start off with a site that has no mirror integration, and then the admin creates one. However, if the admin creates a new integration, the entire site configuration is reloaded, which means a new `CollectionSettingsController` will be created. Move this whole method into the `CollectionSettingsController` constructor and we'll be fine.

api/admin/controller/collection_settings.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def load_libraries(self, collection_dict, collection_object, user, protocol):
+        libraries = []
+        for library in collection_object.libraries:
+            if not user or not user.is_librarian(library):
+                continue
+            libraries.append(self._get_integration_library_info(
+                    collection_object.external_integration, library, protocol))
+
+        collection_dict['libraries'] = libraries
```
leonardr: This is a really minor point, but this method would be more reusable if it didn't take `collection_dict` as an argument and it just returned `libraries`. The caller would be responsible for deciding what to do with `libraries`.

api/admin/controller/collection_settings.py
```diff
@@ -0,0 +1,266 @@
...
+        for protocol_setting in protocol_settings:
+            key = protocol_setting.get("key")
+            if key not in collection_dict.get("settings"):
+                if key == 'mirror_integration_id':
+                    value = collection_object.mirror_integration_id or self.NO_MIRROR_INTEGRATION
+                elif protocol_setting.get("type") == "list":
+                    value = collection_object.external_integration.setting(key).json_value
+                else:
+                    value = collection_object.external_integration.setting(key).value
+                collection_dict["settings"][key] = value
```
leonardr: Similarly for this one. It would be easier to understand this code if `load_settings` was responsible for creating the entire `settings` dictionary (including the `external_account_id=collection_object.external_account_id` bit which is currently defined in collection_to_dict), and then `process_get` assigned it to `collection_dict["settings"]`.

api/admin/controller/collection_settings.py
leonardr: We tried this out and it broke a bunch of tests, because the test setup creates the CirculationManager with no mirror integration in place. We decided to avoid setting self.protocols for this one -- we can always come back to this later.

api/admin/controller/collection_settings.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def find_protocol_class(self, collection_object):
+        protocolClassFound = [p for p in self.PROVIDER_APIS if p.NAME == collection_object.protocol]
+        if len(protocolClassFound) == 1:
+            return protocolClassFound
```
leonardr: You can `return protocolClassFound[0]` and then the caller won't have to remove the single result from its list.

api/admin/controller/collection_settings.py
```diff
@@ -0,0 +1,266 @@
...
+        error = self.validate_form_fields(is_new, **fields)
+        if error:
+            return error
+
+        if protocol and not collection:
+            collection, is_new = get_one_or_create(self._db, Collection, name=name)
+            if not is_new:
+                self._db.rollback()
+                return COLLECTION_NAME_ALREADY_IN_USE
+            collection.create_external_integration(fields.get("protocol"))
```
leonardr: You can just use `protocol` here -- you defined it earlier in the method.

api/admin/controller/collection_settings.py
```diff
@@ -0,0 +1,266 @@
...
+
+
+    # POST
+    def process_post(self):
+        self.require_system_admin()
+        is_new = False
+        collection = None
+
+        name = flask.request.form.get("name")
+        protocol = flask.request.form.get("protocol")
```
leonardr: You didn't change this code, but later in this method we have a protocol dictionary also called `protocol`. For the sake of clarity let's call this `protocol_name` and the other one `protocol_dict`.

api/admin/controller/collection_settings.py
```diff
@@ -0,0 +1,266 @@
...
+
+        if protocol and not collection:
+            collection, is_new = get_one_or_create(self._db, Collection, name=name)
+            if not is_new:
+                self._db.rollback()
+                return COLLECTION_NAME_ALREADY_IN_USE
+            collection.create_external_integration(fields.get("protocol"))
+
+        collection.name = name
+        [protocol] = [p for p in self.protocols if p.get("name") == protocol]
```
leonardr: Here's the thing that should be `protocol_dict` for clarity.

api/admin/controller/collection_settings.py
```diff
@@ -0,0 +1,266 @@
...
+
+        settings = self.validate_parent(protocol, collection)
+        if isinstance(settings, ProblemDetail):
+            return settings
```
leonardr: I think it would be good to move the various calls to `self._db.rollback()` out of the helper methods such as `validate_parent` and put them right before the `return` calls.

The reasoning is that `do_post` is the method that changed the database in the first place, by calling `get_one_or_create`. If it turns out that the database shouldn't have been modified, it's the responsibility of `do_post` to roll back the changes. Then a method like `validate_parent` can just focus on checking whether the incoming data is okay.

api/admin/controller/collection_settings.py
```diff
@@ -0,0 +1,266 @@
...
+            collection, is_new = get_one_or_create(self._db, Collection, name=name)
+            if not is_new:
+                self._db.rollback()
+                return COLLECTION_NAME_ALREADY_IN_USE
+            collection.create_external_integration(fields.get("protocol"))
+
+        collection.name = name
+        [protocol] = [p for p in self.protocols if p.get("name") == protocol]
+
+        settings = self.validate_parent(protocol, collection)
```
leonardr: I was confused by a validation method modifying the settings that are used by `process_settings` below. This turns out to be the right behavior, but  add a comment here explaining what's going on.

Specifically, in addition to verifying that the parent collection is set properly, `validate_parent` figures out whether the settings that need to be validated are settings for a regular collection (e.g. client key and client secret for an Overdrive collection), or the settings for a child collection (e.g. library ID for an Overdrive Advantage collection).

api/admin/controller/collection_settings.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def validate_external_account_id_setting(self, value, setting):
+        if not value and not setting.get("optional"):
+            # Roll back any changes to the collection that have already been made.
+            self._db.rollback()
+            return INCOMPLETE_CONFIGURATION.detailed(
+                _("The collection configuration is missing a required setting: %(setting)s",
+                  setting=setting.get("label")))
+
+    def get_integration_id(self, value):
```
leonardr: Call this `get_mirror_integration_id` -- otherwise it sounds like a general helper method.

api/admin/controller/collection_settings.py
```diff
@@ -0,0 +1,266 @@
...
+
+    def process_libraries(self, protocol, collection):
+        libraries = []
+        if flask.request.form.get("libraries"):
+            libraries = json.loads(flask.request.form.get("libraries"))
+
+        for library_info in libraries:
+            library = get_one(self._db, Library, short_name=library_info.get("short_name"))
+            if not library:
+                return NO_SUCH_LIBRARY.detailed(_("You attempted to add the collection to %(library_short_name)s, but it does not exist.", library_short_name=library_info.get("short_name")))
```
leonardr: This isn't your code, but change this to make it clear that the "it" that doesn't exist is the library, not the collection.

PR #1093, opened 2018-10-30
AdminAuthServicesController

adriana-alter:


api/admin/controller/admin_auth_services.py
```diff
@@ -0,0 +1,84 @@
...
+            return Response(unicode(auth_service.protocol), 200)
+
+    def validate_form_fields(self, **fields):
+        if fields.get("protocol") and fields.get("protocol") not in ExternalIntegration.ADMIN_AUTH_PROTOCOLS:
+            return UNKNOWN_PROTOCOL
+        if fields.get("auth_service"):
+            if fields.get("id") and int(fields.get("id")) != fields.get("auth_service").id:
+                return MISSING_SERVICE
+            if fields.get("protocol") != fields.get("auth_service").protocol:
+                return CANNOT_CHANGE_PROTOCOL
```
leonardr: This code is fine but it's a little awkward to be checking `fields.get` all the time. If you assigned these to variables your code could be more like it was before, e.g. `if id and int(id) != auth_service.id:`

api/admin/controller/admin_auth_services.py
```diff
@@ -0,0 +1,88 @@
...
+from . import SettingsController
+from api.admin.google_oauth_admin_authentication_provider import GoogleOAuthAdminAuthenticationProvider
+from api.admin.problem_details import *
+
+class AdminAuthServicesController(SettingsController):
+
+    def process_admin_auth_services(self):
+        self.require_system_admin()
+        provider_apis = [GoogleOAuthAdminAuthenticationProvider]
+        self.protocols = self._get_integration_protocols(provider_apis, protocol_name_attr="NAME")
```
leonardr: A method like this which is called on every request shouldn't modify instance variables like `.protocols`. The good news is that the protocols only need to be gathered once, and since you're making a separate class for each controller, you can set up the protocols for that controller in the constructor. So move this to a constructor and have it work something like this:

```
    def __init__(self, manager):
        super(AdminAuthServicesController, self).__init__(manager)
        provider_apis = [GoogleOAuthAdminAuthenticationProvider]
        self.protocols = self._get_integration_protocols(provider_apis, protocol_name_attr="NAME")
```

leonardr: Take a look at your existing classes and make sure they don't set `.protocols`. This is something I may have overlooked in previous reviews. If you do this going forward it'll all be good.

PR #1094, opened 2018-10-31
PatronAuthServicesController

adriana-alter:


tests/admin/controller/test_controller.py
```diff
@@ -3475,446 +3475,6 @@ def test_admin_auth_service_delete(self):
...
-        """We're not really testing these arguments, but a value for them
-        is required for all Basic Auth type integrations.
-        """
-        B = BasicAuthenticationProvider
-        return [
-            (B.TEST_IDENTIFIER, "user"),
-            (B.TEST_PASSWORD, "pass"),
-            (B.IDENTIFIER_KEYBOARD, B.DEFAULT_KEYBOARD),
-            (B.PASSWORD_KEYBOARD, B.DEFAULT_KEYBOARD),
-            (B.IDENTIFIER_BARCODE_FORMAT, B.BARCODE_FORMAT_CODABAR),
```
leonardr: This was removed and not added back -- let's figure out what happened.

tests/admin/controller/test_controller.py
```diff
@@ -3475,446 +3475,6 @@ def test_admin_auth_service_delete(self):
...
-            self._db, ExternalIntegration,
-            protocol=MilleniumPatronAPI.__module__,
-            goal=ExternalIntegration.PATRON_AUTH_GOAL
-        )
-
-        common_args = self._common_basic_auth_arguments()
-        with self.request_context_with_admin("/", method="POST"):
-            M = MilleniumPatronAPI
-            flask.request.form = MultiDict([
-                ("name", "some auth name"),
```
leonardr: This isn't present in the new file.

api/admin/controller/patron_auth_services.py
```diff
@@ -0,0 +1,172 @@
...
+
+    def get_name(self, auth_service):
+        """Check that there isn't already an auth service with this name"""
+
+        name = flask.request.form.get("name")
+        if name:
+            if auth_service.name != name:
+                service_with_name = get_one(self._db, ExternalIntegration, name=name)
+                if service_with_name:
+                    self._db.rollback()
```
leonardr: As with the other branch, the calls to rollback should be in the method that modified the database in the first place -- that is, before the return statements in process_post().

tests/admin/controller/test_controller.py
leonardr: This was causing the test failures -- we added it in its new location.

tests/admin/controller/test_controller.py
leonardr: We think Edwin added this just to be tidy. We propagated it to its new location.

PR #1095, opened 2018-11-01
MetadataServicesController (note: tests changed to reflect the fact t…

adriana-alter:
…hat 'name' should be required)


api/admin/controller/metadata_services.py
```diff
@@ -0,0 +1,137 @@
...
+        else:
+            service, is_new = self._create_integration(
+                self.protocols, protocol, ExternalIntegration.METADATA_GOAL
+            )
+
+        if isinstance(service, ProblemDetail):
+            self._db.rollback()
+            return service
+
+        error = self.check_name_unique(service, name) or self.set_protocols(service, protocol) or self.register_with_metadata_wrangler(is_new, service)
```
leonardr: check_name_unique is just a validation check, but set_protocols and register_with_metadata_wrangler have serious side effects. It's fine to use `or` to string together validation checks, but a method call that has side effects should be on its own line, even though this makes the code longer with an `if error:` check for each method call. Otherwise it's easy to overlook that the things with side effects are even happening.

PR #1096, opened 2018-11-01
AnalyticsServicesController (name required now)

adriana-alter:


api/admin/controller/analytics_services.py
```diff
@@ -0,0 +1,122 @@
...
+
+    def check_name_unique(self, service, name, id):
+        """A service cannot be created with, or edited to have, the same name
+        as a service that already exists."""
+
+        service_with_name = get_one(self._db, ExternalIntegration, name=name)
+        if service_with_name and not service_with_name.id == service.id:
+            # Without checking that the IDs are different, you can't save
+            # changes to an existing service unless you've also changed its name.
+            return INTEGRATION_NAME_ALREADY_IN_USE
```
leonardr: Since you changed the behavior slightly it would be good to have a standalone test of this method. Something like this would demonstrate all the functionality, and you wouldn't have to go through calling `process_analytics_services` again.

```
def test_check_name_unique(self):

    kwargs = dict(protocol=GoogleAnalyticsProvider.__module__,
                  goal=ExternalIntegration.ANALYTICS_GOAL)
   service1, ignore = create(self._db, ExternalIntegration, name=self._str,
                             **kwargs)
   service2, ignore = create(self._db, ExternalIntegration, name=self._str,
                             **kwargs)

   # Try to change service2 so that it has the same name as service1
   # -- this is not allowed.
   result = self.controller.check_name_unique(service1, service1.name, service2.id)
   eq_(result, INTEGRATION_NAME_ALREADY_IN_USE)

   # Try to edit service1 without changing its name -- this is fine.
   eq_(
       None,
       self.controller.check_name_unique(service1, service1.name, service1.id)
   )
```

Now that I notice this, I also see that you made the same change in the metadata controller back in https://github.com/NYPL-Simplified/circulation/pull/1095/files. Add a similar test to the metadata controller and we'll be covered.

PR #1098, opened 2018-11-02
Update circ manager to allow cloudwatch logging

jonathangreen:
Brings the changes from https://github.com/NYPL-Simplified/server_core/pull/982 into the circ manager.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -23,6 +23,7 @@ Flask-Babel
 money
 pymarc
 accept-types
+watchtower # for cloudwatch logging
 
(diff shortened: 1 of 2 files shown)


requirements.txt
```diff
@@ -23,6 +23,7 @@ Flask-Babel
 money
 pymarc
 accept-types
+watchtower # for cloudwatch logging
```
jonathangreen: Do the requirements come in recursively, or does this need to be added here?

leonardr: It needs to be added separately. The core requirements are only used when running the unit tests.

PR #1102, opened 2018-11-05
CollectionSelfTestsController

adriana-alter:


api/admin/controller/collection_self_tests.py
```diff
@@ -0,0 +1,77 @@
...
+            return self.process_post(identifier)
+
+    def process_get(self, identifier):
+        collection = self.look_up_collection_by_id(identifier)
+        collection_info = self.get_collection_info(collection)
+        return dict(collection=collection_info)
+
+    def look_up_collection_by_id(self, identifier):
+        """Find the collection to display self test results or run self tests for"""
+        return self._db.query(Collection).filter(Collection.id==int(identifier))[0]
```
leonardr: You've changed the behavior here in a way that needs some extra checking.

In https://github.com/NYPL-Simplified/circulation/pull/1102/files#diff-9386fdb8523ae6e5a969364ab6f26207L2370, if there is no collection with this ID, we iterate over an empty list and return a dictionary containing an empty dictionary. Here, if there is no collection with this ID, we'll access `[][0]` which will raise an exception.

You can replace this with `Collection.by_id(self._db, identifier)`, which will return a Collection if possible and None if not. Then you can change `get_collection_info` to return an empty dictionary `if collection is None`.

Since your tests pass with this problem in place it's an indication that some test coverage needs to be added. The simplest thing would be to add something to `test_collection_self_tests_test_get` where you call collection_self_tests with a negative number, and verify the result you get.

leonardr: Actually, I would expect NO_SUCH_COLLECTION if you pass in a collection ID that doesn't exist.

PR #1103, opened 2018-11-05
Collection library registrations

adriana-alter:


api/admin/controller/discovery_service_library_registrations.py
```diff
@@ -61,7 +61,7 @@ def process_get(self):
                     id=registry.integration.id,
                     libraries=libraries,
                 )
-            )
+
```
leonardr: Did you intend to make this change? Seems like it's going to introduce a syntax error.

adriana-alter: No, that was an accident; fixed it!

PR #1108, opened 2018-11-07
Added script files for two new scripts.

leonardr:
This branch adds script files for the two CoverageProviders defined in https://github.com/NYPL-Simplified/server_core/pull/983.

### DIFF
--- a/bin/informational/adobe-id-for-short-client-token
+++ b/bin/informational/adobe-id-for-short-client-token
@@ -8,7 +8,7 @@ import base64
 import re
 import sys
 
-base_url = "https://circulation.librarysimplified.org"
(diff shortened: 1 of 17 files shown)


bin/informational/adobe-id-for-short-client-token
```diff
@@ -8,7 +8,7 @@ import base64
 import re
 import sys
 
-base_url = "https://circulation.librarysimplified.org"
+base_url = "https://libraryregistry.librarysimplified.org"
```
leonardr: An unrelated change as long as I was messing around with scripts.

circulation.librarysimplified.org no longer manages the Short Client Token. This script wasn't broken, but it was going to the wrong server.

PR #1110, opened 2018-11-07
Email validation

adriana-alter:


api/admin/controller/individual_admin_settings.py
```diff
@@ -83,6 +83,8 @@ def validate_form_fields(self, email):
         # eventually also check whether the input is formatted as a valid email address.
```
leonardr: You can change this comment now.

api/admin/controller/individual_admin_settings.py
```diff
@@ -83,6 +83,8 @@ def validate_form_fields(self, email):
...
         if not email:
             return INCOMPLETE_CONFIGURATION
+        if not self.is_email(email):
+            return INVALID_EMAIL
```
leonardr: Since there are multiple email fields on the form, it would be useful to set a detail explaining which email address is bad. Something like this:

`return INVALID_EMAIL.detailed(_('"%(address)s" is not a valid email address", address=email))`

api/admin/controller/library_settings.py
```diff
@@ -134,8 +135,6 @@ def check_for_missing_settings(self, settings):
             )
 
     def check_input_type(self, settings):
-        # Once there are validations for email address, URL, etc., they'll get called
-        # from here; for now, it's just image type.
```
leonardr: It's OK to remove this comment because it's no longer accurate but as you do this work, you should be moving closer to having a general form validation function that everybody uses.

PR #1112, opened 2018-11-09
Validate url

adriana-alter:


api/admin/controller/__init__.py
```diff
@@ -2483,3 +2501,23 @@ def _is_email(self, email):
...
+            url_inputs = [flask.request.form.get(field.get("key")) for field in url_fields]
+        else:
+            # The SitewideSettingsController passes in a string
+            url_inputs = [settings]
+        for url in url_inputs:
+            if not self._is_url(url):
+                return INVALID_URL.detailed(_('"%(url)s" is not a valid URL.', url=url))
+
+    def _is_url(self, url):
+        url_format = "(localhost|http[s]?://)(?:[a-zA-Z]|[0-9]|[$-_@.&+]|[!*\(\),]|(?:%[0-9a-fA-F][0-9a-fA-F]))+"
```
leonardr: I'm not sure about the `localhost` thing; we'll take a look once I go through the rest.

api/admin/controller/__init__.py
```diff
@@ -2223,7 +2223,7 @@ def _set_integration_setting(self, integration, setting):
                     "The configuration value for %(setting)s is invalid.",
                     setting=setting.get("label"),
                 ))
-        if not value and setting.get("required"):
+        if not value and setting.get("required") and not setting.get("default"):
             return INCOMPLETE_CONFIGURATION.detailed(
                 _("The configuration is missing a required setting: %(setting)s",
                   setting=setting.get("label")))
```
leonardr: "Are all the required settings set?" is a good candidate for moving into validate_formats next. I noticed you added another check that returns INCOMPLETE_CONFIGURATION and this would let you remove all of those.

api/admin/controller/cdn_services.py
```diff
@@ -85,15 +85,20 @@ def process_post(self):
...
         name = fields.get("name")
         protocol = fields.get("protocol")
 
         if not name:
             return INCOMPLETE_CONFIGURATION
-        if protocol and protocol not in [p.get("name") for p in self.protocols]:
-            return UNKNOWN_PROTOCOL
+        if protocol:
+            if protocol not in [p.get("name") for p in self.protocols]:
+                return UNKNOWN_PROTOCOL
```
leonardr: This little bit could also move into validate_formats in the near future -- you use it a lot.

leonardr:
You should do a follow-up branch of `core` that sets the `type` of fields like `BASE_URL_KEY` to `"url"`.

api/admin/problem_details.py
```diff
@@ -304,12 +304,19 @@
...
     detail=_("You must enter a valid email address."),
 )
 
+INVALID_URL = pd(
+    "http://librarysimplified.org/terms/problem/invalid-url",
+    status_code=400,
+    title=_("Invalid URL format"),
+    detail=_("You must enter a valid URL."),
+)
+
```
leonardr: Maybe @aslagle has an opinion, but I'd be okay with all of these "bad format" problem details having the same URI.

api/admin/controller/__init__.py
```diff
@@ -2483,3 +2501,23 @@ def _is_email(self, email):
...
+        else:
+            # The SitewideSettingsController passes in a string
+            url_inputs = [settings]
+        for url in url_inputs:
+            if not self._is_url(url):
+                return INVALID_URL.detailed(_('"%(url)s" is not a valid URL.', url=url))
+
+    def _is_url(self, url):
+        url_format = "(http[s]?://)+"
+        return re.search(url_format, url)
```
leonardr: I think it would be better to remove the regular expression altogether and do something like this:

```
return any([url.startswith(protocol + "://") for protocol in "http", "https"])
```

Because, the other problem I was going to mention -- and this may be the cause of the 'localhost' problem you were having -- is that the regular expression isn't tied to the front of the string. This regular expression will match any string that _contains_ a URL, such as `Hi, Rhonda, the URL to the library's website is http://library.org/. Be sure to get it in the circulation manager ASAP! Thanks!"

To fix this in a regular expression you'd stick a carat on the start of the string, but given that I also asked you to dramatically simplify the regular expression I think using `startswith` is simpler and more explicit.

api/admin/problem_details.py
leonardr: OK, let's use `invalid-format` for all of these going forward.

PR #1115, opened 2018-11-13
Numeric validations

adriana-alter:


api/admin/controller/__init__.py
```diff
@@ -2510,18 +2511,46 @@ def _is_email(self, email):
...
     def _is_url(self, url):
         return any([url.startswith(protocol + "://") for protocol in "http", "https"])
+
+    def validate_number(self, settings):
+        """Find any numbers that the user has submitted, and make sure that they are 1) actually numbers,
+        2) positive, and 3) lower than the specified maximum, if there is one."""
+        # Find the fields that should have numeric input and are not blank.
+        number_fields = filter(
+                            lambda s: (s.get("type") == "number") and
+                            (flask.request.form.get(s.get("key")) or flask.request.form.get("value"))
```
leonardr: I'm not crazy about putting in these special cases for the sitewide settings controller, where the value of the setting is always in `"value"` instead of in a variable taken from the setting's `key`. But I don't have a better idea that's easy to implement. Maybe @aslagle has an idea.

aslagle: I guess you could change the front-end code for sitewide settings to send the value in a parameter with the key instead. Maybe that code could be made more consistent with the other types of settings.

adriana-alter: I agree that it should be consistent with the other types of settings; I'll make a separate branch to look into that!

adriana-alter: @aslagle Was there a particular rationale for the front-end sitewide settings code not being consistent with the other settings in the first place (i.e. is there something that I should be particularly careful not to break by messing around with this!)?

aslagle: Not really, it was just because those settings aren't associated with an integration or a library so there's always a single field in the form. As long as you're careful to test after you change things it should be fine.

api/admin/controller/__init__.py
```diff
@@ -2510,18 +2511,46 @@ def _is_email(self, email):
...
+        input = flask.request.form.get(field.get("key")) or flask.request.form.get("value")
+        max = field.get("max")
+
+        try:
+            input = float(input)
+        except ValueError:
+            return INVALID_NUMBER.detailed(_('"%(input)s" is not a number.', input=input))
+
+        if input < 0:
+            return INVALID_NUMBER.detailed(_('%(field)s must be greater than 0.', field=field.get("label")))
```
leonardr: Based on https://github.com/NYPL-Simplified/server_core/pull/988/files, where zero is a valid value for several settings, I think it would be better to allow a setting to configure a 'min' value as well as a 'max' value, with the default 'min' being zero.

PR #1121, opened 2018-11-20
Fix bug preventing creation of first admin

adriana-alter:


api/admin/controller/individual_admin_settings.py
```diff
@@ -57,25 +55,51 @@ def process_post(self):
...
+
+                # By this point, we know no one is a system admin.
+                if user.is_sitewide_library_manager():
+                    return
+
+                # The user isn't a system admin or a sitewide manager.
+                if admin.is_sitewide_library_manager():
+                    raise AdminNotAuthorized
+
+                # By process of elimination, the admin must be a librarian.
```
aslagle: Not necessarily, they could still be a library manager that's not sitewide.

api/admin/controller/individual_admin_settings.py
```diff
@@ -57,25 +55,51 @@ def process_post(self):
...
+
+                # Is the user a library manager?
+                user_manages_admin_library = library and user.is_library_manager(library)
```
aslagle: This isn't quite right, since the admin may have multiple libraries. If that's the case, the user would have to be a library manager of all libraries with modified roles for the admin. That's likely why it was checked in a different part of the code.

api/admin/controller/individual_admin_settings.py
```diff
@@ -57,25 +55,51 @@ def process_post(self):
...
+                manager_role = (not library and filter(lambda role: role.role == AdminRole.LIBRARY_MANAGER, user.roles))
+                user_is_library_manager = (user_manages_admin_library or manager_role)
+
+                if user_is_library_manager:
+                    return
```
aslagle: I don't think you want to return here, shouldn't you raise `AdminNotAuthorized` in the opposite case?

api/admin/controller/individual_admin_settings.py
```diff
@@ -57,25 +55,51 @@ def process_post(self):
...
-        # User must be a sitewide library manager in order to create/edit
-        # a sitewide library manager.
-        if admin.is_sitewide_library_manager() and not settingUp:
-            self.require_sitewide_library_manager()
-        # User must be a system admin in order to create/edit a system admin.
-        if admin.is_system_admin() and not settingUp:
-            self.require_system_admin()
+        if not settingUp:
+            user = flask.request.admin
+            if not user.id == admin.id:
```
aslagle: I'm not sure if you actually need this check. (You might, I'd have to think through it)

api/admin/controller/individual_admin_settings.py
```diff
@@ -38,8 +38,6 @@ def process_post(self):
         # For readability: the person who is submitting the form is referred to as "user"
         # rather than as something that could be confused with "admin" (the admin
         # which the user is submitting the form in order to create/edit.)
-
```
aslagle: The comment above this should maybe be moved now that user is no longer in this method?

aslagle:
I think you should add at least one test - since the `settingUp` case wasn't working before we must not have good enough test coverage for it.

api/admin/controller/individual_admin_settings.py
adriana-alter: I think that's still being checked in `handle_roles` and `handle_password`

aslagle: That's probably right. I'm not sure it's helpful to have it here when it can't check everything anyway.

api/admin/controller/individual_admin_settings.py
```diff
@@ -57,25 +54,58 @@ def process_post(self):
...
+            if admin.is_sitewide_library_manager():
+                raise AdminNotAuthorized
+
+            # By process of elimination, the admin must be either a librarian or
+            # a (non-sitewide) library manager.
+
+            # Is the user a library manager?
+            user_manages_admin_library = library and user.is_library_manager(library)
+            manager_role = (not library and filter(lambda role: role.role == AdminRole.LIBRARY_MANAGER, user.roles))
+            user_is_library_manager = (user_manages_admin_library or manager_role)
```
aslagle: I think there are some cases where this code won't work. For example: admin is a librarian of library A and library B, and user is a library manager of library B. User tries to remove admin's librarianship from library B only, but gets AdminNotAuthorized because they aren't a manager of library A.

aslagle: I think there is other code that already checks role-related permissions so I would just remove this from here.

api/admin/controller/individual_admin_settings.py
```diff
@@ -57,25 +54,42 @@ def process_post(self):
...
+            user = flask.request.admin
+
+            # System admin has all permissions.
+            if user.is_system_admin():
+                return
+
+            # If we've hit this point, then the user isn't a system admin.  If the
+            # admin is a system admin, the user won't be able to do anything.
+            if admin.is_system_admin():
+                raise AdminNotAuthorized
```
aslagle: Maybe it doesn't matter but I think there are usually parentheses after this - this is raising an exception class rather than an instance of it?

adriana-alter: Good catch :)

PR #1126, opened 2018-11-27
Stop using Bibliotheca's "circulation" api -- it doesn't always have the information we need

leonardr:
This branch fixes a substantial problem (https://jira.nypl.org/browse/SIMPLY-1478) that's been lurking in the background for years. The `/circulation/items` endpoint doesn't do what I thought it does. If there are active loans or holds for a book, it gives detailed information about the current circulation, which is great. If there are no active loans or holds, it gives absolutely no information.

But we've been interpreting "absolutely no information" as meaning "the book was removed from the collection". So unpopular books have been incorrectly dropping out from the collection for a long time.

This branch changes all the code that uses `/circulation/items` to use `/items` instead. `/items` includes both bibliographic and up-to-date circulation information. I changed the `ItemListParser` to parse the circulation information (it's in a slightly different, easier-to-use format than the information  parsed by  the now-obsolete `CirculationParser`), and create a CirculationData object (previously we were creating a proto-CirculationData object using code that predated the metadata layer).

The upshot of this is that the Bibliotheca circulation sweep will also detect if there are any changes to the metadata. Those changes will be applied to the Bibliotheca Edition for the book, and the Work's `classify` WorkCoverageRecord will be removed, ensuring that before too long the Work will be updated with the new metadata from Bibliotheca. I'm running this on NYPL's Bibliotheca collection now and there have been minor improvements over the years, like 'A Novel" being removed from the title of a novel and moved to the subtitle.

I wrote tests for all the brand new code, but to validate my changes I'm mainly relying on the old monolithic tests.

With this branch and recent problems reported with Enki, I think we need a more general strategy for handling updated metadata from sources other than the metadata wrangler. More detailed comments on this topic are in the diff.


api/bibliotheca.py
```diff
@@ -1217,58 +1187,54 @@ def process_items(self, identifiers):
...
-                    if pool.presentation_edition:
-                        self.log.warn("Removing %s (%s) from circulation",
-                                      pool.presentation_edition.title, pool.presentation_edition.author)
-                    else:
-                        self.log.warn(
-                            "Removing unknown work %s from circulation.",
-                            identifier.identifier
-                        )
-                pool.update_availability(0, 0, 0, 0, self.analytics)
-                pool.last_checked = now
```
leonardr: This is duplicative of `LicensePool.update_availability`.

api/bibliotheca.py
```diff
@@ -1217,58 +1187,54 @@ def process_items(self, identifiers):
...
-            # shouldn't happen--how did we know about the
-            # identifier?--but it shouldn't be a big deal to
-            # create one.
-            pool, ignore = LicensePool.for_foreign_id(
-                self._db, self.collection.data_source, identifier.type,
-                identifier.identifier, collection=self.collection
-            )
-
-            # Bibliotheca books are never open-access.
-            pool.open_access = False
```
leonardr: `CirculationData.license_pool` takes care of this.

api/bibliotheca.py
```diff
@@ -1217,58 +1187,54 @@ def process_items(self, identifiers):
...
+        if changed:
+            # Bibliotheca's metadata for this work has changed.
+            work = pool.work
+            if work:
+                # The work's presentation will be completely
+                # recalculated the next time work_classification runs.
+                WorkCoverageRecord.add_for(
+                    work, WorkCoverageRecord.CLASSIFY_OPERATION,
+                    status=WorkCoverageRecord.REGISTERED
+                )
```
leonardr: This is brand new. I think we should come up with a simple way of doing this, similar to `Work.external_index_needs_updating`. We need to do this every time we get updated metadata from a data source. (I believe with the metadata wrangler we recalculate the presentation immediately.)

api/bibliotheca.py
```diff
@@ -1301,8 +1267,11 @@ def __init__(self, _db, collection, api_class=BibliothecaAPI,
             self.api = api_class
         else:
             self.api = api_class(_db, collection)
+        self.replacement_policy = BibliothecaAPI.replacement_policy(
+            _db, self.analytics
+        )
         self.bibliographic_coverage_provider = BibliothecaBibliographicCoverageProvider(
-            collection, self.api
+            collection, self.api, replacement_policy=self.replacement_policy
```
leonardr: Propagating the replacement policy is what causes test_analytics_event to catch all the events.

tests/test_bibliotheca.py
```diff
@@ -1063,11 +1039,13 @@ def test_handle_event(self):
...
+        # Four analytics events were collected: one for the license add
         # event itself, one for the 'checkin' that made the new
-        # license available, and one for the first appearance of a new
-        # LicensePool.
-        eq_(3, analytics.count)
+        # license available, one for the first appearance of a new
+        # LicensePool, and a redundant 'license add' event
+        # which was registered with analytics but which did not
+        # affect the counts.
+        eq_(4, analytics.count)
```
leonardr: I believe 4 analytics events were always issued, but one of them went to the wrong place because when BibliothecaEventMonitor created a BibliothecaBibliothecaCoverageProvider, it didn't propagate a ReplacementPolicy that included the mock analytics. A real analytics object was used instead.

I noticed this because one of my changes made it so 2 of the analytics events are generated by the monitor and 2 are generated by the BibliothecaBibliographicCoverageProvider (through its call to bibliographic_lookup), rather than the 3/1 split I believe we had previously.

The fourth analytics event doesn't actually change anything -- by the time it makes it to update_availability_from_delta, it's out of date, but we decide to ignore it but log it anyway for the sake of completeness.

api/bibliotheca.py
```diff
@@ -805,18 +790,47 @@ def value(bibliotheca_key):
...
+            licenses_available = intvalue("AvailableCopies")
+        except IndexError:
+            logging.warn(
+                "No information on available copies for %s",
+                primary_identifier.identifier
+            )
+            licenses_available = 0
+
+        patrons_in_hold_queue = intvalue("OnHoldCount")
+        licenses_reserved = 0
```
aslagle: Are reserved licenses included in one of the other counts then?

tests/test_bibliotheca.py
aslagle: I'm not sure it's a good idea to log the redundant event - it could be confusing when trying to analyze the analytics events.

api/bibliotheca.py
leonardr: Yeah, I meant to mention this. Reserved licenses are counted in the hold count.

Bibliotheca is the only provider I've ever seen that exposed 'reserved' books separately from 'on hold' books, and it now appears that was a fluke. The only thing we did with licenses_reserved was keep track of it and add it to the hold count, so I think this means we can get rid of that field altogether.

tests/test_bibliotheca.py
leonardr: I filed https://jira.nypl.org/browse/SIMPLY-1484, comment on there and let me know if you agree with the rationale.

api/bibliotheca.py
leonardr: I filed https://jira.nypl.org/browse/SIMPLY-1485 for this and would like your thoughts.

aslagle: That sounds good, I like the idea of handling this in the metadata layer.

api/bibliotheca.py
```diff
@@ -107,8 +107,6 @@ class BibliothecaAPI(BaseCirculationAPI, HasSelfTests):
     AUTHORIZATION_HEADER = "3mcl-Authorization"
     VERSION_HEADER = "3mcl-Version"
 
-    MAX_METADATA_AGE = timedelta(days=180)
```
leonardr: We no longer cache 'metadata' representations, since we also now use them to get circulation information.

api/bibliotheca.py
```diff
@@ -1223,19 +1221,8 @@ def _process_metadata(
...
-            # Bibliotheca's metadata for this work has changed.
-            work = pool.work
-            if work:
-                # The work's presentation will be completely
-                # recalculated the next time work_classification runs.
-                WorkCoverageRecord.add_for(
-                    work, WorkCoverageRecord.CLASSIFY_OPERATION,
-                    status=WorkCoverageRecord.REGISTERED
-                )
-
```
leonardr: https://github.com/NYPL-Simplified/server_core/pull/990 does this better.

tests/test_bibliotheca.py
```diff
@@ -630,6 +630,11 @@ def test_circulation_sweep_discovers_work(self):
         )
         monitor.process_items([identifier])
 
+        # Validate that the HTTP request went to the /items endpoint.
+        request = self.api.requests.pop()
+        url = request[1]
+        eq_(url, self.api.full_url("items") + "/" + identifier.identifier)
+
```
leonardr: These tests are better because they eliminate the possibility that we are still going to the /circulation/items endpoint instead of the /items endpoint.

PR #1128, opened 2018-11-30
MARC export

aslagle:
This is the circulation part of https://github.com/NYPL-Simplified/server_core/pull/993


api/admin/problem_details.py
```diff
@@ -373,6 +373,13 @@
     detail=_("You tried to create a new sitewide service, but a sitewide service of the same type is already configured."),
 )
 
+MULTIPLE_MARC_EXPORT_SERVICES = pd(
+    "http://librarysimplified.org/terms/problem/multiple-marc-export-services",
+    status_code=400,
+    title=_("Multiple MARC export services"),
+    detail=_("Each library can only have one MARC export service."),
+)
+
```
leonardr: I would make this more general, like "Each library can only have one service of this type".

api/routes.py
```diff
@@ -298,6 +298,12 @@ def shared_collection_hold_info(collection_name, hold_id):
 def shared_collection_revoke_hold(collection_name, hold_id):
     return app.manager.shared_collection_controller.revoke_hold(collection_name, hold_id)
 
+@library_route('/marc')
+@has_library
+@returns_problem_detail
+def marc_page():
+    return app.manager.marc_records.download_page()
+
```
leonardr: We're okay that this doesn't require authentication?

(Obviously the S3 resource doesn't require authentication but that would be easy for a library to obscure.)

aslagle: Yeah, there's no sensitive information in here. It's intended to end up in the library's public catalog anyway.

bin/repair/marc_records
```diff
@@ -0,0 +1,12 @@
...
+import os
+import sys
+bin_dir = os.path.split(__file__)[0]
+package_dir = os.path.join(bin_dir, "..", "..")
+sys.path.append(os.path.abspath(package_dir))
+from core.monitor import (
+    MARCRecordCacheMonitor,
+)
+from core.scripts import RunMonitorScript
+RunMonitorScript(MARCRecordCacheMonitor).run()
```
leonardr: This needs to go since you removed the monitor.

aslagle:
@leonardr I made one more small update to improve the download page if you want to take a look.

PR #1130, opened 2018-12-03
Fix bug blocking collection validations

adriana-alter:


api/admin/controller/collection_settings.py
```diff
@@ -174,7 +174,9 @@ def validate_form_fields(self, is_new, protocols, **fields):
             return MISSING_COLLECTION_NAME
         if "collection" in fields:
             if fields.get("collection"):
-                return self.validate_collection(**fields)
+                invalid_collection = self.validate_collection(**fields)
+                if invalid_collection:
+                    return invalid_collection
```
leonardr: The other changes here are tested by your changes to test_collections, but this feels like it needs an extra test. Is that right?

Like, you would test the case where the collection information is valid but the protocol is missing. Previously it would have passed validation but with this change validation will fail.

adriana-alter: It looks like those cases are already being tested, at the beginning of `test_collections_post_errors` (line 190 et seq.)...no?  The bug turned out to be caused by `validate_collection` returning even when there wasn't an error; the rest of the validation methods, including the numeric one, were never getting a chance to run.  So, the only resulting change in functionality is that, previously, collections with invalid information would have passed validation--because the code would always have returned after checking for the protocol, and not gone on to check the information--and now they don't; a collection with no protocol wouldn't ever have passed validation.

aslagle:
The front-end code checks `type` to display a number input: https://github.com/NYPL-Simplified/circulation-web/blob/master/src/components/ProtocolFormField.tsx#L77

Are you going to update that too?

api/admin/controller/collection_settings.py
leonardr: That makes sense.

adriana-alter:
The `type` check and the `format` check do two different things: `format` gives you an error message if you try to submit something other than a number, and `type` prevents you from typing anything other than a number into the input box in the first place.  I think there are some places where it makes sense to leave it as `type`, but in most cases, `type` just seems like it will confuse the user--it makes the form seem broken, and gives no explanation.  (Also, for some weird reason which I have not yet investigated, `type` still lets you write the letter "e"!) . If you think we should go back to `type` everywhere, then let's also add labels under the input boxes, saying something along the lines of "you can only enter numbers in this box," so that the user doesn't have to guess.

aslagle:
That's interesting, I've never heard anything about number inputs being confusing before. I think they're generally recommended so that mobile devices can show an appropriate keyboard, which doesn't matter as much here, but in general I think it's better to prevent the user from entering something invalid than let them submit an invalid value and give them an error message after. I would think the fields that say "(in Days)" already should be fine. If it's not obvious that the others are numbers we should probably update those descriptions whether we use number inputs or not.

PR #1135, opened 2018-12-07
Put the Elasticsearch dependencies in separate files 

leonardr:
This makes it easy to switch back and forth by changing an environment variable.

This makes it easier to do https://jira.nypl.org/browse/SIMPLY-1518 -- you just have to specify the variable when deploying -- though I don't know how to do it.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,6 +27,7 @@ before_install:
 
 install:
   - pip install -r requirements.txt
+  - pip install -r elasticsearch-requirements-${ELASTICSEARCH_VERSION}.txt
   - python -m textblob.download_corpora
 
 before_script:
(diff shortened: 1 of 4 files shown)


leonardr:
@aslagle Revision 3902fac fixes the problem I was having in a way that makes me not want to create a PR for this. Can you look at revision 84c9649 and see if you can figure out why it wasn't working on Travis? A seemingly identical setup worked for me locally.

aslagle:
No idea why it wasn't working, but one possibility might be to use different requirements files for installing locally and installing on travis.

leonardr:
I'm closing this; we can try again when we have more time for Elasticsearch work.

PR #1136, opened 2018-12-10
Updated MARC download page and cache script to handle update files.

aslagle:


api/controller.py
```diff
@@ -898,15 +899,33 @@ def download_page(self):
...
+            CachedMARCFile.library==library
+        ).filter(
+            CachedMARCFile.start_time==None
+        )
+
+        updates = self._db.query(CachedMARCFile).filter(
+            CachedMARCFile.library==library
+        ).filter(
+            CachedMARCFile.start_time!=None
+        ).order_by(CachedMARCFile.end_time.desc())
```
leonardr: My reading of the core branch is that end_time will always be present. If that's true, then the original record would show up in both lists.

But I'm reading this as a different design, where we periodically create full MARC record dumps but also have updates between the full dumps.

Can you explain (to me, not in the code) which one we're doing? I see issues with either approach but I don't know which one to focus on.

Either way, I think the "All records" wording is confusing -- Something like "Full recordsets" would work better. "All records" implies that the data is current.

aslagle: end_time will always be present, but start_time is only present for the updates, not the full dumps. This code uses the presence of start_time to determine where to show a file. You're right that we periodically create full MARC record dumps but also have updates between the full dumps.

leonardr: That sounds like a full record dump supercedes all previous delta dumps. In that case I think it would be better to show a single chronological list of the dumps with the deltas in nested lists:

```
<ul>
<li>Full dump 1</li>
<ul>
<li>Update 1A</li>
<li>Update 1B</li>
</ul>
<li>Full dump 2</li>
<ul>
<li>Update 2A</li>
<li>Update 2B</li>
<li>Update 2C</li>
</ul>
</ul>
```

aslagle: I'm not sure if "Full dump" would mean anything to a cataloger but I'll do something like this.

api/admin/controller/__init__.py
```diff
@@ -2544,11 +2544,10 @@ def validate_number(self, settings):
         2) positive, and 3) lower than the specified maximum, if there is one."""
         # Find the fields that should have numeric input and are not blank.
         number_fields = filter(
-                            lambda s: ("number" in [s.get("type"), s.get("format")]) and
+                            lambda s: s.get("type") == "number" and
                             (flask.request.form.get(s.get("key")) or flask.request.form.get("value"))
                             , settings
                         )
```
leonardr: I don't have an opinion on this but I want to make sure @adriana-alter sees this and confirms it's all right.

leonardr: Adriana told me that this was actually her change and you just merged it in.

api/controller.py
```diff
@@ -899,33 +900,44 @@ def download_page(self):
...
-            CachedMARCFile.start_time==None
-        )
+        files_by_lane = defaultdict(dict)
+        for file in library.cachedmarcfiles:
+            if file.start_time == None:
+                files_by_lane[file.lane]["full"] = file
+            else:
+                if not files_by_lane[file.lane].get("updates"):
+                    files_by_lane[file.lane]["updates"] = []
+                files_by_lane[file.lane]["updates"].append(file)
```
leonardr: This makes sense to me if creating a new full MARC file wipes out the previous full file and the updates. Is that right?

leonardr: OK, I get it, thanks.

PR #1140, opened 2018-12-12
Accept empty string as valid default setting

adriana-alter:


aslagle:
I think this reverted the docker submodule to the wrong version...

adriana-alter:
Oh no!  Here's a branch with the right version: https://github.com/NYPL-Simplified/circulation/pull/1141

PR #1144, opened 2018-12-14
Metadata registration fix

leonardr:


api/admin/controller/metadata_services.py
```diff
@@ -107,17 +109,15 @@ def validate_form_fields(self, **fields):
...
 
-    def register_with_metadata_wrangler(self, is_new, service):
+    def register_with_metadata_wrangler(self, do_get, do_post, is_new, service):
         """Register this site with the Metadata Wrangler."""
 
         if ((is_new or not service.password) and
             service.protocol == ExternalIntegration.METADATA_WRANGLER):
 
-            problem_detail = self.sitewide_registration(
-                service, do_get=do_get, do_post=do_post
```
leonardr: This is the core problem. The method name is process_sitewide_registration, not sitewide_registration, and the arguments being passed in -- do_get and do_post -- were not being propagated from process_post into register_with_metadata_wrangler.

api/admin/controller/__init__.py
```diff
@@ -2584,3 +2583,161 @@ def validate_language_code(self, settings):
     def _is_language(self, language):
         # Check that the input string is in the list of recognized language codes.
         return LanguageCodes.string_to_alpha_3(language)
+
+
+class SitewideRegistrationController(SettingsController):
```
leonardr: This class used to be in sitewide_registration.py, which made it look like a controller that did something specific and needed its own route. As far as Adriana and I can tell, it doesn't do anything specific -- it's just a superclass for classes like `MetadataServicesController`. So I moved it to `__init__.py` with the rest of the helpful superclasses.

I hope that eventually this class can be simplified by replacing the registration dance with `RemoteRegistry.push`, similar to what we did with `DiscoveryServiceLibraryRegistrationsController`, but I decided not to tackle that with this branch.

api/admin/controller/metadata_services.py
```diff
@@ -12,9 +12,9 @@
 from core.util.http import HTTP
 from core.util.problem_detail import ProblemDetail
 
-from . import SettingsController
+from . import SitewideRegistrationController
 
-class MetadataServicesController(SettingsController):
+class MetadataServicesController(SitewideRegistrationController):
```
leonardr: This is the core problem -- register_with_metadata_wrangler is in SitewideRegistrationController, not in SettingsController.

tests/admin/controller/test_metadata_services.py
```diff
@@ -19,20 +22,40 @@
 
 class TestMetadataServices(SettingsControllerTest):
 
-    def test_metadata_services_get_with_no_services(self):
+    def test_process_metadata_services_dispatches_by_request_method(self):
```
leonardr: Adding this test lets me simplify most of the other tests in this class. Those tests can test process_get or process_post instead of process_metadata_services, and they don't have to worry about checking permissions.

tests/admin/controller/test_metadata_services.py
```diff
@@ -193,14 +214,129 @@ def test_metadata_services_post_edit(self):
...
-            response = self.manager.admin_metadata_services_controller.process_metadata_services()
+            response = controller.process_post()
             eq_(response.status_code, 200)
 
-        eq_(novelist_service.id, int(response.response[0]))
-        eq_(ExternalIntegration.NOVELIST, novelist_service.protocol)
-        eq_("user", novelist_service.username)
-        eq_("pass", novelist_service.password)
-        eq_([l2], novelist_service.libraries)
+    def test_metadata_services_post_calls_register_with_metadata_wrangler(self):
```
leonardr: This test doesn't prove that the problem has been fixed -- register_with_metadata_wrangler was being called before, but it was exiting out before the crash.

tests/admin/controller/test_metadata_services.py
```diff
@@ -193,14 +214,129 @@ def test_metadata_services_post_edit(self):
...
+            )
+            assert integration != None
+
+            # It was passed in to register_with_metadata_wrangler
+            # along with the rest of the arguments we expect.
+            eq_(expect_called_with + [integration], controller.called_with)
+            eq_(integration, controller.called_with[-1])
+            eq_(self._db, integration._sa_instance_state.session)
+
+    def test_register_with_metadata_wrangler(self):
```
leonardr: _This_ is the test that proves the problem has been fixed.

PR #1145, opened 2018-12-17
Fix coverage

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-1560. Instead of passing in the DataSource object when making queries, we just use `DataSource.id`, which is what SQLAlchemy would have done behind the scenes.

I don't understand why this problem is happening, so it might also occur in a different context where it's not so easy to avoid.

### DIFF
--- a/bin/metadata_wrangler_collection_registrar
+++ b/bin/metadata_wrangler_collection_registrar
@@ -1,5 +1,9 @@
 #!/usr/bin/env python
-"""Gather information from the metadata wrangler."""
+"""Gather information from the metadata wrangler.
+
+This script is DEPRECATED. Use metadata_wrangler_collection_registrar
+instead -- it's the same code.
+"""
 import os
 import sys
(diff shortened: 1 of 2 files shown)


bin/metadata_wrangler_collection_registrar
```diff
@@ -1,5 +1,9 @@
 #!/usr/bin/env python
-"""Gather information from the metadata wrangler."""
+"""Gather information from the metadata wrangler.
+
+This script is DEPRECATED. Use metadata_wrangler_collection_registrar
```
aslagle: This is metadata_wrangler_collection_registrar, isn’t it?

leonardr: Thank you for the polite way you phrased that comment. :)

leonardr: This happened because _sync is a symlink to _registrar. They're the same file.

PR #1148, opened 2018-12-31
Allow asterisk as url in dev

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-1588.

I made it so that, in development, "*" can be used for any URL field, not just the specific one mentioned in the ticket; is that ok?

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -2536,7 +2536,10 @@ class SettingsController(AdminCirculationManagerController):
                 return INVALID_URL.detailed(_('"%(url)s" is not a valid URL.', url=url))
 
     def _is_url(self, url):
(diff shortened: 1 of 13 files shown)


aslagle:
This actually won't work for the use case where it came up - Robert is going to set up a non-production circulation manager that developers can point to locally when working on the web client, so that they don't have to run their own local circulation manager. We want to allow any patron web client URL, but `localhost` won't be in the circ manager URL there.

I think it's okay to allow the asterisk for any circ manager, but perhaps the description text should be updated to explain that it allows CORS requests from anyone and that it should not be used in production.

This also only applies to the patron web client URL - it's a CORS thing.

EdwinGuzman:
Are the updates in the core and docker submodules part of this PR?

PR #1151, opened 2019-01-07
Web client URLs from MARC exporter integration.

aslagle:
The circulation piece of https://github.com/NYPL-Simplified/server_core/pull/1002

### DIFF
--- a/api/marc.py
+++ b/api/marc.py
@@ -41,21 +41,32 @@ class LibraryAnnotator(Annotator):
             if include_genres:
                 self.add_simplified_genres(record, work)
 
-        self.add_web_client_urls(record, self.library, identifier)
+        self.add_web_client_urls(record, self.library, identifier, integration)
 
-    @classmethod
-    def add_web_client_urls(cls, record, library, identifier):
+    def add_web_client_urls(self, record, library, identifier, integration=None):
         _db = Session.object_session(library)
+        settings = []
+
+        if integration:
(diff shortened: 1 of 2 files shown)


api/marc.py
```diff
@@ -41,21 +41,32 @@ def annotate_work_record(self, work, active_license_pool, edition,
...
+
         from api.registry import Registration
-        settings = _db.query(ConfigurationSetting).filter(ConfigurationSetting.key==Registration.LIBRARY_REGISTRATION_WEB_CLIENT, ConfigurationSetting.library_id==library.id)
+        settings += [s.value for s in _db.query(
+            ConfigurationSetting
+        ).filter(
+            ConfigurationSetting.key==Registration.LIBRARY_REGISTRATION_WEB_CLIENT,
+            ConfigurationSetting.library_id==library.id
+        ) if s.value]
+
```
leonardr: Is this a different way of writing the same code as before? It looks like it to me but I want to make sure.

leonardr: I guess it lets you filter out empty values earlier.

leonardr:
This looks good. If you decide to go up one level of abstraction with something analogous to  MirrorUploader, I'm thinking that object would replace `integration`.

api/marc.py
aslagle: Yeah, it's pretty much the same, just reorganized.

PR #1152, opened 2019-01-07
Fixed * setting for patron web client URLs.

aslagle:
Setting the sitewide patron web client URL to `*` wasn't working, I think I broke it when I made it handle both registry URLs and the sitewide setting.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -244,6 +244,8 @@ class CirculationManager(object):
         patron_web_domains = set()
 
         def get_domain(url):
+            if url == "*":
+                return url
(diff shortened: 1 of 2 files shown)


aslagle:
Yeah, travis seems not to have picked up this PR, I don't know why.

EdwinGuzman:
Weird. I guess create a new one because it can't be merged in without passing on Travis.

aslagle:
I think it's because github disabled the hook that triggers it: https://www.githubstatus.com/incidents/kq75qp7try8t

aslagle:
I did get one to start by re-enabling the service in addition to the webhook in the repo settings (not sure why it was originally disabled, and it's still using courteney's account).

PR #1158, opened 2019-01-14
Make it possible to fulfill an individual part of a book

leonardr:
The motivating issue for this branch is https://jira.nypl.org/browse/SIMPLY-1626

This branch extends the `fulfill` route to take an optional argument `part`. This is used to fulfill a single portion of a book rather than the whole thing. The semantics of `part` are vendor-specific, as are how many and which parts exist for a specific book.

The rubber meets the road in `CirculationAPI.fulfill`. I changed most implementations of this method to use `kwargs` as a collector for arguments they're not using; this will make it possible for us to, in the future, add additional optional arguments to `fulfill` without changing every single implementation again.

However, the RBdigital implemementation of `fulfill` uses the two new arguments, `part` and `fulfill_part_url`.

`part` is used when doing partial fulfillment: it's the part of the book the patron wants right now. In the case of RBdigital, `part` is implemented as a numeric offset against the `readingOrder` defined in the audiobook manifest.

`fulfill_part_url` is a function that is used when _not_ doing partial fulfillment. When generating an audiobook manifest, we call this function on every part number, to create a URL that can be used to fulfill that part.

The overall effect is that if you understand the custom JSON format used by RBdigital, then everything is very fast. (We'll be changing SimplyE so that it understands this format.) If you don't understand this format, then everything still works -- you can fulfill any particular chapter through its `alternates` link -- but it'll be slower, because a) the requests are going through the circ manager, b) fulfilling one chapter of one book requires pulling the patron's entire list of loans.


api/problem_details.py
```diff
@@ -191,13 +191,6 @@
       _("Could not fulfill loan."),
 )
 
-NO_ACCEPTABLE_FORMAT = pd(
-    "http://library-simplified.org/terms/problem/no-acceptable-format",
-    502,
-    _("No acceptable format."),
-    _("This book is not available in the requested format."),
-)
-
```
leonardr: This change ended up being unrelated, but I saw that NO_ACCEPTABLE_FORMAT was defined twice in this file.

api/rbdigital.py
```diff
@@ -801,14 +822,19 @@ def get_patron_checkouts(self, patron_id):
...
-            loan_info = self._make_loan_info(item)
+            loan_info = self._make_loan_info(
+                item, fulfill_part_url=fulfill_part_url
+            )
             if loan_info:
                 loans.append(loan_info)
         return loans
 
-    def _make_loan_info(self, item, fulfill=False):
+    def _make_loan_info(self, item, fulfill_part_url=None):
```
leonardr: The old `fulfill` argument wasn't being used at all. Despite appearances, `fulfill` is not being replaced by something similar; I just removed `fulfill` because it didn't do anything.

api/circulation.py
```diff
@@ -753,17 +767,26 @@ def fulfill(self, patron, pin, licensepool, delivery_mechanism, sync_on_failure=
             )
 
         if licensepool.open_access:
+            # We ignore the vendor-specific arguments when doing
+            # open-access fulfillment, because we just don't support
+            # partial fulfillment of open-access content.
```
aslagle: would this need to change for open-access audiobooks?

leonardr:
Right now if you try to do partial fulfillment of something that's not from RBDigital, it'll fulfill the entire book rather than raising CannotPartiallyFulfill. Only RBDigital gets really picky about what partial fulfillment means. I'm okay with this and it saves a lot of code but I want to state it explicitly.

api/circulation.py
aslagle: or those let you download parts directly once you have the manifest, so it doesn't matter?

leonardr: Exactly. The manifest is publicly available and there's no need to route partial fulfillment requests through the circ manager.

api/routes.py
```diff
@@ -291,6 +291,7 @@ def shared_collection_revoke_loan(collection_name, loan_id):
 
 @app.route("/collections/<collection_name>/loans/<loan_id>/fulfill", defaults=dict(mechanism_id=None))
 @app.route("/collections/<collection_name>/loans/<loan_id>/fulfill/<mechanism_id>")
+@app.route("/collections/<collection_name>/loans/<loan_id>/fulfill/<mechanism_id>")
```
aslagle: I think you accidentally copied this line. Or did you mean to add part to it?

leonardr: I did add part to it, then took it out because it's not necessary for this branch, but I left the copy in place.

PR #1159, opened 2019-01-16
Added route for updating lane order.

aslagle:
This branch goes with a circulation-web branch that lets you drag and drop lanes on the lanes page of the admin interface.

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -1816,12 +1816,10 @@ class LanesController(AdminCirculationManagerController):
         submitted_lanes = json.loads(flask.request.data)
 
         def update_lane_order(lanes):
-            index = 0
-            for lane_data in lanes:
+            for index, lane_data in enumerate(lanes):
                 lane_id = lane_data.get("id")
                 lane = self._db.query(Lane).filter(Lane.id==lane_id).one()
                 lane.priority = index
-                index = index + 1
                 update_lane_order(lane_data.get("sublanes", []))
 
         update_lane_order(submitted_lanes)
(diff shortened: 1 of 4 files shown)


api/admin/controller/__init__.py
```diff
@@ -1810,6 +1810,24 @@ def reset(self):
         create_default_lanes(self._db, flask.request.library)
         return Response(unicode(_("Success")), 200)
 
+    def change_order(self):
+        self.require_library_manager(flask.request.library)
+
+        submitted_lanes = json.loads(flask.request.data)
```
aslagle: The submitted lanes are in order but do not have a priority field.

api/admin/package.json
```diff
@@ -9,6 +9,6 @@
   "author": "NYPL",
   "license": "Apache-2.0",
   "dependencies": {
-    "simplified-circulation-web": "0.0.96"
+    "simplified-circulation-web": "0.0.97"
```
EdwinGuzman: This should eventually be `0.0.98`.

api/admin/controller/__init__.py
```diff
@@ -1810,6 +1810,24 @@ def reset(self):
...
+
+        def update_lane_order(lanes):
+            index = 0
+            for lane_data in lanes:
+                lane_id = lane_data.get("id")
+                lane = self._db.query(Lane).filter(Lane.id==lane_id).one()
+                lane.priority = index
+                index = index + 1
```
leonardr: This is purely stylistic and it's not worth making the change, but I do this with `for index, lane_data in enumerate(lanes)`.

PR #1160, opened 2019-01-18
Use version 0.0.98 of circulation-web

adriana-alter:


EdwinGuzman:
`0.0.97` was never pushed to the circulation admin. @aslagle and I were wondering why not.

EdwinGuzman:
You also have changes in core. Make sure you update that before merging.

aslagle:
Leonard and I had a couple core changes we hadn’t put in circulation yet, it’s fine to pull them in here.

PR #1161, opened 2019-01-18
Use DeliveryMechanism.id instead of LicensePoolDeliveryMechanism.id.

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-1659, which was caused by a confusion between LicensePoolDeliveryMechanism and DeliveryMechanism. The `mechanism_id` that needs to go into URL templates like `/works/<work_id>/fulfill/<mechanism_id>/<part>` is the `DeliveryMechanism.id`.


EdwinGuzman:
Are the core changes suppose to be there, too?

PR #1162, opened 2019-01-23
Changed update_formats to also update the Overdrive edition with fresh metadata, as long as we have it.

leonardr:
This fixes https://jira.nypl.org/browse/SIMPLY-1671. It's a low-effort way to make sure that any problems (past or present) with Overdrive metadata are fixed over time.

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -833,16 +833,6 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI, HasSelfTests):
 
         edition, ignore = self._edition(licensepool)
 
-        # The identifier in the Metadata and CirculationData need to
-        # match the identifier associated with the LicensePool --
-        # otherwise a new LicensePool will be created.
-        #
-        # TODO: It's not clear to me why this code is necessary -- why
-        # would the identifiers not match, given that we looked up
-        # this identifier? -- but removing it without understanding
-        # why it was added seems dangerous.
-        metadata.primary_identifier.identifier = licensepool.identifier.identifier
-        metadata.circulation._primary_identifier.identifier = licensepool.identifier.identifier
         replace = ReplacementPolicy.from_license_source(self._db)
         metadata.apply(edition, self.collection, replace=replace)
 
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -819,23 +819,32 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI, HasSelfTests):
 
     def update_formats(self, licensepool):
         """Update the format information for a single book.
+
+        Incidentally updates the metadata, just in case Overdrive has
+        changed it.
         """
         info = self.metadata_lookup(licensepool.identifier)
 
         metadata = OverdriveRepresentationExtractor.book_info_to_metadata(
-            info, include_bibliographic=False, include_formats=True)
+            info, include_bibliographic=True, include_formats=True)
         if not metadata:
             # No work to be done.
             return
-        circulation_data = metadata.circulation
 
-        # The identifier in the CirculationData needs to match the
-        # identifier associated with the LicensePool -- otherwise
-        # a new LicensePool will be created.
-        circulation_data._primary_identifier.identifier = licensepool.identifier.identifier
-        replace = ReplacementPolicy(formats=True)
-        _db = Session.object_session(licensepool)
-        circulation_data.apply(_db, licensepool.collection, replace)
+        edition, ignore = self._edition(licensepool)
+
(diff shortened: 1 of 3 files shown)


api/overdrive.py
```diff
@@ -819,23 +819,32 @@ def circulation_lookup(self, book):
...
         """Update the format information for a single book.
+
+        Incidentally updates the metadata, just in case Overdrive has
+        changed it.
         """
         info = self.metadata_lookup(licensepool.identifier)
 
         metadata = OverdriveRepresentationExtractor.book_info_to_metadata(
-            info, include_bibliographic=False, include_formats=True)
+            info, include_bibliographic=True, include_formats=True)
```
leonardr: Without this, we'll retrieve the bibliographic information from Overdrive, but we'll throw it away. This buys us a little time, but we're inside a script, not handling an HTTP request, and data quality is more important than time.

api/overdrive.py
```diff
@@ -819,23 +819,32 @@ def circulation_lookup(self, book):
...
+        # The identifier in the Metadata and CirculationData need to
+        # match the identifier associated with the LicensePool --
+        # otherwise a new LicensePool will be created.
+        #
+        # TODO: It's not clear to me why this code is necessary -- why
+        # would the identifiers not match, given that we looked up
+        # this identifier? -- but removing it without understanding
+        # why it was added seems dangerous.
+        metadata.primary_identifier.identifier = licensepool.identifier.identifier
+        metadata.circulation._primary_identifier.identifier = licensepool.identifier.identifier
```
leonardr: @aslagle I've asked Edwin to review this branch as a whole but would like your opinion on this weird piece of code I noticed.

aslagle: It seems you added this code to fix tests - https://github.com/NYPL-Simplified/circulation/commit/46b171020e5bc60bce924c53bf7c5a7ac65c2c46

So if the tests are working now, probably ok to remove it?

leonardr: :blush: Thanks for doing `git blame` when I hadn't done it.

I thought that seemed like test code. Removing the code does break a test, but I'll try fixing the test. This is really ugly and should stay in the test if possible.

api/overdrive.py
```diff
@@ -819,23 +819,22 @@ def circulation_lookup(self, book):
...
-        # The identifier in the CirculationData needs to match the
-        # identifier associated with the LicensePool -- otherwise
-        # a new LicensePool will be created.
-        circulation_data._primary_identifier.identifier = licensepool.identifier.identifier
-        replace = ReplacementPolicy(formats=True)
-        _db = Session.object_session(licensepool)
-        circulation_data.apply(_db, licensepool.collection, replace)
+        edition, ignore = self._edition(licensepool)
+
+        replace = ReplacementPolicy.from_license_source(self._db)
```
EdwinGuzman: What is this replacing?

leonardr: The old code is `replace = ReplacementPolicy(formats=True)`, which will replace an old list of delivery mechanisms with a new list. This replaces basically everything, since a license source is considered authoritative for everything.

EdwinGuzman:
Looks good except for the failing tests.

leonardr:
The test failure was fallout from https://github.com/NYPL-Simplified/server_core/pull/1009/.

leonardr:
The problem is that error_handler_object was created using a specific Flask.app object, using the database connection that was associated with the Flask.app object at the time it was created. If any of the controller tests run before routes.py runs, something weird happens to that database connection.

I experimented with fixing app.manager._db during the test, but we're not actually testing the internal behavior of error_handler_object here, we're just checking the exception_handler() calls one of its methods, so it was easier to fix by mocking the error_handler_object. However this probably points to something we could improve in the controller tests.

tests/test_routes.py
```diff
@@ -713,25 +715,37 @@ class TestExceptionHandler(RouteTest):
...
+            # The exception was passed into MockErrorHandler.handle.
+            eq_(value_error, routes.h.handled)
 
-        # At this point the database session has been ruined, so
-        # we can't do any more checks in this test.
+            # The Response is created was passed along.
+            eq_("handled it", result.data)
+            eq_(500, result.status_code)
 
-    def test_exception_handling_http_exception(self):
```
EdwinGuzman: So this is removed because there are no http exceptions?

leonardr: No code is removed. I tried to split a test into two, which didn't solve the problem, so I merged them back to reduce the overall size of the diff.

EdwinGuzman: I was looking at the latest commit and on the entire PR before.

PR #1164, opened 2019-01-25
Convert some circulation scripts into TimestampScripts

leonardr:
This finishes the work for https://jira.nypl.org/browse/SIMPLY-1649. I'll do metadata scripts as well, but that's less important.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -71,6 +71,7 @@ from core.scripts import (
     LibraryInputScript,
     PatronInputScript,
     RunMonitorScript,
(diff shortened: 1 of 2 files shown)


scripts.py
```diff
@@ -699,6 +700,9 @@ def facets(self, lane):
 
 class CacheMARCFiles(LaneSweeperScript):
```
aslagle: This could be a timestamp script, although it's a little weird since most of the time it runs it doesn't do anything.

PR #1166, opened 2019-01-28
The name of the internal format argument to `fulfill` must always be `internal_format`

leonardr:
This fixes https://jira.nypl.org/browse/SIMPLY-1682, a regression introduced by the partial audiobook fulfillment code.

Automated tests didn't catch this bug because we don't test the through-line from `CirculationAPI` to each specific vendor API. We still don't, but the tests for the two broken vendor APIs now call fulfill() the same way `CirculationAPI` does.

The [code that caused this problem](https://github.com/NYPL-Simplified/circulation/blob/master/api/circulation.py#L784) isn't changed here, because it's actually correct -- the problem was in these two vendor APIs.

### DIFF
--- a/api/axis.py
+++ b/api/axis.py
@@ -320,7 +320,7 @@ class Axis360API(Authenticator, BaseCirculationAPI, HasSelfTests):
         response = self.request(url, data=args, method="POST")
         return response
 
-    def fulfill(self, patron, pin, licensepool, internal_format, **kwargs):
+    def fulfill(self, patron, pin, licensepool, **kwargs):
         """Fulfill a patron's request for a specific book.
 
         :param kwargs: A container for arguments to fulfill()
--- a/api/axis.py
+++ b/api/axis.py
@@ -320,7 +320,7 @@ class Axis360API(Authenticator, BaseCirculationAPI, HasSelfTests):
         response = self.request(url, data=args, method="POST")
         return response
 
-    def fulfill(self, patron, pin, licensepool, format_type, **kwargs):
+    def fulfill(self, patron, pin, licensepool, internal_format, **kwargs):
         """Fulfill a patron's request for a specific book.
 
         :param kwargs: A container for arguments to fulfill()
--- a/tests/test_axis.py
+++ b/tests/test_axis.py
@@ -325,6 +325,46 @@ class TestAxis360API(Axis360Test):
         params = request[-1]['params']
         eq_('notifications@example.com', params['email'])
 
+    def test_fulfill(self):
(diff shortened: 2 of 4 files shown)


api/axis.py
```diff
@@ -320,7 +320,7 @@ def _checkout(self, title_id, patron_id, internal_format):
         response = self.request(url, data=args, method="POST")
         return response
 
-    def fulfill(self, patron, pin, licensepool, format_type, **kwargs):
+    def fulfill(self, patron, pin, licensepool, **kwargs):
```
leonardr: Axis 360 can completely ignore this argument because the format type is locked in on checkout.

tests/test_axis.py
```diff
@@ -325,6 +325,46 @@ def test_place_hold(self):
         params = request[-1]['params']
         eq_('notifications@example.com', params['email'])
 
+    def test_fulfill(self):
```
leonardr: This method has no tests at all, so I added some.

tests/test_axis.py
```diff
@@ -325,6 +325,46 @@ def test_place_hold(self):
...
+            with_license_pool=True
+        )
+
+        patron = self._patron()
+        patron.authorization_identifier = "a barcode"
+
+        def fulfill():
+            return self.api.fulfill(
+                patron, "pin", licensepool=pool,
+                internal_format='irrelevant'
```
leonardr: This is the bit of this test that actually tests the change I made.

tests/test_bibliotheca.py
```diff
@@ -473,7 +473,9 @@ def test_fulfill(self):
             200, headers={"Content-Type": "presumably/an-acsm"},
             content="this is an ACSM"
         )
-        fulfillment = self.api.fulfill(patron, 'password', pool, 'ePub')
+        fulfillment = self.api.fulfill(
+            patron, 'password', pool, internal_format='ePub'
```
leonardr: Here I'm changing the `fulfill` call to make it more closely resemble the call that's actually going to come in from `CirculationAPI`.

api/axis.py
EdwinGuzman: What do you mean that it's `locked in`? Is the type (like ebook) set before it's fulfilled?

leonardr: Yeah, for Axis 360 you choose the content type (EPUB or PDF) as part of taking out the loan. With Overdrive or Bibliotheca you choose the content type the first time you ask for an ACSM file.

In either case, once you make your choice you're locked in for the duration of the loan.

PR #1167, opened 2019-01-29
Allow a library to hide specific content types from OPDS feeds

leonardr:
This addresses https://jira.nypl.org/browse/SIMPLY-1683. This will let us hide PDF delivery mechanisms on Open eBooks, which will stop the iOS client from trying to download books of this type.

### DIFF
--- a/api/opds.py
+++ b/api/opds.py
@@ -60,7 +60,7 @@ class CirculationManagerAnnotator(Annotator):
 
     def __init__(self, lane,
                  active_loans_by_work={}, active_holds_by_work={},
-                 active_fulfillments_by_work={},
+                 active_fulfillments_by_work={}, hidden_content_types=[],
                  test_mode=False):
         if lane:
             logger_name = "Circulation Manager Annotator for %s" % lane.display_name
@@ -71,6 +71,7 @@ class CirculationManagerAnnotator(Annotator):
         self.active_loans_by_work = active_loans_by_work
         self.active_holds_by_work = active_holds_by_work
         self.active_fulfillments_by_work = active_fulfillments_by_work
+        self.hidden_content_types = hidden_content_types
         self.test_mode = test_mode
 
     def _lane_identifier(self, lane):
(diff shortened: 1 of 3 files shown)


tests/test_opds.py
```diff
@@ -892,7 +949,7 @@ def test_active_loan_feed(self):
             x for x in links
             if x['rel'] == 'http://librarysimplified.org/terms/rel/user-profile'
         ]
-        annotator = cls(None, None, patron, test_mode=True)
+        annotator = cls(None, None, library=patron.library, patron=patron, test_mode=True)
```
leonardr: This is an unrelated bug in the test -- the third positional argument is the library, not the patron. Previously it didn't make a different but it does now because we always call _hidden_content_types on the library.

api/opds.py
```diff
@@ -278,7 +295,12 @@ def acquisition_links(
                 # set. There is one link for the delivery mechanism
                 # that was locked in, and links for any streaming
                 # delivery mechanisms.
+                #
+                # Since the delivery mechanism has already been locked in,
+                # we choose not to use visible_delivery_mechanisms --
+                # they already chose it and they're stuck with it.
                 for lpdm in active_license_pool.delivery_mechanisms:
+
                     if lpdm is active_loan.fulfillment or lpdm.delivery_mechanism.is_streaming:
```
leonardr: There's a very small inconsistency here: if one of the streaming content types is hidden, it will still show up at this point, because we're not iterating over `visible_delivery_mechanisms`. I don't think it's worth fixing this problem because hopefully we can remove this entire feature once we fix Open eBooks.

EdwinGuzman: But this is only for streaming content types and not other content types?

api/opds.py
```diff
@@ -158,6 +159,20 @@ def active_licensepool_for(self, work):
...
+        """
+        hidden = self.hidden_content_types
+        for lpdm in licensepool.delivery_mechanisms:
+            mechanism = lpdm.delivery_mechanism
+            if not mechanism:
+                # This shouldn't happen, but just in case.
+                continue
+            if mechanism.content_type in hidden:
+                continue
+            yield lpdm
```
EdwinGuzman: Just curious, why the choice to use a generator for this?

api/opds.py
leonardr: Yeah, the logic is: show a link if it's a streaming content type or if it's the content type that the patron "locked in" when they borrowed the book.

If the "locked in" content type is hidden, we want to show it anyway, because the patron is already committed to it. Hiding it will just make the situation worse.

If one of the streaming types is hidden, it would be better to hide it here as well -- presumably we hid it for a reason, and we don't want it showing up at all.

api/opds.py
leonardr: Just personal preference I guess. Most of the time I find it easier to write a generator than to keep track of a list inside of a method.

PR #1168, opened 2019-01-30
Bring circulation up to date with changes to the Timestamp API

leonardr:
This branch changes the circulation manager's use of `Timestamp` to reflect the changes made by https://github.com/NYPL-Simplified/server_core/pull/1013:

* Timestamps are now identified with a `service_type` as well as a `service`.
* The time at which the service completed is now called `finish`, not `timestamp`.
* A `Monitor` that can't complete its task should raise an exception.

### DIFF
--- a/api/monitor.py
+++ b/api/monitor.py
@@ -126,9 +126,9 @@ class MWCollectionUpdateMonitor(MetadataWranglerCollectionMonitor):
                     if link not in seen_links:
                         queue.append(link)
             if new_timestamp:
-                self.timestamp().timestamp = new_timestamp
+                self.timestamp().finish = new_timestamp
             self._db.commit()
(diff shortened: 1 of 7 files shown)


api/monitor.py
```diff
@@ -90,10 +101,7 @@ def endpoint(self, timestamp):
         return self.lookup.updates(timestamp)
 
     def run_once(self, start, cutoff):
-        if not self.lookup.authenticated:
-            self.keep_timestamp = False
-            return
-
+        self.assert_authenticated()
```
leonardr: The core branch removed `keep_timestamp`, so this won't work anymore. It's better overall to raise an exception--the exception will be logged as `Timestamp.exception`, and since the Monitor didn't complete its task, `start` and `finish` will not be updated.

PR #1169, opened 2019-02-01
Categorize and return information about timestamps

adriana-alter:


api/admin/controller/__init__.py
```diff
@@ -372,6 +374,63 @@ def __call__(self, collection, book, path=None):
...
+
+        return sorted
+
+    def _sort_by_type(self, timestamps):
+        """Takes a list of Timestamp objects.  Returns a dict: each key is a type of service
+        (script, monitor, or coverage provider); each value is a dict in which the keys are the names
+        of services and the values are lists of timestamps."""
+
+        result = {}
+        for type in ["script", "monitor", "coverage_provider"]:
```
leonardr: You can use the `Timestamp` constants here. We don't have a comprehensive list inside `Timestamp`, which would be better, but if we ever add a new type of `Timestamp` we can create one then.

api/admin/controller/__init__.py
```diff
@@ -372,6 +374,63 @@ def __call__(self, collection, book, path=None):
...
+            result.setdefault(timestamp.get("service"), []).append(timestamp)
+        return result
+
+    def _sort_by_collection(self, timestamps):
+        """Takes a list of timestamps; turns it into a dict in which each key is a
+        collection ID and each value is a list of the timestamps associated with that collection."""
+
+        result = {}
+        for timestamp in timestamps:
+            result.setdefault((timestamp.get("collection_id") or -1), []).append(timestamp)
```
leonardr: You can pass `-1` as the second argument to `get()` to make that the default value returned.

api/admin/controller/__init__.py
```diff
@@ -372,6 +374,63 @@ def __call__(self, collection, book, path=None):
...
+        return result
+
+    def _extract_info(self, timestamp):
+        """Takes a Timestamp object and returns a dict"""
+
+        return dict(
+            id=timestamp.id,
+            start=timestamp.start,
+            duration=(str(timestamp.finish - timestamp.start)),
```
leonardr: I don't think this will happen in real life, but it could theoretically. You should be prepared for `start` or `finish` to be null, which will crash this code. Before making the dictionary, calculate `duration` and set it to null if either `start` or `finish` is null.

Also, you can convert this into a number of seconds with `(timestamp.finish - timestamp.start).total_seconds()`. Then on the client side you can format the number of seconds however you like.

api/admin/controller/__init__.py
leonardr: On second look, it shouldn't be necessary to explicitly list the types in here at all. You can iterate over `timestamps` and put each one into a dictionary of lists, keyed off of `ts.service_type`. Something like this should work:

```
result = {}
for ts in timestamps:
    info = self._extract_info(ts)
    result.setdefault(ts.service_type, []).append(info)
```

You don't need to make a comprehensive list of all the types of Timestamps, and you only need to iterate over `timestamps` once.

api/admin/controller/__init__.py
```diff
@@ -372,6 +374,63 @@ def __call__(self, collection, book, path=None):
...
 
+class TimestampsController(AdminCirculationManagerController):
+    """Returns a dict: each key is a type of service (script, monitor, or coverage provider);
+    each value is a nested dict in which timestamps are organized by service name and then by collection ID."""
+
+    def diagnostics(self):
+        self.require_system_admin()
+        timestamps = self._db.query(Timestamp).order_by(Timestamp.start)
+        if isinstance(timestamps, ProblemDetail):
+            return timestamps
```
leonardr: It's not possible for a SQLAlchemy operation to return a ProblemDetail, so you can take this out. If there's a database problem, SQLAlchemy will raise an exception which will be handled by the app server itself.

api/admin/controller/__init__.py
```diff
@@ -372,6 +374,63 @@ def __call__(self, collection, book, path=None):
...
+        sorted = self._sort_by_type(timestamps)
+        for type in sorted.keys():
+            for service in sorted[type]:
```
leonardr: This is a little nicer:

```
for type, services in sorted.items():
    for service in services:
```

api/admin/controller/__init__.py
leonardr: This is maybe not a topic for this branch, but is a collection ID enough for this purpose? It'll let you group timestamps together by collection, but it won't mean anything to the end-user.

api/admin/controller/__init__.py
```diff
@@ -415,20 +417,29 @@ def _sort_by_collection(self, timestamps):
         collection ID and each value is a list of the timestamps associated with that collection."""
 
         result = {}
+        collection_name = None
         for timestamp in timestamps:
-            result.setdefault((timestamp.get("collection_id") or -1), []).append(timestamp)
+            result.setdefault(timestamp.get("collection_name"), []).append(timestamp)
```
leonardr: This is a good solution. I had to check that Collection.name is unique, but it is, so this is good.

api/admin/controller/__init__.py
```diff
@@ -415,20 +417,29 @@ def _sort_by_collection(self, timestamps):
...
         """Takes a Timestamp object and returns a dict"""
 
+        duration = None
+        if timestamp.start and timestamp.finish:
+            duration = (timestamp.finish - timestamp.start).total_seconds()
+
+        collection_name = "No associated collection"
+        if timestamp.collection_id:
+            collection_name = self._db.query(Collection).filter(Collection.id == timestamp.collection_id).one().name
+
```
leonardr: A simpler way to do this is: 

```
if timestamp.collection:
    collection_name = timestamp.collection.name
```

The SQL query will happen automatically.

In fact we can set this up so that the SQL queries aren't necessary at all, because `timestamp.collection` is populated from the database using the same query that gets all the `timestamp`s. I can show you how to do this on Monday. It's not actually necessary here because this controller will be used very rarely.

PR #1173, opened 2019-02-12
Changed the Monitor.run_once() implementations that currently return timestamps so that they return TimestampData objects instead.

leonardr:
This is the circulation portion of https://github.com/NYPL-Simplified/server_core/pull/1021. 

We never override `CoverageProvider.run_once` or have a `Script` whose `run_once` implementation has a return value, so there was no need to change our coverage providers or scripts to reflect the new APIs. However, Monitors use the timestamp differently, so I had to change all the monitor classes.

These monitors became `TimelineMonitor`s. Their `run_once` implementations were converted to `catch_up_from` implementations:

* Axis360CirculationMonitor
* BibliothecaEventMonitor
* EnkiImport
* OdiloCirculationMonitor
* ODLConsolidatedCopiesMonitor
* OverdriveCirculationMonitor

For these monitors I modified `run_once` to change the signature, but didn't make any other changes, because these monitors don't do anything with a timestamp:

* MWAuxiliaryMetadataMonitor
* OPDSForDistributorsReaperMonitor
* RBDigitalSyncMonitor
* RBDigitalCirculationMonitor

I changed this monitor to port its custom behavior (returning a certain meaningful time as a custom value for Timestamp.finish) to the new system:

* MWCollectionUpdateMonitor

`BibliothecaEventMonitor` used to be a "custom behavior" monitor -- it used the timestamp to track the dates of the events it processed. I converted it into a `TimelineMonitor` because doing so dramatically simplified the code, and I don't think we were getting any value from the custom behavior.

### DIFF
--- a/api/axis.py
+++ b/api/axis.py
@@ -535,16 +535,28 @@ class Axis360CirculationMonitor(CollectionMonitor):
             Axis360BibliographicCoverageProvider(collection, api_class=self.api)
         )
 
-    def run_once(self, start, cutoff):
-        # Give us five minutes of overlap because it's very important
-        # we don't miss anything.
-        since = start-self.FIVE_MINUTES
+    def run_once(self, progress):
+        """Find Axis 360 books that changed recently.
+
+        :progress: A TimestampData representing the time previously
+        covered by this Monitor.
+        """
+        # Give us five minutes of overlap with the previous run
+        # because it's very important we don't miss anything.
+        since = progress.finish - self.FIVE_MINUTE
+        cutoff = datetime.datetime.utcnow()
         count = 0
         for bibliographic, circulation in self.api.recent_activity(since):
             self.process_book(bibliographic, circulation)
             count += 1
             if count % self.batch_size == 0:
                 self._db.commit()
+
+        # Update the TimestampData to describe the time span covered
+        # during this run.
+        progress.start = since
+        progress.finish = cutoff
+        return progress
 
     def process_book(self, bibliographic, availability):
 
--- a/api/axis.py
+++ b/api/axis.py
@@ -58,6 +58,7 @@ from core.model import (
 from core.monitor import (
     CollectionMonitor,
     IdentifierSweepMonitor,
+    TimelineMonitor,
 )
 
 from core.opds_import import (
@@ -508,7 +509,7 @@ class Axis360API(Authenticator, BaseCirculationAPI, HasSelfTests):
             params=params, **kwargs
         )
 
-class Axis360CirculationMonitor(CollectionMonitor):
+class Axis360CirculationMonitor(CollectionMonitor, TimelineMonitor):
 
     """Maintain LicensePools for Axis 360 titles.
     """
@@ -519,7 +520,6 @@ class Axis360CirculationMonitor(CollectionMonitor):
     PROTOCOL = ExternalIntegration.AXIS_360
 
     DEFAULT_START_TIME = datetime(1970, 1, 1)
-    FIVE_MINUTES = timedelta(minutes=5)
 
     def __init__(self, _db, collection, api_class=Axis360API):
         super(Axis360CirculationMonitor, self).__init__(_db, collection)
@@ -535,28 +535,18 @@ class Axis360CirculationMonitor(CollectionMonitor):
             Axis360BibliographicCoverageProvider(collection, api_class=self.api)
         )
 
-    def run_once(self, progress):
+    def catch_up_from(self, start, cutoff, progress):
         """Find Axis 360 books that changed recently.
 
         :progress: A TimestampData representing the time previously
         covered by this Monitor.
         """
-        # Give us five minutes of overlap with the previous run
-        # because it's very important we don't miss anything.
-        since = progress.finish - self.FIVE_MINUTE
-        cutoff = datetime.datetime.utcnow()
         count = 0
-        for bibliographic, circulation in self.api.recent_activity(since):
+        for bibliographic, circulation in self.api.recent_activity(start):
             self.process_book(bibliographic, circulation)
             count += 1
             if count % self.batch_size == 0:
                 self._db.commit()
-
-        # Update the TimestampData to describe the time span covered
-        # during this run.
-        progress.start = since
-        progress.finish = cutoff
-        return progress
 
     def process_book(self, bibliographic, availability):
 
--- a/api/bibliotheca.py
+++ b/api/bibliotheca.py
@@ -1324,13 +1324,33 @@ class BibliothecaEventMonitor(CollectionMonitor):
             yield slice_start, slice_cutoff, full_slice
             slice_start = slice_start + increment
 
-    def run_once(self, start, cutoff):
+    def _timestamp(self, progress):
+        start = progress.finish or self.default_start_time
+        
+
+
+    def run_once(self, progress):
+        # We want to start at the time of the last event successfully
+        # processed, and continue up to the current time.
+        start = progress.finish or self.default_start_time
+        now = datetime.datetime.utcnow()
+
+        # We use timestamp.start and timestamp.finish to track the 
+        # first and last events processed during this run.
+        #
+        # If no events are processed during a run, we preserve the
+        # last event processed in the _previous_ run as both the start
+        # and end point.
+        first_event = progress.finish
+        final_event = first_event
+
         added_books = 0
         i = 0
         one_day = timedelta(days=1)
         most_recent_timestamp = start
         for start, cutoff, full_slice in self.slice_timespan(
-                start, cutoff, one_day):
+            start, now, one_day
+        ):
             most_recent_timestamp = start
             self.log.info("Asking for events between %r and %r", start, cutoff)
             try:
@@ -1338,9 +1358,9 @@ class BibliothecaEventMonitor(CollectionMonitor):
                 events = self.api.get_events_between(start, cutoff, full_slice)
                 for event in events:
                     event_timestamp = self.handle_event(*event)
-                    if (not most_recent_timestamp or
-                        (event_timestamp > most_recent_timestamp)):
-                        most_recent_timestamp = event_timestamp
+                    if (not final_event or
+                        (event_timestamp > final_event)):
+                        final_event = event_timestamp
                     i += 1
                     if not i % 1000:
                         self._db.commit()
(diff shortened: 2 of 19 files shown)


api/axis.py
```diff
@@ -535,12 +535,14 @@ def __init__(self, _db, collection, api_class=Axis360API):
...
-        since = start-self.FIVE_MINUTES
+    def catch_up_from(self, start, cutoff, progress):
+        """Find Axis 360 books that changed recently.
+
+        :progress: A TimestampData representing the time previously
+        covered by this Monitor.
+        """
         count = 0
-        for bibliographic, circulation in self.api.recent_activity(since):
+        for bibliographic, circulation in self.api.recent_activity(start):
```
leonardr: This is typical of the simple changes made in this branch. A monitor becomes a `TimeLineMonitor`, its `run_once` becomes `catch_up_from`, and code for creating five minutes of overlap is removed, because `TimeLineMonitor` adds five minutes of overlap to the value it passes in to `start`.

api/bibliotheca.py
```diff
@@ -1323,41 +1325,28 @@ def slice_timespan(self, start, cutoff, increment):
             yield slice_start, slice_cutoff, full_slice
             slice_start = slice_start + increment
 
-    def run_once(self, start, cutoff):
+    def catch_up_from(self, start, cutoff, progress):
```
leonardr: This is a more complicated change, but the end result is the same. This Monitor now fits into the `TimelineMonitor` pattern. We were using the `Timestamp` to keep track of the timestamps on the events that we actually processed, rather than the spans of time we covered, but there was no super-important reason for that, and using `TimelineMonitor` simplifies the code a lot.

api/bibliotheca.py
```diff
@@ -1323,41 +1325,28 @@ def slice_timespan(self, start, cutoff, increment):
...
-            except Exception, e:
-                if event:
-                    self.log.error(
-                        "Fatal error processing Bibliotheca event %r.", event,
-                        exc_info=e
-                    )
-                else:
-                    self.log.error(
-                        "Fatal error getting list of Bibliotheca events.",
-                        exc_info=e
```
leonardr: Also, exception handling is now done by `run_once`, so there's no need to have it in this method.

api/monitor.py
```diff
@@ -100,7 +101,16 @@ class MWCollectionUpdateMonitor(MetadataWranglerCollectionMonitor):
     def endpoint(self, timestamp):
         return self.lookup.updates(timestamp)
 
-    def run_once(self, start, cutoff):
+    def run_once(self, progress):
```
leonardr: This doesn't fit the `TimelineMonitor` mold, so we change `run_once` to take a `TimestampData`. `run_once` used to return a new value for `Timestamp.finish`; now we modify `.finish` on the `TimestampData` we were given, and return it to indicate that this is what we want our timestamp to look like.

api/monitor.py
```diff
@@ -194,7 +209,7 @@ def __init__(self, _db, collection, lookup=None, provider=None):
     def endpoint(self):
         return self.lookup.metadata_needed()
 
-    def run_once(self, start, cutoff):
+    def run_once(self, progress):
```
leonardr: Here, we don't actually use the thing passed in to `run_once`, so it's just a matter of changing the signature.

api/overdrive.py
```diff
@@ -1006,13 +1007,13 @@ def patron_request(self, patron, pin, *args, **kwargs):
...
 
-class OverdriveCirculationMonitor(CollectionMonitor):
+class OverdriveCirculationMonitor(CollectionMonitor, TimelineMonitor):
     """Maintain LicensePools for recently changed Overdrive titles. Create
     basic Editions for any new LicensePools that show up.
     """
     SERVICE_NAME = "Overdrive Circulation Monitor"
-    INTERVAL_SECONDS = 500
     PROTOCOL = ExternalIntegration.OVERDRIVE
+    OVERLAP = datetime.timedelta(minutes=1)
```
leonardr: This fits the `TimelineMonitor` mold, and we previously didn't use any overlap at all, so I set the OVERLAP value to something smaller than five minutes.

I do think having a little overlap will be good here; we sometimes miss events from Overdrive and it's possible they're falling into the cracks in between runs.

api/overdrive.py
```diff
@@ -1109,7 +1118,6 @@ class RecentOverdriveCollectionMonitor(OverdriveCirculationMonitor):
     """Monitor recently changed books in the Overdrive collection."""
 
     SERVICE_NAME = "Reverse Chronological Overdrive Collection Monitor"
-    INTERVAL_SECONDS = 60
```
leonardr: This is a constant that we haven't used for a long time; I'm just cleaning it up here.

tests/test_axis.py
```diff
@@ -597,6 +597,30 @@ class TestCirculationMonitor(Axis360Test):
...
+        # The second time run() is called, catch_up_from() is asked
+        # to find events between five minutes before the last cutoff,
+        # and what is now the current time.
+        monitor.run()
+        new_start, new_cutoff, new_progress = monitor.called_with
+        now = datetime.datetime.utcnow()
+        before_old_cutoff = cutoff - monitor.OVERLAP
+        eq_(before_old_cutoff, new_start)
+        assert (now - new_cutoff).total_seconds() < 2
+
```
leonardr: I added end-to-end tests of `run` or `run_once` to a lot of the monitors I changed, even though the code is technically already tested as part of the `TimelineMonitor` test coverage in core.

First, I'm concerned that unless there's an end-to-end test, future changes to the `Monitor` API will break individual monitors without breaking the tests.

Second, my next branch is going to add status reporting to all of these monitors, and a test like this is often the simplest place to put the code that checks whatever status was reported.

tests/test_odilo.py
```diff
@@ -575,7 +577,31 @@ def test_63_return_hold(self):
...
+        # modification date five minutes earlier than the completion
+        # of the last run.
+        monitor.run()
+        expect = completed-monitor.OVERLAP
+        assert (expect-monitor.called_with).total_seconds() < 2
+
+    def test_all_ids_with_date(self):
+        # TODO: This tests that all_ids doesn't crash when you pass in
+        # a date. It doesn't test that all_ids does anything in
+        # particular.
```
leonardr: I noticed this problem but decided not to fix it here.

tests/test_overdrive.py
```diff
@@ -1093,6 +1094,98 @@ def test_sync_bookshelf_ignores_holds_from_other_collections(self):
         eq_(5, len(patron.holds))
         assert overdrive_hold in patron.holds
 
+class TestOverdriveCircluationMonitor(OverdriveAPITest):
```
leonardr: We were simply missing this test coverage and I had to write a big test to make sure my change didn't break anything.

api/monitor.py
```diff
@@ -133,10 +142,29 @@ def run_once(self, start, cutoff):
...
+        # want that to happen.
+        #
+        # If we found an OPDS feed, the latest timestamp in that feed
+        # should be used as Timestamp.finish.
+        #
+        # Otherwise, the existing timestamp.finish should be used. If
+        # that value happens to be None, we need to set
+        # TimestampData.finish to NO_VALUE to make sure it ends up
+        # as None (rather than the current time).
+        finish = new_timestamp or self.timestamp().finish or Timestamp.NO_VALUE
```
leonardr: Here's the case in mentioned in my core branch where we explicitly want `timestamps.finish` to become `None` (actually, to stay `None` and not be set to a specific value).

api/axis.py
EdwinGuzman: A `TimeLineMonitor` still has a `run_once` method. That still gets triggered or is it just `catch_up_from`?

leonardr: Yeah, the main entry point is `run`, which calls `run_once`, and then for a `TimelineMonitor` `run_once` calls `catch_up_from`.

api/rbdigital.py
```diff
@@ -1940,7 +1940,11 @@ def __init__(self, _db, collection, api_class=RBDigitalAPI,
         super(RBDigitalSyncMonitor, self).__init__(_db, collection)
         self.api = api_class(_db, collection, **api_class_kwargs)
 
-    def run_once(self, start, cutoff):
+    def run_once(self, progress):
+        """Find books in the RBdigital collection that changed recently.
+
+        :param progress: A TimestampData, ignored.
```
EdwinGuzman: Why are some of these `progress` params ignored?

leonardr: It depends on the monitor. If a monitor doesn't need to know anything about the circumstances under which it was last run, it can ignore the incoming `progress`. Even in my working branch, when we need to set `achievements`, we will usually be doing this by creating a new `TimestampData` object and we might still ignore the one that's passed in to them ethod.

tests/test_overdrive.py
```diff
@@ -1093,6 +1094,98 @@ def test_sync_bookshelf_ignores_holds_from_other_collections(self):
...
+        api.licensepools.append(lp4)
+
+        monitor.catch_up_from(object(), object(), object())
+
+        # We called update_licensepool on the first three books,
+        # and got the first three LicensePools from the queue.
+        eq_([(1, lp1),(2, lp2),(3, lp3)], api.update_licensepool_calls)
+
+        # At that point we gave up because we'd gotten two consecutive
+        # books from Overdrive that contained no new information.
```
EdwinGuzman: When you say "we gave up", does that mean that it just stops monitor at that point? What happens to the fourth LicensePool that is still in `api.licensepools`?

EdwinGuzman:
It looks good to me. There are a few things I don't completely grasp but are not necessarily code-related.

tests/test_overdrive.py
leonardr: Nothing happens to that LicensePool. It represents a missed opportunity.

When we run this monitor for real, it will periodically give us 20 or 30 consecutive "recent" books that, upon inspection, haven't actually changed. But then we'll get a chunk of books that _have_ actually changed. So we can't stop looking as soon as we find a book that didn't actually change. The data we get from Overdrive isn't well-ordered, the way it is when we get it from Bibliotheca.

At the same time, we can't keep going on forever just in case there's a book way down in the "recent" stack that has new data. (That does happen, but we use a different monitor to catch those changes.) As a compromise, we stop looking at the list of "recent" books once we see 100 consecutive "recent" books that haven't changed.

PR #1175, opened 2019-02-12
Test search term setting

adriana-alter:


leonardr:
It looks like you're having some trouble getting the right revision The revision you want here is 7c40759542fc3bd902009a9f5b89fc6140b7cd1e, the tip of your core branch. Once you merge https://github.com/NYPL-Simplified/server_core/pull/1025 into master, updating this branch to use the tip of master should work fine.

PR #1176, opened 2019-02-13
Implement the new First Book authentication API

leonardr:
This branch implements the new First Book authentication API. It resolves https://jira.nypl.org/browse/SIMPLY-1719.

The new API is located at https://ebooksuat.firstbook.org/api/, but I assume "uat" stands for User Acceptance Testing and the final name will change. Instead of sending barcode and PIN directly in the URL to the API, we create a JWT which includes barcode and PIN in the payload, sign it with a shared secret, and send the JWT over in the URL. So the final URL we use looks like this:

https://ebooksuat.firstbook.org/api/eyJ0eXAiOiJKV1QiLCJhbGciOiJIUzI1NiJ9.eyJiYXJjb2RlIjoiQUJDREVGIiwicGluIjoiMzU3NCIsImlhdCI6IjE1NDk0OTE4MDkifQ.zJ1QXAJbQExrfbP227mNTm_qYri2PKNw_Fpo1qzBZbw

I made a copy of firstbook.py and test_firstbook.py, and modified it as necessary to accommodate the new API. I'll point out the major changes, but I think it's good to go over the whole thing since I don't think you're familiar with the  First Book API.

For now, the circulation manager will support both versions of the API. Once we get Open Ebooks using the new version, we can remove support for the old API.

### DIFF
--- a/api/firstbook2.py
+++ b/api/firstbook2.py
@@ -1,7 +1,11 @@
 from nose.tools import set_trace
 from flask_babel import lazy_gettext as _
+import jwt
+from jwt.algorithms import HMACAlgorithm
 import requests
 import logging
+import time
+
 from authenticator import (
     BasicAuthenticationProvider,
     PatronData,
@@ -21,15 +25,18 @@ from core.model import (
 
 class FirstBookAuthenticationAPI(BasicAuthenticationProvider):
 
-    NAME = 'First Book'
+    NAME = 'First Book - JWT'
 
     DESCRIPTION = _("""
         An authentication service for Open eBooks that authenticates
-        using access codes and PINs. (This is the old version.)""")
+        using access codes and PINs. (This is the new version.)""")
 
     DISPLAY_NAME = NAME
     DEFAULT_IDENTIFIER_LABEL = _("Access Code")
     LOGIN_BUTTON_IMAGE = "FirstBookLoginButton280.png"
+
+    # The algorithm used to sign JWTs.
+    ALGORITHM = 'HS256'
 
     # If FirstBook sends this message it means they accepted the
     # patron's credentials.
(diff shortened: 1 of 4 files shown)


api/firstbook2.py
```diff
@@ -0,0 +1,196 @@
...
+        return PatronData(
+            permanent_id=username,
+            authorization_identifier=username,
+        )
+
+    # End implementation of BasicAuthenticationProvider abstract methods.
+
+    def remote_pin_test(self, barcode, pin):
+        jwt = self.jwt(barcode, pin)
+        url = self.root + jwt
```
leonardr: This is new -- previously we stuck the barcode and PIN on to the end of `url`; now we call `jwt` and put _that_ on the end.

api/firstbook2.py
```diff
@@ -0,0 +1,196 @@
...
+        if response.status_code != 200:
+            msg = "Got unexpected response code %d. Content: %s" % (
+                response.status_code, response.content
+            )
+            raise RemoteInitiatedServerError(msg, self.NAME)
+        if self.SUCCESS_MESSAGE in response.content:
+            return True
+        return False
+
+    def jwt(self, barcode, pin):
```
leonardr: This whole method is new.

api/firstbook2.py
```diff
@@ -0,0 +1,196 @@
...
+            # Simulate a bad connection.
+            raise requests.exceptions.ConnectionError("Could not connect!")
+        elif self.failure_status_code:
+            # Simulate a server returning an unexpected error code.
+            return MockFirstBookResponse(
+                self.failure_status_code, "Error %s" % self.failure_status_code
+            )
+        parsed = urlparse.urlparse(url)
+        token = parsed.path.split("/")[-1]
+        barcode, pin = self._decode(token)
```
leonardr: The `_decode` call is new, as is `_decode` itself. Previously there was nothing to decode, so we just looked for the barcode and pin in `self.valid`.

api/firstbook2.py
```diff
@@ -0,0 +1,196 @@
...
+
+        # The barcode and pin must be present in self.valid.
+        if barcode in self.valid and self.valid[barcode] == pin:
+            return MockFirstBookResponse(200, self.SUCCESS)
+        else:
+            return MockFirstBookResponse(200, self.FAILURE)
+
+    def _decode(self, token):
```
leonardr: This whole thing is new.

tests/test_firstbook2.py
```diff
@@ -0,0 +1,142 @@
...
+        # Let's see what the mock API had to work with.
+        requested = self.api.request_urls.pop()
+        assert requested.startswith(self.api.root)
+        token = requested[len(self.api.root):]
+
+        # It's a JWT, with the provided barcode and PIN in the
+        # payload.
+        barcode, pin = self.api._decode(token)
+        eq_("ABCD", barcode)
+        eq_("1234", pin)
```
leonardr: Previously we didn't bother testing which request was sent.

tests/test_firstbook2.py
```diff
@@ -0,0 +1,142 @@
...
+        os.environ['AUTOINITIALIZE'] = "False"
+        from api.app import app
+        self.app = app
+        del os.environ['AUTOINITIALIZE']
+        with self.app.test_request_context("/"):
+            doc = self.api.authentication_flow_document(self._db)
+            eq_(self.api.DISPLAY_NAME, doc['description'])
+            eq_(self.api.FLOW_TYPE, doc['type'])
+
+    def test_jwt(self):
```
leonardr: This whole test is new.

api/firstbook2.py
```diff
@@ -0,0 +1,196 @@
...
+        # which is also the authorization identifier.
+        return PatronData(
+            permanent_id=username,
+            authorization_identifier=username,
+        )
+
+    # End implementation of BasicAuthenticationProvider abstract methods.
+
+    def remote_pin_test(self, barcode, pin):
+        jwt = self.jwt(barcode, pin)
```
EdwinGuzman: Where is this `self.jwt` function defined?

EdwinGuzman: Never mind, got slightly confused before.

PR #1179, opened 2019-02-15
Enki library id for library+collection

EdwinGuzman:
Moving the library id setting from the Enki integration setting to be for each associated library in the collection. Adding migration for existing Enki collections.

Fixes [SIMPLY-1220](https://jira.nypl.org/browse/SIMPLY-1220).

### DIFF
--- a/api/enki.py
+++ b/api/enki.py
@@ -70,9 +70,12 @@ class EnkiAPI(BaseCirculationAPI, HasSelfTests):
 
     DESCRIPTION = _("Integrate an Enki collection.")
     SETTINGS = [
-        { "key": Collection.EXTERNAL_ACCOUNT_ID_KEY, "label": _("Library ID"), "required": True },
         { "key": ExternalIntegration.URL, "label": _("URL"), "default": PRODUCTION_BASE_URL, "required": True, "format": "url" },
     ] + BaseCirculationAPI.SETTINGS
+
+    LIBRARY_SETTINGS = [
+        { "key": Collection.EXTERNAL_ACCOUNT_ID_KEY, "label": _("Library ID"), "required": True },
+    ]
 
     list_endpoint = "ListAPI"
     item_endpoint = "ItemAPI"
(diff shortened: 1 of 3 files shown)


api/enki.py
```diff
@@ -70,10 +70,13 @@ class EnkiAPI(BaseCirculationAPI, HasSelfTests):
 
     DESCRIPTION = _("Integrate an Enki collection.")
     SETTINGS = [
-        { "key": Collection.EXTERNAL_ACCOUNT_ID_KEY, "label": _("Library ID"), "required": True },
         { "key": ExternalIntegration.URL, "label": _("URL"), "default": PRODUCTION_BASE_URL, "required": True, "format": "url" },
     ] + BaseCirculationAPI.SETTINGS
 
+    LIBRARY_SETTINGS = [
+        { "key": Collection.EXTERNAL_ACCOUNT_ID_KEY, "label": _("Library ID"), "required": True },
```
leonardr: You can technically call a key-value pair whatever you want, but this is the name of a field in the `collections` table and also using it as the name of a key-value pair is confusing. I'd call it something custom and very specific, like "enki_library_id", just because we've got so many IDs floating around.

migration/20190215-update-enki-collection-integration.sql
```diff
@@ -0,0 +1,18 @@
...
+from (
+    select externalintegration_id, library_id, external_account_id
+    from collections join externalintegrations_libraries as el
+    on collections.external_integration_id=el.externalintegration_id
+    join externalintegrations as e on e.id=el.externalintegration_id where e.protocol='Enki'
+) updatesetting
+where
+    cs.external_integration_id=updatesetting.externalintegration_id
+    and cs.library_id=updatesetting.library_id
+    and key='external_account_id';
```
leonardr: This is pretty close but it needs to be an `insert` statement rather than an `update`. We're going from a situation where there are no Enki library IDs in `configurationsettings` -- they're all stored in `collections.external_account_id` -- to a situation where there is one Enki library ID in `configurationsetting` for every Enki collection. Basically move the value from `collections.external_account_id` to a new row in `configurationsettings`.

EdwinGuzman:
@leonardr when you say clear out the `external_account_id`, do you mean remove the row from `collections` or just the value in that row?

leonardr:
Just set the value to null.

PR #1181, opened 2019-02-17
Fixed missing presentation editions when a book is in multiple Overdrive collections

aslagle:
This hopefully fixes https://jira.nypl.org/browse/SIMPLY-1722.

It uses the same presentation edition and the same work for all the license pools. I think it's better to use the same work but I'm not sure if that will end up having other consequences.

### DIFF
--- a/api/enki.py
+++ b/api/enki.py
@@ -738,7 +738,7 @@ class EnkiImport(CollectionMonitor):
     def update_circulation(self, since):
         """Process circulation events that happened since `since`."""
         for circulation in self.api.recent_activity(since):
-            license_pool, made_changes = circulation.apply(
+            license_pool, is_new = circulation.license_pool(
                 self._db, self.collection
             )
             if not license_pool.work:
@@ -749,6 +749,11 @@ class EnkiImport(CollectionMonitor):
                 metadata = self.api.get_item(license_pool.identifier.identifier)
                 if metadata:
                     self.process_book(metadata)
+            else:
+                license_pool, made_changes = circulation.apply(
+                    self._db, self.collection
+                )
+
 
     def process_book(self, bibliographic):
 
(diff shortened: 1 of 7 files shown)


api/overdrive.py
```diff
@@ -920,11 +920,14 @@ def update_licensepool_with_book_info(self, book, license_pool, is_new_pool):
...
-        # and if this edition is newly made, then associate pool and edition
-        # as presentation_edition
-        if ((not license_pool.presentation_edition) and is_new_edition):
+        # then associate pool and edition as presentation_edition.
+        if not license_pool.presentation_edition:
             edition_changed = license_pool.set_presentation_edition()
 
+        # Make sure the pool also has a work.
+        if not license_pool.work:
+            license_pool.calculate_work()
```
aslagle: Is there a better place to do this? In some cases this will get called when it was already getting called somewhere else, but it wasn't getting called at all for the problematic pools.

leonardr:
What if `CirculationData.apply` called `set_presentation_edition` and `calculate_work`? We didn't do that originally because we were expecting that the bibliographic data wasn't there yet.

If we did that, we'd need to make sure that `set_presentation_edition` and `calculate_work` return `None` if the data isn't there. And we'd want to make sure that when we have both bibliographic and circulation data, we lay down the bibliographic data first. But that seems doable.

leonardr:
This fix also needs a migration script. It feels like we could do it with SQL -- find a work ID associated with each identifier and set licensepool.work_id to that ID for all license pools associated with that identifier that have no work ID. But it might be complicated/slow.

leonardr:
Laura McKeegan mentioned that this problem isn't happening for Enki collections. I think that's because of this code in enki.py:

```
        if not license_pool.work:
            work, is_new = license_pool.calculate_work()
            if work:
                work.set_presentation_ready()
```

That's the sort of code I was thinking we could move to `CirculationData.apply`.

aslagle:
I'm not sure if we do need a migration since this can be corrected by the regular collection monitor scripts, but that would also be slow so maybe a migration is better.

api/enki.py
```diff
@@ -749,6 +749,11 @@ def update_circulation(self, since):
                 metadata = self.api.get_item(license_pool.identifier.identifier)
                 if metadata:
                     self.process_book(metadata)
+            else:
+                license_pool, made_changes = circulation.apply(
+                    self._db, self.collection
+                )
```
aslagle: This way it doesn't call circulation.apply twice - if bibliographic data doesn't exist yet, it will get that first.

api/enki.py
```diff
@@ -762,30 +767,20 @@ def process_book(self, bibliographic):
...
             identifiers=False,
             subjects=True,
             contributions=True,
             formats=True,
         )
-        availability.apply(
-            self._db,
-            license_pool.collection,
-            replace=policy,
-        )
```
aslagle: It's unnecessary to call this separately, bibliographic.apply already calls it.

api/odl.py
```diff
@@ -1409,11 +1409,17 @@ def _detail_for_elementtree_entry(cls, parser, entry_tag, feed_url=None):
                 if copies_tags:
                     copies_tag = copies_tags[0]
                     licenses_available = copies_tag.attrib.get("available")
+                    if licenses_available != None:
+                        licenses_available = int(licenses_available)
```
aslagle: I don't understand why this started failing.

leonardr: I chased this down for a bit and couldn't figure it out either.

leonardr:
This looks good once the core branch is merged.

If the migration script is blocking this, I think it's okay to get a release out pretty soon and let the existing mechanisms fix the data gradually. Then maybe put in a migration script later.

PR #1184, opened 2019-02-19
Implement _run_self_tests for ExternalSearchIndex

adriana-alter:
https://jira.nypl.org/browse/SIMPLY-1714

Goes with https://github.com/NYPL-Simplified/server_core/pull/1030

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -169,6 +169,8 @@ def setup_admin_controllers(manager):
     from api.admin.controller.sitewide_services import *
     manager.admin_sitewide_services_controller = SitewideServicesController(manager)
     manager.admin_logging_services_controller = LoggingServicesController(manager)
+    from api.admin.controller.search_service_self_tests import SearchServiceSelfTestsController
+    manager.admin_search_service_self_tests_controller = SearchServiceSelfTestsController(manager)
     manager.admin_search_services_controller = SearchServicesController(manager)
     manager.admin_storage_services_controller = StorageServicesController(manager)
     from api.admin.controller.catalog_services import *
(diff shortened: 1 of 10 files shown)


api/admin/controller/__init__.py
```diff
@@ -2390,43 +2394,55 @@ def _get_collection_protocols(self, provider_apis):
...
+                provider_apis.append(OPDSImportMonitor)
+
+                if item.protocol == OPDSImportMonitor.PROTOCOL:
+                    protocol_class = OPDSImportMonitor
+
+                if protocol_class in provider_apis and issubclass(protocol_class, HasSelfTests):
+                    if (item.protocol == OPDSImportMonitor.PROTOCOL):
+                        extra_args = (OPDSImporter,)
+                    else:
+                        extra_args = ()
```
leonardr: As you do this work, this code is accumulating special cases -- we do something different if we have an OPDSImportMonitor, we do something different if it's a collection versus a search index.

This isn't something to worry about now, but once you expand this to also cover the self-tests for patron authentication integrations, this code will probably get another `else if` clause, and once that's done it's worth seeing if we can simplify it, maybe by creating specialized subclasses of `HasSelfTests` for collection integrations, patron integrations, and search integrations, and putting the special logic in the subclasses.

api/admin/controller/search_service_self_tests.py
```diff
@@ -0,0 +1,45 @@
...
+
+from core.selftest import HasSelfTests
+from core.util.problem_detail import ProblemDetail
+from api.admin.controller.self_tests import SelfTestsController
+
+class SearchServiceSelfTestsController(SelfTestsController, ExternalSearchTest):
+
+    def __init__(self, manager):
+        super(SearchServiceSelfTestsController, self).__init__(manager)
+        self.missing_id_error = MISSING_SEARCH_SERVICE_IDENTIFIER
```
leonardr: I don't think the specific error is super important here, since this isn't an error a normal user is likely to see. I think you could use MISSING_SERVICE wherever you would use MISSING_SEARCH_SERVICE_IDENTIFIER or MISSING_COLLECTION_IDENTIFIER, and get rid of both of those more specific documents. If you want to be more specific about what's missing, you could return something like `MISSING_SERVICE.detailed(self.missing_id_error)` and then have `missing_id_error` be a localized string like `_("Missing search service identifier")`.

This isn't a big deal, but it would let us get rid of a couple problem detail documents.

leonardr:
This looks good to me now, but since you and I both worked on it I'd like to see what @EdwinGuzman thinks.

api/admin/controller/search_service_self_tests.py
adriana-alter: MISSING_SERVICE seems to me like too different a type of error (it's a 404 error for if the service you looked up couldn't be found; these are 400 errors for if you somehow didn't pass in an identifier, so the lookup couldn't be performed in the first place).  I'm making a generic MISSING_IDENTIFIER error, which we can probably also use in other places at some point, and abandoning the idea of using the detailed `self.missing_id_error` (since the user probably won't see it anyway); what do you think?

leonardr: That sounds right to me.

leonardr:
That's good enough for me -- let's merge this.

PR #1185, opened 2019-02-20
Remove flask monkeypatch

jonathangreen:
This patch is now included in the latest flask releases we are using, so it no longer needs to be monkeypatched in our code.

Flask issue here:
https://github.com/pallets/flask/issues/879

You can see its patched here: 
https://github.com/pallets/flask/blob/master/flask/app.py#L1860-L1874

The newest flask with this patch is being installed as we are not constraining flask version here: 
https://github.com/NYPL-Simplified/circulation/blob/master/requirements.txt#L11


leonardr:
Great catch.

PR #1186, opened 2019-02-21
Fix NoveList library list bug where a non-NoveList library terminates update process

alsrlw:
This fixes the NoveList library list issue as discussed. I did have to add an import for the CannotLoadConfiguration class to the novelist.py file. Otherwise, I've tested the change in v.2.2.14 on our live CM, scheduling a separate novelist_update job with a non-NoveList library and our new NoveList library as parameters. The log indicates an 'info' entry for the non-novelist library and follows with a good connection/update entry for the NoveList library.

### DIFF
--- a/tests/test_lanes.py
+++ b/tests/test_lanes.py
@@ -342,7 +342,7 @@ class TestRelatedBooksLane(DatabaseTest):
         self._db.commit()
 
         assert_raises(
-            CannotLoadConfiguration, RelatedBooksLane, self._default_library, self.work, ""
+            ValueError, RelatedBooksLane, self._default_library, self.work, ""
         )
 
         # A book with a contributor initializes a RelatedBooksLane.
--- a/tests/test_lanes.py
+++ b/tests/test_lanes.py
@@ -29,6 +29,7 @@ from core.model import (
 
 from api.config import (
     Configuration,
+    CannotLoadConfiguration,
     temp_config,
 )
 from api.lanes import (
@@ -341,7 +342,7 @@ class TestRelatedBooksLane(DatabaseTest):
         self._db.commit()
 
         assert_raises(
-            ValueError, RelatedBooksLane, self._default_library, self.work, ""
+            CannotLoadConfiguration, RelatedBooksLane, self._default_library, self.work, ""
         )
 
         # A book with a contributor initializes a RelatedBooksLane.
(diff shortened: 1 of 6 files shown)


leonardr:
The test failures in test_novelist.py are because test_from_config is expecting a ValueError rather than a CannotLoadConfiguration.

There's a similar problem in api/lanes.py, in _recommendation_sublane -- it expects a ValueError when NoveList isn't configured. And there's also one in test_lanes.py, in TestWorkBasedLane.test_initialization.

Changing ValueError to CannotLoadConfiguration in all three places should fix the tests.

alsrlw:
Also, I see a config from/import reference in lanes.py and test_lanes.py, to which I think I should add the CannotLoadConfiguration class. But I don't see a config from/import line in test_novelist.py at all (surprised me). I'll need to add one there, along with the class in the others, right?

Also, the test_initialization you reference in test_lanes is in TestRelatedBooksLane, right?

leonardr:
Yes, you'll need to add `from api.config import CannotLoadConfiguration` to test_lanes.py. If a file already imports something from api.config, you can append `CannotLoadConfiguration` to the list of things imported.

And yes, I do mean TestRelatedBooksLane.test_initialization.

alsrlw:
Rats, failed again. I see that the ValueError reference in line 834 of lanes.py causes a failure. I had actually seen that in my review, but I convinced myself that the ValueError there was of a different stripe, with the self._recommendation_sublane reference in the _get_sublanes function just not appending anything. Is the ValueError there related to the CannotLoadConfiguration error in _recommendation_sublane?

leonardr:
I see. Yes, the `ValueError` raised in `RelatedBooksLane.__init__` is different. It's caused by (among other things) the lack of a NoveList integration, but it's a different type of error, and I think ValueError is okay there.

alsrlw:
Anything else I need to do so it doesn't break the build?

leonardr:
The only other thing I see is there's a `except ValueError` on line 1618 of api/controller.py that should be catching `CannotLoadConfiguration` now.

tests/test_lanes.py
```diff
@@ -341,7 +342,7 @@ def test_initialization(self):
         self._db.commit()
 
         assert_raises(
-            ValueError, RelatedBooksLane, self._default_library, self.work, ""
+            CannotLoadConfiguration, RelatedBooksLane, self._default_library, self.work, ""
```
leonardr: I think this is causing your final failing test. RelatedBooksLane.__init__ in api/lanes.py is raising its own ValueError, so if you change this back to ValueError the tests should all pass.

leonardr:
I filed https://jira.nypl.org/browse/SIMPLY-1739 to track this fix.

PR #1187, opened 2019-02-22
Monitors track their achievements across runs

leonardr:
This branch completes the work for https://jira.nypl.org/browse/SIMPLY-1699.

This branch modifies all circulation manager Monitors so that they report their progress when they run. "Progress" is calculated differently for each monitor, but it's generally the number of database records or books processed according to guidance from some external API.

All monitors gathered this data in their own way to send to the logs. Now the data is recorded to the `timestamps` table as well. In some cases the numbers were calculated by a method called by `run_once` or `catch_up_from`, so I had to change the return value of that method to propagate those numbers, so that `run_once` or `catch_up_from` could record them.

These Monitors are TimelineMonitors so I changed their `catch_up_from` implementations to set `progress.achievements`, which is the simple solution:

* Axis360CirculationMonitor
* BibliothecaEventMonitor
* EnkiImport
* OdiloCirculationMonitor
* ODLConsolidatedCopiesMonitor
* OverdriveCirculationMonitor

These monitors override `Monitor.run_once`, so to add achievement support I needed to make sure they return a brand new `TimestampData` that doesn't have values for `start` or `end`.

* MWAuxiliaryMetadataMonitor
* ODLHoldReaper
* OPDSForDistributorsReaperMonitor
* RBDigitalCirculationMonitor
* RBDigitalSyncMonitor

`MWCollectionUpdateMonitor` overrides `run_once` but it also manages its own timestamp, so it's okay that that one modifies `progress.achievements` -- it's also setting its own values for `start` and `finish`.

### DIFF
--- a/api/axis.py
+++ b/api/axis.py
@@ -547,9 +547,7 @@ class Axis360CirculationMonitor(CollectionMonitor, TimelineMonitor):
             count += 1
             if count % self.batch_size == 0:
                 self._db.commit()
-        progress.achievements = progress.format_achievements(
-            "Caught up with %(number)s %(thing)s.", count, "modified title"
-        )
+        progress.achievements = "Modified titles: %d." % count
 
     def process_book(self, bibliographic, availability):
 
--- a/api/axis.py
+++ b/api/axis.py
@@ -547,6 +547,9 @@ class Axis360CirculationMonitor(CollectionMonitor, TimelineMonitor):
             count += 1
             if count % self.batch_size == 0:
                 self._db.commit()
+        progress.achievements = progress.format_achievements(
+            "Caught up with %(number)s %(thing)s.", count, "modified title"
+        )
 
     def process_book(self, bibliographic, availability):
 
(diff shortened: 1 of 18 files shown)


api/axis.py
```diff
@@ -547,6 +547,7 @@ def catch_up_from(self, start, cutoff, progress):
             count += 1
             if count % self.batch_size == 0:
                 self._db.commit()
+        progress.achievements = "Modified titles: %d." % count
```
leonardr: This is a little stilted but it's very easy to localize (no worries about plurals) and this isn't patron-facing prose.

api/odl.py
```diff
@@ -924,13 +930,21 @@ def run_once(self, progress):
...
             changed_pools.add(hold.license_pool)
             self._db.delete(hold)
+            total_deleted_holds += 1
 
         for pool in changed_pools:
             self.api.update_hold_queue(pool)
 
+        message = "Holds deleted: %d. License pools updated: %d" % (
+            total_deleted_holds,
+            len(changed_pools)
```
aslagle: I think these two numbers are always the same.

aslagle: Never mind, I guess not if one pool has multiple deleted holds.

PR #1189, opened 2019-02-25
Change the process by which RBdigital accounts are looked up/created for circulation manager patrons

leonardr:
This branch implements our side of the changes laid out in https://jira.nypl.org/browse/SIMPLY-1723.

The functional changes are these:

* We will look up an existing RBdigital account for a patron's actual barcode before trying to create a new account for them.
* If the ILS provides us with access to the patron's email address, we will use it, plus their real barcode, to create an RBdigital account that can be used in other contexts, e.g. through the RBdigital web site.
* If we need to create a dummy account for a patron, we will use a patron identifier based on their barcode, rather than one that's totally random. This will get us past library-specific data validation.
* A dummy value for `postalCode` is now provided when creating a patron account.

To implement these changes I refactored the account creation code and added a lot of mocked-up tests.

### DIFF
--- a/api/clever/__init__.py
+++ b/api/clever/__init__.py
@@ -180,7 +180,7 @@ class CleverAuthenticationAPI(OAuthAuthenticationProvider):
             'Authorization': 'Bearer %s' % token
         }
         result = self._get(self.CLEVER_API_BASE_URL + '/me', bearer_headers)
-        data = result.get('data', {})
+        data = result.get('data', {}) or {}
 
         identifier = data.get('id', None)
 
--- a/api/rbdigital.py
+++ b/api/rbdigital.py
@@ -714,7 +714,7 @@ class RBDigitalAPI(BaseCirculationAPI, HasSelfTests):
         credential = Credential.lookup(
             _db, DataSource.RB_DIGITAL,
             Credential.IDENTIFIER_FROM_REMOTE_SERVICE,
-            patron, _refresh_credential, allow_persistent_token=True
+            patron, refresh_credential, allow_persistent_token=True
         )
         return credential.credential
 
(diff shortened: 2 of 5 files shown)


api/clever/__init__.py
```diff
@@ -180,7 +180,7 @@ def remote_patron_lookup(self, token):
             'Authorization': 'Bearer %s' % token
         }
         result = self._get(self.CLEVER_API_BASE_URL + '/me', bearer_headers)
-        data = result.get('data', {})
+        data = result.get('data', {}) or {}
```
leonardr: This is an unrelated change. At one point while testing `patron_email_address` I ended up with a real but misconfigured Clever OAuth provider that made a real request to Clever and got back a dictionary that had `data: None`. If that happens for real, this will at least ensure we get to the point where we can give a good error message.

api/rbdigital.py
```diff
@@ -932,31 +1057,6 @@ def get_patron_holds(self, patron_id):
 
         return holds
 
-    def get_patron_information(self, patron_id):
```
leonardr: This method was never used, so I removed it.

api/rbdigital.py
```diff
@@ -1002,7 +1102,7 @@ def validate_response(self, response, message, action=""):
         if response.status_code not in [200, 201]:
             if not message:
                 message = response.text
-            self.log.warning("%s call failed: %s ", action, message)
+            self.log.info("%s call failed: %s ", action, message)
```
leonardr: This is no longer a 'warning' scenario (if it ever was). It's very common that this call fails because we're looking up the barcode of a patron who doesn't have an account yet.

leonardr:
I filed https://jira.nypl.org/browse/SIMPLY-1741 to get answers to the policy questions.

PR #1190, opened 2019-02-26
Rbdigital test refactoring

leonardr:
This branch refactors some of the RBdigital tests. The functionality is unchanged except for some very minor bug fixes.

Highlights:

* I removed 5 of the books in `response_catalog_all_sample.json`, leaving only 3. This is plenty to test with and it makes `test_populate_all_catalog()` run significantly faster.
* I test `populate_all_catalog` and `populate_delta` in isolation and in separate tests; previously they were being tested in the test of `RBDigitalSyncMonitor`, and one important case -- a book that is added to the collection during `populate_delta` -- was not properly tested.
* I test all of the subclasses of `RBDigitalSyncMonitor`, in isolation, not just `RBDigitalSyncMonitor` itself.


api/rbdigital.py
```diff
@@ -1410,9 +1410,9 @@ def populate_delta(self, months=1):
         delta = self.get_delta(from_date=(today - time_ago), to_date=today)
         if not delta or len(delta) < 1:
             return None, None
-
-        items_added = delta[0].get("addedTitles", 0)
-        items_removed = delta[0].get("removedTitles", 0)
+        [delta] = delta
+        items_added = delta.get("addedTitles", [])
+        items_removed = delta.get("removedTitles", [])
```
leonardr: This is the bug fix. if `addedTitles` or `removedTitles` are ever missing, we would have ended up calling `len(0)` and crashing.

api/rbdigital.py
```diff
@@ -2039,7 +2044,9 @@ def __init__(self, _db, collection, api_class=RBDigitalAPI,
                  api_class_kwargs={}):
         """Constructor."""
         super(RBDigitalSyncMonitor, self).__init__(_db, collection)
-        self.api = api_class(_db, collection, **api_class_kwargs)
+        if not isinstance(api_class, RBDigitalAPI):
+            api_class = api_class(_db, collection, **api_class_kwargs)
+        self.api = api_class
```
leonardr: This is a common pattern with other vendor APIs; it lets me pass in an already-instantiated mock API class.

tests/files/rbdigital/response_catalog_delta.json
```diff
@@ -42,7 +41,7 @@
         "audience": "Childrens",
         "genres": "JUVENILE NONFICTION / Biography & Autobiography / Sports & Recreation",
         "primaryGenre": "biography-autobiography-memoir,history,juvenile-nonfiction,sports",
-        "isbn": "9781615730186",
+        "isbn": "9780590543439",
```
leonardr: This was the same ISBN as "Tricks", which made it impossible to separate what happened to this book from what happened to "Tricks".

tests/test_rbdigital.py
```diff
@@ -664,6 +681,100 @@ def test_get_metadata_by_isbn(self):
...
+        # Test the method that retrieves the entire catalog from RBdigital
+        # and mirrors it locally.
+
+        datastr, datadict = self.get_data("response_catalog_all_sample.json")
+        self.api.queue_response(status_code=200, content=datastr)
+        result = self.api.populate_all_catalog()
+
+        # populate_all_catalog returns two numbers, as required by
+        # RBDigitalSyncMonitor.
+        eq_((3, 3), result)
```
leonardr: This was previously not tested.

PR #1192, opened 2019-02-26
Remove the 'url' validation from the Axis 360 server, since either nicknames or URLs are valid.

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-1724.


adriana-alter:
Why not just set the "allowed" field instead of removing the validation?  We do this with the patron web client URL, which is allowed to be "*"; see line 161 of api/config.py.  That way, "production" and "qa" will be valid, but no other non-URL strings.

leonardr:
What a great feature! I'd forgotten about that.

PR #1197, opened 2019-03-01
Patron auth self tests

adriana-alter:
Goes with https://github.com/NYPL-Simplified/circulation-web/pull/214
Resolves https://jira.nypl.org/browse/SIMPLY-1730


api/admin/controller/self_tests.py
```diff
@@ -20,14 +20,22 @@ def process_get(self, identifier):
...
-        if (value):
+        if value and isinstance(value, ProblemDetail):
+            return value
+        elif value:
             return Response(_("Successfully ran new self tests"), 200)
-        return FAILED_TO_RUN_SELF_TESTS
+        return FAILED_TO_RUN_SELF_TESTS.detailed(_(
+            "Failed to run self tests for this %s."
+            %(self.type)
+        ))
```
leonardr: A couple minor things about `_()`:

* The string interpolation needs to go outside of the `_()` call, because the string in the `_()` call is what gets translated.

* It's better if the whole thing is on one line, because we use an automated tool to grab out the strings to be translated.

So:

```
return FAILED_TO_RUN_SELF_TESTS.detailed(
    _("Failed to run self tests for this %s.") % self.type
)
```

Even this isn't perfect because `self.type` also needs to be translated, but this will at least make us notice that there's some translation happening here.

aslagle: I would do 
```
return FAILED_TO_RUN_SELF_TESTS.detailed(
    _("Failed to run self tests for this %(type)s.", self.type)
)
```

and then also use `_()` on the type names where those are defined.

api/admin/controller/__init__.py
```diff
@@ -2432,20 +2430,31 @@ def _get_prior_test_results(self, item, protocol_class=None):
...
+                library = None
+                if len(item.libraries):
+                    library = item.libraries[0]
+                    self_test_results = protocol_class.prior_test_results(
+                        self._db, None, library, item
+                    )
+                else:
+                    self_test_results = dict(
+                        exception=_("You must associate this service with at least one library before you can run self tests for it."),
+                        disabled=True
```
leonardr: Can you return a problem detail document instead of this custom data structure?

adriana-alter: I think so...but I'm pretty sure that (unless I special-case the front-end rendering in a weird, unintuitive way) that will make it display like an error message.  Is that what we want here?

leonardr: I see, this is being displayed in the context of the test results. I ran into a similar problem myself just now. Leave this alone for now and we can revisit improving the handling of unusual error cases later.

PR #1199, opened 2019-03-04
Make Axis 360 collections check for the presence of Blio titles

leonardr:
This branch gets us most of the way to a fix for https://jira.nypl.org/browse/SIMPLY-1711. Any Axis 360 titles that have no delivery mechanisms are called out in a special self-test.

Since there's nothing Axis 360-specific about a problem where titles have no delivery mechanisms, I put this code into a new mixin class, HasCollectionSelfTests, which we can put into other API classes as appropriate.

I've noticed that when Baker & Taylor fixes these problematic books, we don't seem to pick up on the fact that the problem has been fixed, so I may need another branch to fix that part of the problem.


leonardr:
A book with no delivery mechanism just means that whoever imported the book decided not to create a delivery mechanism for it. This could be due to a bug or because the importing code doesn't recognize the format.

In the Axis 360 case, we know there's no bug (because real EPUBs have EPUB as a delivery mechanisms), and we  know that Blio and AxisNow don't correspond to any known delivery mechanism. So in the case of Axis 360 it's a deliberate decision, and if it happens to a given book, the most likely explanation is that B&T messed up and gave us a Blio version of what ought to be an EPUB.

An alternative would be to explicitly mark down Blio as a delivery mechanism that we don't support. That would make our database more accurate, but it would also make this particular branch more difficult -- we'd have to find books that have Blio but not EPUB.

PR #1201, opened 2019-03-07
Use the Enki ID for the patron's library when in a patron context; otherwise omit it or use dummy value

leonardr:
This addresses https://jira.nypl.org/browse/SIMPLY-1757. The big change is that `library_id` is no longer stored in the `EnkiAPI` constructor, since a single `EnkiAPI` is responsible for managing the Enki collection for all libraries on the circulation manager.

When the API is acting on behalf of a specific patron, we look up the Enki library ID for that patron's library and use it.

When the API is not acting on behalf of a specific patron, we omit the `lib` variable from API requests (if possible) or pass `lib=0` (if it's required), to indicate that we're asking about the Enki collection itself and not any particular library.,

### DIFF
--- a/api/enki.py
+++ b/api/enki.py
@@ -229,7 +229,6 @@ class EnkiAPI(BaseCirculationAPI, HasSelfTests):
             minutes=minutes,
         )
         response = self.request(url, params=args)
-        set_trace()
         data = json.loads(response.content)
         parser = BibliographicParser()
         for element in data['result']['recentactivity']:
--- a/api/enki.py
+++ b/api/enki.py
@@ -31,6 +31,7 @@ from core.util.http import (
 from core.model import (
     CirculationEvent,
     Collection,
+    ConfigurationSetting,
     DataSource,
     DeliveryMechanism,
     Edition,
@@ -112,16 +113,18 @@ class EnkiAPI(BaseCirculationAPI, HasSelfTests):
             )
 
         self.collection_id = collection.id
-        self.library_id = collection.external_integration.setting(self.ENKI_LIBRARY_ID_KEY).value
         self.base_url = collection.external_integration.url or self.PRODUCTION_BASE_URL
-
-        if not self.library_id or not self.base_url:
-            raise CannotLoadConfiguration(
-                "Enki configuration is incomplete."
-            )
 
     def external_integration(self, _db):
         return self.collection.external_integration
+
+    def enki_library_id(self, library):
+        """Find the Enki library ID for the given library."""
+        _db = Session.object_session(library)
+        return ConfigurationSetting.for_library_and_externalintegration(
+            _db, self.ENKI_LIBRARY_ID_KEY, library,
+            self.external_integration(_db)
+        ).value
 
     @property
     def collection(self):
@@ -224,7 +227,7 @@ class EnkiAPI(BaseCirculationAPI, HasSelfTests):
         args = dict(
             method='getRecentActivity',
             minutes=minutes,
-            lib=self.library_id
+            lib='0'
         )
         response = self.request(url, params=args)
         data = json.loads(response.content)
@@ -251,7 +254,7 @@ class EnkiAPI(BaseCirculationAPI, HasSelfTests):
             method='getUpdateTitles',
             minutes=minutes,
             id='secontent',
-            lib=self.library_id
+            lib='0'
         )
         response = self.request(url, params=args)
         for metadata in BibliographicParser().process_all(response.content):
@@ -269,7 +272,7 @@ class EnkiAPI(BaseCirculationAPI, HasSelfTests):
         args = dict(
             method="getItem",
             recordid=enki_id,
-            lib=self.library_id,
+            lib='0',
             size="large",
         )
         response = self.request(url, params=args)
@@ -300,7 +303,7 @@ class EnkiAPI(BaseCirculationAPI, HasSelfTests):
         args['id'] = "secontent"
         args['strt'] = strt
         args['qty'] = qty
-        args['lib'] = self.library_id
+        args['lib'] = '0'
         response = self.request(url, params=args)
         for metadata in BibliographicParser().process_all(response.content):
             yield metadata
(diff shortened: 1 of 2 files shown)


tests/test_enki.py
```diff
@@ -297,18 +314,19 @@ def test_get_all_titles(self):
...
 
+        # Unlike some API calls, it's not necessary to pass 'lib' in here.
+        assert 'lib' not in params
+
     def test__epoch_to_struct(self):
         """Test the _epoch_to_struct helper method."""
         eq_(datetime.datetime(1970, 1, 1), EnkiAPI._epoch_to_struct("0"))
 
-    def test_checkout_open_access(self):
+    def test_checkout_open_access_parser(self):
```
leonardr: These tests tested the parser for the result of the API call, but not what came in and went out of the API calls themselves, so I had to add some more test coverage (e.g. test_checkout_success below).

tests/test_enki.py
```diff
@@ -364,31 +410,76 @@ def test_checkout_not_available(self):
...
+        eq_(fulfillment.identifier, pool.identifier.identifier)
+        eq_(fulfillment.collection_id, pool.collection.id)
+        eq_(DeliveryMechanism.ADOBE_DRM, fulfillment.content_type)
+        assert fulfillment.content_link.startswith(
+            "http://afs.enkilibrary.org/fulfillment/URLLink.acsm"
+        )
+        eq_(fulfillment.content_expires,
+            datetime.datetime(2017, 9, 13, 19, 42, 35, 0))
+
     def test_patron_activity(self):
```
leonardr: This test verified the effects of calling `patron_activity` but not the request that went over the network.

api/enki.py
```diff
@@ -318,7 +323,11 @@ def _epoch_to_struct(cls, epoch_string):
     def checkout(self, patron, pin, licensepool, internal_format):
         identifier = licensepool.identifier
         enki_id = identifier.identifier
-        response = self.loan_request(patron.authorization_identifier, pin, enki_id)
+        enki_library_id = self.enki_library_id(patron.library)
+        response = self.loan_request(
+            patron.authorization_identifier, pin, enki_id,
+            enki_library_id
+        )
```
leonardr: Here the library ID does matter, quite a bit, so we grab it from the patron's library.

api/enki.py
```diff
@@ -368,7 +377,10 @@ def fulfill(self, patron, pin, licensepool, internal_format, **kwargs):
         :return: a FulfillmentInfo object.
         """
         book_id = licensepool.identifier.identifier
-        response = self.loan_request(patron.authorization_identifier, pin, book_id)
+        enki_library_id = self.enki_library_id(patron.library)
+        response = self.loan_request(
+            patron.authorization_identifier, pin, book_id, enki_library_id
+        )
```
leonardr: Library ID also matters here since fulfill and checkout are basically the same operation.

api/enki.py
```diff
@@ -414,7 +426,10 @@ def parse_fulfill_result(self, result):
         return (url, item_type, expires)
 
     def patron_activity(self, patron, pin):
-        response = self.patron_request(patron.authorization_identifier, pin)
+        enki_library_id = self.enki_library_id(patron.library)
+        response = self.patron_request(
+            patron.authorization_identifier, pin, enki_library_id
+        )
```
leonardr: And library ID matters here because we're asking about a patron of a specific Enki library.

api/enki.py
```diff
@@ -346,14 +355,14 @@ def checkout(self, patron, pin, licensepool, internal_format):
         )
         return loan
 
-    def loan_request(self, barcode, pin, book_id):
+    def loan_request(self, barcode, pin, book_id, enki_library_id):
```
EdwinGuzman: Both `checkout` and `fulfill` use this method. Checkout is borrowing the book, but fulfill is actually downloading it, correct?

api/enki.py
```diff
@@ -481,9 +496,16 @@ def __init__(self, _db, collection=None, *args, **kwargs):
                 _db, name="Test Enki Collection", protocol=EnkiAPI.ENKI
             )
             collection.protocol=EnkiAPI.ENKI
-            collection.external_integration.setting(self.ENKI_LIBRARY_ID_KEY).value = u'c'
         if collection not in library.collections:
             library.collections.append(collection)
+
+        # Set the "Enki library ID" variable between the default library
+        # and this Enki collection.
+        ConfigurationSetting.for_library_and_externalintegration(
```
EdwinGuzman: This is also the default value that'll appear in the admin interface?

api/enki.py
leonardr: Yes, that's right. For Enki, checkout and fulfillment are the same API call, but we need slightly different methods to implement the `CirculationAPI` contract.

api/enki.py
leonardr: No, this is only used in unit tests. When you add a library to an Enki collection in the admin interface, no default value will be given for this configuration setting.

tests/test_enki.py
```diff
@@ -364,31 +410,76 @@ def test_checkout_not_available(self):
...
+        # A FulfillmentInfo for the loan was returned.
+        assert isinstance(fulfillment, FulfillmentInfo)
+        eq_(fulfillment.identifier, pool.identifier.identifier)
+        eq_(fulfillment.collection_id, pool.collection.id)
+        eq_(DeliveryMechanism.ADOBE_DRM, fulfillment.content_type)
+        assert fulfillment.content_link.startswith(
+            "http://afs.enkilibrary.org/fulfillment/URLLink.acsm"
+        )
+        eq_(fulfillment.content_expires,
+            datetime.datetime(2017, 9, 13, 19, 42, 35, 0))
```
EdwinGuzman: How'd you get this value?

leonardr: It's the `duedate` from `checked_out_acs.json`. This is supposed to be the time the ACSM file expires. Once an ACSM file has been generated, it's usually good for eight hours. This ACSM file hasn't been generated -- it's generated when you hit `URLLink.acsm` -- so we fall back to using the expiration time of the loan as the expiration.

I don't think this is actually used by anything, but it's up to SimplyE what to do with this information.

PR #1202, opened 2019-03-08
Novelist book info update

EdwinGuzman:
Fixes [SIMPLY-1749](https://jira.nypl.org/browse/SIMPLY-1749).

Adding `narrator`, `publicationDate`, and `distributor` to the book object that gets sent to Novelist.

### DIFF
--- a/api/novelist.py
+++ b/api/novelist.py
@@ -473,6 +473,8 @@ class NoveListAPI(object):
         is being processed. If the next ISBN being processed is new, the existing one
         gets added to the list of items. If the ISBN is the same, then we append
         the Author property since there are multiple contributors.
+
+        :return: a list of Novelist objects to send
         """
         collectionList = []
         for c in library.collections:
@@ -482,14 +484,13 @@ class NoveListAPI(object):
         i1 = aliased(Identifier)
         i2 = aliased(Identifier)
         roles = list(Contributor.AUTHOR_ROLES)
-        # TODO: We should handle the Narrator role properly, by
-        # setting the 'narrator' field in the NoveList API document.
-        # roles.append(Contributor.NARRATOR_ROLE)
+        roles.append(Contributor.NARRATOR_ROLE)
 
         isbnQuery = select(
             [i1.identifier, i1.type, i2.identifier,
-            Edition.title, Edition.medium,
-            Contribution.role, Contributor.sort_name],
+            Edition.title, Edition.medium, Edition.published,
+            Contribution.role, Contributor.sort_name,
+            DataSource.name],
         ).select_from(
             join(LicensePool, i1, i1.id==LicensePool.identifier_id)
             .join(Equivalency, i1.id==Equivalency.input_id, LEFT_OUTER_JOIN)
@@ -500,6 +501,7 @@ class NoveListAPI(object):
             )
             .join(Contribution, Edition.id==Contribution.edition_id)
             .join(Contributor, Contribution.contributor_id==Contributor.id)
+            .join(DataSource, DataSource.id==LicensePool.data_source_id)
         ).where(
             and_(
                 LicensePool.collection_id.in_(collectionList),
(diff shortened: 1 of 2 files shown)


api/novelist.py
```diff
@@ -551,31 +569,50 @@ def create_item_object(self, object, currentIdentifier, existingItem):
...
+
+        role = object[6]
+        author_or_narrator = object[7] if role in roles else ""
+        distributor = object[8]
 
         # If we encounter an existing ISBN and its role is "Primary Author",
         # then that value overrides the existing Author property.
-        #
-        # TODO: add 'narrator' field when we encounter a Narrator role.
+        # If the role is narrator, then we don't need to worry about this step.
```
EdwinGuzman: Here, I am assuming that, unlike the author role, there won't be another narrator role to update the value for. Is that correct to assume? (In this one, if "Primary Author" was found after "Author", that value replaced the first one).

api/novelist.py
```diff
@@ -551,31 +569,50 @@ def create_item_object(self, object, currentIdentifier, existingItem):
...
                 title=title,
                 mediaType=mediaType,
-                author=author,
                 role=role,
+                distributor=distributor
             )
-            return (isbn, existingItem, newItem, True)
+
+            publicationDate = object[5]
+            if publicationDate:
```
EdwinGuzman: This gave me an issue in a test because the value was None. There can be books without a publication date?

api/novelist.py
```diff
@@ -538,6 +545,17 @@ def create_item_object(self, object, currentIdentifier, existingItem):
...
+        :param object: the current item object to process
+        :param currentIdentifier: the current identifier to process
+        :param existingItem: the previously processed item object
+
+        :return: (
+            current identifier,
+            the existing object if available,
+            a new object if the item wasn't found before,
+            if the item is ready to the added to the list of books to send
+        )
```
leonardr: I like this way of talking about a tuple return value.

api/novelist.py
leonardr: Yeah, publication date is just a piece of information that the distributor is supposed to provide, and some distributors don't provide it.

leonardr:
NoveList said the fields should be called "Narrator", "PublicationDate", and "Distributor", but the fields we have now are in camelCase. I think you have the field names right, since you also used camelCase, but we should go back to NoveList to make sure.

api/novelist.py
leonardr: Yeah, there's no equivalent role to "Primary Narrator". If a book has multiple narrators I think picking the first one is fine.

api/novelist.py
```diff
@@ -551,31 +569,50 @@ def create_item_object(self, object, currentIdentifier, existingItem):
             # a data error.
             return (None, None, None, False)
 
-        role = object[5]
-        author = object[6] if role in Contributor.AUTHOR_ROLES else ""
+        roles = list(Contributor.AUTHOR_ROLES)
+        roles.append(Contributor.NARRATOR_ROLE)
+
+        role = object[6]
+        author_or_narrator = object[7] if role in roles else ""
```
leonardr: NoveList asked that narrator be provided separately from author, in a field called 'narrator'. So I think you need to keep track of the last author you saw _and_ the last narrator you saw.

leonardr:
My NoveList contact says that the field names are case-insensitive.

EdwinGuzman:
That's good about the case; I remember the docs had them all capitalized and vaguely remember having to change that to camelcase.

EdwinGuzman:
If a book has both author and narrator, wouldn't it appear in the same `Contribution.role` and `Contribution.sort_name` values? If it's like the the case when the same book appears again and there are different values, I could update the object with the new author or narrator value?

leonardr:
Looking at the full code again, I think you can get this to work without changing the query more than you already have.

Let's say a book has an author, a primary author, and two narrators. As you said, it's going to end up with 4 rows in the search results. So I think you can get this to work by keeping track of both an author and a narrator.

With author, you might tentatively decide that someone is 'the' author, but then encounter a primary author and have to change it. There's no "primary narrator" role, so the first narrator you encounter will be 'the' narrator and there's no need to change it.

Once you iterate over all the rows for a book, you'll have author and narrator (or one or the other, or neither). Stick the author in 'author' (if you have it) and the narrator in 'narrator (if you have it). Then we should be good.

EdwinGuzman:
I'd like to see if I can refactor some of the logic for the loop since the use case changed before merging this in.

leonardr:
Sounds good.

PR #1204, opened 2019-03-08
If there's an exception initializing the search index, the search controller will return a problem detail document rather than faking it with an empty list of search results.

leonardr:
This is a very simple way of addressing https://jira.nypl.org/browse/SIMPLY-1214 -- simpler than the solution proposed in the issue.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -122,6 +122,7 @@ from authenticator import (
 from config import (
     Configuration,
     CannotLoadConfiguration,
+    IntegrationException,
 )
 
 from lanes import (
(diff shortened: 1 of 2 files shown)


api/controller.py
```diff
@@ -122,6 +122,7 @@
 from config import (
     Configuration,
     CannotLoadConfiguration,
+    IntegrationException,
```
EdwinGuzman: It looks like this need to be imported for the REMOTE_INTEGRATION_FAILED problem detail to work?

leonardr: This is a mistake on my part. I tried a version of the code where trying to run a search when there was no search index configured would raise IntegrationException. What I ended up with was simpler but I didn't get rid of the import statement.

PR #1205, opened 2019-03-10
ODL with individual copies

aslagle:
This branch adds support for tracking individual copies in ODL, using https://github.com/NYPL-Simplified/server_core/pull/1043

It can't be merged yet. Once we have an updated ODL spec, we'll need to make sure both this code and Feedbooks' ODL feed match it. There's also one outstanding bug Feedbooks needs to fix.


migration/20190320-1-add-license-to-loans.sql
```diff
@@ -0,0 +1,2 @@
+alter table loans add column license_id integer;
+alter table loans add constraint loans_license_id_key foreign key (license_id) references licenses(id);
```
leonardr: These migrations should be in core (even though circ is the only component that uses these fields) because the model code is also in core.

migration/20190320-2-remove-odl-consolidated-licenses.sql
```diff
@@ -0,0 +1,8 @@
...
+
+update licensepools set licenses_owned = 0 from collections join externalintegrations on collections.external_integration_id = externalintegrations.id where licensepools.collection_id = collections.id and externalintegrations.protocol = 'ODL with Consolidated Copies';
+update licensepools set licenses_available = 0 from collections join externalintegrations on collections.external_integration_id = externalintegrations.id where licensepools.collection_id = collections.id and externalintegrations.protocol = 'ODL with Consolidated Copies';
+
+delete from coveragerecords where coveragerecords.identifier_id in (select identifiers.id from identifiers join licensepools on licensepools.identifier_id = identifiers.id join collections on licensepools.collection_id = collections.id join externalintegrations on collections.external_integration_id = externalintegrations.id where externalintegrations.protocol = 'ODL With Consolidated Copies') and coveragerecords.operation = 'import';
+
+update externalintegrations set protocol = 'ODL' where protocol = 'ODL with Consolidated Copies';
```
leonardr: These migrations are fine to stay in circulation.

api/odl.py
```diff
@@ -268,16 +243,28 @@ def get_license_status_document(self, loan):
...
+                url, query_params = match.groups()
+                query_params = query_params.split(",")
+                if query_params:
+                    url += "?"
+
+                for param, value in params.items():
+                    if param in query_params:
+                        url += "%s=%s&" % (param, value)
+                if url[-1] == "&":
+                    url = url[:-1]
```
leonardr: Is it worth bringing in a library such as uritemplate?

api/odl.py
```diff
@@ -788,22 +758,31 @@ class ODLXMLParser(OPDSXMLParser):
...
     format information from 'odl:license' tags.
     """
-    NAME = ODLWithConsolidatedCopiesAPI.NAME
+    NAME = ODLAPI.NAME
     PARSER_CLASS = ODLXMLParser
 
     @classmethod
-    def _detail_for_elementtree_entry(cls, parser, entry_tag, feed_url=None):
+    def _detail_for_elementtree_entry(cls, parser, entry_tag, feed_url=None, do_get=None):
+        do_get = do_get or Representation.cautious_http_get
```
leonardr: Do you have a reason to be cautious here?

aslagle: I wanted to be able to pass `do_get` into the parser constructor from a test, but it already had a `do_get`, so I used that. The regular OPDSImporter does need to be cautious, and since I'm using the same one that's what normally gets used here. I can't use `HTTP.get_with_timeout` as the default since it's return values are different. I don't think the cautiousness hurts anything.

leonardr: You're right, it doesn't hurt anything, I just wanted to understand why you chose that here when you're not getting anything from Google Books or Gutenberg. It makes sense that you'd want to do things the same way as OPDSImporter.

api/odl.py
```diff
@@ -423,8 +434,16 @@ def _fulfill(self, loan):
...
+
+        links = doc.get("links", [])
+        content_link = None
+        content_type = None
+        for link in links:
+            if link.get("rel") == "license":
+                content_link = link.get("href")
+                content_type = link.get("type")
+                break
+
```
leonardr: This seems like a great case for a helper method similar to `OPDSImporter.extract_link`. But the only reason I'd say this is useful, as opposed to a good way to save code, is if these `href`s might be relative links that need to be made into absolute URLs.

Are relative links allowed in ODL? The closest thing I can find to an answer is that Readium Web Publication Manifest says that in a standalone manifest, "all resources in readingOrder, resources and links must be referenced relatively to the manifest".

leonardr:
Is there a published ODL spec? I only know about the Google Doc.

aslagle:
@leonardr https://drafts.opds.io/odl-1.0

api/odl.py
aslagle: This part is actually the LCP License Status Document. I'm not sure if it allows relative links.

aslagle: Sorry, this is actually ODL - but other places in the file that do this same thing are for the LCP License Status Document.

leonardr: Maybe this is a task for later, when we want a library that can turn all of these similar formats -- OPDS 2, ODL, LCP LSD -- into a Python data model.

PR #1206, opened 2019-03-12
Writing self tests for SIP2 connection and test patron testing.

EdwinGuzman:
Fixes [SIMPLY-1756](https://jira.nypl.org/browse/SIMPLY-1756).

Writing self tests for testing the SIP2 connection as well as logging in with a username and password. Once those pass, the test patron's self tests also run.

Samples of failing tests:
Screen Shot 2019-03-12 at 5 12 07 PM
Screen Shot 2019-03-12 at 5 12 33 PM

### DIFF
--- a/api/sip/client.py
+++ b/api/sip/client.py
@@ -844,10 +844,6 @@ class MockSIPClient(SIPClient):
         self.status.append("Creating new socket connection.")
         self.reset_connection_state()
         return None
-    
-    def login(self):
-        if not self.login_user_id and not self.login_password:
-            raise IOError("Error logging in")
 
     def do_send(self, data):
         self.requests.append(data)
--- a/api/sip/client.py
+++ b/api/sip/client.py
@@ -32,6 +32,7 @@ import re
 import socket
 import ssl
 import tempfile
+from nose.tools import set_trace
 
 # SIP2 defines a large number of fields which are used in request and
 # response messages. This library focuses on defining the response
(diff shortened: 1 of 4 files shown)


api/sip/client.py
```diff
@@ -315,14 +316,16 @@ def connect(self):
                 self.connection = self.make_secure_connection()
             else:
                 self.connection = self.make_insecure_connection()
+            
+            self.connection.settimeout(12)
```
EdwinGuzman: I moved this from outside the try statement to before the `connect()` call otherwise it would just stall and the tests would never run. Also, is this suppose to represent 12 seconds? The docs say so but it always seems to timeout at twice the number input.

leonardr: That's weird. Maybe what you're seeing is that there are two requests and they both have to time out for an error to be raised out of the SIP2 client?

Setting the timeout immediately makes sense.

EdwinGuzman:
Added another test to output the raw test patron's information. Screenshot truncated in case test information shouldn't be public.

Screen Shot 2019-03-13 at 3 49 56 PM

leonardr:
This is not what I was expecting. I forgot that SIP2Client automatically parses the data sent back from the SIP2 server and returns it in parsed form. I thought it would be easier to get the raw message.

However, this is definitely good enough to see whether or not SIP2Client understands an incoming SIP2 message, so let's keep it like this, try it in a debugging situation and see if we need to go further.

If we do want to go further, maybe we could have the SIP2 message parser stick the original message into that dictionary.

EdwinGuzman:
Alright, I added one more test for the test patron request data that gets sent.

PR #1208, opened 2019-03-15
Geographic areas

adriana-alter:


api/admin/controller/__init__.py
```diff
@@ -2609,7 +2609,15 @@ def validate_formats(self, settings=None):
     def _value(self, field):
         # Extract the user's input for this field. If this is a sitewide setting,
         # then the input needs to be accessed via "value" rather than via the setting's key.
-        return flask.request.form.get(field.get("key")) or flask.request.form.get("value")
+        # We use getlist instead of get so that, if the field is such that the user can input multiple values
+        # (e.g. language codes), we'll extract all the values, not just the first one.
+
+        value = flask.request.form.getlist(field.get("key"))
+        if not value:
+            return flask.request.form.get("value")
```
leonardr: Would it make sense to also call `getlist` here?

adriana-alter: Unnecessary; this is only called if a sitewide setting is being saved, and none of the sitewide settings accepts multiple values.  (There might be a case to be made for changing the "excluded audiobook sources" setting to make it accept multiple values, but that seems like a conversation for another branch...)

api/admin/controller/library_settings.py
```diff
@@ -235,3 +250,60 @@ def image_setting(self, setting):
             image.save(buffer, format="PNG")
             b64 = base64.b64encode(buffer.getvalue())
             return "data:image/png;base64,%s" % b64
+
+    def validate_geographic_areas(self, values):
```
leonardr: We talked about this in person; I'm recording the results of our conversation here.

This code looks correct, but there are a number of improvements you could make. This is a standalone method that will work even when there is no active HTTP request, and that gives you a lot of useful options that will make it more testable and more maintanable in the long run.

We decided not to do this immediately because there are a lot of validation methods in the controller module and it would be better to move them all out as part of the same project. I filed https://jira.nypl.org/browse/SIMPLY-1800 to cover this.

In the medium term we're going to change this so that if the validation code doesn't recognize a place, it calls out to the library registry and sees if _it_ recognizes the place. The local validation will then just be a quick way to avoid having to call out to the registry in most cases.

PR #1210, opened 2019-03-19
Open eBooks authentication updates

leonardr:
This branch makes two minor changes to the way we authenticate students through Open Ebooks

First, it finalizes the URL for the JWT-based FirstBook authentication service -- we're going from user acceptance testing to production later this week.

Second, it adds the NCES IDs for three newly constructed Title I schools in Texas, so that students from those schools can get access through Clever. The NCES website doesn't list information about these schools, but we confirmed with the school district that they are eligible:

https://nces.ed.gov/ccd/schoolsearch/school_detail.asp?Search=1&DistrictID=4828890&ID=482889013218
https://nces.ed.gov/ccd/schoolsearch/school_detail.asp?Search=1&DistrictID=4828890&ID=482889013527
https://nces.ed.gov/ccd/schoolsearch/school_detail.asp?Search=1&DistrictID=4828890&ID=482889013222

The 2017-2018 NCES data isn't out yet, but the last time I see us changing it is about a year ago when you brought in the 2014-2015 data. We do now have 2016-2017 data, so I'll do a separate project to get some updates in.


aslagle:
This looks good but you should also update the script in the operations repo with the whitelist changes.

PR #1211, opened 2019-03-19
Language values

adriana-alter:
Now that we have a feature (from the new geographic input stuff) for displaying extra information in a tooltip next to the saved input in an edit form: why not also show the full names of languages next to the saved language code strings?

Screen Shot 2019-03-19 at 10 31 42 AM


adriana-alter:
Goes with https://github.com/NYPL-Simplified/circulation-web/pull/218

aslagle:
There are other places where we have the language autocomplete and it shows the language name in the box instead of the code. Sorry don’t remember exactly where and I’m on my phone right now. I think it would be better to do that, but the tooltip is at least better than just the code.

adriana-alter:
That sounds like a really good feature!  (I'm having trouble finding it, but if you could point me in the right direction when you can--no rush, obviously!--that would be great.)  I also think it might be worth making it so that users have the option of typing in the language name instead of the code.  In the meantime, though...?

leonardr:
I don't know where this is in the code, but the language autocomplete feature shows up when you're editing the metadata for a book.

aslagle:
https://github.com/NYPL-Simplified/circulation-web/blob/master/src/components/BookEditForm.tsx#L165

adriana-alter:
@aslagle Thank you!!  Working on implementing it now.

PR #1218, opened 2019-04-03
Check whether the registry knows about locations

adriana-alter:
Note: once the validation refactoring is complete, this method will live in a validation file rather than in `library_settings`.

### DIFF
--- a/api/admin/controller/library_settings.py
+++ b/api/admin/controller/library_settings.py
@@ -16,16 +16,19 @@ from api.lanes import create_default_lanes
 from core.model import (
     ConfigurationSetting,
     create,
+    ExternalIntegration,
     get_one,
     Library,
     Representation,
 )
+from core.util.http import HTTP
 from PIL import Image
 from api.admin.exceptions import *
 from api.admin.problem_details import *
 from core.util.problem_detail import ProblemDetail
 from core.util import LanguageCodes
 from nose.tools import set_trace
+from api.registry import RemoteRegistry
 
 class LibrarySettingsController(SettingsController):
 
@@ -294,8 +297,8 @@ class LibrarySettingsController(SettingsController):
                     elif len([x for x in us_search.query(state=state, returns=None) if x.county == city_or_county]):
                         locations["US"].append(value);
                     else:
-                        return UNKNOWN_LOCATION.detailed(_('Unable to locate "%(value)s".', value=value))
-
+                        if self.ask_registry(value):
+                            return self.ask_registry(value)
                 elif value.isdigit():
                     # Is it a US zipcode?
                     info = us_search.by_zipcode(value)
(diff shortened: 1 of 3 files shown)


api/admin/controller/library_settings.py
```diff
@@ -294,19 +297,42 @@ def validate_geographic_areas(self, values):
                     elif len([x for x in us_search.query(state=state, returns=None) if x.county == city_or_county]):
                         locations["US"].append(value);
                     else:
-                        return UNKNOWN_LOCATION.detailed(_('Unable to locate "%(value)s".', value=value))
-
+                        if self.ask_registry(value):
+                            return self.ask_registry(value)
```
leonardr: This is going to call `ask_registry` twice, potentially making an unnecessary HTTP request. You can assign the result of `ask_registry` to a variable and then check it.

api/admin/controller/library_settings.py
```diff
@@ -294,19 +297,42 @@ def validate_geographic_areas(self, values):
...
+
+        for registry in RemoteRegistry.for_protocol_and_goal(
+            self._db, ExternalIntegration.OPDS_REGISTRATION, ExternalIntegration.DISCOVERY_GOAL
+        ):
+
+            base_url = registry.integration.url + "/coverage?coverage="
+
+            for nation in ["US", "CA"]:
+                search_term = '{"%s": "%s"}' % (nation, value)
+                response = HTTP.debuggable_get(base_url + search_term)
```
leonardr: Some terminology for the sake of this discussion: the `{"US": "x"}` thing is a [service area object](https://github.com/NYPL-Simplified/Simplified/wiki/Authentication-For-OPDS-Extensions#service_area). There's a wide variety of strings that become valid service area objects when converted to JSON.

I think `ask_registry` needs to take an actual service area object (or maybe a JSON-dumped string describing a service area object), rather than taking a string that has failed validation and seeing if `{"US": "x"}` or `{"CA": "x"}` works. This is the value that's going to be included in the library's authentication document for consumption by the library registry, so if the library registry understands `{"US": "x"}` but not `x`, we need to store `{"US": "x"}` in the database.

Since `ask_registry` gives a yes-or-no answer, any changes to the service area object need to happen outside of `ask_registry`.

One way of doing this is to do the `for nation in ['US', 'CA']` thing in `validate_geographic_areas`, not in `ask_registry`. That way you can test out various potential service area objects, and pick the one that ends up working (assuming any of them end up working).

api/admin/controller/library_settings.py
```diff
@@ -294,19 +297,42 @@ def validate_geographic_areas(self, values):
...
                 elif value.isdigit():
                     # Is it a US zipcode?
                     info = us_search.by_zipcode(value)
                     if not info:
                         return UNKNOWN_LOCATION.detailed(_('"%(value)s" is not a valid U.S. zipcode.', value=value))
                     locations["US"].append(self.format_place(value, info.major_city, info.state));
                 else:
-                    return UNKNOWN_LOCATION.detailed(_('Unable to locate "%(value)s".', value=value))
+                    if self.ask_registry(value):
+                        return self.ask_registry(value)
```
leonardr: `ask_registry` gives either a problem detail or `None`, and you'll want to do something other than return if the return value is `None`. I'm not sure exactly how you'll end up there, but if `ask_registry` says everything is okay, you'll eventually want to end up down at `return json.dumps(locations)`.

api/admin/controller/library_settings.py
```diff
@@ -297,44 +296,43 @@ def validate_geographic_areas(self, values):
...
-                unknown_place = json.loads(response.content).get("unknown")
-                # If the registry hasn't found the place, it will return a value for "unknown";
-                # otherwise, :unknown_place will be None.
-                responses.append(unknown_place)
-
-        if all(responses):
-            # If every registry has reported that the place is unknown for both the US and Canada, return an error message.
-            return UNKNOWN_LOCATION.detailed(_('Unable to locate "%(value)s".', value=value))
+            response = HTTP.debuggable_get(base_url + service_area_object)
+            unknown_place = json.loads(response.content).get("unknown")
```
leonardr: I think you need to call `urllib.quote` on `service_area_object`.

leonardr:
This now looks nice overall, I just have some minor concerns about the HTTP requests you're making.

api/admin/controller/library_settings.py
```diff
@@ -294,19 +297,50 @@ def validate_geographic_areas(self, values):
...
+            base_url = registry.integration.url + "/coverage?coverage="
+
+            response = HTTP.debuggable_get(base_url + service_area_object)
+            if not response.status_code == 200:
+                return REMOTE_INTEGRATION_FAILED.detailed(_("Unable to contact the registry at %(url)s.", url=registry.integration.url))
+
+            content = json.loads(response.content)
+            found_place = not (content.get("unknown") or content.get("ambiguous"))
+            if found_place:
+                return True
```
leonardr: Don't bail out immediately if one of the registries can't be contacted, but save the problem detail and at the end of the method, if there's a saved problem detail, return it rather than returning None.

tests/admin/controller/test_library.py
```diff
@@ -444,3 +491,86 @@ def test_library_delete(self):
...
+
+    def test_find_location_through_registry(self):
+        library = self._default_library
+        controller = self.manager.admin_library_settings_controller
+        original_ask_registry = controller.ask_registry
+        get = self.do_request
+        test = self
+        class Mock(LibrarySettingsController):
+            called_with = []
+            def mock_ask_registry(self, service_area_object):
```
leonardr: You should be able to override `ask_registry` rather than defining a method with a different name and assigning it to `ask_registry`.

PR #1220, opened 2019-04-08
Delete collections

EdwinGuzman:
Helps fix [SIMPLY-1846](https://jira.nypl.org/browse/SIMPLY-1846). This add the `marked_for_deletion` flag to the collection object that is sent to the UI. It also marks a collection for deletion instead of outright deleting it.


EdwinGuzman:
Front-end PR - https://github.com/NYPL-Simplified/circulation-web/pull/224

leonardr:
I was about to say "I'm surprised this didn't break a test" but then I saw it did break a test. It looks good once the test is fixed.

PR #1222, opened 2019-04-17
Refactor validation

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-1800


api/admin/controller/__init__.py
```diff
@@ -2591,7 +2591,7 @@ def _get_settings(self):
         [protocol] = [p for p in self.protocols if p.get("name") == flask.request.form.get("protocol")]
         return protocol.get("settings")
 
-    def validate_formats(self, settings=None):
+    def validate_formats(self, settings=None, validator=Validator()):
```
leonardr: This is a big improvement. My only remaining comment is really minor; this is going to create a useless `Validator` object in some cases. I usually do one of these two things:

* `validator_class=Validator` and then `validator = validator_class()` inside the method body.
* `validator=None` and then `validator = validator or Validator()` inside the method body.

Really doesn't matter much though.

PR #1226, opened 2019-04-25
Filter out invisible delivery mechanisms for distributors where mechanism is set at the point of first fulfillment.

leonardr:
This fixes https://jira.nypl.org/browse/SIMPLY-1889, which is some missing work I didn't do for https://jira.nypl.org/browse/SIMPLY-1680.

### DIFF
--- a/api/opds.py
+++ b/api/opds.py
@@ -239,14 +239,15 @@ class CirculationManagerAnnotator(Annotator):
         borrow_links = []
         if can_borrow:
             # Borrowing a book gives you an OPDS entry that gives you
-            # fulfillment links.
+            # fulfillment links for every visible delivery mechanism.
+            visible_mechanisms = self.visible_delivery_mechanisms(
+                active_license_pool
+            )
             if set_mechanism_at_borrow and active_license_pool:
                 # The ebook distributor requires that the delivery
                 # mechanism be set at the point of checkout. This means
                 # a separate borrow link for each mechanism.
-                for mechanism in self.visible_delivery_mechanisms(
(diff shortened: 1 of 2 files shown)


api/opds.py
```diff
@@ -258,13 +259,13 @@ def acquisition_links(
...
                 # checkout. This means a single borrow link with
-                # indirectAcquisition tags for every delivery
+                # indirectAcquisition tags for every visible delivery
                 # mechanism. If a delivery mechanism must be set, it
                 # will be set at the point of fulfillment.
                 borrow_links.append(
                     self.borrow_link(
                         active_license_pool,
-                        None, active_license_pool.delivery_mechanisms,
+                        None, visible_mechanisms,
```
leonardr: This is the bug fix.

tests/test_opds.py
```diff
@@ -1358,14 +1358,43 @@ def test_acquisition_links(self):
...
+        hidden = self._default_library.setting(
+            Configuration.HIDDEN_CONTENT_TYPES
+        )
+        available_types = [
+            lpdm.delivery_mechanism.content_type
+            for lpdm in loan2.license_pool.delivery_mechanisms
+        ]
+        hidden.value = json.dumps(available_types)
+
+        # The list of hidden content types is stored in the Annotator
```
EdwinGuzman: Is this list of hidden content per library and similar to the `hidden` value above?

leonardr: Yes, the constructor grabs the same ConfigurationSetting I modified above and uses its value to configure the Annotator.

PR #1229, opened 2019-05-02
Look up zip codes when library form loads

adriana-alter:
Good news: we've gone back to storing geographical input in the right format, without sacrificing the hover functionality in the interface.

Bad news: apparently that wasn't the only problem.  Trying to update the library's registration is still throwing an error.  I will continue to investigate (possibly in a different branch). 
Screen Shot 2019-05-02 at 4 35 43 PM

### DIFF
--- a/api/admin/geographic_validator.py
+++ b/api/admin/geographic_validator.py
@@ -22,7 +22,21 @@ class GeographicValidator(Validator):
 
         us_search = uszipcode.SearchEngine(simple_zipcode=True)
         ca_search = PostalCodeDatabase()
-        CA_PROVINCES = ["AB", "BC", "MB", "NB", "NL", "NT", "NS", "NU", "ON", "PE", "QC", "SK", "YT"]
+        CA_PROVINCES = {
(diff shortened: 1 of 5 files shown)


api/admin/geographic_validator.py
```diff
@@ -73,12 +74,28 @@ def validate_geographic_areas(self, values, db):
...
+        if country == "US":
+            info = uszipcode.SearchEngine(simple_zipcode=True).by_zipcode(zip)
+            if formatted:
+                info = self.format_place(zip, info.major_city, info.state)
+        elif country == "CA":
+            info = PostalCodeDatabase()[zip]
+            if formatted:
+                info = self.format_place(zip, info.city, info.province)
+        return info
+
```
leonardr: These new methods need simple standalone tests.

adriana-alter:
The rationale for adding the province names (and changing the tests accordingly): in the previous implementation, the circ manager had to ask the registry about province names, which takes a long time and doesn't seem worth it (given that entering a province name seems like a really common use case).

PR #1232, opened 2019-05-15
Reformat old geographic information

adriana-alter:


migration/20190515-reformat-existing-geographic-data.py
```diff
@@ -0,0 +1,31 @@
...
+from core.model.configuration import ConfigurationSetting
+
+_db = production_session()
+
+area_settings = _db.query(ConfigurationSetting).filter(
+    or_(ConfigurationSetting.key == "service_area", ConfigurationSetting.key == "focus_area")).filter(
+    or_(ConfigurationSetting._value != None, ConfigurationSetting._value == "")
+    ).all()
+
+settings_with_wrong_format = [s for s in area_settings if isinstance(json.loads(s._value), (list,))]
```
leonardr: I think this will work if the bad value is a JSON list like `["Waterford, CT"]`, but the value I saw in a real circ manager was `Waterford, CT`. Trying to load that string as JSON will cause a `ValueError`.

So maybe try to load the setting as JSON, and treat it as needing repair if it raises a `ValueError` or the result isn't a dictionary.

leonardr:
We've gone back and forth on this over Slack, and I've given inconsistent advice because it's a confusing situation and we can't really add unit tests to a migration script -- my usual technique for proving that behavior is correct.

TBH I'm still confused and that's not a good state to approve a PR in, so I'd like to suggest putting some simple tests _inside_ this script. They won't be run as part of our continuous integration, but you can run them during development and they'll stop the migration script from running in a real situation if something's wrong.

So, define a function called `fix` which takes the value of `setting._value` and turns it into the new value `json.dumps({"US": format(setting._value)})`. Then you can add these tests just before iterating over `area_settings`:

```
expect = json.dumps({"US": ["Waterford, CT"]})
assert fix("Waterford, CT") == expect
assert fix(json.dumps("Waterford, CT") == expect)
assert fix(json.dumps(["Waterford, CT"]) == expect)
assert fix(expect) == expect)
```

I think this should cover ever case, because `ConfigurationSetting._value` is always going to be a string -- either a regular string or a JSON-encoded string. If it's a JSON-encoded string, it's either going to be one place name, a list of place names, or a dictionary that's already been fixed and doesn't need to be "fixed" again.

PR #1233, opened 2019-05-16
Don't show a refresh metadata button if no metadata wrangler is configured.

aslagle:
This removes the "Refresh Metadata" button in the admin interface if the metadata wrangler is not configured, since it would just give an error anyway. That button is a bit misleading since librarians using it may not be aware of the metadata wrangler - it's not clear that it uses an external source. But that's a separate issue. I also added some test coverage.

### DIFF
--- a/api/admin/opds.py
+++ b/api/admin/opds.py
@@ -14,6 +14,8 @@ from core.model import (
 from core.opds import AcquisitionFeed
 from core.util.opds_writer import AtomFeed
 from core.mirror import MirrorUploader
+from api.coverage import MetadataWranglerCollectionRegistrar
+from api.config import CannotLoadConfiguration
 
 class AdminAnnotator(LibraryAnnotator):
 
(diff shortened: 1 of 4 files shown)


api/admin/opds.py
```diff
@@ -31,14 +33,20 @@ def annotate_work_entry(self, work, active_license_pool, edition, identifier, fe
...
-        feed.add_link_to_entry(
-            entry,
-            rel="http://librarysimplified.org/terms/rel/refresh",
-            href=self.url_for(
-                "refresh",
-                identifier_type=identifier.type,
-                identifier=identifier.identifier, _external=True)
-        )
+        try:
+            MetadataWranglerCollectionRegistrar(work.license_pools[0].collection)
```
EdwinGuzman: Do you not have to assign this to any variable?

aslagle: I don't actually need it - I only need to make sure it doesn't throw an exception. There's code in the controller that actually uses this, and I don't want to show the button if that code would just fail.

EdwinGuzman: Makes sense but looks weird to just have this floating around like this.

aslagle: Yeah it does. Do you have a better idea?

EdwinGuzman: No :(

EdwinGuzman:
Where exactly does the button show up in the Admin?

aslagle:
It's on the book edit tab - next to the hide/show button.

PR #1234, opened 2019-05-17
Kansas patron authentication service

jonathangreen:
## JIRA Ticket
https://jira.nypl.org/browse/SIMPLY-1960

## What does this Pull Request do?

The Kansas State library uses a custom authentication method authentication with their Bibliotecha (Cloud Library) and Enki ebook collections and possibly its Freading and RB Digital ebook/audiobook services. This custom authentication method is described below. This PR integrates this custom authentication method so that Kansas state can access their ebook collections through the Simply-E app.

## API Detail

The Authentication is done using a REST interface with the endpoint: 
https://ks-kansaslibrary3m.civicplus.com/api/UserDetails

The client wishing to authenticate sends a POST request with the header `Content-Type: text/xml`. The body of the post request is the following XML. 

```xml
<?xml version="1.0"?>
<AuthorizeRequest>
  <UserID> {{ barcode }} </UserID>
  <Password> {{ pin }} </Password>
</AuthorizeRequest>
```

The response body of the message will contain the following XML:

```xml
<?xml version="1.0"?>
<AuthorizeResponse>
  <Status>1</Status>
  <LibraryID>-2</LibraryID>
  <FirstName>Montgomery</FirstName>
  <LastName>Burns</LastName>
</AuthorizeResponse>
```

In the response the `Status` will be `1` if the patron is authenticated. Any other value for `Status` will be rejected as an invalid patron. `FirstName` contains the first name of the patron, and `LastName` contains the last name of the patron. `LibraryID` contains an identifier for the library that the user is associated with.

## How should this be tested?

A number of test cases have been added to the code to exercise the new authentication service.

## Interested parties
@leonardr @aslagle


leonardr:
This looks like a good implementation of the API. Let's talk about the 'library identifier' -- That's the major piece that _might_ need work. But I think we're okay.

Presumably the 'library identifier' tells the circulation manager whether this person is a patron of the Topeka library or the Lawrence library. Is that distinction important? Does Topeka have access to different resources than Lawrence or are they all sharing a collection? Are there even going to be multiple Kansas libraries in SimplyE or is it just going to be one big library covering the whole state?

If all these libraries share a collection then I think there's nothing more to do. If they have different collections then I think there's no more _code_ that needs to be written, but in the admin interface you'll need to set up a "library identifier restriction" for each library. That way someone from Topeka can't authenticate with the Lawrence library.

Sorry if you already knew this -- I had to load it back in while going over this code.

api/kansas_patron.py
```diff
@@ -0,0 +1,122 @@
...
+        except requests.exceptions.ConnectionError, e:
+            raise RemoteInitiatedServerError(
+                str(e.message),
+                self.NAME
+            )
+        if response.status_code != 200:
+            msg = "Got unexpected response code %d. Content: %s" % (
+                response.status_code, response.content
+            )
+            raise RemoteInitiatedServerError(msg, self.NAME)
```
aslagle: This is fine, but we also have a module for making requests that could handle some of this: https://github.com/NYPL-Simplified/server_core/blob/master/util/http.py

jce1028:
The libraries within Kansas share a statewide collection.    For now, this is a State Library implementation.  The virtual card is issued from state for from the specific library, hence the identifier.  This lets the State know how much others use the collection.  To our understanding there is no Overdrive Advantage like set up within these vendor platforms managed by the state that allows the library to have local only items.  I think the future does hold that individual libraries Like Topeka will eventually have a SimplyE Instance where this state collection would be merged.  This virtual library card should continue to work because to our understanding its also established within the local libraries ILS (if they have one)

jonathangreen:
Thanks for the feedback @leonardr. As @jce1028 mentioned, I don't think we need to any filtering on `library_identifier` currently, there is just going to be one large state-wide collection. I just thought I'd add it to the patron data so its there if we need it in the future.

api/kansas_patron.py
jonathangreen: Thanks @aslagle! I didn't realize this existed. I'll see if I can use that instead.

jonathangreen:
Thanks for the suggestion @aslagle! I updated the PR to use `HTTP.post_with_timeout` instead.

aslagle:
:+1:

jonathangreen:
@leonardr @aslagle I've completed my real world testing, so this is good to go from my perspective, unless you guys have more feedback. 

It looks like travis started once changed the status so hopefully that will be green soon.

PR #1235, opened 2019-05-20
Added option to configure a custom CSS link for the web catalog.

aslagle:


leonardr:
My gut reaction is that the web application should provide this link. How does the circ manager know what HTML tags and classes a web application will use to visualize an OPDS feed?

But if the goal is to have one web application that can talk to multiple libraries then I understand why you want a CSS file that works against the _combination_ of a web application and a circulation manager. And the link to that file would have to be in the circulation manager.

Is there anything we can do to distinguish between this stylesheet and other stylesheets that an Authentication For OPDS document might serve? In particular, does the circ manager know the hostname of the web application it wants to configure?

aslagle:
Yeah, the circulation manager does not know about the web application's HTML, but the person who writes the CSS file does. Your second paragraph is correct, and I think it would be very unlikely for a library to use multiple web applications with different CSS files.

The circ manager does know the hostname since that also must be configured, but I'm not sure how that would make a difference here. The CSS file does not need to be on the same host. I could put it in a different place instead of links or use a different rel if you think that would be better.

leonardr:
Since you know the hostname, I was thinking you could do a link that looked like this:

`
{
 "rel": "stylesheet",
 "href": "https://stylesheet/",
 "type": "text/css",
 "use-with": "https://web-application/"
}
`

Basically, make up an extension that gives the link a _second_ target. Now the link joins three resources instead of two.

What do you think?

aslagle:
That would mean I'd also have to change the configuration to be for a web catalog and a library instead of just a library, since the circ manager can have multiple.

leonardr:
Yeah, that's annoying. I think what you have is fine for the forseeable future.

PR #1243, opened 2019-05-31
Feature branch: Use ElasticSearch 6 to generate nearly all OPDS feeds

leonardr:
This is the start of a feature branch that will eventually address all the subtasks of https://jira.nypl.org/browse/SIMPLY-1904.

The first one is https://jira.nypl.org/browse/SIMPLY-1982 -- changing the special "books in a series" lane to get its list via the search index rather than the materialized view.

There are two parts to this initla branch -- the `SeriesLane` work, and the work I had to do to merge in the new `core`. Since all "normal" Lanes now get their results from a search index rather than the materialized view, so a number of tests stopped working -- they added books to the materialized view, built an OPDS feed, and checked for the presence or absence of books in the feed.


api/controller.py
```diff
@@ -667,19 +667,14 @@ def public_key_document(self):
 
 class OPDSFeedController(CirculationManagerController):
 
-    def groups(self, lane_identifier):
+    def groups(self, lane_identifier, feed_class=AcquisitionFeed):
```
leonardr: The methods in this class were streamlined a little bit and given the ability to mock the feed class whose methods we call.

We had really complicated end-to-end tests that stopped working when I merged in the core branch -- all these `AcquisitionFeed` methods used to end up going against the materialized view, and now they go against the search index. So without creating a real search index, we can't do an end-to-end test that verifies that e.g. the French book doesn't show up in the "Adult English Fiction" lane.

Fortunately, all that stuff is tested lower down, ultimately in `core/tests/test_external_search.py`, where we test that a `Filter` constructed a certain way will match certain books; and in `core/tests/test_lane.py`, where we test that a certain faceting object will modify a `Filter` in a certain way. So I changed the tests of these methods to mock the feed class and verify that the appropriate faceting objects (among other things) are passed in.

Overall I think this is a great change, but it does mean we don't have a full end-to-end test where we get an OPDS feed of all the books in a lane.

api/controller.py
```diff
@@ -828,14 +830,40 @@ def _load_search_facets(self, lane):
...
+            return lane
+
+        # TODO: Is this still necessary? Most other feeds now come
+        # from the search index, and those methods let lower-down code
+        # find or create the ExternalSearchIndex object.
+        #
+        # The argument against is that it's inefficient to look up
+        # the ExternalSearchIndex object every time. I think this is only
+        # valid if creating the object requires a back-and-forth with the
+        # search server.
```
aslagle: Yeah, I think this was because the initial setup was slow. It is still necessary to handle the case where the search engine is misconfigured, but maybe not like this.

leonardr:
OK, I've got it so that every method that (currently) might need to use the search index uses the one associated with the CirculationManager. I was worried I'd have to change a bunch of method signatures, but I'd already changed them.

aslagle:
This looks good. I don't know if I should approve since it's not actually ready to merge.

leonardr:
I'll leave it as is, but it's good to know you're okay with this, so I can use it as the target for future PRs.

leonardr:
I'll be merging this into the `3.0beta` branch, not directly into `master`.

PR #1245, opened 2019-06-07
Deleting a custom lane whenever all its lists are deleted.

EdwinGuzman:
Server-side fix for [SIMPLY-579](https://jira.nypl.org/browse/SIMPLY-579). Deleting a lane once all of its custom lists have been deleted.

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -1740,11 +1740,16 @@ class CustomListsController(AdminCirculationManagerController):
             # Build the list of affected lanes before modifying the
             # CustomList.
             affected_lanes = Lane.affected_by_customlist(list)
+            for lane in affected_lanes:
+                lane.update_size(self._db)
+                # There's only one custom list in the lane and it's going
+                # to be deleted.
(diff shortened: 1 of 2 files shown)


api/admin/controller/__init__.py
```diff
@@ -1740,11 +1740,16 @@ def custom_list(self, list_id):
             # Build the list of affected lanes before modifying the
             # CustomList.
             affected_lanes = Lane.affected_by_customlist(list)
+            for lane in affected_lanes:
```
EdwinGuzman: I moved this above the other delete functions because it didn't seem to ever reach this otherwise. I'm not sure if it's the right move.

EdwinGuzman:
Also, this was discussed over slack but it's good to point out that this will delete any future custom lanes created with custom lists. Any existing lanes, however, will still not be editable.

api/admin/controller/__init__.py
adriana-alter: Would there be an easier way to do this via setting a cascade delete constraint in the model?

aslagle: I don't think that's possible for this type of relationship.

PR #1248, opened 2019-06-10
Book cover

adriana-alter:
I moved `WorkController` out of `controller/__init__` and into its own file (in keeping with the pattern we implemented for the rest of the controller module, where each controller has its own file), and refactored the methods involving previewing and changing cover images.  (Much of this is just to make it easier for me to track down the Book Edit Form bugs that I'm looking into.)


package-lock.json
```diff
@@ -0,0 +1,7682 @@
+{
```
leonardr: This whole file is new? Is this right, or is this the sort of file that's generated on the local computer and should be in .gitignore?

adriana-alter: It is generated on the local computer (by running `npm install`), but it **is** supposed to be committed.  (In and of itself, it has nothing to do with this PR; it got generated because I was running `npm install` to bring in the newest version of circ-web.)

PR #1249, opened 2019-06-13
Add link for device management endpoint to profile doc

adriana-alter:


api/authenticator.py
```diff
@@ -425,12 +426,14 @@ def set_authorization_identifier(self, authorization_identifier):
         self.authorization_identifier = authorization_identifier
         self.authorization_identifiers = authorization_identifiers
 
-class CirculationPatronProfileStorage(PatronProfileStorage):
+class CirculationPatronProfileStorage(PatronProfileStorage, LibraryAnnotator):
```
leonardr: This works, but I think it's more complex than necessary, and I'm a little worried about it because `LibraryAnnotator` was only designed to be used to make OPDS feeds. What happens if you use the normal `url_for` from the `flask` package instead of `self.url_for`?

adriana-alter: I had tried the normal `url_for` originally, but it throws an error message--something about taking the wrong number of arguments, I think?  I will take another look, though!

leonardr:
The Travis problem is fixed but you have some test failures related to your change in the constructor.

adriana-alter:
The tests pass in local.  I think the problem is just that one of the changes is to `core`, and that's not getting picked up on here.  I've made a separate `core` PR: https://github.com/NYPL-Simplified/server_core/pull/1072

PR #1255, opened 2019-06-18
Get crawlable feeds from the search index, not the materialized view

leonardr:
This branch converts all of the crawlable feeds--library, colleciton, and custom list--to get search results from the search index, not the materialized view. The search index is about 2-4 times slower (because the list of results is sorted by a script field), but crawlable feeds aren't used by human beings, and it'll be worth it to eventually get rid of the materialized view.

It's pretty much the same as when I converted over `SeriesLane` to use a search index. `WorksFromDatabase` subclasses become `DynamicLane` subclasses.  Controller tests that previously made real OPDS feeds, now just verify that and that appropriate `WorkList` and `Facet` objects are created and passed into `AcquisitionFeed.feed`.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -783,7 +783,8 @@ class OPDSFeedController(CirculationManagerController):
             library_short_name=library_short_name,
         )
         title = library.name
-        lane = CrawlableCollectionBasedLane(library)
+        lane = CrawlableCollectionBasedLane()
+        lane.initialize(library)
         return self._crawlable_feed(library, title, url, lane)
 
     def crawlable_collection_feed(self, collection_name):
@@ -798,7 +799,8 @@ class OPDSFeedController(CirculationManagerController):
             "crawlable_collection_feed",
             collection_name=collection.name
         )
-        lane = CrawlableCollectionBasedLane(None, [collection])
+        lane = CrawlableCollectionBasedLane()
+        lane.initialize([collection])
         if collection.protocol in [ODLAPI.NAME]:
             annotator = SharedCollectionAnnotator(collection, lane)
         else:
@@ -826,6 +828,7 @@ class OPDSFeedController(CirculationManagerController):
     def _crawlable_feed(self, library, title, url, lane, annotator=None):
         annotator = annotator or self.manager.annotator(lane)
         facets = CrawlableFacets.default(library)
+        facets.order_ascending = CrawlableFacets.ORDER_DESCENDING
         pagination = load_pagination_from_request()
         if isinstance(pagination, ProblemDetail):
             return pagination
@@ -833,6 +836,7 @@ class OPDSFeedController(CirculationManagerController):
             self._db, title, url, lane, annotator=annotator,
             facets=facets,
             pagination=pagination,
+            cache_type="crawlable"
         )
         return feed_response(feed)
 
--- a/api/controller.py
+++ b/api/controller.py
@@ -821,6 +821,9 @@ class OPDSFeedController(CirculationManagerController):
         """Build or retrieve a crawlable, paginated acquisition feed for the
         named CustomList, sorted by update date.
         """
+        # TODO: A library is not strictly required here, since some
+        # CustomLists aren't associated with a library, but this isn't
+        # a use case we need to support now.
         library = flask.request.library
         list = CustomList.find(self._db, list_name, library=library)
         if not list:
--- a/api/controller.py
+++ b/api/controller.py
@@ -829,7 +829,9 @@ class OPDSFeedController(CirculationManagerController):
         annotator = annotator or self.manager.annotator(lane)
         facets = CrawlableFacets.default(library)
         facets.order_ascending = CrawlableFacets.ORDER_DESCENDING
-        pagination = load_pagination_from_request()
+        pagination = load_pagination_from_request(
+            default_size=Pagination.DEFAULT_CRAWLABLE_SIZE
+        )
         if isinstance(pagination, ProblemDetail):
             return pagination
         feed = AcquisitionFeed.page(
--- a/api/controller.py
+++ b/api/controller.py
@@ -421,28 +421,40 @@ class CirculationManager(object):
         return authdata
 
     def annotator(self, lane, facets=None, *args, **kwargs):
-        """Create an appropriate OPDS annotator for the given lane."""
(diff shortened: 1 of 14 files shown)


api/opds.py
```diff
@@ -592,9 +596,23 @@ def lane_url(self, lane, facets=None):
...
+                search_update = datetime.datetime.utcfromtimestamp(
+                    last_updates[0]
+                )
+                # There's a chance that work.last_updated has been
+                # modified but the change hasn't made it to the search
+                # engine yet.
+                #
+                # TODO: Is it better to use the later value, which is more
+                # accurate, or to ensure that the feed stays sorted?
+                updated = max(updated, search_update)
```
leonardr: I would like your take on this.

api/controller.py
```diff
@@ -821,18 +834,40 @@ def crawlable_list_feed(self, list_name):
...
+
+        annotator = annotator or self.manager.annotator(lane)
+
+        # A crawlable feed has only one possible set of Facets,
+        # so library settings are irrelevant.
+        facets = CrawlableFacets.default(None)
+
+        feed = feed_class.page(
+            _db=self._db, title=title, url=url, lane=lane, annotator=annotator,
+            facets=facets, pagination=pagination, cache_type="crawlable",
```
leonardr: These used to be 'page' feeds, and it's not technically necessary for crawlable feeds to have a distinctive type, because they have a distinctive sort order, but it was really useful to be able to just delete all the crawlable feeds from the database while testing against a real site.

api/controller.py
```diff
@@ -429,6 +429,11 @@ def annotator(self, lane, facets=None, *args, **kwargs):
         else:
             library = flask.request.library
 
+        # If all else fails, use an OPDS annotator for the system
+        # default library. TODO: This is less than ideal, but it only
+        # actually happens within a unit test of a helper method.
```
aslagle: Which test is that? It does seem like there must be a nicer way to do this.

leonardr: It's test__crawlable_feed: https://github.com/NYPL-Simplified/circulation/pull/1255/files#diff-a4d1d2853355ff76196c2e33fccbe761R3327

leonardr: I was a little surprised that we don't end up doing this for the 'feed based on a collection' case. I didn't look into it in huge detail, but I think we do use the system default library for some purposes; it's just that the time the code gets into this method the decision has already been made.

I would like this whole system to work even in a situation where there is no library involved in the request, but there are a lot of places where we assume OPDS feed == library context.

leonardr: Of course I can get rid of this by changing the test, but I like being able to handle this case, even if it's handled in a way I think is awkward.

aslagle: The "feed based on a collection" case uses a different annotator that doesn't require a library - maybe that code could be moved in here instead. It does need to know if it's a shared collection, and the collection feed for a non-shared collection was never actually used as far as I know and may not even work.

api/controller.py
```diff
@@ -798,17 +805,23 @@ def crawlable_collection_feed(self, collection_name):
             "crawlable_collection_feed",
             collection_name=collection.name
         )
-        lane = CrawlableCollectionBasedLane(None, [collection])
+        lane = CrawlableCollectionBasedLane()
+        lane.initialize([collection])
         if collection.protocol in [ODLAPI.NAME]:
             annotator = SharedCollectionAnnotator(collection, lane)
         else:
             annotator = CirculationManagerAnnotator(lane)
```
aslagle: This is the code to create annotators without a library.

api/controller.py
aslagle: This seems fine but probably needs some changes to CachedFeed in the model.

api/opds.py
aslagle: If you use the search engine value, someone who imports the feed will update this work, and the next time they pull the feed they'll have to update it again even if there are no changes.

If you use the later value, someone who imports the feed may record that timestamp as their last update time, but there may also be new works that have not made it to the search index yet and have timestamps earlier than this one. Next time they pull the feed, they may not reach those works since they'll stop at this later one they've already seen.

The second case seems worse so I would go with the earlier value from the search engine.

tests/test_controller.py
```diff
@@ -552,6 +560,10 @@ def test_sitewide_key_pair(self):
         # Calling it again will do nothing.
         eq_((new_public, new_private), self.manager.sitewide_key_pair)
 
+    def test_annotator(self):
+        # TODO.
+        pass
```
aslagle: Did you mean to leave this?

api/controller.py
leonardr: I took a look at core and here are the changes we might want to make:

* In core, we might want to add a constant for 'crawlable' so other people know not to use that string.
* The Crawlable*Lane classes ought to define MAX_CACHE_AGE. Previously they were using the site's default cache time for nongrouped page feeds, and now the default cache time is zero. What's a good cache time for crawlable feeds?

tests/test_controller.py
leonardr: Yeah, I noticed there was no coverage for this method but originally intended to leave it out. However since I'm changing this method (to move the CirculationManagerAnnotator code into it) I will add the coverage.

api/controller.py
leonardr: FWIW it seems to work in this branch; I've been testing with a crawlable feed for NYPL's Overdrive collection.

api/controller.py
aslagle: For the list feeds, I think the import script only checks once a day so the cache time could probably be a few hours. For a shared collection feed it checks every 6 hours and should probably be shorter.

api/lanes.py
```diff
@@ -1139,7 +1141,13 @@ def default_facet(cls, config, facet_group_name):
         return cls.SETTINGS[facet_group_name]
 
 
-class CrawlableCollectionBasedLane(DynamicLane):
+class CrawlableLane(DynamicLane):
+
+    # Crawlable feeds are cached for 12 hours.
+    MAX_CACHE_AGE = 12 * 60 * 60
```
aslagle: This should be fine in most cases, but it's too long for a shared ODL collection feed so maybe CrawlableCollectionBasedLane needs to override this.

PR #1258, opened 2019-06-21
Remove materialized view from contributor lane generation

leonardr:
Once I got the Elasticsearch part of this working and moved the "do your best to find this person" code [into core](https://github.com/NYPL-Simplified/server_core/pull/1076), this was basically the same as my SeriesLane code. There's a ContributorFacets which changes the search filter (this is clumsy and I'd like to refactor it if possible), a ContributorLane which manages the Lane configuration, and a controller method that ties everything together.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -1552,8 +1552,8 @@ class WorkController(CirculationManagerController):
         if not contributor_name:
             return NO_SUCH_LANE.detailed(_("No contributor provided"))
 
-        # contributor_name is probably a sort_name, but it could be a
-        # display_name. Pass it in for both fields and
+        # contributor_name is probably a display_name, but it could be a
+        # sort_name. Pass it in for both fields and
         # ContributorData.lookup() will do its best to figure it out.
         contributor = ContributorData.lookup(
             self._db, sort_name=contributor_name, display_name=contributor_name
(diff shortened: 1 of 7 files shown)


api/controller.py
```diff
@@ -1591,35 +1593,58 @@ def _lane_details(self, languages, audiences):
...
-        annotator = self.manager.annotator(lane)
-        facets = load_facets_from_request(worklist=lane)
         if isinstance(facets, ProblemDetail):
             return facets
+
         pagination = load_pagination_from_request()
         if isinstance(pagination, ProblemDetail):
             return pagination
+
+        annotator = self.manager.annotator(lane, facets)
```
leonardr: Previously `facets` was not being passed in, which is definitely a problem, but I don't know what effects it would have.

api/lanes.py
```diff
@@ -1053,71 +1046,73 @@ def url_arguments(self):
...
         return self.ROUTE, kwargs
 
-    @property
-    def contributor_name_clauses(self):
-        return [
-            Contributor.display_name==self.contributor_name,
-            Contributor.sort_name==self.contributor_name
-        ]
-
-    def apply_filters(self, _db, qu, facets, pagination, featured=False):
```
leonardr: Some of this code went into `ContributorData.lookup` and the rest went into `Filter.author_filter`.

PR #1260, opened 2019-06-27
Avoid materialized view in recommendation lane

leonardr:
This branch completes https://jira.nypl.org/browse/SIMPLY-1981. It converts RecommendationLane (which gets works from NoveList) into a `DatabaseExclusiveWorkList` -- a `DatabaseBackedWorkList` that (unlike `Lane`) uses the database as its `works()` implementation.

This branch takes advantage of recent changes in core. Two custom faceting classes are now simple subclasses of `DefaultSortOrderFacets`. Elasticsearch filters are modified in lane-specific hook methods rather than in the filter classes.

Custom lanes define `overview_facets` to convert a `FeaturedFacets` object into an appropriate faceting object. This ensures that the swim lanes generated by the `RelatedBooksLane` are sorted appropriately for the sublane.


api/lanes.py
```diff
@@ -907,115 +917,100 @@ def _contributor_sublanes(self, _db):
...
         """Get identifiers of recommendations for this LicensePool"""
-
-        metadata = self.api.lookup(self.edition.primary_identifier)
+        metadata = self.novelist_api.lookup(self.edition.primary_identifier)
         if metadata:
             metadata.filter_recommendations(_db)
             return metadata.recommendations
         return []
 
-    def apply_filters(self, _db, qu, facets, pagination, featured=False):
```
leonardr: Unlike with most other lanes, this code wasn't removed -- it went into `modify_database_query_hook` below.

tests/test_controller.py
```diff
@@ -2423,257 +2429,282 @@ def test_recommendations(self):
...
             )
-
-        eq_(200, response.status_code)
-        feed = feedparser.parse(response.data)
-        eq_(1, len(feed['entries']))
-        [entry] = feed['entries']
-        eq_(self.work.title, entry['title'])
+        eq_(kwargs.pop('url'), expect_url)
 
     def test_related_books(self):
```
leonardr: Now that all of the sublanes used by RelatedBooksLane have been converted over, I converted over its controller test as well.

api/admin/controller/__init__.py
```diff
@@ -1740,16 +1740,24 @@ def custom_list(self, list_id):
             # Build the list of affected lanes before modifying the
             # CustomList.
             affected_lanes = Lane.affected_by_customlist(list)
+            surviving_lanes = []
             for lane in affected_lanes:
-                lane.update_size(self._db)
-                # There's only one custom list in the lane and it's going
-                # to be deleted.
-                if lane.size == 0 and len(lane._customlist_ids) == 1:
```
leonardr: There's a bug in this line exposed by my core changes. Even for a Lane based solely on one custom list, the lane won't be empty until after the list is deleted. The test was working because `update_size` used to use the materialized view, which wasn't being updated. Now it uses the database, which gives the right answer, so test lists weren't being deleted when they should have been.

api/controller.py
```diff
@@ -1643,7 +1642,7 @@ def contributor(
         feed = feed_class.page(
             _db=self._db, title=lane.display_name, url=url, lane=lane,
             facets=facets, pagination=pagination,
-            annotator=annotator, cache_type=CachedFeed.CONTRIBUTOR_TYPE,
+            annotator=annotator, cache_type=lane.CACHED_FEED_TYPE,
```
aslagle: If you're no longer going to use CONTRIBUTOR_TYPE, should it be removed from core?

aslagle: Oh sorry, that's in the Lane classes now.

api/lanes.py
```diff
@@ -854,8 +854,11 @@ class RelatedBooksLane(WorkBasedLane):
     * RecommendationLane: Works provided by a third-party recommendation
       service.
     """
+    CACHED_FEED_TYPE = CachedFeed.RELATED_TYPE
     DISPLAY_NAME = "Related Books"
     ROUTE = 'related_books'
+    MAX_CACHE_AGE = 48*60*60 # 48 hours, the shortest time of any component
+                             # sublane.
```
aslagle: Why is that? It doesn't seem like these would change very much.

leonardr: That's just what it was before. I bumped up the component cache times to 96 hours and turned this into an expression rather than a hard-coded value. (This required that I move RelatedBooksLane to lower in lanes.py, after the classes for its sublanes.)

leonardr: Working on my bugfix branch I realized it's more complicated than this. The "recommendations" part of this feed is set up to show books that are currently available. That information is going to go stale a lot faster than the rest of the feed.

If we had a way of caching the _recommendation data_ from NoveList, we could cache _that_ for a really long time, and cache the _feeds_ (which take availability into account) for a relatively short time.

PR #1261, opened 2019-06-28
Bug fixes and minor improvements for ElasticSearch

leonardr:
I wrote a script that requested various special (non-Lane) OPDS feeds from a circ manager, parsed out the pagination and faceting links, and displayed that information along with the works in the feed and the information used to sort the work lists.

I ran this script against NYPL's QA circulation manager using the current (materialized view) code and using the new (ElasticSearch) code. Then I manually compared the outputs. While doing this I found some bugs and inconsistencies, most of which are resolved in this branch. (The main bug not yet resolved is https://jira.nypl.org/browse/SIMPLY-2087.)

Here are the inconsistencies that remain -- these are all situations where either the new code is better than the old code, or there's no way to get the functionality of the old code.

* All crawlable feeds are properly sorted by update time. (But because of this, generating those feeds is much slower.)

* Pages of crawlable feeds are twice as large as they used to be. (I bumped up the size to compensate for the slowness.)

* The portion of a contributor feed that shows up in the "related books" feed is now sorted by title. I can't tell how those books were sorted previously.

* The various "books like this one" feeds (contributor, series, recommendations, and "related books") now include books from the full collection and books that have a holds queue. I don't think it worked this way before. It's debatable whether we want to show books with a holds queue, but I think this makes sense -- especially given how long we cache these feeds -- and it's now very clear how to change this if we want.

* Recommendation feeds are no longer sortable by 'recently added', since that's only available for feeds that go through ElasticSearch.

* Series feeds now propagate entry point, like author and recommendation feeds; previously they didn't.

* On a series feed, 'series position' is the first facet in the facet group. Previously it was selected but not the first one in the group.

* Previously generic series feeds had an extra slash ("{controller}//{faceting info}") in their pagination/faceting URLs. That extra slash is still there for generic contributor feeds, and I can't figure out what the difference is, but it's not hurting anybody. By "generic feeds", I mean feeds for a series or contributor that don't also have a language or audience restriction. That makes me think it has something to do with the routing.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -803,9 +803,7 @@ class OPDSFeedController(CirculationManagerController):
         title = library.name
         lane = CrawlableCollectionBasedLane()
         lane.initialize(library)
-        return self._crawlable_feed(
-            library=library, title=title, url=url, lane=lane
-        )
+        return self._crawlable_feed(title=title, url=url, lane=lane)
 
     def crawlable_collection_feed(self, collection_name):
         """Build or retrieve a crawlable acquisition feed for the
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -844,96 +844,6 @@ class WorkBasedLane(DynamicLane):
         worklist.audiences = self.audiences
 
 
-class RelatedBooksLane(WorkBasedLane):
-    """A lane of Works all related to a given Work by various criteria.
-
-    Each criterion is represented by another WorkBaseLane class:
-
-    * ContributorLane: Works by one of the contributors to this work.
-    * SeriesLane: Works in the same series.
-    * RecommendationLane: Works provided by a third-party recommendation
-      service.
-    """
-    CACHED_FEED_TYPE = CachedFeed.RELATED_TYPE
-    DISPLAY_NAME = "Related Books"
-    ROUTE = 'related_books'
-    MAX_CACHE_AGE = 48*60*60 # 48 hours, the shortest time of any component
-                             # sublane.
-
-    def __init__(self, library, work, display_name=None,
-                 novelist_api=None):
-        super(RelatedBooksLane, self).__init__(
-            library, work, display_name=display_name,
-        )
-        _db = Session.object_session(library)
-        sublanes = self._get_sublanes(_db, novelist_api)
-        if not sublanes:
-            raise ValueError(
-                "No related books for %s by %s" % (self.work.title, self.work.author)
-            )
-        self.children = sublanes
-
-    def works(self, _db, *args, **kwargs):
-        """This lane never has works of its own.
-
-        Only its sublanes have works.
-        """
-        return []
-
-    def _get_sublanes(self, _db, novelist_api):
-        sublanes = list()
-
-        for contributor_lane in self._contributor_sublanes(_db):
-            sublanes.append(contributor_lane)
-
-        for recommendation_lane in self._recommendation_sublane(_db, novelist_api):
-            sublanes.append(recommendation_lane)
-
-        # Create a series sublane.
-        series_name = self.edition.series
-        if series_name:
-            sublanes.append(SeriesLane(self.get_library(_db), series_name, parent=self, languages=self.languages))
-
-        return sublanes
-
-    def _contributor_sublanes(self, _db):
-        """Create contributor sublanes"""
-        viable_contributors = list()
-        roles_by_priority = list(Contributor.author_contributor_tiers())[1:]
-
-        while roles_by_priority and not viable_contributors:
-            author_roles = roles_by_priority.pop(0)
-            viable_contributors = [c.contributor
-                                   for c in self.edition.contributions
-                                   if c.role in author_roles]
-
-        library = self.get_library(_db)
-        for contributor in viable_contributors:
-            contributor_lane = ContributorLane(
-                library, contributor, parent=self,
-                languages=self.languages, audiences=self.audiences,
-            )
-            yield contributor_lane
-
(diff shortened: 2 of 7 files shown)


leonardr:
For the record, here's my script:

```
import time
from bs4 import BeautifulSoup
from pdb import set_trace
from collections import defaultdict
import requests
import feedparser

class Client(object):

    def __init__(self, server, url, order):
        self.server = server
        self.url = server + url
        self.short_url = url
        self.order = order
        self.start = time.time()
        self.response = requests.get(self.url)
        self.end = time.time()
        self.parsed = feedparser.parse(self.response.content)
        self.soup = BeautifulSoup(self.response.content, 'xml')

    def links_by_rel(self, links):
        by_rel = defaultdict(list)
        for link in links:
            by_rel[link['rel']].append(link)
        return by_rel

    def pagination(self, links):
        return links.get('next', [])

    def facet_groups(self, links):
        built = defaultdict(list)
        facets = links['http://opds-spec.org/facet']
        if not facets:
            return built
        for link in facets:
            group = link['facetgroup']
            built[group].append(link)
        return built

    def shorten(self, link, base=None):
        base = base or self.server
        return link['href'][len(base):]

    def explain(self):
        feed = self.parsed['feed']
        feed_title = feed['title']
        url = self.short_url
        feed_by_rel = self.links_by_rel(self.parsed['feed']['links'])

        entries = []
        for e in self.soup.find_all('entry'):
            title = e.find('title').string
            entry_by_rel = self.links_by_rel(e.find_all('link'))
            collection = entry_by_rel['collection']
            if collection:
                collection = self.shorten(collection[0])
            series = e.find('schema:Series')
            if series:
                series_position = series.get('schema:position', 'NO POSITION')
            else:
                series_position = 'NO SERIES'
            updated = e.find('updated').string

            if self.order == 'update':
                fields = [updated, title]
            elif self.order == 'title':
                fields = [title]
            elif self.order == 'series_position':
                fields = [series_position, title]
            if collection: 
               fields.append(collection)
            entries.append(fields)

        print '%.2f: "%s" (%s)' % (self.end-self.start, feed_title, url)
        for link in self.pagination(feed_by_rel):
            print " PAGE"
            print "  ",self.shorten(link, self.url)

        for group, facets in sorted(self.facet_groups(feed_by_rel).items()):
            print " FACETS: %s" % group
            for link in facets:
                if link.get('activefacet') == 'true':
                    active = "ACTIVE "
                else:
                    active = ""
                print "  ", active, self.shorten(link, self.url)
        print " ENTRIES"
        for e in entries:
            print "  ", ("|".join(e).encode("utf8"))
        print

update_order = [
    'QANYPL/crawlable',
    'QANYPL/lists/Staff%20Picks/crawlable',
    'collections/Overdrive/crawlable',
]

title_order = [
    'QANYPL/works/contributor/Jude%20Watson',
    'QANYPL/works/contributor/Jude%20Watson/eng/Children',
    'QANYPL/works/Overdrive%20ID/601d87b8-61ee-4800-a8f3-3bcb3adc8261/recommendations',
    'QANYPL/works/Overdrive%20ID/601d87b8-61ee-4800-a8f3-3bcb3adc8261/related_books',
]

series_position_order = [
    'QANYPL/works/series/Star%20Wars:%20The%20Last%20of%20the%20Jedi',
    'QANYPL/works/series/Star%20Wars:%20The%20Last%20of%20the%20Jedi/eng/Children',
]

server = 'http://qa.circulation.librarysimplified.org/'
server = 'http://localhost:6500/'
for l, order in (
    (update_order, 'update'),
    (title_order, 'title'),
    (series_position_order, 'series_position')
):
    for url in l:
        Client(server, url, order).explain()
```

api/controller.py
```diff
@@ -803,9 +803,7 @@ def crawlable_library_feed(self):
         title = library.name
         lane = CrawlableCollectionBasedLane()
         lane.initialize(library)
-        return self._crawlable_feed(
-            library=library, title=title, url=url, lane=lane
-        )
+        return self._crawlable_feed(title=title, url=url, lane=lane)
```
leonardr: `library` isn't in this method signature anymore. The test didn't catch it because the mock `_crawlable_feed` would take whatever arguments you gave it.

api/opds.py
```diff
@@ -176,9 +176,28 @@ def visible_delivery_mechanisms(self, licensepool):
...
+                # last_update is seconds-since epoch; convert to UTC datetime.
+                updated = datetime.datetime.utcfromtimestamp(last_updates[0])
+
+                # There's a chance that work.last_updated has been
+                # modified but the change hasn't made it to the search
+                # engine yet. Even then, we stick with the search
+                # engine value, because a sorted list is more
+                # important to the import process than an up-to-date
+                # 'last update' value.
+
```
leonardr: This was previously in `CirculationManagerAnnotator`, which meant the functionality wasn't available for `CrawlableCollectionLane` -- last update time was being calculated and ignored.

api/lanes.py
```diff
@@ -849,7 +849,10 @@ class RecommendationLane(WorkBasedLane, DatabaseExclusiveWorkList):
 
     DISPLAY_NAME = "Recommended Books"
     ROUTE = "recommendations"
-    MAX_CACHE_AGE = 7*24*60*60      # one week
+
+    # Cache for 24 hours -- would ideally be much longer but availability
+    # information goes stale.
+    MAX_CACHE_AGE = 24*60*60
```
leonardr: Per my comment here: https://github.com/NYPL-Simplified/circulation/pull/1260#discussion_r298627864

PR #1266, opened 2019-07-01
Added missing shebang and made Python migration script executable.

leonardr:
This branch fixes a problem I found while migrating a real server to 2.3.5. The Python script isn't executable, which causes the migration script to fail.

I verified the fix by migrating a different server using this branch. The migration script ran without a problem.

I also added a script in bin/informational that was missing -- the script class was written but there was no file in `bin` to invoke it, so it was impossible to run.


leonardr:
Paul found an error in the script, so that won't be ready yet.

PR #1267, opened 2019-07-02
Fixed minor errors in the Overdrive Advantage script.

leonardr:
This script was refactored but the variable names weren't changed. I worked with Paul Swanson to get something that works in a real scenario.

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -1159,8 +1159,8 @@ class OverdriveAdvantageAccountListScript(Script):
         print "%d associated Overdrive Advantage account(s)." % len(
             advantage_accounts
         )
-        for collection in advantage_accounts:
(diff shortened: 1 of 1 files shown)


api/overdrive.py
```diff
@@ -1159,17 +1159,17 @@ def explain_main_collection(self, collection):
         print "%d associated Overdrive Advantage account(s)." % len(
             advantage_accounts
         )
-        for i in advantage_accounts:
+        for collection in advantage_accounts:
```
EdwinGuzman: This `collection` is scoped to the for loop, right? Because there is another `collection` in this function’s param.

leonardr: Yeah, this is masking the other `collection` variable. I'll change the name to make it more clear.

PR #1268, opened 2019-07-02
Lane size is now calculated through search engine

leonardr:
Dealing with fallout from the core branch that fixed https://jira.nypl.org/browse/SIMPLY-2088.

This branch changes the way controller code tells a lane to update its `.size`. Since this now happens through the search engine, the controller's `.search_engine` is passed in to `Lane.update_size`. As a side effect, this makes it easy to set up mocks to ensure that the size is updated.

### DIFF
--- a/api/admin/controller/work_editor.py
+++ b/api/admin/controller/work_editor.py
@@ -961,7 +961,9 @@ class WorkController(AdminCirculationManagerController):
                         affected_lanes.add(lane)
 
             # If any list changes affected lanes, update their sizes.
+            # NOTE: This may not make a difference until the
+            # works are actually re-indexed.
             for lane in affected_lanes:
-                lane.update_size(self._db)
+                lane.update_size(self._db, self.search_engine)
 
             return Response(unicode(_("Success")), 200)
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -951,7 +951,7 @@ class LanesController(AdminCirculationManagerController):
             for list in lane.customlists:
                 if list.id not in custom_list_ids:
                     lane.customlists.remove(list)
-            lane.update_size(self._db)
+            lane.update_size(self._db, self.search_engine)
 
             if is_new:
                 return Response(unicode(lane.id), 201)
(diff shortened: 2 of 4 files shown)


api/admin/controller/work_editor.py
```diff
@@ -961,7 +961,9 @@ def custom_lists(self, identifier_type, identifier):
                         affected_lanes.add(lane)
 
             # If any list changes affected lanes, update their sizes.
+            # NOTE: This may not make a difference until the
+            # works are actually re-indexed.
             for lane in affected_lanes:
-                lane.update_size(self._db)
+                lane.update_size(self._db, self.search_engine)
```
leonardr: I don't think this was working before. For lane size to be affected by a work being added to or removed from a list, the materialized view would have to be updated.  The mechanism has changed from "materialized view" to "search index", but the data that would actually change the count still hasn't been propagated to that mechanism. It's something that happens in the background.

(This isn't the case for when a list is added to or removed from a lane, so the rest of this branch isn't affected.)

I didn't remove the code because I'm not sure I'm right and this is in the middle of a much bigger change.

aslagle: It's not a huge deal for this because the page where you add and remove lists from an individual work does not show lane sizes itself. But in general it is important for the admin interface to have a way to get updated data.

leonardr: Long-term we'll be in a better position, because the search index is updated much more frequently than the materialized view. For example, we can flag a lane as needing a size refresh, and have a script that runs once an hour recalculating the size of any affected lanes.

aslagle: Once an hour is still not enough for this use case though. If you make a change in the admin interface, you expect it to display immediately on the page you're looking at.

leonardr: I don't think we disagree.

If you remove a work from a custom list and need it to immediately be reflected in `lane.size` then you need to reindex the work and then call Lane.update_size(). That's the bit that I don't think was working before. You can force one work to be reindexed right now, but you can't force an update to the materialized view.

If you remove a custom list from a lane, then calling Lane.update_size() (which happens now) is enough to update the size. The list of custom lists goes into the `Filter`, so it takes effect immediately. The presence of a work on a custom list goes into the search index, so the work has to be reindexed before Lane.update_size() will notice.

PR #1273, opened 2019-07-08
Add annotations endpoint link to patron profile doc

adriana-alter:


leonardr:
Confirmed with Mark that this looks good.

PR #1278, opened 2019-07-13
Replace the Overdrive 'full' monitor with one that only looks for new books

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-2121 by replacing FullOverdriveCollectionMonitor with a monitor that looks in the recent past for books that were added to the collection. Everything else that monitor did, is done better by OverdriveCollectionReaper or OverdriveFormatSweep.

I implemented this by taking the code that decides to stop based on how many unchanged works were seen, and moving it out of the superclass and into `RecentOverdriveCollectionMonitor`. There's now a hook method called `should_stop` which lets a subclass decide when to stop processing works.


api/overdrive.py
```diff
@@ -1009,21 +1011,11 @@ class OverdriveCirculationMonitor(CollectionMonitor, TimelineMonitor):
...
-    def __init__(self, _db, collection, api_class=OverdriveAPI):
+    def __init__(self, _db, collection, api_class=OverdriveAPI, analytics_class=Analytics):
         """Constructor."""
         super(OverdriveCirculationMonitor, self).__init__(_db, collection)
         self.api = api_class(_db, collection)
-        self.maximum_consecutive_unchanged_books = (
-            self.MAXIMUM_CONSECUTIVE_UNCHANGED_BOOKS
-        )
-        self.analytics = Analytics(_db)
+        self.analytics = analytics_class(_db)
```
leonardr: Previously we didn't test that anything happened to the Analytics object; this makes it easy to test.

api/overdrive.py
```diff
@@ -1055,40 +1044,59 @@ def catch_up_from(self, start, cutoff, progress):
...
-                if (self.maximum_consecutive_unchanged_books
-                    and consecutive_unchanged_books >=
-                    self.maximum_consecutive_unchanged_books):
-                    # We're supposed to stop this run after finding a
-                    # run of books that have not changed, and we have
-                    # in fact seen that many consecutive unchanged
-                    # books.
-                    self.log.info("Stopping at %d unchanged books.",
-                                  consecutive_unchanged_books)
-                    break
```
leonardr: All this stuff moved to `RecentOverdriveCollectionMonitor`.

leonardr:
@aslagle I asked Edwin to do the code review but I'd also like you to think about whether this strategy might miss any data from Overdrive.

aslagle:
It sounds fine to me.

tests/test_overdrive.py
```diff
@@ -1182,48 +1193,174 @@ def update_licensepool(self, book_id):
...
+        )
+
+        m = monitor.should_stop
+
+        # If the monitor has never run before, we need to keep going
+        # until we run out of books.
+        eq_(False, m(None, object(), object()))
+        eq_(False, m(monitor.NEVER, object(), object()))
+
+        # If information is missing or invalid, we assume that we
```
EdwinGuzman: Is this because the missing or invalid data is what is being updated? Or just safe to keep going because there are more books to process?

leonardr: More like "there are more books to process". This shouldn't happen, but if Overdrive provides bad information for a book, I want this script to keep going until it either gets good information, or it reaches the end of the list (which will make it act like the old script -- inefficient but complete).

PR #1283, opened 2019-07-19
Improve instrumentation on search integration tests, update them, and add new ones

leonardr:
This branch updates the search integration tests to take into account changes in NYPL's collection since October, and changes to the search index itself, which go in https://github.com/NYPL-Simplified/server_core/pull/1095.

The primary ticket here is https://jira.nypl.org/browse/SIMPLY-2133, but there are some smaller ones, which I reference inline.

I added a decorator `@known_to_fail`, which inverts the normal sense of a unit test. An exception will be raised if one of these tests succeeds, and the test will 'pass' so long as a known problem persists.

This makes it possible to represent problems we know we have without having to fix them immediately. We can also make tradeoffs, fixing some tests while breaking others. If we only count the number of tests that fail, we don't have the information necessary to make decisions.

I ran this test suite against the current 3.0 branch and against the new branch I'm working on, which overhauls our query algorithm. In both cases I was using an index created by the new branch. I used this to make further improvements to the query algorithm.

There are now 274 tests in the suite.

In 3.0 beta branch, I see:
 49 expected failures
 55 unexpected failures
 7 unexpected successes
 177 total successes

In the branch I'm working on, I se:
 57 expected failures
 0 errors
 0 unexpected successes
217 total successes

So the new system is doing much better and the code is less complicated. The queries we're sending out are much _more_ complicated, and there is a performance hit, which I talk about in the [core branch](https://github.com/NYPL-Simplified/server_core/pull/1095).

### DIFF
--- a/integration_tests/test_search.py
+++ b/integration_tests/test_search.py
@@ -177,7 +177,7 @@ class Evaluator(object):
 
     def _match_scalar(self, value, expect):
         if hasattr(expect, 'search'):
-            if expect:
+            if expect and value is not None:
                 success = expect.search(value)
             else:
                 success = False
@@ -248,6 +248,7 @@ class Evaluator(object):
         """Does the given result match these criteria?"""
 
         for field, expect in self.kwargs.items():
+            fields = None
             if field == 'subject':
                 success, value, expect_str = self._match_subject(expect, result)
             elif field == 'genre':
@@ -256,9 +257,16 @@ class Evaluator(object):
                 success, value, expect_str = self._match_target_age(expect, result)
             elif field == 'author':
                 success, value, expect_str = self._match_author(expect, result)
+            elif field == 'title_or_subtitle':
+                fields = ['title', 'subtitle']
             else:
-                value = self._field(field, result)
-                success, expect_str = self._match_scalar(value, expect)
+                fields = [field]
+            if fields:
+                for field in fields:
+                    value = self._field(field, result)
+                    success, expect_str = self._match_scalar(value, expect)
+                    if success:
+                        break
             if not success:
                 return False, value, expect_str
         return True, value, expect_str
@@ -413,7 +421,8 @@ class SpecificAuthor(FirstMatch):
             return True
 
         title = self._field('title', first)
-        if self.accept_title and self.accept_title in title:
+        subtitle = self._field('subtitle', first)
+        if self.accept_title and (self.accept_title in title or self.accept_title in subtitle):
             return True
 
         # We have failed.
@@ -483,11 +492,10 @@ class TestGibberish(SearchTest):
             ReturnsNothing()
         )
 
-    @known_to_fail
     def test_wordlike_junk(self):
         # To a human eye this is obviously gibberish, but it's close
-        # enough to English words that it picks up a few results on
-        # a fuzzy match.
+        # enough to English words that it might pick up a few results
+        # on a fuzzy match.
         self.search(
             "asdfza oiagher ofnalqk",
             ReturnsNothing()
@@ -695,8 +703,8 @@ class TestUnownedTitle(SearchTest):
         # search query but doesn't say "spy" anywhere.
         self.search(
             "My life as a spy",
-            Common(title=re.compile("life|spy"), first_must_match=False,
-                   threshold=0.9)
+            Common(title_or_subtitle=re.compile("life|spy"),
+                   threshold=0.5)
         )
 
     @known_to_fail
@@ -745,14 +753,13 @@ class TestMisspelledTitleSearch(SearchTest):
             FirstMatch(title="The Restless Wave")
         )
 
-    @known_to_fail
     def test_kingdom_of_the_blind(self):
         # The first word, which is a fairly common word, is slightly misspelled.
         #
         # The desired book is not on the first page.
         self.search(
             "Kngdom of the blind",
-            FirstMatch(title="The Kingdom of the Blind")
+            FirstMatch(title="Kingdom of the Blind")
         )
 
     def test_seven_husbands(self):
@@ -765,10 +772,10 @@ class TestMisspelledTitleSearch(SearchTest):
 
     @known_to_fail
     def test_nightingale(self):
-        # The top results are works like "Modern Warfare, Intelligence,
-        # and Deterrence."
-
-        # Unusual word, misspelled
+        # Unusual word misspelled.
+        #
+        # This isn't so bad -- the book we're looking for is the second result,
+        # with the first being a title by Florence Nightingale.
         self.search(
             "The nightenale",
             FirstMatch(title="The Nightingale")
@@ -873,12 +880,8 @@ class TestPartialTitleSearch(SearchTest):
                 )
             )
 
-    @known_to_fail
     def test_open_wide(self):
         # Search query cuts off midway through the second word of the subtitle.
-        #
-        # The book we're looking for is on the first page, but underneath two
-        # other books called "open wide" or "wide open".
         self.search(
             "Open wide a radical",
             FirstMatch(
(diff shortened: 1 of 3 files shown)


leonardr:
Here are the tests that work in 3.0beta and fail in the branch I'm working on. Many of these are caused by inappropriate title matches.

```
 #641: test_divorce
 #857: test_i_funnyest
 #892: test_open_wide
 #1131: test_sparks
 #1147: test_singh
 #1521: test_misspelling_1
 #1525: test_misspelling_2
 #1574: test_with_spaces
 #1601: test_sf
 #1628: test_christian_authors
 #1637: test_christian_lust
 #1663: test_graphic_novel
 #1679: test_scary_stories
 #1789: test_deep_poems
 #1883: test_da_vinci
 #2195: test_maisie_dobbs
 #2217: test_gossip_girl_misspelled
 #2299: test_foundation
 #2449: test_incorrect_2
```

integration_tests/test_search.py
```diff
@@ -209,19 +232,46 @@ def _match_target_age(self, how_old_is_the_kid, result):
             return True, how_old_is_the_kid, (result_lower, result_upper)
         return False, how_old_is_the_kid, (result_lower, result_upper)
 
+    def _match_author(self, author, result):
```
leonardr: This encapsulates the logic for checking that a particular author was a major contributor to a book.

integration_tests/test_search.py
```diff
@@ -250,8 +300,8 @@ class Common(Evaluator):
     """It must be common for the results to match certain criteria.
     """
     def __init__(self, threshold=0.5, minimum=None, first_must_match=True,
-                 **kwargs):
-        """Constructor.
+                 negate=False, **kwargs):
```
leonardr: This is used by the `Uncommon` evaluator, which checks that certain results don't dominate the search results (though a few can show up).

integration_tests/test_search.py
```diff
@@ -323,34 +400,116 @@ class SpecificAuthor(FirstMatch):
...
+        for hit in hits:
+            role = self.author_role(author, hit)
+            author_matches.append(role is not None)
+            last_role = role
+            last_title = hit.title
+        self.assert_ratio(author_matches, authors, self.threshold)
+
+
+class SpecificSeries(Common):
+    """Verify that results come from a certain series of books."""
```
leonardr: We have a lot of tests for books in a series, but we don't have a lot of good .series data. Fortunately the series frequently appears in the title of a book, and we can check that. If necessary we can combine this with an author match to make sure we're not getting unrelated books with the series name in the title.

integration_tests/test_search.py
```diff
@@ -323,34 +400,116 @@ class SpecificAuthor(FirstMatch):
...
+        return (series_match or title_match) and author_match, actual
 
 
 class SearchTest(object):
     """A test suite that runs searches and compares the actual results
     to some expected state.
     """
 
+    expected_failures = []
+    unexpected_successes = []
```
leonardr: We keep track of these and summarize the results during module teardown.

integration_tests/test_search.py
```diff
@@ -323,34 +400,116 @@ class SpecificAuthor(FirstMatch):
...
+
```
leonardr: Now we get into changes to specific tests -- mostly changing the way we evaluate tests -- and new tests.

integration_tests/test_search.py
```diff
@@ -578,169 +737,160 @@ def test_unowned_partial_title_rosetta_stone(self):
...
-        # NOTE: this fails in both versions of ES, but the top ES1 results are reasonable
-        # (titles containing "nightfall"), whereas the top ES6 result is entitled
-        # "Modern Warfare, Intelligence, and Deterrence."
-
-        # Unusual word, misspelled
+    def test_nightingale(self):
+        # Unusual word misspelled.
+        #
+        # This might fail because a book by Florence Nightingale is
+        # seen as a better match.
```
leonardr: Sometimes a test passes right now but I've seen it go back and forth as I tweaked the search algorithm. In that case I put a note like this one explaining why a test might _start_ failing.

integration_tests/test_search.py
```diff
@@ -855,21 +1011,30 @@ def test_title_match_with_genre_name_romance(self):
...
             "modern romance aziz ansari",
             FirstMatch(title="Modern Romance", author="Aziz Ansari")
         )
 
+    def test_partial_title_match_with_genre_name_education(self):
+        self.search(
+            "education henry adams",
+            FirstMatch(title="The Education of Henry Adams"),
+        )
+
```
leonardr: For https://jira.nypl.org/browse/SIMPLY-548

integration_tests/test_search.py
```diff
@@ -878,20 +1043,13 @@ def test_law_of_the_mountain_man_with_author(self):
             ]
         )
 
-    def test_title_match_with_genre_name_spy_unowned(self):
```
leonardr: I moved this test -- it's earlier in the diff.

integration_tests/test_search.py
```diff
@@ -323,34 +400,116 @@ class SpecificAuthor(FirstMatch):
     Most of the results must also be by that author.
     """
 
-    def __init__(self, author, accept_book_about_author=False, threshold=0):
+    def __init__(self, author, accept_title=None, threshold=0):
```
leonardr: accept_title lets me customize what kind of title I'll accept as a book about the author. It doesn't have to be that the title of the biography is literally the author's full name.

integration_tests/test_search.py
```diff
@@ -1239,31 +1433,73 @@ def test_danielle_steel(self):
             ]
         )
 
-    def test_nabokov(self):
+    def test_primary_author_with_coauthors(self):
```
leonardr: These tests, especially test_narrator, check our new ability to search contributors with different roles.

integration_tests/test_search.py
```diff
@@ -1239,31 +1433,73 @@ def test_danielle_steel(self):
...
         self.search(
             "Griffiths elly", SpecificAuthor("Elly Griffiths")
         )
 
-    def test_mankel(self):
-        # The search query lists the author's last name before his first name.
-        self.search(
-            "mankel henning", SpecificAuthor("Henning Mankel")
-        )
-
```
leonardr: All the Henning Mankel tests were moved to their own class.

integration_tests/test_search.py
```diff
@@ -1324,114 +1560,196 @@ def test_incorrect_2(self):
...
+    @known_to_fail
     def test_with_no_periods_or_spaces(self):
+        # The author name is indexed as "m j", and without a space
+        # between the "m" and the "j" Elasticsearch won't match the
+        # tokens.
         self.search("mj rose")
 
 
+class TestPublisherMatch(SearchTest):
```
leonardr: We searched these fields before but didn't have any tests for them.

integration_tests/test_search.py
```diff
@@ -1854,6 +2190,76 @@ def test_witches(self):
             Common(subject=re.compile('witch'))
         )
 
+class TestFuzzyConfounders(SearchTest):
+    """Test searches on very distinct terms that are near each other in
+    Levenstein distance.
+    """
```
leonardr: This verifies that it's okay to remove the special fuzzy regexp matching in core.

integration_tests/test_search.py
```diff
@@ -2310,8 +2837,60 @@ def test_grade_and_subject(self):
         )
 
 
+class TestSearchOnStopwords(SearchTest):
+    # These tests verify our ability to search, when necessary, using
+    # words that are normally stripped out as stopwords.
```
leonardr: These tests verify the fix to https://jira.nypl.org/browse/SIMPLY-1398

integration_tests/test_search.py
```diff
@@ -59,6 +63,19 @@
 #     self.book_by_someone_else,      # match across fields (no 'biography')
 # ]
 
+def known_to_fail(f):
+    @wraps(f)
+    def decorated(*args, **kwargs):
+        try:
+            ignore = f(*args, **kwargs)
+        except Exception, e:
+            SearchTest.expected_failures.append(f)
```
aslagle: Appending the test function seems a bit weird, but I guess you can get its name?

integration_tests/test_search.py
```diff
@@ -578,166 +737,155 @@ def test_unowned_partial_title_rosetta_stone(self):
...
         )
 
-    def test_misspelled_title_match_kingdom(self):
-        # NOTE: this works in ES1, but not in ES6!  In ES6, the target title is not
-        # in the top search results.
-
+    def test_kingdom_of_the_blind(self):
         # The first word, which is a fairly common word, is slightly misspelled.
+        #
+        # The desired book is not on the first page.
```
aslagle: How does the test pass then?

aslagle: Never mind, saw your next comment.

aslagle: Maybe better to change this to say "The test might fail because..."

integration_tests/test_search.py
leonardr: Yes, it's used at the end to show the names and line numbers of the tests that need attention -- the sort of service provided by nose when a test actually fails.

PR #1284, opened 2019-07-25
Add a maximum number of times a SIP message can be resent

jonathangreen:
I hit a case where a ILS was causing the SIP client to go into an infinite loop of retries because the message we were sending was missing a value they required. This caused a 96 message, which causes us to resend. Since we didn't add the required data they responded with a retry message. 

This will eliminate the infinite loop of retries should this situation arise again in the future.

### DIFF
--- a/api/sip/client.py
+++ b/api/sip/client.py
@@ -395,7 +395,7 @@ class SIPClient(Constants):
         parsed = None
         retries = 0
         while not parsed:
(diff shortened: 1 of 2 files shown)


jonathangreen:
There are 3 hard problems in computer science: naming things, and off-by-1 errors. 😃 

Tests should pass now.

PR #1286, opened 2019-07-31
Validate lists of urls and emails

aslagle:
This fixes a bug that prevents adding external library URLs to an ODL collection. I did the same thing for emails, just in case, though I don't know of anywhere we use a list of emails.

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -1709,8 +1709,10 @@ class SettingsController(AdminCirculationManagerController):
             return UNKNOWN_PROTOCOL
 
     def _get_settings(self):
-        [protocol] = [p for p in self.protocols if p.get("name") == flask.request.form.get("protocol")]
-        return protocol.get("settings")
+        if hasattr(self, 'protocols'):
+            [protocol] = [p for p in self.protocols if p.get("name") == flask.request.form.get("protocol")]
+            return protocol.get("settings")
+        return []
 
     def validate_formats(self, settings=None, validator=None):
         # If the service has self.protocols set, we can extract the list of settings here;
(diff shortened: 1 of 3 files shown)


api/admin/controller/__init__.py
```diff
@@ -1709,8 +1709,10 @@ def validate_protocol(self, protocols=None):
             return UNKNOWN_PROTOCOL
 
     def _get_settings(self):
-        [protocol] = [p for p in self.protocols if p.get("name") == flask.request.form.get("protocol")]
-        return protocol.get("settings")
+        if hasattr(self, 'protocols'):
+            [protocol] = [p for p in self.protocols if p.get("name") == flask.request.form.get("protocol")]
+            return protocol.get("settings")
+        return []
```
aslagle: This last commit fixes an error when creating manual import collection, which does not have any settings. It tries to look them up using this method but self.protocols is not defined either.

tests/admin/test_validator.py
```diff
@@ -41,6 +42,22 @@ def test_validate_email(self):
...
 
+        # Two valid in a list
+        form = MultiDict([('help-email', valid), ('help-email', 'valid2@email.com')])
+        response = Validator().validate_email(Configuration.LIBRARY_SETTINGS, {"form": form})
+        eq_(response, None)
+
+        # One valid and one empty in a list
+        form = MultiDict([('help-email', valid), ('help-email', '')])
+        response = Validator().validate_email(Configuration.LIBRARY_SETTINGS, {"form": form})
+        eq_(response, None)
```
adriana-alter: What about situations in which everything in the list is required (i.e. an empty input should make the whole validation fail)...or does that never happen?

aslagle: I don't think there's a case when that would happen, since for a list you can have any number of items. The way the UI is, I think it's pretty common to leave the last one blank- at least, that's what I personally do. It's not clear if you have to click add on the last one for it to be included, so I do, and then I get an empty one that also gets submitted. The only problem I could think of was if empty string was actually a valid list item for some reason, but we don't have anything like that.

PR #1287, opened 2019-07-31
Add ao field value to requests

alsrlw:
Addresses Jira ticket [https://jira.nypl.org/browse/SIMPLY-2138](https://jira.nypl.org/browse/SIMPLY-2138). Adds an optional Institution ID field to the Admin interface patron authentication integration form, as a per-integration, per-library setting, following Slack discussion. This enables specification of an institution_id value for an ILS that requires a value for the AO field. Adding it in the per-library section facilitates creating one patron auth integration to manage the case where a shared/consortial ILS uses separate IDs to distinguish libraries.

### DIFF
--- a/api/sip/client.py
+++ b/api/sip/client.py
@@ -235,7 +235,7 @@ class SIPClient(Constants):
     ]
 
     def __init__(self, target_server, target_port, login_user_id=None,
-                 login_password=None, location_code=None, institution_id=None, separator=None,
+                 login_password=None, location_code=None, institution_id='', separator=None,
                  use_ssl=False, ssl_cert=None, ssl_key=None
     ):
         """Initialize a client for (but do not connect to) a SIP2 server.
@@ -826,7 +826,7 @@ class MockSIPClient(SIPClient):
     """
 
     def __init__(self, login_user_id=None, login_password=None, separator="|",
-                 target_server=None, target_port=None, location_code=None, institution_id=None):
+                 target_server=None, target_port=None, location_code=None, institution_id=''):
         super(MockSIPClient, self).__init__(
             None, None, login_user_id=login_user_id,
             login_password=login_password, separator=separator
--- a/api/sip/client.py
+++ b/api/sip/client.py
@@ -235,7 +235,7 @@ class SIPClient(Constants):
     ]
 
     def __init__(self, target_server, target_port, login_user_id=None,
-                 login_password=None, location_code=None, separator=None,
+                 login_password=None, location_code=None, institution_id=None, separator=None,
                  use_ssl=False, ssl_cert=None, ssl_key=None
     ):
         """Initialize a client for (but do not connect to) a SIP2 server.
@@ -253,6 +253,7 @@ class SIPClient(Constants):
         if target_port:
             self.target_port = int(target_port)
         self.location_code = location_code
+        self.institution_id = institution_id
         self.separator = separator or '|'
 
         self.use_ssl = use_ssl or ssl_cert or ssl_key
@@ -425,7 +426,7 @@ class SIPClient(Constants):
         )
 
     def end_session_message(
-            self, patron_identifier, patron_password="", institution_id="",
+            self, patron_identifier, patron_password="",
             terminal_password="",
     ):
         """
@@ -447,7 +448,7 @@ class SIPClient(Constants):
         timestamp = self.now()
 
         message = (code + timestamp +
-                   "AO" + institution_id + self.separator +
+                   "AO" + self.institution_id + self.separator +
                    "AA" + patron_identifier + self.separator +
                    "AC" + terminal_password
         )
@@ -469,7 +470,7 @@ class SIPClient(Constants):
         )
 
     def patron_information_request(
-            self, patron_identifier, patron_password="", institution_id="",
+            self, patron_identifier, patron_password="",
             terminal_password="",
             language=None, summary=None
     ):
@@ -495,7 +496,7 @@ class SIPClient(Constants):
         summary = summary or self.summary()
 
         message = (code + language + timestamp + summary
-                   + "AO" + institution_id + self.separator +
+                   + "AO" + self.institution_id + self.separator +
                    "AA" + patron_identifier + self.separator +
                    "AC" + terminal_password
         )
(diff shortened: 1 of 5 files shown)


api/sip/client.py
```diff
@@ -235,7 +235,7 @@ class SIPClient(Constants):
     ]
 
     def __init__(self, target_server, target_port, login_user_id=None,
-                 login_password=None, location_code=None, separator=None,
+                 login_password=None, location_code=None, institution_id=None, separator=None,
```
leonardr: Note that institution_id is set to None by default...

api/sip/client.py
```diff
@@ -447,7 +448,7 @@ def end_session_message(
         timestamp = self.now()
 
         message = (code + timestamp +
-                   "AO" + institution_id + self.separator +
+                   "AO" + self.institution_id + self.separator +
```
leonardr: That's what's causing the `cannot concatenate 'str' and 'NoneType' objects` errors here. `None + "|"` is giving you an error.

Take a look at login_message:

```
        if location_code:
            message = message + self.separator + "CP" + location_code
```

If location_code is None then we don't even send that field. I think that's the right solution here. If you want to send this field even when the value is empty, then the default should be the empty string rather than None.

tests/sip/test_client.py
```diff
@@ -317,6 +317,22 @@ def test_location_code_is_optional(self):
         )
         assert with_code.endswith("COlogin_password|CPlocation_code")
 
+    def test_institution_id_field_is_required(self):
+        without_institution_arg = self.sip.patron_information_request(
+            "patron_identifier", "patron_password"
+        )
+        assert without_institution_arg.startswith('AO|', 31)
+
+    def test_institution_id_field_value_is_optional(self):
```
leonardr: You took `institution_id` out of the signature of this method, so I don't think the code you're testing exists anymore -- that's why you get a test failure. And the institution_id field value is _not_ optional -- it's required, though it can be empty. So I think you can just remove this test.

api/sip/client.py
alsrlw: Ah, I missed the implication of the initial value: None is an appropriate default setting for an optional _field_, but when a _field_ is required, it's default setting must be an appropriate _value_ (''/blank in this case).

api/sip/client.py
alsrlw: I did see that this was likely the initial cause of the error, and I see that change takes care of my questions in Slack about needing to assign a blank/specific value in the Mock sip object. (My conditional assignment of a blank in `authenticator.py L1273` also takes care of that issue when values are captured from the DB patron auth/library configuration settings.)

tests/sip/test_client.py
alsrlw: > You took `institution_id` out of the signature of this method, so I don't think the code you're testing exists anymore -- that's why you get a test failure. 

Sorry, I'm not sure what you mean here. The institution_id = '' reference (and = 'MAIN') was a mental error when I wrote the test. I'm thinking that shouldn't be there. 

I see changing the default value of institution_id in the MockSIPClient's __init__ method to match the actual class is needed (I think that's what you were referring to above). And that should satisfy the value setting for the "blank" test. I'm assuming **to test the specific value, I would add an assignment `self.institution_id = 'MAIN'` before the 'with_institution_provided` line?**

And, by naming conventions, I may have muddied the waters a bit. It looks like we need either two or three tests here. For the first, we just need to test the presence of the _field_, it specifier, not its value. In the code, we hard-code the inclusion of the AO field, but we didn't have a test to check that that hard-coding remains intact.

Second, we need to test the field's value (which I called field_value, and used optional to specify that the _field value_ is optional--not the field itself. There's maybe a better name for that so it's not so confusing (on the other hand, having a required field that doesn't require a non-zero/non-blank setting seems odd, but there we are). I didn't quite know what to call the test, but field_value_is_optional seemed right when I wrote it :-) . Maybe we don't need a value test; but it seemed like it would be good to test whether a provided value actually gets transmitted in the message.

I'm thinking my test of a blank value is superfluous now.

How would you handle the test(s)/naming?

leonardr: I think testing an empty-string value is a good idea, because it will verify that the AO field is sent out even when "no value" is provided.

alsrlw:
To your points above:

1. I think I was distracted by the setup method in this class's tests, where no parameters were passed. I see now in the test `test_different_separator` that a new client is created in order to pass the parameter for a specific test. So I'll replace my assignment line with: `sip = MockSIPClient(institution_id='MAIN')` and then call sip.patron_information_request() instead.

2. I'll take a look and create one.

3. If I understand correctly, here you're wanting a test of the value of `self.sip.institution_id` when no institution_id parameter is passed, like: `eq_("", self.sip.institution_id)`? I see the value of an explicit test. Just clarifying, though: Is the same state implicitly tested in `test_institution_id_field_is_required`? (Or did you want the test I took out, where I would pass an empty string explicitly to a new client?)

leonardr:
You're right, `test_institution_id_field_is_required` tests the case I was thinking of. I was confused by the name. Can you call it `test_institution_id_field_is_always_provided`?  "required" makes it sound like our code is the code imposing the requirement; actually it's imposed by the SIP2 spec.

alsrlw:
That makes sense. I'll change the name.

alsrlw:
Well, that's not a failure I was expecting (in `test_institution_id_field_value_provided`). Obviously, either the line creating the new client is wrong or the call to `sip.patron_information_request` is. I'm missing something (it looks right to me based on `test_different_separator`).

api/sip/client.py
```diff
@@ -825,7 +826,7 @@ class MockSIPClient(SIPClient):
     """
 
     def __init__(self, login_user_id=None, login_password=None, separator="|",
-                 target_server=None, target_port=None, location_code=None):
+                 target_server=None, target_port=None, location_code=None, institution_id=''):
```
leonardr: Here you take `institution_id` as an argument but you don't do anything with it. It should be passed into the superclass constructor along with login_user_id and separator.

alsrlw: Ah, I saw I needed the value in the constructor, but missed passing it to the parent class. About that, an educational point: are we essentially overriding __init__ to mask the need for target_server and target_port in the calls to MockSIPClient? And, if so, are the parameters `target_server=None, target_port=None,` needed in the constructor?

leonardr: Right on both counts. We don't want MockSIPClient to actually make any calls, and it's probably not necessary for the constructor to take the arguments.

PR #1288, opened 2019-07-31
Sphinx documentation

EdwinGuzman:
Resolves [SIMPLY-2161](https://jira.nypl.org/browse/SIMPLY-2161).

The documentation for this branch: https://circulation.readthedocs.io/en/sphinx-documentation/index.html

It's a manual process to select which branches to generate documentation for on readthedocs but it's set up to build whenever the `master` and this PR branch have new updates. There is more information on how to build the docs in the readme.

The PR mostly consists of updating docstring syntax so that Sphinx can parse them:
Screen Shot 2019-07-29 at 2 02 45 PM


.gitmodules
```diff
@@ -1,6 +1,6 @@
 [submodule "core"]
 	path = core
-	url = git@github.com:NYPL-Simplified/server_core.git
+	url = https://github.com/NYPL-Simplified/server_core.git
```
EdwinGuzman: readthedocs could not install the submodules through ssh so switched to https.

api/admin/controller/search_service_self_tests.py
```diff
@@ -7,7 +7,7 @@
     ExternalIntegration
 )
 from core.external_search import ExternalSearchIndex
-from core.tests.test_external_search import ExternalSearchTest
```
EdwinGuzman: This creates a reference to the test function which doesn't seem necessary and breaks the build process. `core/tests/test_external_search` calls `core/testing` so nothing has changed.

api/circulation.py
```diff
@@ -325,16 +325,15 @@ def __repr__(self):
...
-    :param identifier_type Ex.: Identifier.RBDIGITAL_ID.
-    :param identifier Expected to be the unicode string of the isbn, etc..
-    :param start_date When the patron made the reservation.
-    :param end_date When reserved book is expected to become available.  Expected to be passed in
-        date, not unicode format.
-    :param hold_position  Patron's place in the hold line.
-        When not available, default to be passed is None, which is equivalent to "first in line".
+    :param identifier_type: Ex. Identifier.RBDIGITAL_ID.
+    :param identifier: Expected to be the unicode string of the isbn, etc.
+    :param start_date: When the patron made the reservation.
```
EdwinGuzman: Param doc syntax needs a colon after the variable name.

EdwinGuzman:
Not all the documentation is perfect or nicely printed, but this first step was to get it to work and deployed.

.gitmodules
leonardr: This is going to cause problems, but I don't remember what the problems are. I do remember we tried doing this once or twice and switched back. Here are my guesses as to where the problems lie:

* This might screw up the Puppet deployment used by NYPL. Possibly also the system for building the Docker images.
* This might make it more difficult to do normal development, since your commits and branches on core won't be associated with your SSH key.

@aslagle Do you remember this at all?

leonardr: I found it: it's a Puppet problem: https://github.com/NYPL-Simplified/circulation/commit/bf7c9ab726ca9792b6c2c091d558ab51f467061c#diff-8903239df476d7401cf9e76af0252622

Our Puppet setup hasn't changed much since 2015, so it's probably still a live problem, but I'll check.

aslagle: Yeah, I think this will cause problems. You might need to give it git credentials so it can install through ssh.

README.md
```diff
@@ -11,6 +11,24 @@ It depends on the [LS Server Core](https://github.com/NYPL-Simplified/server_cor
...
+
+There are two steps to building the documentation: generate the `.rst` files from the codebase and generate the html static site (not to be committed). Inside the `/docs` directory, run:
+
+```
+   $ sphinx-apidoc -f -o source/ .. ../tests/* ../core/tests/* ../integration_tests/*
+```
+
+```
+   $ make html
+```
```
aslagle: It would be nice to put these commands in an executable file.

README.md
```diff
@@ -11,6 +11,24 @@ It depends on the [LS Server Core](https://github.com/NYPL-Simplified/server_cor
 * [How to set up a development environment](https://github.com/NYPL-Simplified/Simplified/wiki/Deployment-Instructions)
 * Two sets of Ansible playbooks are in development: [One developed by Minitex](https://github.com/Minitex/ansible-playbook-libsimple) and [a derivative developed by Amigos Library Services](https://github.com/alsrlw/ansible-playbook-libsimple)
 
+## Generating Documenation
+
+Code documentation using Sphinx can be found on [readthedocs.org](). It currently documents this repo's `api` directory, `scripts` file, and the `core` directory submodule. The configuration for the documentation can be found in `/docs`.
+
+There are two steps to building the documentation: generate the `.rst` files from the codebase and generate the html static site (not to be committed). Inside the `/docs` directory, run:
```
aslagle: Why are the `.rst` files committed?

README.md
```diff
@@ -11,6 +11,24 @@ It depends on the [LS Server Core](https://github.com/NYPL-Simplified/server_cor
...
+
+```
+   $ sphinx-apidoc -f -o source/ .. ../tests/* ../core/tests/* ../integration_tests/*
+```
+
+```
+   $ make html
+```
+
+The `sphinx-apidoc` command will generate a new set of `.rst` files. Currently, there is no automatic way to generate them using continuous integration so they will need to be generated manually.
```
aslagle: Is it possible to have something that runs during CI to ensure that the docstrings are formatted properly, so we don't inadvertently break the docs?

docs/source/conf.py
```diff
@@ -0,0 +1,180 @@
...
+# documentation root, use os.path.abspath to make it absolute, like shown here.
+#
+import os
+import sys
+sys.path.insert(0, os.path.abspath('../../'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = u'circulation-admin'
```
aslagle: Does this control the URL on readthedocs? Something like `simplified-circulation` would probably make more sense.

docs/source/conf.py
```diff
@@ -0,0 +1,180 @@
...
+copyright = u'2019, ed'
+author = u'ed'
+
+# The short X.Y version
+version = u''
+# The full version, including alpha/beta/rc tags
+release = u''
```
aslagle: Does this have to be update for every release in order to have docs for different releases? That would be annoying.

.gitmodules
leonardr: Ho-Ling said the problem is that the Puppet server is on an old OS that doesn't support TLS 1.2. So SSH will work but HTTPS won't.

However, _HTTP_ would also work, and all of this code is public, so security of the code isn't an issue. I ran an experiment: if you clone a GitHub repo via HTTP, and then try to push a new branch, the push will go over HTTPS. So we could make these into HTTP submodules without leaking our Github credentials (although _keeping track_ of the credentials is a pain, at least on Linux -- I had to [build something from source](https://askubuntu.com/questions/773455/what-is-the-correct-way-to-use-git-with-gnome-keyring-and-https-repos).)

aslagle: That won't mess up local development?

README.md
leonardr: I agree. We could even turn it into a target of docs/Makefile.

.gitignore
```diff
@@ -51,6 +51,8 @@ coverage.xml
 
 # Sphinx documentation
 docs/_build/
+docs/doctrees/
+docs/html/
```
leonardr: I think this should be `docs/*/`. We don't want anything except the Makefile. In particular, I agree with Amy that we don't want `docs/source` in the repo -- it's a byproduct of the documentation build process.

README.md
leonardr: pylint is the obvious candidate, but I don't see anything that does exactly this. By default pylint will complain about _missing_ docstrings, and you can [get it to complain](https://docs.pylint.org/en/1.6.0/extensions.html) about missing _parameters_ in docstrings, but we're in the middle and I don't see a way to do that.

We could also run sphinx during CI. We'd need to have a way of running it just on core. Otherwise we wouldn't find out about a core problem until we brought that revision into circulation.

.gitmodules
leonardr: I don't think so; worst case we'll all have to take an extra step to make Git cache our Github credentials locally.

leonardr:
Edwin, can you do a circulation branch and a core branch that is _just_ the reformatting changes? We can get those out of the way immediately. That will make it a lot easier for me to talk about changes we need to make to the build process, and it will reduce the chances that we'll get merge conflicts due to other ongoing work while you're working on the build part.

.gitmodules
EdwinGuzman: This is a problem because readthedocs (.org) is meant for open source projects and giving credentials to install using ssh doesn't seem to be possible. We could use the .com version which is meant for private projects but the .org version is more desirable for us.

Is there a plan on updating the Puppet server? The other alternative I can think of is finding if there is a way to prevent the git submodules to install but we will then not have documentation for `core` .

README.md
EdwinGuzman: Readthedocs doesn't generate them, it only builds the static website from those files.

README.md
EdwinGuzman: I have also not found a way to deploy to readthedocs from Travis CI and there are no webhooks for it. So when a push is made, both readthedocs and Travis build and deploy separately.

docs/source/conf.py
EdwinGuzman: It does not. The name of the repo controls the url. I can change this to `simplified-circulation` for the title of the doc site.

docs/source/conf.py
EdwinGuzman: It would be. There is a `stable` and `latest` convention on readthedocs. I imagine we would use `latest` for master` but `stable` for when we have a new tag release.

docs/source/conf.py
leonardr: :+1: on changing the title

README.md
leonardr: That's fine; I'm only suggesting we run sphinx because sphinx is the piece of software that can catch these particular errors.

README.md
leonardr: I'm unhappy about this. I'd like to think about ways to avoid keeping these files in the repo, or at least putting the CI process in charge of keeping them updated.

.gitmodules
leonardr: Brett knows more about the Puppet situation, and he's out right now. He and Ho-Ling are looking at options but haven't decided on anything.

I think the best options are:

* Change the .gitmodule file to `https` for the general public, and then change it back to `ssh` in the branches we use for Puppet deployments.
* Change the .gitmodule file to `http` and we all configure the [git credential helper](https://help.github.com/en/articles/caching-your-github-password-in-git) so we don't have to keep typing our password.

EdwinGuzman: I prefer the first option but from experience, it is hard to keep track of and easy to override a file with different configurations in different branches. In this case, it seems we'll end up having to use the git credential with either option and it's something else to add to a developer's local workflow.

leonardr: Can we agree to try it with HTTP submodules, with a comment explaining the constraints (NYPL Puppet can't handle HTTPS or SSH, Read The Docs can't handle SSH)?

EdwinGuzman: I wouldn't be opposed to this but I'm trying to figure out if there's a different way to solve this. Just to update something similar to this, ssh also doesn't work on github pages but https does.

EdwinGuzman:
Closing this. I managed to work around the webhooks and use travis to build and then deploy to a separate branch for gh-pages. This means building everything on travis and not committing the generated `.rst` files.
It can be viewed here: http://nypl-simplified.github.io/circulation/

leonardr:
That's really cool. I think this needs some documentation, maybe on the wiki, even if there's no code for it. I see a readthedocs.org web hook set up in Github, but I don't see how the Travis part works -- https://github.com/NYPL-Simplified/circulation/pull/1289/files doesn't touch travis.yml, so I don't understand how Travie knows what to do.

leonardr:
Also, this is something we have no control over, but it would be great if a Github webhook had a field for WHY DID WE SET THIS UP???

EdwinGuzman:
Yeah, I'm going to create a new PR branch for it based on a test branch: https://github.com/NYPL-Simplified/circulation/commit/4c1f8fbe268017ed7a21a0ecdec4249078057ea2

That's how the Travis part works.

PR #1289, opened 2019-08-01
Docstring update

EdwinGuzman:
Based on PR #1288 , this branch only contains the updates to the docstring syntax.


api/circulation.py
```diff
@@ -162,7 +162,7 @@ def apply(self, loan, autocommit=True):
 
 
 class FulfillmentInfo(CirculationInfo):
-    """A record of a technique that can be used _right now_ to fulfill
+    """A record of a technique that can be *used right* now to fulfill
```
aslagle: I think you accidentally changed the emphasized words here.

aslagle:
Seems like there might still be a problem with the core submodule based on the failing tests.

EdwinGuzman:
It seemed to have been with core and not docker.

PR #1291, opened 2019-08-01
Updating local way to build sphinx documentation and Travis command…

EdwinGuzman:
…to build and deploy to github pages.

Documentation can be viewed on the [Github Pages](http://nypl-simplified.github.io/circulation/index.html) for this repo.

This works off the closed PR #1288 . It resolves discussions we had in the previous PR:
* use ssh for git submodules
* does not commit any generated `.rst` source files

It does commit three necessary files that are needed at a minimum to generate the documentation. The make command now also generates the source files as well as build the static site. 

There's no automatic webhook from Travis CI to readthedocs and we were having issues with ssh/https. Instead of using readthedocs, Travis will build the documentation and then deploy it to the `gh-pages` branch (as a force push but still keep history). Travis does have a special command for that which is nice. The only downside is that I had to create a Github Personal Access Token and add it as an environment variable to Travis. This is used by the travis file to push to the `gh-pages` branch.


.travis.yml
```diff
@@ -41,3 +41,14 @@ env:
...
   - ./verbose-test
+  - cd docs && make html
+
+deploy:
+  provider: pages
+  skip_cleanup: true
+  keep-history: true
+  github_token: $GITHUB_TOKEN
+  on:
+    branch: travis-sphinx-doc
```
EdwinGuzman: This needs to be updated to either `3.0beta` or `master`. `master` should be the correct branch since we only want to generate new documentation whenever an update goes into master, but if that's the case then we may not see any new updates until `3.0beta` is merged into `master`.

docs/source/conf.py
```diff
@@ -0,0 +1,179 @@
...
+#
+import os
+import sys
+sys.path.insert(0, os.path.abspath('../../'))
+
+
+# -- Project information -----------------------------------------------------
+
+project = u'Simplified Circulation'
+copyright = u'2019, NYPL Simplified'
```
leonardr: The copyright should be the same as the one in the README: "The New York Public Library, Astor, Lenox, and Tilden Foundations"

docs/source/conf.py
```diff
@@ -0,0 +1,179 @@
...
+author = u'NYPL Simplified'
```
leonardr: The author doesn't have to be that long thing; you can credit it to the Library Simplified team.

docs/source/conf.py
```diff
@@ -0,0 +1,179 @@
...
+# for a list of supported languages.
+#
+# This is also used if you do content translation via gettext catalogs.
+# Usually you set "language" from the command line for these cases.
+language = None
+
+# List of patterns, relative to source directory, that match files and
+# directories to ignore when looking for source files.
+# This pattern also affects html_static_path and html_extra_path.
+exclude_patterns = ['../../core/tests/', '../../tests/']
```
leonardr: Is it not necessary to exclude integration_tests?

docs/source/conf.py
```diff
@@ -0,0 +1,179 @@
...
+    # Latex figure (float) alignment
+    #
+    # 'figure_align': 'htbp',
+}
+
+# Grouping the document tree into LaTeX files. List of tuples
+# (source start file, target name, title,
+#  author, documentclass [howto, manual, or own class]).
+latex_documents = [
+    (master_doc, 'SimplifiedCirculation.tex', u'Simplified Circulation Documentation',
```
leonardr: In general I would call this product the "Library Simplified Circulation Manager" rather than "Simplified Circulation".

docs/source/conf.py
EdwinGuzman: It's not, I'll remove.

aslagle:
I agree that github pages and master are fine. It would be cool if we could put docs for different branches/tags in different directories, but not necessary.

EdwinGuzman:
I can try to figure out a way to have multiple doc directories later... that was something readthedocs handled and may be hard to do manually.

I updated the configuration and also the travis command so that if there are any rst errors in the docstrings, the travis build will error out before running tests (which take longer). Example:
Screen Shot 2019-08-02 at 12 26 03 PM

EdwinGuzman:
Alright, the failing example works and now it's passing. It's not deploying right now since the target branch is now set to `master`. I think this is good to go now.

PR #1298, opened 2019-08-12
Make Python 2 code forwards-compatible with Python 3

leonardr:
Like https://github.com/NYPL-Simplified/server_core/pull/1102, this branch rewrites a bunch of code that can't be automatically converted to Python 3, into code that either works in both versions or can be converted to Python 3.

The pitfalls are mostly the same as in the core branch, but there are two problems we encountered here for the first time.

* In Python 3, you can't `import *` from inside a function.
* Some tests had some problems with the Python 3 version of werkzeug, because we were mocking `flask.request.form` without also touching `flask.request.stream` and `flask.request.files`. This put the request into an inconsistent state and triggered an infinite loop. The simplest solution was to write a new context manager which makes sure all three values are set.

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -1788,6 +1788,8 @@ class SitewideRegistrationController(SettingsController):
 
         ignore, private_key = self.manager.sitewide_key_pair
         decryptor = Configuration.cipher(private_key)
+        # We use the standard library's b64decode because decrypt()
+        # only takes a bytestring.
         shared_secret = decryptor.decrypt(
             stdlib_base64.b64decode(shared_secret)
         )
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -1,4 +1,5 @@
 from nose.tools import set_trace
+import base64 as stdlib_base64
 import logging
 import sys
 import os
@@ -1787,7 +1788,9 @@ class SitewideRegistrationController(SettingsController):
 
         ignore, private_key = self.manager.sitewide_key_pair
         decryptor = Configuration.cipher(private_key)
-        shared_secret = decryptor.decrypt(base64.b64decode(shared_secret))
+        shared_secret = decryptor.decrypt(
+            stdlib_base64.b64decode(shared_secret)
+        )
         integration.password = unicode(shared_secret)
 
     def get_catalog(self, do_get, url):
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -1583,7 +1583,7 @@ class SettingsController(AdminCirculationManagerController):
             # making the configuration changes necessary to fix
             # this problem.
             message = _("Exception getting self-test results for %s %s: %s")
-            args = (self.type, item.name, str(e))
+            args = (self.type, item.name, unicode(e))
             logging.warn(message, *args, exc_info=e)
             self_test_results = dict(exception=message % args)
 
@@ -1799,7 +1799,7 @@ class SitewideRegistrationController(SettingsController):
         try:
             response = do_get(url)
         except Exception as e:
-            return REMOTE_INTEGRATION_FAILED.detailed(str(e))
+            return REMOTE_INTEGRATION_FAILED.detailed(unicode(e))
 
         if isinstance(response, ProblemDetail):
             return response
@@ -1859,7 +1859,7 @@ class SitewideRegistrationController(SettingsController):
                 headers=headers
             )
         except Exception as e:
-            return REMOTE_INTEGRATION_FAILED.detailed(str(e))
+            return REMOTE_INTEGRATION_FAILED.detailed(unicode(e))
         return response
 
     def get_shared_secret(self, response):
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -2,7 +2,6 @@ from nose.tools import set_trace
 import logging
 import sys
 import os
-import base64
 import random
 import json
 import jwt
@@ -63,6 +62,7 @@ from core.model import (
 from core.lane import (Lane, WorkList)
 from core.log import (LogConfiguration, SysLogger, Loggly, CloudwatchLogs)
 from core.util.problem_detail import ProblemDetail
+from core.util.string_helpers import base64
 from core.metadata_layer import (
     Metadata,
     LinkData,
@@ -298,7 +298,8 @@ class AdminController(object):
 
     def generate_csrf_token(self):
         """Generate a random CSRF token."""
-        return base64.b64encode(os.urandom(24)).decode("utf8")
+        # NOTE: We don't use random_string here -- maybe we could.
+        return base64.b64encode(os.urandom(24))
 
 class AdminCirculationManagerController(CirculationManagerController):
     """Parent class that provides methods for verifying an admin's roles."""
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -2,8 +2,6 @@ from nose.tools import set_trace
 import base64 as stdlib_base64
 import logging
 import sys
-import os
-import random
 import json
 import jwt
 import re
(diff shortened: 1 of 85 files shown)


api/admin/controller/__init__.py
```diff
@@ -173,14 +174,19 @@ def setup_admin_controllers(manager):
     manager.admin_library_settings_controller = LibrarySettingsController(manager)
     from api.admin.controller.individual_admin_settings import IndividualAdminSettingsController
     manager.admin_individual_admin_settings_controller = IndividualAdminSettingsController(manager)
-    from api.admin.controller.sitewide_services import *
+    from api.admin.controller.sitewide_services import (
+        SitewideServicesController,
+        LoggingServicesController,
+        SearchServicesController,
+        StorageServicesController,
+    )
```
leonardr: In Python 3, you can't `import *` from inside a function.

api/admin/controller/__init__.py
```diff
@@ -293,6 +299,7 @@ def get_csrf_token(self):
 
     def generate_csrf_token(self):
         """Generate a random CSRF token."""
+        # NOTE: We don't use random_string here -- maybe we could.
         return base64.b64encode(os.urandom(24))
```
leonardr: I didn't change this because random_string(24) is going to encode the random string as hex digits, which will give a significantly longer string than base64-encoding.  I don't know if that distinction is important.

tests/admin/controller/test_controller.py
```diff
@@ -862,39 +858,39 @@ def test_diagnostics(self):
         eq_(set(response.keys()), set(["coverage_provider", "monitor", "script", "other"]))
 
         cp_service = response["coverage_provider"]
-        cp_name, cp_collection = cp_service.items()[0]
+        [(cp_name, cp_collection)] = cp_service.items()
```
leonardr: In Python 3, `dict.items()` returns an instance of `dict_items`, which can't be indexed numerically. Fortunately there's only one key-value pair in these dicts, so unrolling them is easy.

tests/test_adobe_vendor_id.py
```diff
@@ -749,17 +753,34 @@ def test_decode_round_trip_with_intermediate_mischief(self):
...
     def test_encode(self):
-        """Test that _encode gives a known value with known input."""
+        # Test that _encode gives a known value with known input.
+
+        # Three pieces of data go into our JWT.
         patron_identifier = "Patron identifier"
         now = datetime.datetime(2016, 1, 1, 12, 0, 0)
         expires = datetime.datetime(2018, 1, 1, 12, 0, 0)
+
+        raw_jwt = 'eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJpc3MiOiJodHRwOi8vbXktbGlicmFyeS5vcmcvIiwic3ViIjoiUGF0cm9uIGlkZW50aWZpZXIiLCJpYXQiOjE0NTE2NDk2MDAuMCwiZXhwIjoxNTE0ODA4MDAwLjB9.wKAnFfJVfJP55CIyD7PntFZrtWVTwDcXHjL-quTndzc'
```
leonardr: The JWT changes here because using Python 3 (or using OrderedDict) changes the order of the key-value pairs in the dict we're serializing, which changes the payload.

This took a long time to debug so I included the code we used to debug as part of the test.

tests/test_axis.py
```diff
@@ -165,15 +165,15 @@ def patron_activity(self, patron, pin):
...
+            str(no_patron_credential.exception))
 
         eq_("Asking for circulation events for the last five minutes",
             recent_circulation_events.name)
         eq_(True, recent_circulation_events.success)
         eq_("Found 3 event(s)", recent_circulation_events.result)
         since = api.recent_activity_called_with
         five_minutes_ago = datetime.datetime.utcnow() - datetime.timedelta(minutes=5)
-        assert (five_minutes_ago-since).total_seconds() < 2
+        assert (five_minutes_ago-since).total_seconds() < 5
```
leonardr: For some reason this test reliably took longer in Python 3.

tests/test_lanes.py
```diff
@@ -84,11 +84,11 @@ def test_create_lanes_for_large_collection(self):
         for lane in lanes:
             eq_(self._default_library, lane.library)
             # They all are restricted to English and Spanish.
-            eq_(x.languages, languages)
+            eq_(lane.languages, languages)
```
leonardr: `x` was the wrong variable name. Python 3 caught this error because `x` goes out of scope once the list comprehension completes.

tests/test_overdrive.py
```diff
@@ -1015,7 +1015,7 @@ def test_sync_bookshelf_creates_local_loans(self):
         self.api.queue_response(200, content=holds_data)
         loans, holds = self.circulation.sync_bookshelf(patron, "dummy pin")
         eq_(4, len(loans))
-        eq_(loans.sort(), patron.loans.sort())
+        eq_(loans, patron.loans)
```
leonardr: Apparently sorting these lists was not strictly necessary.

api/bibliotheca.py
```diff
@@ -534,7 +541,7 @@ def findaway_license_to_webpub_manifest(
             license document via Bibliotheca, or a dictionary
             representing such a document loaded into JSON form.
         """
-        if isinstance(findaway_license, basestring):
+        if isinstance(findaway_license, (bytes, unicode)):
```
leonardr: [The porting guide](https://portingguide.readthedocs.io/en/latest/strings.html?highlight=__repr__#type-checking) mentions `six.string_types` as a substitute for basestring, but it doesn't do what we want in this case, because in Python 3, `bytes` isn't considered a "string type".

api/sip/client.py
```diff
@@ -352,12 +352,12 @@ def make_secure_connection(self):
         tmp_ssl_cert_path = None
         tmp_ssl_key_path = None
         if self.ssl_cert:
-            fd, tmp_ssl_cert_path = tempfile.mkstemp()
-            os.write(fd, self.ssl_cert)
+            fd, tmp_ssl_cert_path = tempfile.mkstemp('wb')
+            os.write(fd, self.ssl_cert.encode("ascii"))
```
leonardr: AFAICT SSL certs are always supposed to be ASCII, but we're storing them in the database as Unicode.

api/util/patron.py
```diff
@@ -64,7 +65,8 @@ def assert_borrowing_privileges(cls, patron):
 
         if patron.fines:
             max_fines = Configuration.max_outstanding_fines(patron.library)
-            if max_fines is not None and patron.fines > max_fines.amount:
+            fines = MoneyUtility.parse(patron.fines)
+            if max_fines is not None and fines.amount > max_fines.amount:
```
leonardr: This is a bug caught by the 3.0 conversion -- we were comparing a string (patron.fines) to a number (max_fines.amount).

tests/test_authenticator.py
```diff
@@ -1068,11 +1069,25 @@ def test_create_bearer_token(self):
...
+        # it's not a proper JWT and the signature verification will
+        # fail.  But if you're debugging this test and you need to
+        # look at the pieces that make up expect_jwt, you can call
+        # PyJWT._load(), like so:
+        from jwt.api_jwt import _jwt_global_obj
+        loaded = _jwt_global_obj._load(expect_jwtlike)
+        payload = loaded[0]
+        payload_dict = json.loads(payload)
+        eq_("some token", payload_dict['token'])
+        eq_(oauth1.NAME, payload_dict['iss'])
```
leonardr: Similar to the adobe_vendor_id JWT case. The string we generate changes because it's an encoded version of the JSON representation of a dictionary, and an OrderedDict gives a different representation than a dict. I added look-inside code to the test to make it easier to debug in the future.

tests/test_controller.py
```diff
@@ -326,9 +325,44 @@ def request_context_with_library(self, route, *args, **kwargs):
...
+            # we look up one of them, and we can put in mock values
+            # afterwards.
+            if flask.request.stream is None:
+                flask.request.stream = BytesIO()
+            if flask.request.form is None:
+                flask.request.form = {}
+            if flask.request.files is None:
+                flask.request.files = {}
+
+            yield ctx
```
leonardr: This is the code to avoid the infinite loop I mentioned in the PR description.

api/admin/controller/__init__.py
aslagle: This token ends up in a cookie and also passed down to the front-end code. As long as it's not exceeding the maximum cookie size I think it would be fine.

PR #1304, opened 2019-08-30
Bring in cover images from the Bibliotheca API to use while waiting for metadata wrangler results

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-2263. It's just like https://github.com/NYPL-Simplified/circulation/pull/1303, but for the Bibliotheca API instead of Axis 360.

The URL hack I used in this branch is less official than the one used in the Axis 360 branch, but I'll email Bibliotheca about it to confirm.

Existing sites will pick up the thumbnail images the next time they run bin/bibliotheca_circulation_sweep.

### DIFF
--- a/api/bibliotheca.py
+++ b/api/bibliotheca.py
@@ -247,7 +247,8 @@ class BibliothecaAPI(BaseCirculationAPI, HasSelfTests):
         else:
             return self._request_with_timeout(
                 method, url, data=body, headers=headers,
-                allow_redirects=False)
+                allow_redirects=False, timeout=60
+            )
 
     def get_bibliographic_info_for(self, editions, max_age=None):
         results = dict()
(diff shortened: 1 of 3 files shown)


api/bibliotheca.py
```diff
@@ -247,7 +247,8 @@ def request(self, path, body=None, method="GET", identifier=None,
         else:
             return self._request_with_timeout(
                 method, url, data=body, headers=headers,
-                allow_redirects=False)
+                allow_redirects=False, timeout=60
+            )
```
leonardr: The Bibliotheca API was giving me a lot of timeouts when I tested this, so I boosted the default timeout from 20 minutes to 60. Hopefully this will resolve other transient Bibliotheca API failures going forward.

leonardr:
Confirmed from Bibliotheca that this is a supported feature and using it won't cause any problems on their end.

api/bibliotheca.py
```diff
@@ -760,8 +761,29 @@ def value(bibliotheca_key):
...
+        )
+
+        # Unless the URL format has drastically changed, we should be
+        # able to generate a thumbnail URL based on the full-size
+        # cover URL found in the response document.
+        #
+        # NOTE: this is an undocumented feature of the Bibliotheca API
+        # which was discovered by investigating the BookLinkURL.
+        if '/delivery/img' in cover_url:
+            thumbnail_url = cover_url + "&size=NORMAL"
```
EdwinGuzman: Is the `cover_url` size "large"?

leonardr: Yeah, that's the default size.

PR #1307, opened 2019-09-04
Moved code for exporting analytics events so it can be used from a script.

aslagle:
This moves some of the admin interface code for exporting a CSV of analytics events so that it can be called from a script with an arbitrary date range, instead of a single date.

### DIFF
--- a/api/local_analytics_exporter.py
+++ b/api/local_analytics_exporter.py
@@ -1,0 +1,108 @@
+import csv, codecs, cStringIO
+from StringIO import StringIO
+
+from sqlalchemy.sql import func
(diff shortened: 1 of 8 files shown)


api/local_analytics_exporter.py
```diff
@@ -0,0 +1,108 @@
...
+
+        data = [header] + map(result_to_row, results)
+
+        output = StringIO()
+        writer = UnicodeWriter(output)
+        writer.writerows(data)
+        return output.getvalue()
+
+
+class UnicodeWriter:
```
leonardr: Not necessary to make this change now, but I discovered there's a unicodecsv package that works really well.

PR #1308, opened 2019-09-05
Optimize analytics export

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-2275. It makes the generation of CSV analytics reports 15-16 times faster.

I got most of this improvement by optimizing the query to only ask the database for fields that will actually be used in the CSV. Most of the rest comes from the fact that the genre lookup is done in a subquery -- this is modelled after the subqueries in `Work.to_search_documents`.

While I was at it I put _all_ of the data massage in the database query, making it possible to simply dump the result rows into a CSV writer. And I replaced `UnicodeCSVWriter` with the `unicodecsv` package, which lets us get rid of code and speeds up the actual writing of the CSV by about ten percent.

Functional differences from before:

* I added an 'imprint' field after the 'publisher' field. If you think it's bad to alter the ordering of these fields (for backwards compatibility) I'll put it at the end instead. We'll be adding a couple more fields in the near future -- "library" and "location".
* LocalAnalyticsExporter.export() now outputs a bytestring, not a string object. This doesn't make a difference in Python 2 and should improve Python 3 compatibility, since the CSV will be dumped right into a bytestream -- either standard output or an HTTP response body.

### DIFF
--- a/tests/test_local_analytics_exporter.py
+++ b/tests/test_local_analytics_exporter.py
@@ -26,14 +26,21 @@ class TestLocalAnalyticsExporter(DatabaseTest):
         [lp1] = w1.license_pools
         [lp2] = w2.license_pools
         edition1 = w1.presentation_edition
+        edition1.publisher = "A publisher"
+        edition1.imprint = "An imprint"
         edition2 = w2.presentation_edition
         identifier1 = w1.presentation_edition.primary_identifier
         identifier2 = w2.presentation_edition.primary_identifier
-        genres = self._db.query(Genre).all()
+        genres = self._db.query(Genre).order_by(Genre.name).all()
         get_one_or_create(self._db, WorkGenre, work=w1, genre=genres[0], affinity=0.2)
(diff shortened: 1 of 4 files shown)


tests/test_local_analytics_exporter.py
```diff
@@ -26,14 +26,21 @@ def test_export(self):
         [lp1] = w1.license_pools
         [lp2] = w2.license_pools
         edition1 = w1.presentation_edition
+        edition1.publisher = "A publisher"
+        edition1.imprint = "An imprint"
         edition2 = w2.presentation_edition
         identifier1 = w1.presentation_edition.primary_identifier
         identifier2 = w2.presentation_edition.primary_identifier
-        genres = self._db.query(Genre).all()
+        genres = self._db.query(Genre).order_by(Genre.name).all()
```
leonardr: This made it easier to debug the affinity change, since there was an obvious right order for the genres -- reverse alphabetical order.

PR #1311, opened 2019-09-09
Make the patron's neighborhood available to analytics code for the duration of a request

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-2270. It allows an AuthenticationProvider to communicate a patron's geographic neighborhood by setting the new field `PatronData.neighborhood` when creating a `PatronData` object. This value will be available as `flask.request.patron.neighborhood` for the duration of the request, but since there is no `Patron.neighborhood` field in the model, the information will not be written to the database.

`flask.request.patron.neighborhood` should always be set, but if this feature is disabled or the ILS in use doesn't support it, the value will be `None`.

A future branch will associate this information with appropriate circulation events; this branch is solely about moving the information from the ILS to `flask.request.patron`.

I implemented neighborhood gathering for two of our authentication providers:

* You can set a 'test neighborhood' when configuring the `SimpleAuthenticationProvider`. All authenticated patrons will be given that `.neighborhood`. This is used in the test setup in `test_controller.py`.

* When configuring the `MilleniumPatronAPI` you may choose to extract neighborhood data from the `HOME LIBR[p53]` field (which names the library branch the patron has designated their 'home' branch). You may instead choose to try and extract a ZIP code from the `ADDRESS[pa]` field. And of course you can turn the whole feature off.

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -1243,7 +1243,6 @@ class DashboardController(AdminCirculationManagerController):
         events = map(lambda result: {
             "id": result.id,
             "type": result.type,
-            "patron_id": result.foreign_patron_id,
             "time": result.start,
             "book": {
                 "title": result.license_pool.work.title,
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -120,6 +120,7 @@ class PatronData(object):
                  fines=None,
                  block_reason=None,
                  library_identifier=None,
+                 neighborhood=None,
                  complete=True,
     ):
         """Store basic information about a patron.
@@ -177,11 +178,19 @@ class PatronData(object):
         :param library_identifier: A string pulled from the ILS that
         is used to determine if this user belongs to the current library.
 
+        :param neighborhood: A string pulled from the ILS that
+        identifies the patron's geographic location in a deliberately
+        imprecise way that makes sense to the library -- maybe the
+        patron's ZIP code or the name of their home branch. This data
+        is never stored in a way that can be associated with an
+        individual patron. Depending on library policy, this data may
+        be associated with circulation events -- but a circulation
+        event is not associated with the patron who triggered it.
+
         :param complete: Does this PatronData represent the most
         complete data we are likely to get for this patron from this
         data source, or is it an abbreviated version of more complete
         data we could get some other way?
-
         """
         self.permanent_id = permanent_id
 
@@ -201,6 +210,11 @@ class PatronData(object):
         # We do not store email address in the database, but we need
         # to have it available for notifications.
         self.email_address = email_address
+
+        # We do not store the patron's neighborhood in the database
+        # record, but we may need it to store in the database records
+        # for circulation events triggered by this patron.
+        self.neighborhood = neighborhood
 
     def __repr__(self):
         return "<PatronData permanent_id=%r authorization_identifier=%r username=%r>" % (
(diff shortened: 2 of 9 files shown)


api/admin/controller/__init__.py
```diff
@@ -1243,7 +1243,6 @@ def circulation_events(self):
         events = map(lambda result: {
             "id": result.id,
             "type": result.type,
-            "patron_id": result.foreign_patron_id,
```
leonardr: Because I removed the `foreign_patron_id` field from the `patrons` table, we need to stop gathering this information and sending it out to the controllers. It looks like this information was never actually used in the display (which makes sense since we never really had it), but I'd like you to verify this.

api/authenticator.py
```diff
@@ -202,6 +211,11 @@ def __init__(self,
         # to have it available for notifications.
         self.email_address = email_address
 
+        # We do not store the patron's neighborhood in the database
+        # record, but we may need it to store in the database records
+        # for circulation events triggered by this patron.
+        self.neighborhood = neighborhood
+
```
leonardr: I felt a need to explain every that this is not going into the database every time there was an assignment that might cause someone scanning the code to be alarmed.

api/admin/controller/__init__.py
EdwinGuzman: Yes, `patron_id` is not displayed in the web admin. We would have to make a minor change to the front-end since it expects `patron_id` in this object (not a big deal but just making a note).

tests/test_millenium_patron.py
```diff
@@ -382,6 +392,27 @@ def test_patron_dump_to_patrondata_restriction_field(self):
         patrondata = api.patron_dump_to_patrondata('alice', content)
         eq_("10", patrondata.library_identifier)
 
+    def test_neighborhood(self):
+        # The value of PatronData.neighborhood depends on the 'neighborhood mode' setting.
+
+        # Default behavior is not to gather neighborhood information at all.
+        api = self.mock_api()
+        content = api.sample_data("dump.success.html")
+        patrondata = api.patron_dump_to_patrondata('alice', content)
```
EdwinGuzman: This is an example of what the ILS returns?

leonardr: Yep, it's a small HTML document with a bunch of lines separated by <BR> tags.

tests/test_authenticator.py
```diff
@@ -958,7 +966,7 @@ def test_authenticated_patron_basic(self):
         patrondata = PatronData(
             permanent_id=patron.external_identifier,
             authorization_identifier=patron.authorization_identifier,
-            username=patron.username
+            username=patron.username, neighborhood="Achewood"
```
EdwinGuzman: Just to make sure, right here the `patron` already exists and is found, so what's happening is this value is added to the `patron` object in `get_or_create_patron` (this is similar to the test above). From here until the test in line 992, does the `PatronData.authenticate` method get called? That's the other place where `patron` is assigned the neighborhood value.

leonardr: Yes, in this case the patron is looked up by get_or_create_patron, and at that point Patron.neighborhood is set. 

`apply_patrondata` is called when the patron is found through `local_patron_lookup`, which can't use `get_or_create_patron` because the query is too complicated. (`local_patron_lookup` is actually the most common way a Patron object is created in the real world.)

PR #1312, opened 2019-09-09
Propagate the neighborhood of a user-triggered circulation event to the analytics handler

leonardr:
This branch completes the work of https://jira.nypl.org/browse/SIMPLY-2271.

There are two ways for the circ manager to trigger a circulation event: it can happen in `CirculationAPI` because a patron did something like check out a book, or it can happen in `AnalyticsController` because a client such as SimplyE explicitly reported an event that happened clientside.

This branch changes both `CirculationAPI` and `AnalyticsController` to look in `flask.request.patron.neighborhood` and if it's there, to pass it on to the analytics handlers. 

https://github.com/NYPL-Simplified/circulation/pull/1311 is in charge of making `flask.request.patron.neighborhood` show up when appropriate, and https://github.com/NYPL-Simplified/server_core/pull/1114 is in charge of taking that information and putting it in the database. This is the bit in the middle.

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -459,6 +459,65 @@ class CirculationAPI(object):
             return True
         return False
 
+    def _collect_event(self, patron, licensepool, name,
+                       include_neighborhood=False):
+        """Collect an analytics event.
+
+        :param patron: The Patron associated with the event. If this
+            is not specified, the current request's authenticated
+            patron will be used.
+        :param licensepool: The LicensePool associated with the event.
+        :param name: The name of the event.
+        :param include_neighborhood: If this is True, _and_ the
+            current request's authenticated patron is the same as the
+            patron in `patron`, _and_ the authenticated patron has
+            associated neighborhood information obtained from the ILS,
+            then that neighborhood information (but not the patron's
+            identity) will be associated with the circulation event.
+        """
+        if not self.analytics:
+            return
+
+        # It would be really useful to know which patron caused this
+        # this event -- this will help us get a library and
+        # potentially a neighborhood.
+        if flask.request:
+            request_patron = getattr(flask.request, 'patron', None)
+        else:
+            request_patron = None
+        patron = patron or request_patron
+
+        # We need to figure out which library is associated with
+        # this circulation event.
+        if patron:
+            # The library of the patron who caused the event.
+            library = patron.library
+        elif flask.request:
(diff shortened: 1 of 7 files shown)


api/circulation.py
```diff
@@ -459,6 +459,65 @@ def can_revoke_hold(self, licensepool, hold):
             return True
         return False
 
+    def _collect_event(self, patron, licensepool, name,
```
leonardr: There are a lot of rules in `fulfill` about where to get `patron` and `library` under different scenarios. That's because you can fulfill a book without authenticating (this happens in The SimplyE Collection(tm) for example). I centralized rules for analytics collection in this method, so that I could centralize the rules about pulling the 'neighborhood' out of the request.

However, the rules for finding the 'neighborhood' depend on having `patron` first. It was easiest to _also_ move in the complicated rules for finding `patron` and `library` in various places, and run that logic all the time, even though `fulfill` is the only time we currently need that functionality.

api/circulation.py
```diff
@@ -459,6 +459,65 @@ def can_revoke_hold(self, licensepool, hold):
...
+        elif flask.request:
+            # The library associated with the current request.
+            library = flask.request.library
+        else:
+            # The library associated with the CirculationAPI itself.
+            library = self.library
+
+        neighborhood = None
+        if (include_neighborhood and flask.request
+            and request_patron and request_patron == patron):
```
EdwinGuzman: It's possible that no patron is passed and so patron becomes `request_patron`. In that case, `request_patron` is just being checked against itself. Is that fine?

leonardr: Yes, that's fine -- in fact that's the most common case.

EdwinGuzman:
Is there a difference between events in `CirculationAPI` and `AnalyticsController`? How would a patron be able to check out a book (`CirculationAPI`) and not be through a client such as SimplyE?

leonardr:
In `CirculationAPI`, the patron wants to do something that (usually) requires we go out to a third-party vendor and take some action on their behalf. There's some "real work" that needs to be done -- a new loan created, a hold deleted, etc -- and the analytics event is a side effect of that real-world action.

In `AnalyticsController`, we're logging analytics for an event that the server would otherwise not know about -- for example, someone opened a book. The only purpose of `AnalyticsController` is to log the analytics event -- there's no "real work" that needs to be done.

PR #1313, opened 2019-09-10
Fetch TOS information from discovery services before displaying information about them

leonardr:
This branch represents work towards https://jira.nypl.org/browse/SIMPLY-2283. I thought most of this work was done a while ago, but no big deal.

Each discovery service (e.g. the library registry for the SimplyE mobile app) has the ability to provide a link to a terms-of-service document in its registration document. As of https://jira.nypl.org/browse/SIMPLY-2277, a discovery service may provide a snippet of HTML that includes a terms-of-service link and provides human-readable context.

However, the `DiscoveryServiceLibraryRegistrationsController` never actually grabs a discovery service's registration document, and so the TOS information is never made available to the admin interface UI. The text you see during library registration that says "I have read and agree to the terms and conditions" is hard-coded.

This branch changes `DiscoveryServiceLibraryRegistrationsController` so that a discovery service's TOS link and TOS HTML snippet are made available in the dictionary sent out to the admin interface for that service. If there's a problem accessing the discovery service (probably because the server is down), a serialized problem detail document is also sent out -- this will allow the admin interface to flag discovery services that are currently not working.

The main underlying change I made is to move the code that fetches a discovery service's root catalog and registration document from `Registration` to `RemoteRegistry`. We now need this information twice: during the registration process, but also _before_ the registration process, when we're displaying the form.

### DIFF
--- a/api/registry.py
+++ b/api/registry.py
@@ -114,16 +114,16 @@ class RemoteRegistry(object):
     def fetch_registration_document(self, do_get=HTTP.debuggable_get):
         catalog = self.fetch_catalog(do_get=do_get)
         if isinstance(catalog, ProblemDetail):
-            return catalog, None, None
+            return catalog
         registration_url, vendor_id = catalog
 
         response = do_get(registration_url)
         if isinstance(response, ProblemDetail):
-            return response, None, None
+            return response
         terms_of_service_link, terms_of_service_html = (
             self._extract_registration_information(response)
         )
-        return None, terms_of_service_link, terms_of_service_html
+        return terms_of_service_link, terms_of_service_html
 
     @classmethod
     def _extract_catalog_information(cls, response):
@@ -132,11 +132,14 @@ class RemoteRegistry(object):
 
         :param response: A requests-style Response object.
 
-        :return A ProblemDetail if there's a problem parsing the
+        :return A ProblemDetail if there's a problem accessing the
         catalog; otherwise a 2-tuple (registration URL, Adobe vendor
         ID).
         """
-        catalog, links = cls._extract_links(response)
+        result = cls._extract_links(response)
+        if isinstance(result, ProblemDetail):
+            return result
+        catalog, links = result
         if catalog:
             vendor_id = catalog.get("metadata", {}).get("adobe_vendor_id")
         else:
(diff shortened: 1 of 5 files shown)


api/registry.py
```diff
@@ -94,6 +98,131 @@ def registrations(self):
         for x in self.integration.libraries:
             yield Registration(self, x)
 
+    def fetch_catalog(self, catalog_url=None, do_get=HTTP.debuggable_get):
```
leonardr: This code was moved in from Registration, and it properly belongs here. The main catalog and the registration document are located on the remote service, so the code for accessing them should be in RemoteRegistry.

I moved it because we now need to get the registration document at two different times: once when we're about to  register the circulation manager with a discovery service, and once when we're about to display the registration _form_ and we need to see 1) if the server's up and 2) if there are any terms of service we need to display.

tests/test_registry.py
```diff
@@ -119,6 +125,265 @@ def test_registrations(self):
         eq_(registry, registration.registry)
         eq_(self._default_library, registration.library)
 
+    def test_fetch_catalog(self):
```
leonardr: This code existed before, in RemoteRegistry, but was not tested. Since I moved it to another class I decided it was a good opportunity to add test coverage.

tests/test_registry.py
```diff
@@ -231,23 +502,21 @@ def _process_registration_result(self, catalog, encryptor, stage):
                 )
                 return "all done!"
 
-            def mock_do_get(self, url):
-                self.do_get_called_with = url
-                return "A fake catalog"
-
```
leonardr: The code that made the HTTP request used to be in push() (https://github.com/NYPL-Simplified/circulation/pull/1313/files#diff-4251fb997120a022b5b6c87638511e60L233), so we had to mock the request. Now push() calls Registry.fetch_catalog, and we're mocking that method, so there's no longer any reason to mock the request-making process here.

tests/test_registry.py
```diff
@@ -355,60 +621,10 @@ def fail(*args, **kwargs):
...
-            return INVALID_REGISTRATION.detailed(
-                "could not extract catalog information"
-            )
-        registration._extract_catalog_information = fail
+            return INVALID_REGISTRATION.detailed("could not fetch catalog")
+        registry.fetch_catalog = fail
         problem = cause_problem()
-        eq_("could not extract catalog information", problem.detail)
-
-    def test__extract_catalog_information(self):
```
leonardr: This test was moved to TestRemoteRegistry.

api/registry.py
```diff
@@ -94,6 +98,158 @@ def registrations(self):
...
+
+        :return: A 2-tuple (parsed_catalog, links),
+           with `links` being a list of dictionaries, each containing
+           one OPDS link.
+        """
+        # The response must contain either an OPDS 2 catalog or an OPDS 1 feed.
+        type = response.headers.get("Content-Type")
+        if type and type.startswith(cls.OPDS_2_TYPE):
+            # This is an OPDS 2 catalog.
+            catalog = json.loads(response.content)
```
EdwinGuzman: OPDS2 is in json format?

leonardr: Yep.

tests/test_registry.py
```diff
@@ -261,19 +530,17 @@ def push():
         )
         public_key, private_key = Configuration.key_pair(key_pair_setting)
         result = registration.push(
```
EdwinGuzman: Can't this be `result = push()` like the above since the function does this same thing?

PR #1316, opened 2019-09-11
Updating the analytics query to filter by event type and locations.

EdwinGuzman:
This partly resolves [SIMPLY-2274](https://jira.nypl.org/browse/SIMPLY-2274). The other part is the UI. This PR is only for the site-wide dashboard events report and doesn't take a library's id into account.

### DIFF
--- a/api/local_analytics_exporter.py
+++ b/api/local_analytics_exporter.py
@@ -28,10 +28,10 @@ from core.model import (
 class LocalAnalyticsExporter(object):
     """Export large numbers of analytics events in CSV format."""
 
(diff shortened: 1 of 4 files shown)


api/local_analytics_exporter.py
```diff
@@ -59,6 +59,20 @@ def analytics_query(self, start, end):
             directly to a CSV file.
         """
 
+        events_type_filter = True
```
EdwinGuzman: Is there a better way to conditionally add an `and_` clause?

PR #1322, opened 2019-09-18
Updating circulation-web to 0.3.16.

EdwinGuzman:


EdwinGuzman:
@leonardr are the `core` submodule updates okay to merge in?

leonardr:
Yes, thanks for merging them.

In general, if there are unmerged updates in core it's fine to update circulation. The only downside is you might have test failures in circulation you'll have to fix.

PR #1326, opened 2019-09-24
No overdrive hold notification address

leonardr:
This branch resolves https://jira.nypl.org/browse/SIMPLY-2295 by changing the behavior of the Overdrive API client when a patron places a hold. If the patron has not previously registered an email notification address with Overdrive, we tell Overdrive not to send a hold availability notification at all. Previously we sent them the site-wide default notification email address, which is supposed to bounce or trash all incoming email.

The site-wide default address still exists, since other vendor API clients use it, but the Overdrive client ignores it even if it is set.

Most of the code in this branch is tests for the previously untested place_hold method. For easier testing I split it into two methods, one to make the request and one to process the response.

I tested this against the live Overdrive API and it basically works, with one big caveat.

The first time you place a hold through the Overdrive API, and provide a hold notification address, that address is permanently associated with your Overdrive patron account. So there are a lot of NYPL patrons whose Overdrive hold notification address is `holdnotifications@librarysimplified.org`, and similarly for other libraries.

When we're about to place a hold on your behalf, we ask Overdrive if you have a hold notification address. If you do, we pass that address right back to Overdrive and don't set `ignoreHoldEmail`.

Even if we do set `ignoreHoldEmail`, if the patron has a hold notification address on the Overdrive side, Overdrive will ignore the value of `ignoreHoldEmail` and use the patron's hold notification address anyway. Theoretically, we could detect this scenario and tell Overdrive to avoid it, but Overdrive would do it anyway.

Basically, anyone who's ever placed an Overdrive hold before will keep using the "trash everything" email address as their hold notification address. Only patrons who've never placed an Overdrive hold before will trigger the `ignoreHoldEmail` code path.

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -763,7 +763,7 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI, HasSelfTests):
             document
         )
         return self.process_place_hold_response(
-            response, patron, pin, overdrive_id
+            response, patron, pin, licensepool
         )
 
     def process_place_hold_response(self, response, patron, pin, licensepool):
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -710,39 +710,68 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI, HasSelfTests):
         )
 
     def default_notification_email_address(self, patron, pin):
-        site_default = super(OverdriveAPI, self).default_notification_email_address(
-            patron, pin
-        )
+        """Find the email address this patron wants to use for hold
+        notifications.
+
+        :return: The email address Overdrive has on record for
+           this patron's hold notifications, or None if there is
+           no such address.
+        """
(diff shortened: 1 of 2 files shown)


api/overdrive.py
```diff
@@ -710,58 +710,103 @@ def process_checkout_data(cls, checkout, collection):
...
             )
-
         overdrive_id = licensepool.identifier.identifier
-        headers, document = self.fill_out_form(
-            reserveId=overdrive_id, emailAddress=notification_email_address)
+        form_fields = dict(reserveId=overdrive_id)
+        if notification_email_address:
+            form_fields['emailAddress'] = notification_email_address
+        else:
+            form_fields['ignoreHoldEmail'] = True
```
leonardr: This is the fundamental change. `ignoreHoldEmail` is a recent change to the Overdrive API that lets us avoid specifying an email address.

PR #1327, opened 2019-09-24
Updating scripts to add a storage name for the marc uploader script, …

EdwinGuzman:
…and updating the storage integration from sitewide to multiple.

Resolves [SIMPLY-2296](https://jira.nypl.org/browse/SIMPLY-2296).

### DIFF
--- a/api/admin/controller/sitewide_services.py
+++ b/api/admin/controller/sitewide_services.py
@@ -137,14 +137,60 @@ class SearchServicesController(SitewideServicesController):
             service_id, ExternalIntegration.SEARCH_GOAL
         )
 
-class StorageServicesController(SitewideServicesController):
-    def process_services(self):
-        detail = _("You tried to create a new storage service, but a storage service is already configured.")
-        return self._manage_sitewide_service(
-            ExternalIntegration.STORAGE_GOAL,
+class StorageServicesController(SettingsController):
+
+    def __init__(self, manager):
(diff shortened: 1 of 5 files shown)


api/admin/controller/sitewide_services.py
```diff
@@ -137,15 +137,61 @@ def process_delete(self, service_id):
             service_id, ExternalIntegration.SEARCH_GOAL
         )
 
-class StorageServicesController(SitewideServicesController):
-    def process_services(self):
-        detail = _("You tried to create a new storage service, but a storage service is already configured.")
-        return self._manage_sitewide_service(
-            ExternalIntegration.STORAGE_GOAL,
+class StorageServicesController(SettingsController):
```
EdwinGuzman: This should be moved to its own file or with other controllers but leaving it temporarily.

api/admin/controller/sitewide_services.py
```diff
@@ -137,15 +137,61 @@ def process_delete(self, service_id):
...
+        else:
+            # Create a new service
+            storage_service, is_new = self._create_integration(
+                self.protocols, protocol, self.goal
+            )
+            if isinstance(storage_service, ProblemDetail):
+                self._db.rollback()
+                return storage_service
+
+        storage_service.name = name
```
EdwinGuzman: Haven't yet updated the actual values for the storage integration.

PR #1328, opened 2019-09-25
Fix analytics query

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-2306.

In addition to the end-to-end test, I added a test that mocks `LocalAnalyticsExporter` to verify that the controller is passing the right arguments into the `export()` method. That's where the disconnect lay that allowed this bug to slip through the test process.

While testing this I found a couple small problems with the admin interface code, which I filed as https://jira.nypl.org/browse/SIMPLY-2307.

### DIFF
--- a/tests/admin/controller/test_controller.py
+++ b/tests/admin/controller/test_controller.py
@@ -1752,14 +1752,21 @@ class TestDashboardController(AdminControllerTest):
         get_one_or_create(self._db, WorkGenre, work=self.english_1, genre=genres[0], affinity=0.2)
 
         time = datetime.now() - timedelta(minutes=1)
-        get_one_or_create(
+        event, ignore = get_one_or_create(
             self._db, CirculationEvent,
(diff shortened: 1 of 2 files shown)


tests/admin/controller/test_controller.py
```diff
@@ -1752,14 +1752,21 @@ def test_bulk_circulation_events(self):
...
 
+        # Try an end-to-end test, getting all circulation events for
+        # the current day.
         with self.app.test_request_context("/"):
             response, requested_date, date_end, library_short_name = self.manager.admin_dashboard_controller.bulk_circulation_events()
-        reader = csv.reader([row for row in response.split("\r\n") if row], dialect=csv.excel)
+        reader = csv.reader(
+            [row for row in response.split("\r\n") if row],
+            dialect=csv.excel
+        )
```
leonardr: These changes don't matter -- I just did them to make the code easier for me to read and test.

PR #1332, opened 2019-09-30
Rebalance subject weights

leonardr:
This branch fixes the parts of https://jira.nypl.org/browse/SIMPLY-2320 that must be fixed in circulation. The core PR is https://github.com/NYPL-Simplified/server_core/pull/1122.

### DIFF
--- a/api/odilo.py
+++ b/api/odilo.py
@@ -229,15 +229,17 @@ class OdiloRepresentationExtractor(object):
         language = book.get('language', 'spa')
 
         subjects = []
+        trusted_weight = Classification.TRUSTED_DISTRIBUTOR_WEIGHT
         for subject in book.get('subjects', []):
-            subjects.append(SubjectData(type=Subject.TAG, identifier=subject, weight=100))
+            subjects.append(SubjectData(type=Subject.TAG, identifier=subject, weight=trusted_weight))
 
         for subjectBisacCode in book.get('subjectsBisacCodes', []):
-            subjects.append(SubjectData(type=Subject.BISAC, identifier=subjectBisacCode, weight=100))
+            subjects.append(SubjectData(type=Subject.BISAC, identifier=subjectBisacCode, weight=trusted_weight))
 
         grade_level = book.get('gradeLevel')
         if grade_level:
-            subject = SubjectData(type=Subject.GRADE_LEVEL, identifier=grade_level, weight=10)
+            # TODO: It's not clear why we trust this data less.
+            subject = SubjectData(type=Subject.GRADE_LEVEL, identifier=grade_level, weight=trusted_weight / 10)
             subjects.append(subject)
 
         medium = None
--- a/api/odilo.py
+++ b/api/odilo.py
@@ -239,8 +239,7 @@ class OdiloRepresentationExtractor(object):
 
         grade_level = book.get('gradeLevel')
         if grade_level:
-            # TODO: It's not clear why we trust this data less.
-            subject = SubjectData(type=Subject.GRADE_LEVEL, identifier=grade_level, weight=trusted_weight / 10)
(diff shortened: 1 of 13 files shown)


api/overdrive.py
```diff
@@ -836,7 +836,7 @@ def release_hold(self, patron, pin, licensepool):
         url = self.HOLD_ENDPOINT % dict(
             product_id=licensepool.identifier.identifier)
         response = self.patron_request(patron, pin, url, method='DELETE')
-        if response.status_code / 100 == 2 or response.status_code == 404:
+        if response.status_code // 100 == 2 or response.status_code == 404:
```
leonardr: Unrelated change for Python 3 forward-compatibility, found while grepping for "100".

leonardr:
I found another thing that needed changing (Enki tag classifications) when looking for places that should use the new TRUSTED_DISTRIBUTOR_WEIGHT constant.

There are also a couple more places -- Novelist, Onix -- where we should maybe use TRUSTED_DISTRIBUTOR_WEIGHT but we don't. And there are a couple places -- Odilo, RBdigital -- where we use fractions or multiples of TRUSTED_DISTRIBUTOR_WEIGHT for reasons I don't understand. I think we'd probably be safe using TRUSTED_DISTRIBUTOR_WEIGHT in all of these cases but I'm not sure. Any thoughts?

leonardr:
The big disconnect is "data from the distributor" versus "data from the metadata wrangler", and even without making any more changes I think we're reliably privileging data from the distributor over data from the metadata wrangler, so I'm okay with this even if there are some rough edges.

api/odilo.py
```diff
@@ -229,15 +230,17 @@ def record_info_to_metadata(cls, book, availability):
...
+            subjects.append(SubjectData(type=Subject.TAG, identifier=subject, weight=trusted_weight))
 
         for subjectBisacCode in book.get('subjectsBisacCodes', []):
-            subjects.append(SubjectData(type=Subject.BISAC, identifier=subjectBisacCode, weight=100))
+            subjects.append(SubjectData(type=Subject.BISAC, identifier=subjectBisacCode, weight=trusted_weight))
 
         grade_level = book.get('gradeLevel')
         if grade_level:
-            subject = SubjectData(type=Subject.GRADE_LEVEL, identifier=grade_level, weight=10)
+            # TODO: It's not clear why we trust this data less.
```
aslagle: Seems likely this code was copied from another distributor without thinking about it.

api/onix.py
```diff
@@ -200,6 +200,10 @@ def parse(cls, file, data_source_name):
 
             subject_tags = parser._xpath(record, 'descriptivedetail/subject')
             subjects = []
+
+            # TODO: How much do we trust this subject data? We could
+            # bump up the weight quite a bit if it comes right from
+            # the publisher.
```
aslagle: Yeah, I think this is safe to bump up.

leonardr: Unlike the other changes, we can't migrate this change in an existing database because the subject types aren't distinctive and we don't know the data source.

aslagle: As far as I know I'm the only person who has used this so far, and only for one data source, so I can manually update mine.

EdwinGuzman:
So the end result of this is that now all the weights are the same constant but that is also lower than the weight given from the metadata wrangler?

leonardr:
> So the end result of this is that now all the weights are the same constant but that is also lower than the weight given from the metadata wrangler?

Not quite -- (almost) all the weights are the same constant, 100, and the metadata wrangler weights are whatever weight the metadata wrangler gives. This could be higher than 100 (if more than 100 libraries say the same thing), but it's usually lower, and if the metadata wrangler doesn't provide a weight, we assume the default weight is 1 (much lower than what we'd get from a distributor), where it used to be 100.

PR #1333, opened 2019-09-30
Multiple storage integrations

EdwinGuzman:
This is a working version of adding an external integration link between integrations. There are no tests yet but I wanted to make sure this was going the right way before continuing.

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -1591,12 +1591,15 @@ class SettingsController(AdminCirculationManagerController):
         if not integrations.all():
             return
 
-        mirror_integration_settings = ExternalIntegrationLink._get_settings()
-        for integration in integrations:
-            for setting in mirror_integration_settings:
-                setting['options'].append(
-                    dict(key=str(integration.id), label=integration.name)
-                )
+        mirror_integration_settings = list(ExternalIntegrationLink.COLLECTION_MIRROR_SETTINGS)
+        for setting in mirror_integration_settings:
+            for integration in integrations:
+                # set_trace()
+                existing = [option for option in setting['options'] if option['key']==str(integration.id)]
+                if not existing: 
+                    setting['options'].append(
+                        dict(key=str(integration.id), label=integration.name)
+                    )
         return mirror_integration_settings
 
     def _create_integration(self, protocol_definitions, protocol, goal):
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -1595,11 +1595,9 @@ class SettingsController(AdminCirculationManagerController):
         mirror_integration_settings = copy.deepcopy(ExternalIntegrationLink.COLLECTION_MIRROR_SETTINGS)
         for setting in mirror_integration_settings:
             for integration in integrations:
-                existing = [option for option in setting['options'] if option['key']==str(integration.id)]
-                if not existing: 
-                    setting['options'].append(
-                        dict(key=str(integration.id), label=integration.name)
-                    )
+                setting['options'].append(
+                    dict(key=str(integration.id), label=integration.name)
+                )
         return mirror_integration_settings
 
     def _create_integration(self, protocol_definitions, protocol, goal):
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -1393,8 +1393,9 @@ class SettingsController(AdminCirculationManagerController):
             for setting in protocol.get("settings", []):
                 key = setting.get("key")
 
-                # If the setting is a mirror, need to get the value from
-                # ExternalIntegrationLink and not ConfigurationSetting
+                # If the setting is a covers or books mirror, we need to get
+                # the value from ExternalIntegrationLink and
+                # not from a ConfigurationSetting.
                 if key.endswith('mirror_integration_id'):
                     storage_integration = get_one(
                         self._db, ExternalIntegrationLink, external_integration_id=service.id
@@ -1577,7 +1578,7 @@ class SettingsController(AdminCirculationManagerController):
 
         return self_test_results
 
-    def _mirror_integration_setting(self, type=""):
+    def _mirror_integration_setting(self, type):
         """Create a setting interface for selecting a storage integration to
         be used when mirroring items from a collection.
         """
@@ -1587,12 +1588,12 @@ class SettingsController(AdminCirculationManagerController):
             ExternalIntegration.name
         )
 
-        if not integrations:
+        if not integrations.all():
             return
-        key = "%s_mirror_integration_id" % type.lower() if type else "mirror_integration_id"
+        key = "%s_mirror_integration_id" % type.lower()
         mirror_integration_setting = {
             "key": key,
-            "label": _("%s Mirror" % type),
+            "label": _("%s Mirror" % type.capitalize()),
             "description": _("Any cover images or free books encountered while importing content from this collection can be mirrored to a server you control."),
             "type": "select",
             "options" : [
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -181,6 +181,7 @@ def setup_admin_controllers(manager):
     from api.admin.controller.search_service_self_tests import SearchServiceSelfTestsController
     manager.admin_search_service_self_tests_controller = SearchServiceSelfTestsController(manager)
     manager.admin_search_services_controller = SearchServicesController(manager)
+    from api.admin.controller.storage_services import StorageServicesController
     manager.admin_storage_services_controller = StorageServicesController(manager)
     from api.admin.controller.catalog_services import *
     manager.admin_catalog_services_controller = CatalogServicesController(manager)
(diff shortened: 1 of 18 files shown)


api/admin/controller/__init__.py
```diff
@@ -1563,7 +1578,7 @@ def _mirror_integration_setting(self):
         }
         for integration in integrations:
             mirror_integration_setting['options'].append(
-                dict(key=integration.id, label=integration.name)
+                dict(key=str(integration.id), label=integration.name)
```
leonardr: This looks like you're setting up a ConfigurationSetting where `key` is "{type}_mirror_integration_id" and `value` is the database ID of another ExternalIntegration. But this shouldn't be a ConfigurationSetting at all -- it's an ExternalIntegration link where `purpose` is "{type}_mirror".

leonardr: No, I guess this is handled specially earlier in this file. But earlier in the file you need to check `.endswith("mirror_integration_id")`.

api/admin/controller/__init__.py
```diff
@@ -1462,9 +1476,10 @@ def _set_integration_library(self, integration, library_info, protocol):
     def _set_integration_settings_and_libraries(self, integration, protocol):
         settings = protocol.get("settings")
         for setting in settings:
-            result = self._set_integration_setting(integration, setting)
-            if isinstance(result, ProblemDetail):
-                return result
+            if setting.get('key') != 'mirror_integration_id':
```
leonardr: In other places (e.g. line 1398) it's possible for there to be multiple integrations of this type, all ending with "mirror_integration_id", but here, no value other than "mirror_integration_id" will work. I suspect the other code is right and this needs to be changed to use `endswith`.

api/admin/controller/__init__.py
```diff
@@ -1562,20 +1577,22 @@ def _get_prior_test_results(self, item, protocol_class=None):
...
+
         if not integrations:
             return
+        key = "%s_mirror_integration_id" % type.lower() if type else "mirror_integration_id"
         mirror_integration_setting = {
-            "key": "mirror_integration_id",
-            "label": _("Mirror"),
+            "key": key,
+            "label": _("%s Mirror" % type),
             "description": _("Any cover images or free books encountered while importing content from this collection can be mirrored to a server you control."),
```
leonardr: The description also needs to be changed to match the type of thing being mirrored. Maybe `label` and `description` should be separate arguments to this method; then the caller would have complete control over how the UI is presented.

api/admin/controller/catalog_services.py
```diff
@@ -20,6 +23,9 @@ def __init__(self, manager):
         super(CatalogServicesController, self).__init__(manager)
         service_apis = [MARCExporter]
         self.protocols = self._get_integration_protocols(service_apis, protocol_name_attr="NAME")
+        self.protocols[0]['settings'].append(
+            MARCExporter.get_storage_settings(self._db)
+        )
```
leonardr: I'm suspicious any time `[0]` shows up in non-test code but I checked and this is okay -- you put in one API and you're guaranteed to get one dictionary back.

api/admin/controller/catalog_services.py
```diff
@@ -67,24 +73,65 @@ def process_post(self):
...
+    def _set_external_integration_link(self, service):
+        mirror_integration_id = flask.request.form.get('mirror_integration_id')
+        if mirror_integration_id == self.NO_MIRROR_INTEGRATION:
+            current_integration_link = get_one(
+                self._db,
+                ExternalIntegrationLink,
+                library_id=None,
+                external_integration_id=service.id,
+                purpose="MARC"
+            )
```
leonardr: This bit is repeated in _update_external_integration_link -- it should be refactored. Since this would leave _update_external_integration_link almost empty, you could maybe replace it with _get_external_integration_link and move the code that sets the new value into this method.

api/admin/controller/catalog_services.py
```diff
@@ -67,24 +73,65 @@ def process_post(self):
...
+        """
+        external_integration_link, ignore = get_one_or_create(
+            _db, ExternalIntegrationLink,
+            library_id=None,
+            external_integration_id=external_integration.id,
+            purpose="MARC"
+        )
+
+        external_integration_link.other_integration_id=other_external_integration.id
+        _db.commit()
```
leonardr: I don't think you need to commit here -- it's pretty rare. Try it without the `commit` and see if it works.

scripts.py
```diff
@@ -1381,22 +1402,26 @@ def load_collection(self, collection_name, data_source_name):
...
                 )
             except CannotLoadConfiguration, e:
-                # There is no sitewide mirror configuration, or else
-                # there is more than one. Either way, we can't
-                # associate a mirror integration with the new collection.
+                # We can't associate any mirror integration with the new collection.
                 mirror_integration = None
 
             if mirror_integration:
                 collection.mirror_integration = mirror_integration
```
EdwinGuzman: I don't believe this is needed anymore and can be removed, but should the `mirror_integration_id` in the Collections model be removed since now there can be multiple integrations?

api/admin/controller/collection_settings.py
```diff
@@ -24,10 +25,11 @@ def _get_collection_protocols(self):
         protocols = super(CollectionSettingsController, self)._get_collection_protocols(self.PROVIDER_APIS)
         # If there are storage integrations, add a mirror integration
         # setting to every protocol's 'settings' block.
-        mirror_integration_setting = self._mirror_integration_setting()
-        if mirror_integration_setting:
+        mirror_type = ['Covers', 'Books']
```
leonardr: These should be class constants so that it's easy to see where the configuration UI comes from. This is even more important if you end up sending the label and description into `_mirror_integration_setting`.

api/admin/controller/collection_settings.py
```diff
@@ -93,15 +95,25 @@ def load_settings(self, protocol_settings, collection_object, collection_setting
...
-            key = protocol_setting.get("key")
-            if not collection_settings or key not in collection_settings:
-                if key == 'mirror_integration_id':
-                    value = collection_object.mirror_integration_id or self.NO_MIRROR_INTEGRATION
-                elif protocol_setting.get("type") == "list":
-                    value = collection_object.external_integration.setting(key).json_value
-                else:
-                    value = collection_object.external_integration.setting(key).value
-                settings[key] = value
+            if protocol_setting:
```
leonardr: This isn't wrong, but you can make a diff like this a lot smaller and easier to read by doing it this way:

```
if not protocol_setting:
    continue
[all the other stuff]
```

That way you avoid adding an extra indent.

scripts.py
leonardr: Yes, we can and should remove `collections.mirror_integration_id`. We'll need a migration script that creates two ExternalIntegrationLinks for every collection with `mirror_integration_id` set -- one to say that book covers use the S3 integration and another one to say that books use the same S3 integration.

api/admin/controller/__init__.py
leonardr: In fact, it might be smart to put all the UI configuration in ExternalIntegrationLink, since thats' where we keep the constants COVERS and BOOKS.

scripts.py
```diff
@@ -1260,6 +1278,11 @@ def arg_parser(cls, _db):
             help=u'Titles will be imported into a collection with this name. The collection will be created if it does not already exist.',
             required=True
         )
+        parser.add_argument(
+            '--storage-name',
+            help=u'Storage service to use.',
+            required=True
+        )
```
leonardr: Since this script stores both books and covers, I don't think it makes sense to ask for a single storage integration.

Either we should take separate `--book-storage-name` and `--cover-storage-name` arguments, or we should expect everything to be set up beforehand (so no auto-creating a collection).

Since you have to set up the storage integrations ahead of time, I think the simplest thing is to also require that the collection be set up ahead of time. Basically take out this code and the ability to auto-create a collection. If a collection is specified and it's missing one or both storage integrations, then complain.

scripts.py
```diff
@@ -1381,22 +1406,26 @@ def load_collection(self, collection_name, data_source_name):
...
-                mirror_integration = MirrorUploader.sitewide_integration(
-                    self._db
+                mirror_integration = MirrorUploader.integration_by_name(
+                    self._db, storage_name
                 )
             except CannotLoadConfiguration, e:
-                # There is no sitewide mirror configuration, or else
-                # there is more than one. Either way, we can't
-                # associate a mirror integration with the new collection.
+                # We can't associate any mirror integration with the new collection.
```
leonardr: Yeah, looking at this I don't think it makes sense anymore. The two calls to `MirrorUploader.for_collection` below are more what we need.

scripts.py
```diff
@@ -1381,22 +1406,26 @@ def load_collection(self, collection_name, data_source_name):
...
+                    collection.name,
+                    storage_name
                 )
-        mirror = MirrorUploader.for_collection(collection)
+        mirror = dict(
+            covers=MirrorUploader.for_collection(
+                collection, ExternalIntegrationLink.COVERS),
+            books=MirrorUploader.for_collection(
+                collection, ExternalIntegrationLink.BOOKS)
+        )
```
leonardr: This is good, but after looking at the above code I think it doesn't go far enough. We can't complete the job of this script if there's nowhere to put the books, and if there's nowhere to put the covers we can only complete the job if it turns out there are no images. This would be a good time to raise an exception either MirrorUploader ends up None.

leonardr:
Just to reiterate, I think both these branches are good to go.

EdwinGuzman:
Yes, thanks. The Travis branch build kept building fine, but the "PR branch" kept breaking. I updated the unit test so now I'll begin to merge.

PR #1337, opened 2019-10-11
Prohibit multiple discovery mechanisms with the same URL

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-2342. It introduces a new data validation method, `SettingsController.check_url_unique`, which returns a problem detail document if the user is trying to create a situation where there are multiple ExternalIntegrations with the same protocol, goal, and URL. It also catches cases where the URLs are very slightly different (e.g. HTTP vs. HTTPS versions of the same URL).

I call this method in discovery_services.py to put a stop to situations (which I now have three reports of) where the SimplyE library registry is registered multiple times. This can cause a lot of problems including seemingly "duplicate" libraries showing up in SimplyE.

I initially tried to add this to `metadata_services.py` as well, but the point is moot with metadata integrations because you can only have one metadata integration, period.

It might make sense to call this data validation method from other controllers, to stop other kinds of duplicate services from being created, but I think you'd have a better feel for that than I do.

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -1660,6 +1660,10 @@ class SettingsController(AdminCirculationManagerController):
         differently. Since this isn't an exact science, this doesn't
         need to catch all variant URLs, only the most common ones.
         """
+        if not url:
+            # None has no variants.
+            return
+
         # A URL is a 'variant' of itself.
         yield url
 
@@ -1690,8 +1694,11 @@ class SettingsController(AdminCirculationManagerController):
         is up to you -- it's a good general rule but there are
         conceivable exceptions.
 
-        This method is used by discovery_services and metadata_services.
+        This method is used by metadata_services.
         """
+        if not url:
+            return
+
         # Look for the given URL as well as minor variations.
         #
         # We can't use urlparse to ignore minor differences in URLs
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -1660,8 +1660,8 @@ class SettingsController(AdminCirculationManagerController):
         differently. Since this isn't an exact science, this doesn't
         need to catch all variant URLs, only the most common ones.
         """
-        if not url:
-            # None has no variants.
+        if not Validator()._is_url(url, []):
+            # An invalid URL has no variants.
             return
 
         # A URL is a 'variant' of itself.
@@ -1694,7 +1694,7 @@ class SettingsController(AdminCirculationManagerController):
         is up to you -- it's a good general rule but there are
         conceivable exceptions.
 
-        This method is used by metadata_services.
+        This method is used by discovery_services.
         """
         if not url:
             return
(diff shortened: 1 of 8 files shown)


tests/admin/controller/test_discovery_services.py
```diff
@@ -45,28 +45,31 @@ def test_discovery_services_get_with_one_service(self):
         )
         discovery_service.url = self._str
 
+        controller = self.manager.admin_discovery_services_controller
+
```
leonardr: This is just to make the code shorter.

api/admin/controller/discovery_services.py
```diff
@@ -93,6 +93,12 @@ def process_post(self):
             self._db.rollback()
             return name_error
 
+        url = flask.request.form.get("url")
+        url_error = self.check_url_unique(service, url, protocol, self.goal)
+        if url_error:
```
adriana-alter: Maybe call this something other than `url_error`?  That sounds like the errors we throw when a URL isn't in a valid format.

api/admin/controller/__init__.py
```diff
@@ -1650,6 +1650,81 @@ def check_name_unique(self, new_service, name):
...
+    def check_url_unique(self, new_service, url, protocol, goal):
+        """Enforce a rule that a given circulation manager can only have
+        one integration that uses a given URL for a certain purpose.
+
+        Whether to enforce this rule for a given type of integration
+        is up to you -- it's a good general rule but there are
+        conceivable exceptions.
+
+        This method is used by metadata_services.
+        """
```
adriana-alter: The comment should say "used by **discovery**_services", no?

api/admin/controller/__init__.py
```diff
@@ -1650,6 +1650,81 @@ def check_name_unique(self, new_service, name):
             # changes to an existing service unless you've also changed its name.
             return INTEGRATION_NAME_ALREADY_IN_USE
 
+    @classmethod
+    def url_variants(cls, url, check_protocol_variant=True):
+        """Generate minor variants of a URL -- HTTP vs HTTPS, trailing slash
```
adriana-alter: It's okay for now because `discovery_services` calls `validate_formats` before calling this, but in general: it might be worth putting the url through `Validator._is_url` right at the top.  (If the URL isn't valid, then there's no point going ahead with generating variants on it.)

adriana-alter:
I don't think we need to call it from other controllers; for services where the limit is one and the admin has already created one, the "create new thing" button disappears from the UI.

Is something up with the metadata configuration, though?  It does let you create multiple metadata service configurations, which, if I'm reading your comment above right, shouldn't be the case.  (Only one of them takes a URL, though, so we still don't need to use this there.)

leonardr:
I misspoke about "metadata integrations"; I should have said "metadata _wrangler_ integrations". You can have multiple metadata integrations -- metadata wrangler, NYT best-sellers, NoveList -- but you can only have one "metadata wrangler" integration. It doesn't make sense to get metadata from two different wranglers, the way it might make sense to be registered with two different library registries.

PR #1340, opened 2019-10-22
Setting up a Local Analytics service if it doesn't exist when the app starts up.

EdwinGuzman:
Resolves [SIMPLY-2346](https://jira.nypl.org/browse/SIMPLY-2346).
Get one or create a Local Analytics service so one will always exist, even if it's the default one.

### DIFF
--- a/tests/admin/controller/test_analytics_services.py
+++ b/tests/admin/controller/test_analytics_services.py
@@ -21,10 +21,14 @@ from test_controller import SettingsControllerTest
 
 class TestAnalyticsServices(SettingsControllerTest):
 
-    def test_analytics_services_get_with_no_services(self):
+    def test_analytics_services_get_with_one_default_service(self):
         with self.request_context_with_admin("/"):
             response = self.manager.admin_analytics_services_controller.process_analytics_services()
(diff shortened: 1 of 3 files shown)


tests/admin/controller/test_analytics_services.py
```diff
@@ -204,24 +211,18 @@ def test_analytics_services_post_create(self):
...
+            protocol=GoogleAnalyticsProvider.__module__
+        )
         eq_(service.id, int(response.response[0]))
         eq_(GoogleAnalyticsProvider.__module__, service.protocol)
         eq_("http://test", service.url)
         eq_([library], service.libraries)
         eq_("trackingid", ConfigurationSetting.for_library_and_externalintegration(
                 self._db, GoogleAnalyticsProvider.TRACKING_ID, library, service).value)
 
-        # Creating a local analytics service doesn't require a URL.
```
leonardr: Since we already have these tests, it seems useful to keep them in (by deleting the local analytics service beforehand). That way if we ever go back to not _always_ having an analytics service, we'll still have the code coverage.

PR #1344, opened 2019-10-28
API Admin Routes test

EdwinGuzman:
Resolves [SIMPLY-1353](https://jira.nypl.org/browse/SIMPLY-1353).

Adding tests for `api/admin/routes.py`.

### DIFF
--- a/api/admin/routes.py
+++ b/api/admin/routes.py
@@ -23,7 +23,7 @@ from templates import (
     admin_sign_in_again as sign_in_again_template,
 )
 from api.routes import (
-    # has_library,
+    has_library,
     library_route,
     allows_library
 )
@@ -78,21 +78,6 @@ def requires_admin(f):
                 return admin
 
         return f(*args, **kwargs)
-    return decorated
-
-def has_library(f):
-    """Decorator to extract the library short name from the arguments."""
-    @wraps(f)
-    def decorated(*args, **kwargs):
-        if 'library_short_name' in kwargs:
-            library_short_name = kwargs.pop("library_short_name")
-        else:
-            library_short_name = None
-        library = app.manager.index_controller.library_for_request(library_short_name)
-        if isinstance(library, ProblemDetail):
-            return library.response
-        else:
-            return f(*args, **kwargs)
     return decorated
 
 def requires_csrf_token(f):
--- a/api/admin/routes.py
+++ b/api/admin/routes.py
@@ -321,10 +321,7 @@ def stats():
 @requires_admin
 @requires_csrf_token
 def libraries():
-    if flask.request.method == 'GET':
-        return app.manager.admin_library_settings_controller.process_get()
-    else:
-        return app.manager.admin_library_settings_controller.process_post()
+    return app.manager.admin_library_settings_controller.process_libraries()
 
 @app.route("/admin/library/<library_uuid>", methods=["DELETE"])
 @returns_json_or_response_or_problem_detail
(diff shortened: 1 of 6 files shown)


api/admin/routes.py
```diff
@@ -81,6 +80,21 @@ def decorated(*args, **kwargs):
         return f(*args, **kwargs)
     return decorated
 
+def has_library(f):
```
EdwinGuzman: I tried to reuse `api/routes.py`'s `has_library` decorator but it doesn't seem to work. I am getting an `AttributeError: 'Flask' object has no attribute 'manager'` error and it seems to be because on line 91, `app` is coming from `api/app` but in `api/routes.py` it is coming from the base. I'm getting stuck here because `app` is not being mocked as I expect it to, and as how the tests for the other api routes do mock `app` correctly.

api/admin/routes.py
```diff
@@ -367,10 +378,7 @@ def admin_auth_service(protocol):
 @requires_admin
 @requires_csrf_token
 def individual_admins():
-    if flask.request.method == 'GET':
-        return app.manager.admin_individual_admin_settings_controller.process_get()
-    else:
-        return app.manager.admin_individual_admin_settings_controller.process_post()
+    return app.manager.admin_individual_admin_settings_controller.process_individual_admins()
```
EdwinGuzman: I couldn't find an existing pattern or a way to update `flask.request.method` to be either GET or POST. It seems other tests don't care about it at this point and just test that the main function was called and that the appropriate methods are allowed. So, I switched it to the that pattern here.

tests/admin/test_routes.py
```diff
@@ -0,0 +1,794 @@
...
+
+class TestAdminStatic(AdminRouteTest):
+
+    CONTROLLER_NAME = "static_files"
+
+    def test_static_file(self):
+        url = "/admin/static/circulation-web.js"
+
+        local_path = os.path.abspath(os.path.dirname(__file__))
+        local_path = local_path.split('/tests')[0] + "/api/admin/node_modules/simplified-circulation-web/dist"
```
EdwinGuzman: Is there a better way I can mock this or is this fine?

EdwinGuzman:
@leonardr set this up as a draft because I have some questions.

api/admin/routes.py
leonardr: Yeah, putting the code inside the controller is better.

tests/admin/test_routes.py
```diff
@@ -0,0 +1,794 @@
...
+        http_method = kwargs.pop('http_method', 'GET')
+        response = self.request(url, http_method)
+
+        # A Flask template string is returned.
+        assert "You are now logged in" in response
+
+        # Either not being authenticated or not having a csrf token fails
+        # but setting both to false.
+        self.manager.admin_sign_in_controller.csrf_token = False
+        self.manager.admin_sign_in_controller.authenticated = False
```
leonardr: What's supposed to happen if one of these is set but not both?

tests/admin/test_routes.py
leonardr: Maybe something like this example from `test_controller.py`:

```
directory = os.path.join(os.path.abspath(os.path.dirname(__file__)), "..", "resources", "images")
```

Using ".." on the current directory is the key improvement I think.

api/admin/routes.py
leonardr: I think this is what's happening:

1. api/routes.py defines `app`.
2. api/admin/routes.py imports `app`, copying the object into its namespace.
3. Your test setup mocks `api/admin/routes.py:app` (i.e. changes which object it refers to) but leaves  `api/routes.py:app` alone.
4. Code in `api/admin/routes.py` calls decorator code in `api/routes.py` which references the un-mocked `api/routes.py:app`, causing the crash.

You should be able to fix this by mocking `api.routes.app` at the same time you mock `api.admin.routes.app`. They start the test as two references to the same real object, they should both be mocked (so they are references to the same mock object) and then they should both be unmocked (so they end up references to the same real object).

tests/admin/test_routes.py
EdwinGuzman: It's close but I'm not actually opening up a file so the `".."` are still there. Since it's doing a string comparison, it doesn't seem to help.

tests/admin/test_routes.py
EdwinGuzman: Hmm yes, I should add those cases.

api/admin/routes.py
EdwinGuzman: That was it. Thanks.

tests/admin/test_routes.py
leonardr: I see. Calling `os.path.abspath` on a path that contains ".." will give you a path without the "..". Does that help?

EdwinGuzman: Yes, it does, thanks.

PR #1347, opened 2019-10-31
Explain in English what CleverAuthenticationAPI.remote_patron_lookup does with the personal patron data it has available from Clever.

leonardr:
This is by Clever's request to have a plain-English explanation of this. For ease of verification I decided the best place to put this information was next to the code it describes.

Let me know if this looks right to you.

### DIFF
--- a/api/clever/__init__.py
+++ b/api/clever/__init__.py
@@ -194,14 +194,13 @@ class CleverAuthenticationAPI(OAuthAuthenticationProvider):
           PatronData class in authenticator.py and the Patron class in
           core/model/patron.py.
 
-          Upon login, the patron's personal name is sent via HTTP to the
-          authorized client, running on a device being operated by the
-          patron. (See the OAuthController.oauth_authentication_callback()
-          method for this.) The Open eBooks client displays the patron's
-          personal name to show who is logged in. This is important in
-          an environment where devices are shared across a
-          classroom. When the patron logs out, the Open eBooks client
-          destroys this information.
+          Upon login, the patron's personal name is sent via HTTP to
+          the authorized client, running on a device being operated by
+          the patron. (See the OAuthController.oauth_authentication_callback()
+          method for this.) Access to this information is important in an
+          environment where devices are shared across a classroom. When
+          the patron logs out, the Open eBooks client destroys this
+          information.
 
         * If the patron is a student, their grade level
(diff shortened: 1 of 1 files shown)


api/clever/__init__.py
```diff
@@ -171,9 +171,64 @@ def _remote_exchange_payload(self, _db, code):
...
+          personal information about them.
+
+        * The patron's personal name is passed out of this method
+          through the PatronData object. This information is available
+          for the duration of the active HTTP request (e.g. it could
+          be used to display a welcome message) but it cannot be
+          persisted to the database, because there's no place to store
+          it. You can verify this by looking at the PatronData class
+          in authenticator.py and the Patron class in
+          core/model/patron.py.
```
aslagle: Don't know if it needs to be mentioned here, but I believe the name is persisted by the mobile apps so they can show who is logged in.

leonardr: Hmm, I don't know how that would work. @kyles-ep Do you have access to the personal name of a patron who signed in to Open eBooks through Clever?

aslagle: The server-side part is https://github.com/NYPL-Simplified/circulation/blob/master/api/authenticator.py#L2466

leonardr: Hmm, OK, it's definitely possible then, and relevant. Once I hear from Kyle I'll update the text.

kyles-ep: It seems like there's code to fetch a patron's name from Clever login, yes. I haven't verified it works.

leonardr: I'm going to write "The Open eBooks client displays the patron's personal name to show who is logged in. This is important in an environment where devices are shared across a classroom." Let me know if it goes beyond that.

kyles-ep: There's a method to extract the full name but I dont see it used anywhere.
I don't think we even display the user's name to be honest.

kyles-ep: And yes, I believe we do wipe the user's name/info when we wipe the account details. So for example when we sign out.

leonardr: I bet that was a requirement at one point but we dropped the ball. We should either stop sending this information to the client or stop using it. I filed https://jira.nypl.org/browse/SIMPLY-2372 to cover this.

PR #1349, opened 2019-11-06
Stop exporting patron's personal name from Clever login process

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-2372. Open eBooks doesn't actually use the patron's personal name, and we can't make Open eBooks consistently greet patrons by name (since FirstBook doesn't know the patron's name), so I'm removing the code that sends it out, as a way of reducing the system's privacy footprint.

Theoretically, some other OAuth provider could still send the patron's name to the client via `oauth_callback`, and the client could use it to greet the patron by name. I just removed the code that does this for Clever.

This code also adds a more specific error message for a problem we're seeing, where we ask Clever for information on a school that definitely has an NCES ID, but Clever doesn't provide the NCES ID. Without this change, people who are eligible for Open eBooks are being told that they're not eligible.

### DIFF
--- a/api/clever/__init__.py
+++ b/api/clever/__init__.py
@@ -40,6 +40,13 @@ CLEVER_NOT_ELIGIBLE = pd(
     _("Your Clever account is not eligible to access this application."),
 )
 
+CLEVER_UNKNOWN_SCHOOL = pd(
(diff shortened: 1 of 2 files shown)


leonardr:
image

PR #1350, opened 2019-11-06
Derive OPDS For Distributors "supported media types" from list of media types renderable by SimplyE

leonardr:
This branch hopefully completes the work on https://jira.nypl.org/browse/SIMPLY-2370, in combination with https://github.com/NYPL-Simplified/server_core/pull/1134.

Previously, the only format you could import from OPDS For Distributors was EPUB. Now you can import any format that's listed alongside the BEARER_TOKEN access control scheme in DeliveryMechanism.default_client_can_fulfill. Thanks to 1134, this means EPUB, PDF, and Readium Audiobook Manifest. This also means that we don't have to remember to add to the OPDS For Distributors list every time we add a new rendering mechanism to SimplyE; we just have to add it to DeliveryMechanism.default_client_can_fulfill.

### DIFF
--- a/api/opds_for_distributors.py
+++ b/api/opds_for_distributors.py
@@ -60,10 +60,23 @@ class OPDSForDistributorsAPI(BaseCirculationAPI, HasSelfTests):
         }
     ]
 
-    SUPPORTED_MEDIA_TYPES = [Representation.EPUB_MEDIA_TYPE]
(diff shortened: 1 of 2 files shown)


leonardr:
No idea why this is failing; I'll look into it today.

leonardr:
Yeah, that fixed it.

PR #1353, opened 2019-11-08
Added distributor and medium to the Google Analytics output.

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-2340. It adds two more custom dimensions to the Google Analytics output: the distributor of the LicensePool (so we can compare usage of books we get from different sources) and the medium (so we can compare audiobooks vs ebooks).

This change will need a special release note explaining how to do the migration on the Google Analytics side; this is the subject of https://jira.nypl.org/browse/SIMPLY-2382.


leonardr:
These test failures are weird; I was getting the same ones on another branch but they went away when i updated core.

leonardr:
I believe both of these test failures were caused by implicit assumptions about the order in which objects came out of lists, which I have hopefully fixed. If not I'll pick it up on Monday.

PR #1355, opened 2019-11-14
Look at recent Enki circulation events using getRecentActivityTime endpoint instead of getRecentActivity

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-1221 and https://jira.nypl.org/browse/SIMPLY-1848. Instead of using the `getRecentActivity` endpoint to get all activity since the last time the EnkiImporter monitor ran (which can crash the Enki API), we split up the timeline since that time into two-hour chunks and ask the Enki API about one chunk at a time -- similar to what we do with Bibliotheca -- using the `getRecentActivityTime` endpoint.

### DIFF
--- a/api/bibliotheca.py
+++ b/api/bibliotheca.py
@@ -1330,22 +1330,6 @@ class BibliothecaEventMonitor(CollectionMonitor, TimelineMonitor):
             return default_start_time
         return None
 
-    def slice_timespan(self, start, cutoff, increment):
-        """Slice a span of time into segements no large than [increment].
-
-        This lets you divide up a task like "gather the entire
-        circulation history for a collection" into chunks of one day.
-        """
-        slice_start = start
-        while slice_start < cutoff:
-            full_slice = True
-            slice_cutoff = slice_start + increment
-            if slice_cutoff > cutoff:
-                slice_cutoff = cutoff
-                full_slice = False
-            yield slice_start, slice_cutoff, full_slice
-            slice_start = slice_start + increment
-
     def catch_up_from(self, start, cutoff, progress):
         added_books = 0
         i = 0
--- a/api/enki.py
+++ b/api/enki.py
@@ -206,7 +206,7 @@ class EnkiAPI(BaseCirculationAPI, HasSelfTests):
 
         # Look for the error indicator and raise
         # RemoteIntegrationException if it appears.
-        if self.ERROR_INDICATOR in response.content:
+        if response.content and self.ERROR_INDICATOR in response.content:
             raise RemoteIntegrationException(url, "An unknown error occured")
         return response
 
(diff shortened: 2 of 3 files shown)


api/bibliotheca.py
```diff
@@ -1330,22 +1330,6 @@ def create_default_start_time(self, _db, cli_date):
...
-        slice_start = start
-        while slice_start < cutoff:
-            full_slice = True
-            slice_cutoff = slice_start + increment
-            if slice_cutoff > cutoff:
-                slice_cutoff = cutoff
-                full_slice = False
-            yield slice_start, slice_cutoff, full_slice
-            slice_start = slice_start + increment
-
```
leonardr: This code moved to core.

api/enki.py
```diff
@@ -101,6 +102,12 @@ class EnkiAPI(BaseCirculationAPI, HasSelfTests):
         (epub, adobe_drm): 'acs',
     }
 
+    # Enki API serves all responses with a 200 error code and a
+    # text/html Content-Type. However, there's a string that
+    # reliably shows up in error pages which is unlikely to show up
+    # in normal API operation.
+    ERROR_INDICATOR = '<h1>Oops, an error occurred</h1>'
+
```
leonardr: This is the core of the fix to https://jira.nypl.org/browse/SIMPLY-1848. I'm not super happy with it, but it's okay.

api/enki.py
```diff
@@ -784,9 +803,31 @@ def incremental_import(self, since):
         return new_titles, circulation_changes
 
     def update_circulation(self, since):
-        """Process circulation events that happened since `since`."""
+        """Process circulation events that happened since `since`.
```
leonardr: The behavior of this method hasn't changed -- we still process all the events between `since` and the current time. But we now may be making multiple API calls.

api/enki.py
EdwinGuzman: Yeah, this sounds annoying to get all responses as 200.

tests/test_enki.py
```diff
@@ -740,6 +762,47 @@ def update_circulation(self, since):
...
+        # slice_timespan() sliced up the timeline into two-hour
+        # chunks. It yielded up two chunks: "three hours ago" to "one
+        # hour ago" and "one hour ago" to "now".
+        #
+        # _update_circulation() was called on each chunk, and in each
+        # case the return value was an item popped from monitor.sizes.
+        chunk1, chunk2 = monitor._update_circulation_called_with
+        eq_((three_hours_ago, one_hour_ago), chunk1)
+        eq_(one_hour_ago, chunk2[0])
+        assert (chunk2[1] - now).total_seconds() < 2
```
EdwinGuzman: Is this because it's hard to test that the "now" from the test is the current "now?

leonardr: It's impossible -- the two times will always be slightly off. So yeah, I just check that they're basically the same time.

PR #1361, opened 2019-11-22
Add self-test controller for metadata integrations

leonardr:
This branch is progress towards https://jira.nypl.org/browse/SIMPLY-2420. I created a new controller, `MetadataServiceSelfTestsController`, based on the `SearchServiceSelfTestsController`, which manages the self-tests for metadata integrations that implement `HasSelfTests`. Right now that's the NYT best-seller API and the metadata wrangler.

I'm not able to test this controller directly, but I refactored `SearchServiceSelfTestsController` to work almost the same way as `MetadataServiceSelfTestsController` does, and I tested that in a real scenario, so it should be pretty close to working.

In the future I'd like to see all (or almost all) self-test code moved into `SelfTestsController`. In particular, `SettingsController._get_prior_test_results` should be in `SelfTestsController` -- it's got a lot of custom code for different types of self-test results that I think we can get rid of once we give every `SelfTestsController` subclass a good implementation of `_find_protocol_class`.


api/admin/controller/__init__.py
```diff
@@ -1561,7 +1563,10 @@ def _get_prior_test_results(self, item, protocol_class=None):
                 self_test_results = ExternalSearchIndex.prior_test_results(
                     self._db, None, self._db, item
                 )
-
+            elif self.type == "metadata service":
+                self_test_results = protocol_class.prior_test_results(
+                    self._db, *extra_args
+                )
```
leonardr: I'm envisioning a future where `_get_prior_test_results` can just call `find_protocol_class` to get the arguments, and pass those arguments into `prior_test_results`. This little bit would become basically the whole method.

api/admin/controller/self_tests.py
```diff
@@ -8,29 +8,62 @@
...
     def process_get(self, identifier):
-        item = self.look_up_by_id(identifier)
-        if isinstance(item, ProblemDetail):
-            return item
-        info = self.get_info(item)
-        protocol_class = None
-        if hasattr(self, "_find_protocol_class"):
-            protocol_class = self._find_protocol_class(item)
-        info["self_test_results"] = self._get_prior_test_results(item, protocol_class)
+        integration = self.look_up_by_id(identifier)
```
leonardr: I know this isn't always going to be an `ExternalIntegration`, but the case where it's something else (a Collection) actually turns out to be the weirdo case. It's just that that's the case we implemented first.

tests/admin/controller/test_metadata_service_self_tests.py
```diff
@@ -0,0 +1,95 @@
...
+    ExternalIntegration,
+)
+from core.opds_import import MetadataWranglerOPDSLookup
+
+from api.admin.problem_details import *
+from api.nyt import NYTBestSellerAPI
+
+class TestMetadataServiceSelfTests(SettingsControllerTest):
+
+    def test_metadata_service_self_tests_with_no_identifier(self):
```
leonardr: I copied these tests from `test_search_service_self_tests.py`. All these tests for the different types of self-test providers take a long time to run, and one great side effect of just testing `SelfTestsController` is almost all of them will go away.

PR #1362, opened 2019-11-22
Add library name as an analytics dimension

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-514 by adding the library's short name as a Google Analytics dimension. When you have a whole bunch of libraries sharing a single GA account, this makes it possible to group analytics events by which library they happened to.

This is based off of my earlier branch, https://github.com/NYPL-Simplified/circulation/pull/1353/files. I noticed a couple more tests where I should have changed that branch to look at `cd13` and `cd14`. I added the missing tests here along with tests that look at `cd15`. Note that `cd15` is pretty much guaranteed to be present, even if lots of other analytics dimensions are missing.


killerpuppytails:
💃

PR #1363, opened 2019-12-04
Add test of URNLookupController.

leonardr:
This is the circulation portion of https://jira.nypl.org/browse/SIMPLY-2432.
It's basically a submodule update plus a test of a previously untested class, to make sure we didn't break anything.


EdwinGuzman:
I think this is what threw me off before when asking what test to write for the controller: https://github.com/NYPL-Simplified/circulation/blob/master/tests/test_controller.py#L4500

There is a test for `url_lookup.work_lookup` but it's under `TestURLLookupController` which is a typo. I think both can be combined.

leonardr:
I see! I've combined the two tests into a best-of.

PR #1365, opened 2019-12-10
No language lanes

EdwinGuzman:
Resolves [SIMPLY-458](https://jira.nypl.org/browse/SIMPLY-458) along with [PR in core](https://github.com/NYPL-Simplified/server_core/pull/1142).

When a language code is encountered with a name, creating lanes breaks. Instead of creating no lanes, skip over the bad language code and missing name.

### DIFF
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -50,7 +50,6 @@ from core.model import (
 
 from core.util import LanguageCodes
 from novelist import NoveListAPI
-from core.util.problem_detail import ProblemDetail
 
 def load_lanes(_db, library):
     """Return a WorkList that reflects the current lane structure of the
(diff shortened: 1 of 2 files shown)


api/lanes.py
```diff
@@ -699,6 +698,8 @@ def create_lane_for_small_collection(_db, library, parent, languages, priority=0
         genres=[],
     )
     language_identifier = LanguageCodes.name_for_languageset(languages)
+    if not language_identifier:
+        return 0
```
leonardr: In addition to my comment on the core branch, I'd suggest logging an error if this happens--otherwise it could be mysterious behavior.

EdwinGuzman:
Updated the PR to catch the ValueError and log a warning.

PR #1371, opened 2020-01-06
Updating OPDS url based on new audiences.

EdwinGuzman:
I don't think this is necessary to the [core PR](https://github.com/NYPL-Simplified/server_core/pull/1147), but I ran into this and thought it would be good to update it.

### DIFF
--- a/api/opds.py
+++ b/api/opds.py
@@ -719,10 +719,11 @@ class LibraryAnnotator(CirculationManagerAnnotator):
             audiences = [Classifier.AUDIENCE_CHILDREN]
         elif work.audience == Classifier.AUDIENCE_YOUNG_ADULT:
             audiences = Classifier.AUDIENCES_JUVENILE
+        elif work.audience == Classifier.AUDIENCE_ALL_AGES:
+            audiences = [Classifier.AUDIENCE_CHILDREN,
+                Classifier.AUDIENCE_ALL_AGES]
         elif work.audience in Classifier.AUDIENCES_ADULT:
             audiences = list(Classifier.AUDIENCES_NO_RESEARCH)
-        elif work.audience == Classifier.AUDIENCE_ALL_AGES:
-            audiences = [Classifier.AUDIENCE_ALL_AGES]
         elif work.audience == Classifier.AUDIENCE_RESEARCH:
             audiences = list(Classifier.AUDIENCES)
         else:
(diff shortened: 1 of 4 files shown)


api/opds.py
```diff
@@ -721,6 +721,10 @@ def language_and_audience_key_from_work(self, work):
             audiences = Classifier.AUDIENCES_JUVENILE
         elif work.audience in Classifier.AUDIENCES_ADULT:
             audiences = list(Classifier.AUDIENCES)
```
leonardr: Should this  be list(AUDIENCES_NO_RESEARCH)? I don't think we want research titles showing up in a "by this author" list that's not research-related. But maybe we do!

leonardr: We can always change it later.

api/opds.py
```diff
@@ -721,6 +721,10 @@ def language_and_audience_key_from_work(self, work):
...
+        elif work.audience == Classifier.AUDIENCE_ALL_AGES:
+            audiences = [Classifier.AUDIENCE_ALL_AGES]
+        elif work.audience == Classifier.AUDIENCE_RESEARCH:
+            audiences = [Classifier.AUDIENCE_RESEARCH]
```
leonardr: And this should probably be AUDIENCES, since if you're doing research you probably want to see everything this person wrote.

api/opds.py
```diff
@@ -721,6 +721,10 @@ def language_and_audience_key_from_work(self, work):
             audiences = Classifier.AUDIENCES_JUVENILE
         elif work.audience in Classifier.AUDIENCES_ADULT:
             audiences = list(Classifier.AUDIENCES)
+        elif work.audience == Classifier.AUDIENCE_ALL_AGES:
+            audiences = [Classifier.AUDIENCE_ALL_AGES]
```
leonardr: I think this should also include AUDIENCE_CHILDREN, but not AUDIENCE_ADULT.

api/opds.py
EdwinGuzman: `AUDIENCES_NO_RESEARCH` makes more sense then.

api/opds.py
EdwinGuzman: I don't understand why this that's the case for this.

leonardr: Here's my thinking:  we never want to show books that are a jump up in age level. In this case, if you are in the children's section and you look at an all-ages book by Conan Doyle, then ask for more books by the same author, we don't want to show his audience="Adult" books. But we do want to show his audience="Children" books, not just the other "All Ages" stuff.

This is the same logic behind showing all of the audiences when you're looking at works by this author for audience="Adult". It's okay to show an All Ages book in that context, and even a Children book, because there's no risk of showing age-inappropriate material. But not vice versa.

EdwinGuzman: Thanks, that helps.

PR #1374, opened 2020-01-16
Add a missing endpoint() call.

leonardr:
This branch fixes a pretty bad bug caused by the fix to https://jira.nypl.org/browse/SIMPLY-2486. Requests to Overdrive that act on behalf of a patron won't actually go through. I didn't catch this before because `patron_request` makes live HTTP requests and isn't tested.

I don't understand why this bug didn't show up when I ran QA. It looks like I released 3.0.3 without updating the submodule, which was its own mistake, but I wouldn't expect this to mean everything works.

At any rate, I'm going to release a 3.0.4 ASAP with this fix.

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -193,6 +193,7 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI, HasSelfTests):
                 method = 'post'
             else:
                 method = 'get'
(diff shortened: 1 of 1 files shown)


leonardr:
It actually calls HTTP.request_with_timeout -- not OverdriveAPI._do_get or some other method that gets mocked.

PR #1378, opened 2020-01-29
ODL fixes for Feedbooks audiobooks

leonardr:
This branch makes some tweaks to ODL import and fulfillment to support audiobooks protected with the Feedbooks access-control scheme, because the representation of such audiobooks has changed slightly, The behavior is unchanged for other formats or DRM schemes. These changes were validated on a live integration.

### DIFF
--- a/api/odl.py
+++ b/api/odl.py
@@ -862,10 +862,9 @@ class ODLImporter(OPDSImporter):
                     break
             if not odl_status_link:
                 # TODO: This is a hack necessary because Feedbooks
-                # doesn't provide the status link yet.
-                #
-                # It's also not clear whether it really makes sense
-                # for us to do the identifier.replace thing.
+                # doesn't provide the status link yet. Once that's fixed,
+                # this link will be present with rel="self", just like
+                # in the unit tests.
                 odl_status_link = "https://license.feedbooks.net/copy/status/?uuid=" + identifier.replace("urn:uuid:", "")
 
             if odl_status_link:
--- a/api/odl.py
+++ b/api/odl.py
@@ -802,7 +802,7 @@ class ODLImporter(OPDSImporter):
             # By default, dcterms:format includes the media type of a
             # DRM-free resource.
             content_type = full_content_type
-            drm_scheme = None
+            drm_schemes = []
 
             # But it may instead describe an audiobook protected with
             # the Feedbooks access-control scheme.
@@ -812,17 +812,29 @@ class ODLImporter(OPDSImporter):
             )
             if full_content_type == feedbooks_audio:
                 content_type = MediaTypes.AUDIOBOOK_MANIFEST_MEDIA_TYPE
-                drm_scheme = DeliveryMechanism.FEEDBOOKS_AUDIOBOOK_DRM
+                drm_schemes.append(DeliveryMechanism.FEEDBOOKS_AUDIOBOOK_DRM)
+
+            # Additional DRM schemes may be described in <odl:protection>
+            # tags.
+            protection_tags = parser._xpath(
+                odl_license_tag, 'odl:protection'
+            ) or []
+            for protection_tag in protection_tags:
+                drm_scheme = subtag(protection_tag, 'dcterms:format')
+                if drm_scheme:
+                    drm_schemes.append(drm_scheme)
+
+            for drm_scheme in (drm_schemes or [None]):
+                formats.append(
+                    FormatData(
+                        content_type=content_type,
+                        drm_scheme=drm_scheme,
+                        rights_uri=RightsStatus.IN_COPYRIGHT,
+                    )
+                )
 
             if content_type == MediaTypes.AUDIOBOOK_MANIFEST_MEDIA_TYPE:
                 data['medium'] = Edition.AUDIO_MEDIUM
-            formats.append(
-                FormatData(
-                    content_type=content_type,
-                    drm_scheme=drm_scheme,
-                    rights_uri=RightsStatus.IN_COPYRIGHT,
-                )
-            )
 
             expires = None
             remaining_checkouts = None
--- a/api/odl.py
+++ b/api/odl.py
@@ -799,13 +799,15 @@ class ODLImporter(OPDSImporter):
             content_type = subtag(odl_license_tag, 'dcterms:format')
             drm_schemes = []
             protection_tags = parser._xpath(odl_license_tag, 'odl:protection') or []
+            if protection_tags:
+                set_trace()
             for protection_tag in protection_tags:
                 drm_scheme = subtag(protection_tag, 'dcterms:format')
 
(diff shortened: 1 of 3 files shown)


api/odl.py
```diff
@@ -447,7 +448,13 @@ def _fulfill(self, loan):
...
         content_type = None
         for link in links:
-            if link.get("rel") == "license":
+            # Depending on the format being served, the crucial information
+            # may be in 'manifest' or in 'license'.
+            #
+            # TODO: when both links are present, access to the
+            # DeliveryMechanism would be great for figuring out which
+            # one to use.
+            if link.get("rel") in ("manifest", "license"):
```
leonardr: ACSM files use rel="license"; audiobook manifests use rel="manifest".

api/odl.py
```diff
@@ -796,28 +803,44 @@ def _detail_for_elementtree_entry(cls, parser, entry_tag, feed_url=None, do_get=
...
+
+            # But it may instead describe an audiobook protected with
+            # the Feedbooks access-control scheme.
+            feedbooks_audio = "%s; protection=%s"  % (
+                MediaTypes.AUDIOBOOK_MANIFEST_MEDIA_TYPE,
+                DeliveryMechanism.FEEDBOOKS_AUDIOBOOK_DRM
+            )
+            if full_content_type == feedbooks_audio:
+                content_type = MediaTypes.AUDIOBOOK_MANIFEST_MEDIA_TYPE
+                drm_schemes.append(DeliveryMechanism.FEEDBOOKS_AUDIOBOOK_DRM)
```
leonardr: Audiobooks with the Feedbooks access-control scheme no longer have an <odl:protection> tag. All the necessary information is in the <dcterms:format> tag, thanks to the "protection" MIME type parameter.

api/odl.py
```diff
@@ -796,28 +803,44 @@ def _detail_for_elementtree_entry(cls, parser, entry_tag, feed_url=None, do_get=
...
+                    drm_schemes.append(drm_scheme)
+
+            for drm_scheme in (drm_schemes or [None]):
+                formats.append(
+                    FormatData(
+                        content_type=content_type,
+                        drm_scheme=drm_scheme,
+                        rights_uri=RightsStatus.IN_COPYRIGHT,
+                    )
+                )
```
leonardr: This is the same code as below, just refactored to remove the `if not drm_schemes` path.

api/odl.py
```diff
@@ -796,28 +803,44 @@ def _detail_for_elementtree_entry(cls, parser, entry_tag, feed_url=None, do_get=
...
-                    rights_uri=RightsStatus.IN_COPYRIGHT,
-                ))
-            for drm_scheme in drm_schemes:
-                formats.append(FormatData(
-                    content_type=content_type,
-                    drm_scheme=drm_scheme,
-                    rights_uri=RightsStatus.IN_COPYRIGHT,
-                ))
+            if content_type == MediaTypes.AUDIOBOOK_MANIFEST_MEDIA_TYPE:
+                data['medium'] = Edition.AUDIO_MEDIUM
```
leonardr: Without this, audiobooks of whatever type don't get Edition.medium="Audiobook"

api/odl.py
```diff
@@ -837,12 +860,19 @@ def _detail_for_elementtree_entry(cls, parser, entry_tag, feed_url=None, do_get=
                 if rel == "self":
                     odl_status_link = link_tag.attrib.get("href")
                     break
+            if not odl_status_link:
+                # TODO: This is a hack necessary because Feedbooks
+                # doesn't provide the status link yet. Once that's fixed,
+                # this link will be present with rel="self", just like
+                # in the unit tests.
+                odl_status_link = "https://license.feedbooks.net/copy/status/?uuid=" + identifier.replace("urn:uuid:", "")
```
leonardr: The need to do this `identifier.replace()` thing myself is the thing I dislike most about this hack, but hopefully we can get rid of it before release.

api/odl.py
```diff
@@ -837,12 +860,19 @@ def _detail_for_elementtree_entry(cls, parser, entry_tag, feed_url=None, do_get=
...
 
             if odl_status_link:
                 ignore, ignore, response = do_get(odl_status_link, headers={})
                 status = json.loads(response)
-                remaining_checkouts = status.get("checkouts", {}).get("left")
-                available_checkouts = status.get("checkouts", {}).get("available")
+                checkouts = status.get("checkouts", {})
+                remaining_checkouts = checkouts.get("left", 0)
```
leonardr: I figured defaulting these values to 0 was better than `None`.

tests/files/odl/feedbooks_bibliographic.atom
```diff
@@ -254,46 +254,54 @@
   <category term="florida everglades" label="florida everglades"/>
   <category term="united states department of the interior" label="united states department of the interior"/>
 </entry>
-<entry>
```
leonardr: I replaced this entire OPDS entry (Feedbooks-protected audiobook) with one from a real feed.

api/odl.py
aslagle: Hmm, I think if the book has unlimited checkouts, those values won't be in the status document and it's supposed to be `None`.

tests/files/odl/feedbooks_bibliographic.atom
```diff
@@ -254,46 +254,54 @@
...
+
+&lt;p&gt;But after a group of rogue dragons, the Raptors, tried to take over Earth, all dragons were banished to another realm.&lt;/p&gt;
+
+&lt;p&gt;Most humans forgot about the dragons, claiming they never existed. Eleven-year-old Sirin knows the truth -- she grew up with stories passed down through the generations. However, when her mother falls ill, even Sirin has trouble believing in magic . . . until she sees a mysterious streak of silver in the night sky.&lt;/p&gt;
+
+&lt;p&gt;Sirin becomes the first child to "lock" with a dragon in centuries -- forming a deep friendship unlike anything she's ever imagined. But Sirin learns that not all dragons returned with good intentions, and soon she finds herself at the center of a battle between the dragons who want to protect the humans . . . and those who want to destroy them.&lt;/p&gt;</content>
+<odl:license>
+  <dcterms:identifier xsi:type="dcterms:URI">urn:uuid:551c7b6c-3e45-11ea-a2da-a26cdbaff453</dcterms:identifier>
```
aslagle: I think the real UUIDs should not be in github. I don't remember exactly why so you might want to check with Hadrien if it matters for this. Or maybe this is fake, but you made a more realistic fake one than I did :)

api/odl.py
leonardr: Good to know, thanks.

tests/files/odl/feedbooks_bibliographic.atom
HadrienGardeur: It's better to replace it with a random UUID indeed.

PR #1379, opened 2020-01-29
Small bug fixes for storage services.

EdwinGuzman:
This goes along with [core 1154](https://github.com/NYPL-Simplified/server_core/pull/1154), but also added minor fixes for issues that probably came up before but didn't seem too big.

These bugs came up when writing documentation for [SIMPLY-2522](https://jira.nypl.org/browse/SIMPLY-2522).

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -1564,12 +1564,9 @@ class SettingsController(AdminCirculationManagerController):
                     self._db, None, self._db, item
                 )
             elif self.type == "metadata service" and protocol_class:
-                try:
-                    self_test_results = protocol_class.prior_test_results(
-                        self._db, *extra_args
-                    )
-                except:
-                    self_test_results = None
+                self_test_results = protocol_class.prior_test_results(
+                    self._db, *extra_args
+                )
             elif self.type == "patron authentication service":
                 library = None
                 if len(item.libraries):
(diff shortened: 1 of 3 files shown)


api/admin/controller/__init__.py
```diff
@@ -1564,9 +1564,12 @@ def _get_prior_test_results(self, item, protocol_class=None, *extra_args):
                     self._db, None, self._db, item
                 )
             elif self.type == "metadata service" and protocol_class:
-                self_test_results = protocol_class.prior_test_results(
-                    self._db, *extra_args
-                )
+                try:
+                    self_test_results = protocol_class.prior_test_results(
```
EdwinGuzman: This caused an error for the NovelistAPI (and a different class which I am forgetting right now), but the error was that it did not have the `prior_test_results` function.

api/admin/controller/storage_services.py
```diff
@@ -5,7 +5,8 @@
 from flask_babel import lazy_gettext as _
 from api.admin.problem_details import *
 from core.model import (
-    ExternalIntegration
+    ExternalIntegration,
+    get_one
```
EdwinGuzman: I couldn't update the name of a Storage integration because it was missing this import.

api/admin/controller/catalog_services.py
```diff
@@ -37,6 +38,7 @@ def process_catalog_services(self):
 
     def process_get(self):
         services = self._get_integration_info(ExternalIntegration.CATALOG_GOAL, self.protocols)
+        self.update_protocol_settings()
```
EdwinGuzman: Added this here because the list of Storage services became stale and was never updated. So if a new Storage integration was added or if an existing Storage integration updated its name, it did not appear correctly in the dropdown menu.

PR #1382, opened 2020-02-04
Eliminate the status link hack.

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-2533 and updates the sample ODL to be in sync with what the Feedbooks ODL server now serves. Specifically, the media type of the License Info Document has been finalized, and we now check for a link of that type, not just any link with rel="self".

The only changes to the tests are changes in the data, which trigger the modified code paths. The audiobook ("Rise of the Dragons, Part 1") gets a brand new License Info Document link. Previously, it didn't have that link, and relied on the now-removed hack. The other books have the media type of their License Info Document links corrected; otherwise the new code will think they're of the wrong type and ignore them.

### DIFF
--- a/api/odl.py
+++ b/api/odl.py
@@ -863,8 +863,9 @@ class ODLImporter(OPDSImporter):
             for link_tag in parser._xpath(odl_license_tag, 'atom:link') or []:
                 attrib = link_tag.attrib
                 rel = attrib.get("rel")
-                type = attrib.get("type")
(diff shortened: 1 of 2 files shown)


aslagle:
My only concern would be backwards compatibility if the ODL server has changed, but if we weren't checking the type at all before that code should still work. However, going forward they'll have to make sure the type doesn't change again or this code would break.

leonardr:
I don't think the media type name will change, since it went from having a version number to not having one (rather than the version number being incremented). To add some forward-compatibility I changed the code so that it will work if parameters are added onto the media type in the future.

PR #1386, opened 2020-02-07
Initial circulation support for Overdrive audio manifests

leonardr:
This is the circulation branch for https://jira.nypl.org/browse/SIMPLY-2546.

This branch allows the `fulfill` method of a `CirculationAPI` to return a special kind of `FulfillmentInfo` that can be translated directly into a Flask `Response` object. If that happens, the `FulfillmentInfo` is converted into a `Response` and returned directly, bypassing the normal fulfillment process, which can get quite complicated and may not be relevant to a given situation.

Second, this branch introduces the `OverdriveManifestFulfillmentInfo` class, which takes advantage of this new feature. This class can be converted to a `Response` that simply does a redirect to the location (obtained through the Overdrive API) of an application/vnd.overdrive.circulation.api+json file. 

In all other cases, a circulation manager that's asked for a manifest file is able to grab the manifest itself and pass it on to the client. (this is why the rest of the fulfillment process is so complicated). Here, the circulation manager is not authorized to access the manifest; all it can do is pass along the link. So there's no need to go through the whole fulfillment process.

The only other thing `OverdriveManifestFulfillmentInfo` does is pass along the "scope string" for the active library and collection's Overdrive integration. SimplyE needs this information to set up its own authentication with the Overdrive API and download the manifest file. This is passed in the X-Overdrive-Scope response header.

Finally, this branch changes the `OverdriveAPI` (the `CirculationAPI` implementation for Overdrive), defining special behavior for the "audiobook-overdrive-manifest" format. "audiobook-overdrive-manifest" is not a format recognized by Overdrive; it goes over the wire to the Overdrive API as a request for the "audiobook-overdrive" format. However, the request is slightly different -- we ask for a manifest file instead of a link that we can use to get an Overdrive Listen link. And the response handling is quite different -- we return an `OverdriveManifestFulfillmentInfo` containing the manifest link, instead of a `FulfillmentInfo` containing an Overdrive Listen link.

This lets us distinguish between "give me a link to a downloadable manifest file" and "give me a link to an Overdrive Read web application", even though Overdrive uses "audiobook-overdrive" to refer to both types of links.


api/controller.py
```diff
@@ -1354,6 +1354,12 @@ def fulfill_part_url(part):
                 unicode(e), status_code=e.status_code
             )
 
+        # A subclass of FulfillmentInfo may want to bypass the whole
+        # response creation process.
+        response = fulfillment.as_response
+        if response:
+            return response
+
```
leonardr: Everything below this point assumes that the circ manager can go from a manifest or ACSM link to the manifest or ACSM itself. For Overdrive audio, that's not true.

api/overdrive.py
```diff
@@ -106,19 +108,41 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI, HasSelfTests):
...
         (pdf, adobe_drm): 'ebook-pdf-adobe',
-        (streaming_text, overdrive_drm): 'ebook-overdrive',
+        (streaming_text, streaming_drm): 'ebook-overdrive',
+        (streaming_audio, streaming_drm): 'audiobook-overdrive',
+        (overdrive_audiobook_manifest, libby_drm): 'audiobook-overdrive-manifest'
     }
 
+    # Once you choose an Adobe format you're locked into it and can't
+    # use other formats.
+    LOCK_IN_FORMATS = [ x for x in BaseOverdriveAPI.FORMATS if 'adobe' in x ]
```
leonardr: Previously, we had a rule that once you chose a format, you were locked in, _unless _that was one of the the 'streaming' formats (e.g. Overdrive Read). But 'lock in' appears to be a feature specific to Adobe's ACS. So I flipped the logic: now only the formats with "adobe" in the name lock you in.

api/overdrive.py
```diff
@@ -528,19 +570,28 @@ def get_fulfillment_link(self, patron, pin, overdrive_id, format_type):
...
 
         if download_link:
+            if format_type in self.MANIFEST_INTERNAL_FORMATS:
+                # The client must authenticate using its own
+                # credentials to fulfill this URL; we can't do it.
+                scope_string = self.scope_string(patron.library)
+                return OverdriveManifestFulfillmentInfo(
+                    self.collection, download_link,
+                    overdrive_id, scope_string
+                )
```
leonardr: Here's where we create an `OverdriveManifestFulfillmentInfo`, short-circuiting the rest of the fulfillment process.

api/overdrive.py
```diff
@@ -1008,13 +1064,31 @@ def update_licensepool_with_book_info(self, book, license_pool, is_new_pool):
...
+           and manifests.
+        :param error_url: Value to interpolate for the {errorpageurl}
+           URI template value.
+        """
         link = None
         format = None
         available_formats = []
+        if format_type in self.MANIFEST_INTERNAL_FORMATS:
+            use_format_type = format_type.replace("-manifest", "")
+            fetch_manifest = True
```
leonardr: This is where we convert from "overdrive-audiobook-manifest" (our internal name) to "overdrive-audiobook" (Overdrive's internal name).

leonardr:
Great, in the future feel free to grab me for an in-person talk if it will help.

kpo-overdrive:
After a couple more days, I'm feeling comfortable with my understanding of the PR (and more of the circulation and server_core projects).

**Positives:**
Overall, there are many positives in the code. Thanks for adding the support.
I like short circuiting the code when it's an audiobook-overdrive, but as a "manifest link". Hopefully, it's ok from an "architecure" standpoint.
I agree with building support ebook-overdrive-manifest too just in case (even though it's not being used). It could help prevent future confusion.

**Caveats with Questions:** My Flask inexperience may be showing, perhaps it's relying on the Android/iOS client, maybe it's hiding in another SimplyE project, or I could be blind. So, I apologize for my likely "Hi, I'm new" questions.

1. Does the (android) client pass the ``scope_string`` input? (``scope_string`` consists of ``website ID`` & ``ILS Name``)? Or is it something or somewhere else?
1A. How are these inputs (for ``scope_string``) validated?
(Feel free to point me to code in the circulation project or elsewhere in a different project.)

2. If the content type is different (mismatched, wrong, etc...), what happens?
Example: A (malicious user) passes an ``OverDrive Listen link`` but with the ``Manifest`` type (ie: ``application/vnd.overdrive.circulation.api+json`` or the ``LIBBY_DRM = u"Libby DRM"``)?
Example: A (malicious user) passes an ``"OverDrive Direct Download link"`` (manifest) but with the ``"OverDrive Listen"`` type (ie: ``application/x-od-media`` or the ``OVERDRIVE_DRM = u"Overdrive DRM"``)?
2A. Is it related to the TODO comment in the ``circulation project -> get_fulfillment_link()`` method:
```
# TODO: Verify that the asked-for format type is the same as the
# one in the loan.
```
	
3. In ```circulation project -> overdrive.py -> def make_direct_download_link(cls, link)```, I saw this comment:
```	"""Convert an Overdrive Read link template to a direct-download link for the manifest.```
I'm confused by the comment. Should this really be "OverDrive Listen link" to "direct download" instead of "OverDrive Read link"?
Or are you using "OverDrive Read" because of the "readauthurl" parameter which appends the additional "contentfile=true"?

Thanks in advance.

File: api/overdrive.py [hunk 5]
```diff
@@ -1030,25 +1104,50 @@ def get_download_link(self, checkout_response, format_type, error_url):
...
+            if fetch_manifest:
+                download_link = cls.make_direct_download_link(download_link)
+            return download_link
         else:
             return None
 
+    @classmethod
+    def make_direct_download_link(cls, link):
+        """Convert an Overdrive Read link template to a direct-download link
+        for the manifest.
```
kpo-overdrive: I'm confused by the comment. Should this really be "OverDrive Listen link" to "direct download" instead of "OverDrive Read link"?
Or are you using "OverDrive Read" because of the "readauthurl" parameter which appends the additional "contentfile=true"?
(This pertains to #3 in my "conversation".)

kpo-overdrive:
Leonard,

Thanks for your responses. I'll follow your lead of using inline responses.

> > ```
> > 1. Does the (android) client pass the `scope_string` input? (`scope_string` consists of `website ID` & `ILS Name`)? Or is it something or somewhere else?
> > ```
> 
> Yes, the mobile client uses `scope_string` to go through Overdrive's Patron Authentication process a _second_ time. The first time was on the server side. **The secret key** was the one associated with the library, and the result from the Overdrive API was the link to the manifest document.
> 
> **The second time, the secret key** is the one associated with the SimplyE mobile application, and the result from the Overdrive API is the manifest document itself.
> 
> The SimplyE mobile application has the patron's library card and PIN, but it doesn't know anything about a given library's Overdrive configuration. The server side _does_ know this information, so it's passed to the client at the same time as the link. We used a preassembled `scope_string`, instead of passing the individual pieces of information, to reduce the amount of code on the mobile side (because the middleware already knows how to make a scope string).

**Ken: ``secret key`` is mentioned for the 1st (server side) and 2nd (mobile client) time `scope_string` is used in OverDrive's Patron Authentication process. I bolded both references inline.
Is this Flask's ``secret key`` (1 unique key for each initialized project; so, 1 for server side and 1 for mobile client)? Or, is this a different secret key (with different initialization properties, inputs, and algorithm)?**

> 
> > ```
> >    1A. How are these inputs (for `scope_string`) validated?
> >    (Feel free to point me to code in the circulation project or elsewhere in a different project.)
> > ```
>
> The mobile client will send the `scope_string` directly to oauth-patron.overdrive.com, so it'll be whatever validation you have on that side. We never take `scope_string` as an input.

**Ken: Ok, this makes sense.**

> 
> > ```
> > 2. If the content type is different (mismatched, wrong, etc...), what happens?
> >    Example: A (malicious user) passes an `OverDrive Listen link` but with the `Manifest` type (ie: `application/vnd.overdrive.circulation.api+json` or the `LIBBY_DRM = u"Libby DRM"`)?
> >    Example: A (malicious user) passes an `"OverDrive Direct Download link"` (manifest) but with the `"OverDrive Listen"` type (ie: `application/x-od-media` or the `OVERDRIVE_DRM = u"Overdrive DRM"`)?
> > ```
> 
> This is tough for me to answer because I'm not sure which party you envision being malicious, but I'll answer all the cases.
> 
> When the patron tries to fulfill a book, they ask for a specific delivery mechanism (media type + DRM scheme). It's possible for a patron to hack the circulation manager URL and ask for a delivery mechanism that a given vendor doesn't support, but this will give a 400 error once the server sees that the vendor never advertised that format as supported.
> 
> It's also possible for a vendor to advertise a delivery mechanism that is unsupported by SimplyE or makes no sense. For "unsupported", we do our best to record the fact of the matter, in case that delivery mechanism becomes supported later. For "makes no sense", the delivery mechanism will end up unrecorded because we can't understand what the vendor is saying.
> 
> A malicious piece of middleware can lie (or just be mistaken) about what it's serving to the client. If this happens the client is generally unable to render the content and gives an error message. I've seen this happen when the client was expecting an unencrypted EPUB but got an ACSM file. In this particular case, the middleware is just serving a link. Regardless of any hints the middleware passed along, the client will be fetching the link and making its own decisions.
> 
> And finally, the _vendor_ might serve something that violates the expectations previously conveyed by the middleware. In this case the client's behavior will be the same as if the middleware had served the content directly: an error message due to an unexpected content type.

**Ken: Ok. This makes much more sense. I think this is just my Flask inexperience showing and knowing where / how input enters, flows, and exits through the circulation project. I did not realize that middleware is another avenue, so I'm thankful that you thought of and listed those scenarios. (It explains why I only listed 2 cases, not 4 as you iterated.)
Yes, either malicious user or misconfigured client would fit the same purpose.**

> > ```
> >    2A. Is it related to the TODO comment in the `circulation project -> get_fulfillment_link()` method:
> > ```
> > 
> > 
> > ```
> > # TODO: Verify that the asked-for format type is the same as the
> > # one in the loan.
> > ```
> 
> Yes, this is related. There are cases where a patron can ask for a format that _is_ supported but get an error anyway. Basically, we have some vendors (including Overdrive) where you need to 'lock' a format in a DRM system before it can be delivered. If you lock in format A, and then ask for format B, you'll get an error.
> 
> The complication we found is that 'locking' is not a general feature of DRM systems; it seems to be solely a feature of Adobe ACS. If you 'lock in' ACS+EPUB, then ACS+PDF is closed to you. But 'Overdrive Read' is still open. So the 'lock in' idea needs to be revisited as a quirk of Adobe ACS, not a general feature of DRM systems.
> 
> The upshot of this is that if you 'lock in' ACS+EPUB and then try to fulfill ACS+PDF (this would only happen if you hacked the URL), the middleware will try to carry out the request and fail due to an error message from the vendor API. The TODO refers to changing the code so that the middleware is able to determine _ahead of time_ that this request will fail, so that there's no need to even try it.
> 
> Determining this ahead of time turns out to be more complicated than we thought (because locking in ACS+EPUB excludes some formats and not others). So for now we just pass along the patron's request to the vendor, and propagate the error message if it turns out we asked for something that's not allowed.
> 
> > ```
> > 3. In `circulation project -> overdrive.py -> def make_direct_download_link(cls, link)`, I saw this comment:
> >    `	"""Convert an Overdrive Read link template to a direct-download link for the manifest.`
> >    I'm confused by the comment. Should this really be `"OverDrive Listen link"` to `"direct download"` instead of `"OverDrive Read link"`?
> >    Or are you using `"OverDrive Read"` because of the `"readauthurl"` parameter which appends the additional `"contentfile=true"`?
> > ```
> 
> The method was designed to work both on Overdrive Read links (which include the "readauthurl" parameter) and on Overdrive Listen links (which don't). I've changed the docstring to clarify this.

**Ken: Thanks for adding the brief comment to clarify. It makes more sense now.**

leonardr:
> **Ken: `secret key` is mentioned for the 1st (server side) and 2nd (mobile client) time `scope_string` is used in OverDrive's Patron Authentication process. I bolded both references inline. Is this Flask's `secret key` (1 unique key for each initialized project; so, 1 for server side and 1 for mobile client)? Or, is this a different secret key (with different initialization properties, inputs, and algorithm)?**

In both cases "secret key" refers to a secret issued by Overdrive. The Patron Authentication documentation refers to this piece of data as the "clientSecret".

kpo-overdrive:
Leonard,

Your response makes sense.
I'm sorry to ask 1 more question (likely due to my Flask inexperience again).

I've been trying (ie: failing) to trace through (via source code) where the "secret key" (ie: ``clientKey`` and ``clientSecret``) are stored/loaded. Put another way, I'm trying to create a simple UML sequence diagram. That is the goal. 

I see the references in ``circulation project`` -> ``overdrive.py`` in the class initiation, which I think is correct for the client key and secret (ie: "secret key"):
```
    { "key": ExternalIntegration.USERNAME, "label": _("Client Key"), "required": True },
    { "key": ExternalIntegration.PASSWORD, "label": _("Client Secret"), "required": True }
```

1. Is the data stored in the credentials table (source: ``server_core`` -> ``credential.py``)? Can you show me the source code location?
2. Is there a corresponding SQL statement for how that DB table is created?
3. Can you show me the file location where it's loaded and trace it through to where the external configuration (for ``OverDriveAPI``) is loaded?
If I understand it correctly, circulation starts with ``app.py`` (ie: ``app = Flask(__name__)``). (Feel free to correct me if wrong.)

Thanks in advance.

leonardr:
> I've been trying (ie: failing) to trace through (via source code) where the "secret key" (ie: `clientKey` and `clientSecret`) are stored/loaded. Put another way, I'm trying to create a simple UML sequence diagram. That is the goal.
> 
> I see the references in `circulation project` -> `overdrive.py` in the class initiation, which I think is correct for the client key and secret (ie: "secret key"):
> 
> ```
>     { "key": ExternalIntegration.USERNAME, "label": _("Client Key"), "required": True },
>     { "key": ExternalIntegration.PASSWORD, "label": _("Client Secret"), "required": True }
> ```

Yes, this is configuration for the web-based administrative interface:

image

This is saying that the Client Key is stored in a `ConfigurationSetting` called 'username' and the Client Secret is stored in a `ConfigurationSetting` called 'password'. This will be important at the end when these values are retrieved from the `configurationsettings` database table.
 
>     1. Is the data stored in the credentials table (source: `server_core` -> `credential.py`)? Can you show me the source code location?

This data is stored in the`configurationsettings` table. The `ConfigurationSetting` data model class is in [model/configuration.py](https://github.com/NYPL-Simplified/server_core/blob/master/model/configuration.py#L491). 

We have some data model diagrams [on the Confluence site](https://confluence.nypl.org/display/SIM/Flow+Diagrams).

>     2. Is there a corresponding SQL statement for how that DB table is created? I saw a lot of SQL, but nothing seemed to match up. I think model is somewhere in `Credential.py`, but I'm not 100% sure.

We use [SQLAlchemy](https://www.sqlalchemy.org/) as an ORM, which means we only need to write SQL in migration scripts. SQLAlchemy maps between operations on `ConfigurationSetting` objects (including initialization) and SQL statements (including CREATE TABLE) against the `configurationsettings` table.

Here's the CREATE TABLE statement for `configurationsettings`:

```
CREATE TABLE public.configurationsettings (
    id integer NOT NULL,
    external_integration_id integer,
    library_id integer,
    key character varying,
    value character varying
);
```

`external_integration_id` is important here; it points to an `ExternalIntegration` object representing the Overdrive account itself. This is how one circulation manager can manage different Overdrive integrations for different libraries.

>     3. Can you show me the file location where it's loaded and trace it through to where the external configuration (for `OverDriveAPI`) is loaded?
>        If I understand it correctly, circulation starts with `app.py` (ie: `app = Flask(__name__)`). (Feel free to correct me if wrong.)

Yes, the Flask application `app` includes everything else. Each incoming HTTP request is routed to a function (in api/routes,py) which has access to `app` and calls some method on one of the members of `app`.

However, the Overdrive configuration is loaded before any requests come in. This happens in api/routes.py, in the function `initialize_circulation_manager`. (right at the top, below the imports).

This function is decorated with `@app.before_first_request`, meaning that it runs immediately. Its main purpose is to instantiate a `CirculationManager` object and store it as `app.manager`.

The `CirculationManager` class is defined in api/controller.py. For our purposes, the main thing the constructor does is call the `load_settings` method, defined later on in the same file.

`load_settings` iterates over all the libraries defined on the system, and calls `setup_circulation` on each one. The `setup_circulation` method instantiates a `CirculationAPI` object which is capable of carrying out any kind of circulation request (place hold, checkout, fulfill, etc.) against any collection that library has licensed from any vendor.

The `CirculationAPI` object is defined in api/circulation.py. Its constructor takes a library and iterates over each collection associated with that library. Each collection has a class instantiated for it based on the type of collection. For an Overdrive collection, an `OverdriveAPI` will be instantiated. For a Bibliotheca collection, a `BibliothecaAPI` will be instantiated. (This mapping is defined in `CirculationAPI.default_api_map`). All of these classes implement a common interface, `BaseCirculationAPI`, which knows how to do common operations (place hold, checkout, fulfill, etc).

Anyway, for an Overdrive collection the class instantiated is `OverdriveAPI`. That's defined in api/overdrive.py, as you saw. This class subclasses `BaseOverdriveAPI`, which is defined in server_core/overdrive.py, and the `OverdriveAPI` constructor just calls the `BaseOverdriveAPI` constructor, so let's go over there.

The `BaseOverdriveAPI` constructor pulls the two pieces of information out of the `ExternalIntegration` associated with the Overdrive account with these two lines of code:

```
        self.client_key = integration.username
        self.client_secret = integration.password
```

`username` and `password` are shortcuts for `ConfigurationSetting`s called 'username' and 'password'. We set up these shortcuts because most integrations have some kind of username and password, and this makes the code easier to understand. For an Overdrive integration, we decided to store the Client Key in 'username' and the Client Secret in 'password'

In server_core/model/configuration.py you can see the helper methods we defined to make it easy to say `integration.username`:

```
    @hybrid_property
    def username(self):
        return self.setting(self.USERNAME).value

    @hybrid_property
    def password(self):
        return self.setting(self.PASSWORD).value
```

They're just retrieving a `ConfigurationSetting` called 'username' or 'password', and then pulling out the value -- the value that was inserted into the table through the admin interface as configured by the code you originally found.

So, that's how the data gets out of the database and into the `OverdriveAPI` object. The Overdrive API client is stored alongside other collections' API clients in a `CirculationAPI` object for the library. That is stored alongside the `CirculationAPI` objects for other libraries in a single `CirculationManager` object. This is stored in the `Flask` object as `app.manager`.

When a request comes in, some function defined in `api/routes.py` will run and it will almost always call some method of `app.manager` or one of its sub-members. If that method needs to carry out some kind of transaction on behalf of a patron of a given library, it will tell the `CirculationAPI` for that library to carry it out. If the `CirculationAPI` determines that the transaction needs to go through Overdrive, it will trigger a request to the Overdrive API using that library's credentials.

leonardr:
BTW the `credentials` table is used to store, among other things, the bearer tokens obtained for patrons through Overdrive Patron Authentication. That's all handled in api/overdrive.py, `get_patron_credential` and `refresh_patron_access_token`.

The essential difference between `Credential` and `ConfigurationSetting` is that credentials are obtained automatically and can automatically expire. Configuration settings are set by administrators and don't expire.

kpo-overdrive:
**Summary: I'm approving both pull requests.**
Thanks for providing responses to these questions. I was able to trace through it, and everything makes sense.

Some ancillary notes to close out the conversation:
1. I feel bad for not realizing the data model diagrams existed. Thank you for pointing them out.
2. I saw the ``SQLAlchemy`` references, but I apologize for not having investigated further for how tables and such were created with the ORM. Thanks for sending them.
2A. After reading your reply, I found the ConfigurationSettings ORM table within ``configuration.py``
``class ConfigurationSetting(Base, HasFullTableCache):``
3. Thanks for differentiating between ``Credentials`` and ``ConfigurationSetting`` table.

leonardr:
Thanks for working with me on these branches. I hope the experience has been as good on your side as it has been for me.

I didn't mention the data model diagrams initially because I thought of this as a branch dealing with network code, not with the database. Had I known where you would go with your investigation I would have mentioned them earlier.

PR #1388, opened 2020-02-11
Check each dcterms:format tag we discover to see if we can find a medium

leonardr:
This branch builds on https://github.com/NYPL-Simplified/server_core/pull/1158 to fix https://jira.nypl.org/browse/SIMPLY-2555.

I thought this branch would be cleanup, but it's new functionality based on the core code. The code in core/opds_import.py looks for a single `<dcterms:format>` tag inside the `<entry>` tag, but in circulation/odl.py, that method is overridden, and we iterate over a number of `<odl:licence>` tag tags, each of which must have a `<dcterms:format>` tag. Fortunately, the `Edition.medium_from_media_type` method works equally well in both situations.

### DIFF
--- a/api/odl.py
+++ b/api/odl.py
@@ -805,9 +805,13 @@ class ODLImporter(OPDSImporter):
         licenses_owned = 0
         licenses_available = 0
         odl_license_tags = parser._xpath(entry_tag, 'odl:license') or []
+        medium = None
         for odl_license_tag in odl_license_tags:
             identifier = subtag(odl_license_tag, 'dcterms:identifier')
             full_content_type = subtag(odl_license_tag, 'dcterms:format')
+
+            if not medium:
+                medium = Edition.medium_from_media_type(full_content_type)
 
             # By default, dcterms:format includes the media type of a
             # DRM-free resource.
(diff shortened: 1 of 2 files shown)


tests/test_odl.py
```diff
@@ -1304,7 +1304,9 @@ def do_get(url, headers):
         # it extracts format information from 'odl:license' tags and creates
         # LicensePoolDeliveryMechanisms.
 
+
         # The importer created 6 editions, pools, and works.
+        eq_({}, failures)
```
leonardr: Without this line, if there's a problem in the import code, the test will fail because fewer than 6 editions were imported. It'll be impossible to see what the actual problem was. If the failure dictionary is non-empty, then the actual failure messages will be output when nosetests explains why the test failed.

api/odl.py
```diff
@@ -843,8 +847,7 @@ def _detail_for_elementtree_entry(cls, parser, entry_tag, feed_url=None, do_get=
                     )
                 )
 
-            if content_type == MediaTypes.AUDIOBOOK_MANIFEST_MEDIA_TYPE:
-                data['medium'] = Edition.AUDIO_MEDIUM
```
leonardr: This solved the problem for audiobooks, but not in general.

PR #1389, opened 2020-02-11
CachedFeed.fetch now handles everything relating to OPDS feed generation, using Credential as a model

leonardr:
This branch finishes the https://jira.nypl.org/browse/SIMPLY-2532 work started in https://github.com/NYPL-Simplified/server_core/pull/1159. 

* The controller method that generates a grouped feed for a lane now ends up calling `AcquisitionFeed.feed` instead of `AcquisitionFeed.groups` if the lane has no children. This lets us make the implementation of `AcquisitionFeed.groups` a whole lot simpler. Previously we relied on _that_ code to handle the case of a lane with no children, even though this controller is the only place someone might ask for a grouped feed of a lane with no children.
* `AcquisitionFeed.page` and `AcquisitionFeed.groups` take a `WorkList` argument, but the argument used to be called 'lane', implying that only a `Lane` could be passed into that method. I changed it to `worklist`. This is a bit of a dangerous change since we have a lot of mock `AcquisitionFeed.page` implementations in tests that could conceivably keep on accepting `lane` quite happily, masking the fact that somebody trying to call `AcquisitionFeed.page` is still passing in `lane=worklist`. But I think I got all of them. There's only one place where we really do mean `lane` (searches), and I left that one along and added a comment for it.
* The `force_refresh` argument was removed from `AcquisitionFeed.page`, replaced by setting `max_age=0`.
* `WorkList.CACHED_FEED_TYPE` and `Facets.CACHED_FEED_TYPE` are now very important for deciding how a CachedFeed should be filed in the database, so I added tests for all subclasses that set a specific value there. They were important before, but those values got passed into `AcquisitionFeed.page`, so it was possible to check those values by verifying they were passed in at the appropriate moment. Now that's all handled inside `AcquisitionFeed.page`, so it's important to check that stuff some other way.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -657,7 +657,7 @@ class CacheOPDSGroupFeedPerLane(CacheRepresentationPerLane):
     name = "Cache OPDS grouped feed for each lane"
 
     def should_process_lane(self, lane):
-        # OPDS group feeds are only generated for lanes that have sublanes.
+        # OPDS grouped feeds are only generated for lanes that have sublanes.
         if not lane.children:
             return False
         if self.max_depth is not None and lane.depth > self.max_depth:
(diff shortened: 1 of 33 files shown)


scripts.py
```diff
@@ -633,6 +633,10 @@ def pagination(self, lane):
         for pagenum in range(0, self.pages):
             yield page
             page = page.next_page
+            if not page:
+                # There aren't enough books to fill `self.pages`
+                # pages. Stop working.
+                break
```
leonardr: This is an unrelated problem I found when running `bin/cache_opds_lane_facets` to force out problems with the code. If you ask to cache more pages than there are books in the lane, it was crashing.

I don't think this is a big deal since cache_opds_lane_facets hasn't been run on a regular basis for quite a while; it's more efficient to generate paginated feeds as people request them.

PR #1393, opened 2020-02-20
Add ability to ignore SIP2 patron status block

jonathangreen:
In some cases the library may want to ignore the patron status block message sent from the ILS and let the patron borrow anyway. This adds an optional setting that allows the configuration of:
 - the current blocking behaviour
 - not set the block based on the patron status message. 

I would have liked to make this more configurable, I was envisioning a set of check boxes where you could toggle each patron block reason, but this would require some UI changes in circulation-web which I wasn't comfortable with. This setting meets the use case of the hosting client I'm working with at LYRASIS, so I'm hoping this is a good improvement for now. 

Tagging @leonardr

### DIFF
--- a/api/sip/__init__.py
+++ b/api/sip/__init__.py
@@ -145,9 +145,9 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
         self.client = client
         patron_status_block = integration.setting(self.PATRON_STATUS_BLOCK).json_value
         if patron_status_block is None or patron_status_block:
-            self.patron_status_fields_that_deny_borrowing = SIPClient.PATRON_STATUS_FIELDS_THAT_DENY_BORROWING_PRIVILEGES
+            self.fields_that_deny_borrowing = SIPClient.PATRON_STATUS_FIELDS_THAT_DENY_BORROWING_PRIVILEGES
         else:
-            self.patron_status_fields_that_deny_borrowing = []
+            self.fields_that_deny_borrowing = []
 
     def patron_information(self, username, password):
         try:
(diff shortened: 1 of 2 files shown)


api/sip/__init__.py
```diff
@@ -281,20 +296,23 @@ def info_to_patrondata(cls, info, validate_password=True):
...
-                )
-                if block_reason not in (PatronData.NO_VALUE,
-                                        PatronData.UNKNOWN_BLOCK):
-                    # Even if there are multiple problems with this
-                    # patron's account, we can now present a specific
-                    # error message. There's no need to look through
-                    # more fields.
-                    break
-        patrondata.block_reason = block_reason
+        if patron_blocks:
```
leonardr: This would be simpler if `patron_blocks` controlled a list of fields that was either `SIPClient.PATRON_STATUS_FIELDS_THAT_DENY_BORROWING_PRIVILEGES` or blank. You wouldn't need to indent this big section, and it would also show a path forward for a potential future expansion where the library administrator can basically customize their own list of "patron status fields that deny borrowing privileges".

leonardr: Also, at this point I think it makes sense to turn `info_to_patrondata` from a class method to an instance method, so you can access `self.patron_status_block` instead of having to pass it in.

A lot of these class methods are only that way to make testing easier, but let me know if it turns out there's a problem with this idea.

jonathangreen: I think these should be resolved now.

jonathangreen:
Code review feedback integrated.

PR #1395, opened 2020-02-26
Use the new, more sophisticated Representation caching system when sending requests to the NoveList API.

leonardr:
This branch should complete https://jira.nypl.org/browse/SIMPLY-2532. It also includes a couple very small unrelated fixes I made while investigating a couple other problems.

The NoveListAPI no longer has its own rules for checking the `representations` table for cached representations. Instead, it uses the standard `Representation.post` method, but passes in different versions of the URL for making the actual HTTP request, versus for caching the Representation in the database.

I ran my "five people simultaneously trying to get the same feed" script and got no more errors.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -714,8 +714,8 @@ class OPDSFeedController(CirculationManagerController):
         if isinstance(lane, ProblemDetail):
             return lane
 
-        if not lane.sublanes:
-            # This lane has no sublanes. Although we can technically
+        if not lane.children:
+            # This lane has no children. Although we can technically
             # create a grouped feed, it would be an unsatisfying
             # gateway to a paginated feed. We should just serve the
             # paginated feed.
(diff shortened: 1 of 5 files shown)


api/controller.py
```diff
@@ -714,8 +714,8 @@ def groups(self, lane_identifier, feed_class=AcquisitionFeed):
         if isinstance(lane, ProblemDetail):
             return lane
 
-        if not lane.sublanes:
-            # This lane has no sublanes. Although we can technically
+        if not lane.children:
+            # This lane has no children. Although we can technically
```
leonardr: I could not figure out how to test this, but this change crashes the main web page in a real application.

api/odilo.py
```diff
@@ -191,8 +191,8 @@ def record_info_to_metadata(cls, book, availability):
 
         title = book.get('title')
         subtitle = book.get('subtitle')
-        series = book.get('series')
-        series_position = book.get('seriesPosition')
+        series = book.get('series').strip() or None
+        series_position = book.get('seriesPosition').strip() or None
```
leonardr: This is totally random -- I was asked to investigate some Odilo problems, and the problems were much bigger than this -- to the point that there's currently no solution -- but these are some relatively simple problems with incoming data that I was able to solve.

PR #1404, opened 2020-03-11
Avoid a situation where Elasticsearch is told to look for audience=None

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-2644.

### DIFF
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -955,7 +955,7 @@ class SeriesLane(DynamicLane):
         self.series = series_name
         if parent:
             parent.append_child(self)
-            if isinstance(parent, WorkBasedLane):
+            if isinstance(parent, WorkBasedLane) and parent.source_audience:
                 # WorkBasedLane forces self.audiences to values
                 # compatible with the work in the WorkBasedLane, but
                 # that's not enough for us. We want to force
--- a/tests/test_lanes.py
+++ b/tests/test_lanes.py
@@ -642,6 +642,15 @@ class TestSeriesLane(LaneTest):
         eq_([work_based_lane.source_audience], child.audiences)
         eq_(work_based_lane.languages, child.languages)
 
+        # If for some reason there's no audience for the work used as
+        # a basis for the parent lane, the parent lane's audience
+        # filter is used as a basis for the child lane's audience filter.
+        work_based_lane.source_audience = None
+        child = SeriesLane(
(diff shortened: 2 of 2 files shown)


leonardr:
That's a good question that opens up a window into a lot of other questions, some of them very complicated.

Basically, I don't think it's necessary to chase this all the way down to Elasticsearch, because this system interacts with Elasticsearch via the `Filter` object, which is really well tested.

The class method `Filter.from_worklist` (core/external_search.py) pulls `audiences` from the WorkList object and sets `Filter.audiences` to the same value, by calling the `Filter` constructor. (This is tested in `TestFilter.test_from_worklist` and `TestFilter.test_constructor`, in core/tests/test_external_search.py).  So we know that whatever you stick in `WorkList.audiences` ends up in `Filter.audiences`. 

The problem we had is that sometimes a 'bad' value ended up in `WorkList.audiences. That's the bit that's fixed and tested by this branch. We know that a 'good' value is now used even in that weird situation. And from already-tested code we know that this value is propagated to `Filter.audiences`.

So now it's a matter of verifying that any 'good' value in `Filter.audiences` is properly handled by Elasticsearch. We do this in two ways, both in core/tests/test_external_search. We run tests against a live Elasticsearch instance in `TestExternalSearchWithWorks.test_query_works` (starting on line 816--it's a huge test). And we verify that a reasonable value for `Filter.audiences` becomes a reasonable-looking Elasticsearch document in `TestFilter.test_build` (line 3687 -- agian, a big test).

It's true that this style of testing (where we decompose the system into tiny bits and then test all the bits and how they fit together) can allow problems to slip through the cracks. Usually this happens because there was some hidden assumption in "how the bits fit together". So bolstering these fragmented tests with end-to-end tests is often useful.

I don't want to do that in this case because setting up a new `EndToEndSearchTest` is a huge pain and it slows down the test suite. The counterargument is, this is not the first bug we've seen in circulation/api/lanes.py because one of the lanes was doing something weird, and end-to-end tests might expose those bugs earlier. The counterargument to _that_ is that we have the same cognitive biases whether we're writing unit tests or end-to-end tests. I overlooked this 'weird' situation when I was writing unit tests and I would have overlooked it writing end-to-end tests.

EdwinGuzman:
Thanks for the very clear and in-depth explanation!

PR #1410, opened 2020-03-23
Enforce loan and hold limits before trying to borrow a book, not during

leonardr:
This branch addresses https://jira.nypl.org/browse/SIMPLY-2657 by moving the check for "is this patron about to exceed their loan or hold limit?" to a single method call near the beginning of `CirculationAPI.borrow`. Previously this code was mixed throughout `borrow` and it was difficult to know whether a given error should be reported as "loan limit exceeded" or "hold limit exceeded".

The big advantage of this branch is that the code is all in one place (easy to understand and test), and the rules are simpler. There are two disadvantages:

First, if the patron is at their loan or hold limit (but not both) we need to fetch up-to-date circulation information for the book before we can see whether or not to send an error. This hurts user-visible response time.

Second, there's a possible race condition. Let's say you're at your hold limit but not your loan limit. You try to borrow a book. We check with the vendor and it's available, so `enforce_limits` does nothing.

At this point, someone else jumps in and borrows the last available copy of the book. 

Then we try to borrow the book. It's not available, so we place a hold, exceeding the patron's hold limit. `enforce_limits` has already run, and it let you through, because back then the relevant question was whether you've exceeded your _loan_ limit, which you haven't.

The way around this would be to run `enforce_limits` again, at the point that a loan fails and we decide to go for a hold again. I don't have a strong opinion on this one way or another; a race condition where someone can get one more hold than they're supposed to doesn't seem like a serious problem.

It's quite possible there were race conditions before, and the only thing that's changed is, the code is now simple enough that I can see the one that remains.

There's more test infrastructure in this PR than usual because `borrow` isn't currently tested very well.

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -551,12 +551,6 @@ class CirculationAPI(object):
 
         # Okay, it's not an open-access book. This means we need to go
         # to an external service to get the book.
-        #
-        # This also means that our internal model of whether this book
-        # is currently on loan or on hold might be wrong. But it's good
-        # enough to see whether the patron _thinks_ they're going for a
-        # loan or a hold, and to enforce any limits on loans or holds.
-        title_seems_available = (licensepool.licenses_available > 0)
 
         api = self.api_for_license_pool(licensepool)
         if not api:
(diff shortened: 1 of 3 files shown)


api/circulation.py
```diff
@@ -601,76 +599,81 @@ def borrow(self, patron, pin, licensepool, delivery_mechanism,
...
+
+        # We try to check out the book even if we believe it's not
+        # available -- someone else may have checked it in since we
+        # last looked.
+        try:
+            loan_info = api.checkout(
+                patron, pin, licensepool, internal_format
+            )
 
-                if isinstance(loan_info, HoldInfo):
```
leonardr: For the most part, this code wasn't removed; it was dedented because the `if not at_loan_limit:` if statement was removed.

PR #1417, opened 2020-03-27
Send appropriate Cache-Control headers for all OPDS feeds

leonardr:
to https://github.com/NYPL-Simplified/server_core/pull/1170, circulation portion of https://jira.nypl.org/browse/SIMPLY-2633.

In this branch, all controllers that serve OPDS feeds expect to get OPDSFeedResponse or OPDSEntryResponse objects from the methods they call, rather than getting OPDS feeds/entries and converting them into Response objects. This change was necessary because the logic that generates the OPDS feed itself is also the bit that knows how long the feed should be cached -- an important piece of information for generating the response as a whole.

As far as the controllers themselves are concerned, this means only that methods which used to call `feed_response` and `entry_response` don't need to do that anymore. Whatever method they call now returns an appropriate Response.

Inside the methods called by the controllers, things are a little more complicated. Some of these methods are covered in the core branch, but I had to change the methods that are circulation-specific -- the bookshelf feed, entries for individual loans or holds, etc.

To make this easier, I unified a bunch of methods whose job was to create a response containing a single OPDS entry for a loan or hold. These became `LibraryLoanAndHoldAnnotator.single_item_feed` and `SharedCollectionLoanAndHoldAnnotator.single_item_feed`. These methods were previously tested only indirectly, through controller tests. These both call the same helper method (`CirculationManagerAnnotator._single_entry_response`), which gives me three methods to test instead of six.

### DIFF
--- a/api/admin/controller/work_editor.py
+++ b/api/admin/controller/work_editor.py
@@ -69,6 +69,8 @@ class WorkController(AdminCirculationManagerController):
         """Return an OPDS entry with detailed information for admins.
 
         This includes relevant links for editing the book.
+
+        :return: An OPDSEntryResponse
         """
         self.require_librarian(flask.request.library)
 
@@ -77,11 +79,11 @@ class WorkController(AdminCirculationManagerController):
             return work
 
         annotator = AdminAnnotator(self.circulation, flask.request.library)
-        # Don't cache these OPDS entries - they should update immediately
-        # in the admin interface when an admin makes a change.
-        return AcquisitionFeed.single_entry(
-            self._db, work, annotator, max_age=0
-        )
+
+        # single_entry returns an OPDSEntryResponse that will not be
+        # cached, which is perfect. We want the admin interface
+        # to update immediately when an admin makes a change.
+        return AcquisitionFeed.single_entry(self._db, work, annotator)
 
     def complaints(self, identifier_type, identifier):
         """Return detailed complaint information for admins."""
--- a/api/admin/controller/work_editor.py
+++ b/api/admin/controller/work_editor.py
@@ -14,8 +14,6 @@ from api.config import (
 from api.metadata_wrangler import MetadataWranglerCollectionRegistrar
 from api.admin.validator import Validator
 from core.app_server import (
-    entry_response,
-    feed_response,
     load_pagination_from_request,
 )
 from core.classifier import (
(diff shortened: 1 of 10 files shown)


api/opds.py
```diff
@@ -1381,60 +1422,60 @@ def active_loans_for(cls, circulation, patron, test_mode=False):
...
-            identifier=identifier.identifier,
-            library_short_name=loan.library.short_name,
-            _external=True
-        )
-        if not work:
-            return AcquisitionFeed(
-                db, "Active loan for unknown work", url, [], annotator)
-        return AcquisitionFeed.single_entry(db, work, annotator)
+    def single_item_feed(cls, circulation, item, fulfillment=None, test_mode=False,
+                         feed_class=AcquisitionFeed, **response_kwargs):
```
leonardr, line 1429: Previously, when generating a feed for a specific FulfillmentInfo, the FulfillmentInfo was passed in as `item`. Now the FulfillmentInfo is the optional `fulfillment` argument, and `item` is always a loan or hold. I think this makes more sense because it makes the signature to this method look more like the signature for `acquisition_links`, which is where the loan/hold/fulfillment end up. And in fact, in `test_acquisition_links`, we do test a case where both `loan` and `fulfillment` is being passed in, which previously never happened.

leonardr:
@aslagle Just a heads-up that this branch refactors some methods from SharedCollectionLoanAndHoldAnnotator into a single method. This  was previously tested only indirectly through controller tests; it now has its own unit test.

api/admin/controller/work_editor.py
```diff
@@ -81,8 +79,8 @@ def details(self, identifier_type, identifier):
         annotator = AdminAnnotator(self.circulation, flask.request.library)
         # Don't cache these OPDS entries - they should update immediately
         # in the admin interface when an admin makes a change.
-        return entry_response(
-            AcquisitionFeed.single_entry(self._db, work, annotator), cache_for=0,
+        return AcquisitionFeed.single_entry(
+            self._db, work, annotator, max_age=0
```
EdwinGuzman, line 83: Does this need `max_age=0`? I thought `single_entry` sets max_age to 0 by default.

tests/test_opds.py
```diff
@@ -1732,3 +1895,121 @@ def test_acquisition_links(self):
...
+            assert isinstance(result, MockAnnotator)
+
+            # Let's examine the MockAnnotator itself.
+            eq_(self.collection, result.collection)
+            eq_(test_mode, result.test_mode)
+
+            # Now let's see what we did with it after calling its
+            # constructor.
+
+            # First, we generated a URL to a controller controller for
```
EdwinGuzman, line 1937: Small typo - "controller" twice.

api/admin/controller/work_editor.py
leonardr: Yes, you're right.

EdwinGuzman:
This is fine but tests are failling.

leonardr:
Just hadn't pushed the latest version of core; it works now.

PR #1418, opened 2020-03-31
Make patron fine checking bulletproof.

leonardr:
This branch fixes https://jira.nypl.org/browse/SIMPLY-2676 by refactoring the code that checks whether a patron's fines are excessive, making it handle more different cases, and testing it much more thoroughly.

In particular, `Patron.fines` is stored as a string, but it was being compared against a number. This branch fixes that problem and makes sure there are no related problems.

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -341,7 +341,7 @@ class HoldInfo(CirculationInfo):
     :param end_date: When reserved book is expected to become available.
         Expected to be passed in date, not unicode format.
     :param hold_position:  Patron's place in the hold line. When not available,
-        default to be passed is None, which is equivalent to "first in line". 
+        default to be passed is None, which is equivalent to "first in line".
     """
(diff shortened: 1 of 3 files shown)


api/circulation.py
```diff
@@ -568,11 +569,6 @@ def borrow(self, patron, pin, licensepool, delivery_mechanism,
 
         internal_format = api.internal_format(delivery_mechanism)
 
-        if patron.fines:
-            max_fines = Configuration.max_outstanding_fines(patron.library)
-            if max_fines is not None and patron.fines > max_fines.amount:
-                raise OutstandingFines()
-
```
leonardr: This code is redundant with what happens inside PatronUtility.assert_borrowing_privileges, and it has the same bug that assert_borrowing_privileges used to have, so the best thing is to remove it.

PR #1428, opened 2020-05-01
Fix RBDigital catalog delta sync

tdilauro:
## JIRA Ticket

https://jira.nypl.org/browse/SIMPLY-2747

## What does this PR do?

This PR addresses an issue caused by changes to the RBDigital API.

Specifically, the `/v1/libraries/:libraryId/media/delta` endpoint, has been deprecated. It previously worked by accepting `begin` and `end` date query parameters, from which it would calculate and return the delta between those dates and return cataloging metadata for the items involved.

In the new version,
- the `begin` and `end` dates are selected from a list of snapshots from the `/v1/libraries/:libraryId/book-holdings/delta/available-dates` endpoint,
- those dates are then given as query parameters to the `/v1/libraries/:libraryId/book-holdings/delta` endpoint (the format of this response has changed, as well), and finally
- the ISBN for each added item is used as the `:token` for the `/v1/libraries/:libraryId/media/:token` endpoint to retrieve the catalog metadata in the correct format.

The PR also fixes a couple of bugs on apparently seldomly traversed paths.

## How should this be tested?

Run the `bin/rbdigital_collection_delta` script in QA, test, and/or dev environment and ensure that it completes successfully.

Run test suite, which has been updated with new tests, assertions, and content that reflect added and changed methods and new API endpoint responses.

## Who has tested?
@tdilauro - In two dev environments from dumps of different Circulation Manager instances.

### DIFF
--- a/api/rbdigital.py
+++ b/api/rbdigital.py
@@ -1288,6 +1288,63 @@ class RBDigitalAPI(BaseCirculationAPI, HasSelfTests):
         response = self.request(url, params=args, verbosity=verbosity)
         return response.json()
 
+    class _FuzzyBinarySearcher(object):
+        """
+        A fuzzy binary searcher sorts an array by its key, and then must either:
+          - find an exact match, if one exists; or
+          - return an "adjacent" index and the direction in which a match
+            would have been found, had one existed.
+        """
+        INDEXED_GREATER_THAN_MATCH = -1
+        INDEXED_EQUALS_MATCH = 0
+        INDEXED_LESS_THAN_MATCH = 1
+
+        def __init__(self, array, key=None):
+            """
+            Initialize the object with a sorted array.
+
+            :param array: An array
+            :param key: A function of one argument that is used to extract
+                a comparison key from each element in array and by which
+                the array is sorted. The default value is None (compare
+                value to complete array element).
+            """
+
+            self.key = key or (lambda e: e)
+            if not callable(self.key):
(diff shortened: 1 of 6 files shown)


tdilauro:
Pinging @leonardr

api/rbdigital.py
```diff
@@ -1316,6 +1288,106 @@ def get_delta(self, from_date=None, to_date=None, verbosity=None):
...
+        INDEXED_EQUALS_MATCH = 0
+        INDEXED_GREATER_THAN_MATCH = -1
+        INDEXED_LESS_THAN_MATCH = 1
+
+        url = "%s/libraries/%s/book-holdings/delta/available-dates" % (self.base_url, str(self.library_id))
+
+        response = self.request(url)
+        try:
+            snapshots = response.json()
+        except Exception as e:
```
leonardr, line 1319: I'm wary of any kind of blanket `except`; I think this should just be ValueError.

tdilauro: Yeah, that makes sense. I should add a test for that, too.

tdilauro:
I ended up refactoring the fuzzy search into a nested class, as it has some constants that I wanted to scope and it made for both cleaner setup and cleaner call sites. And I wrote some specific tests for it.

leonardr:
This looks good. Please file another ticket for optimizing API usage by converting `RBDigitalDeltaMonitor` to a `TimelineMonitor`.

PR #1430, opened 2020-05-08
All Overdrive formats except streaming and manifest-based formats must be locked in before use.

leonardr:
## Description

This branch fixes https://jira.nypl.org/browse/SIMPLY-2755.

## How Has This Been Tested?

Used a local circulation manager install against NYPL's QA database to check out an affected book

https://localhost:6500/NYNYPL/works/Overdrive%20ID/d246264b-1a10-44dd-b273-b96c676da3b7/borrow

and then fulfill its DRM-free offering.

https://localhost:6500/NYNYPL/works/248488/fulfill/1

## Checklist:

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -134,15 +134,19 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI, HasSelfTests):
         (overdrive_audiobook_manifest, libby_drm): 'audiobook-overdrive-manifest'
     }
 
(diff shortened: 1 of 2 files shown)


leonardr:
Yeah, exactly. I thought this was a requirement imposed by ACS similar to only being able to return a fulfilled book through ACS. But it's an Overdrive requirement: you can "lock in" DRM-free EPUB and then not be able to download the DRM-free PDF.

PR #1432, opened 2020-05-11
Fix marc record generator

tdilauro:
## Description

This PR fixes some issues related to MARC export:
- completes some changes started on [SIMPLY-2296](https://jira.nypl.org/browse/SIMPLY-2296) in [3.0.2cm](https://github.com/NYPL-Simplified/circulation/releases/tag/3.0.2):
  - switches from literal "MARC" to constant `ExternalIntegrationLink.MARC` for `ExternalIntegrationLink`'s purpose,
  - drops use of `MARC_Exporter.STORAGE_PROTOCOL`, and
  - fixes some tests; and
- fixes an error that occurs when attempting to lookup the storage integration for a MARC Export and renames a variable for clarity.
  - The query for the MARC Export integration's `ExternalIntegrationLink` was malformed, resulting in a None, rather than an `ExternalIntegrationLink`.

## Prerequisite

Complete fix depends on changes in [server_core PR #1176](https://github.com/NYPL-Simplified/server_core/pull/1176).

## Motivation and Context

See JIRA issue https://jira.nypl.org/browse/SIMPLY-2759

MARC record export/mirroring was not working.

## How Has This Been Tested?

Set a breakpoint at the `ExternalIntegrationLink` and stepped through the code. Once the error was corrected, stepping through revealed a correct `ExternalIntegrationLink`.

## Checklist:

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -795,21 +795,21 @@ class CacheMARCFiles(LaneSweeperScript):
         # integration.
         integration_link = get_one(
             self._db, ExternalIntegrationLink,
-            other_integration_id=exporter.integration.id,
+            external_integration_id=exporter.integration.id,
             purpose=ExternalIntegrationLink.MARC
         )
         # Then use the "other" integration value to find the storage integration.
-        integration = get_one(self._db, ExternalIntegration,
+        storage_integration = get_one(self._db, ExternalIntegration,
             id=integration_link.other_integration_id
         )
 
-        if not integration:
+        if not storage_integration:
             self.log.info("No storage External Integration was found.")
             return
 
         # First update the file with ALL the records.
         records = exporter.records(
-            lane, annotator, integration
+            lane, annotator, storage_integration
         )
 
         # Then create a new file with changes since the last update.
@@ -819,7 +819,7 @@ class CacheMARCFiles(LaneSweeperScript):
             start_time = last_update - timedelta(days=1)
 
             records = exporter.records(
-                lane, annotator, integration, start_time=start_time
+                lane, annotator, storage_integration, start_time=start_time
             )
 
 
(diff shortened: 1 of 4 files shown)


scripts.py
```diff
@@ -795,21 +795,21 @@ def process_lane(self, lane, exporter=None):
         # integration.
         integration_link = get_one(
             self._db, ExternalIntegrationLink,
-            other_integration_id=exporter.integration.id,
+            external_integration_id=exporter.integration.id,
```
leonardr, line 798: The fact that you changed this without needing to change a test tells me there's a problem in test_scripts.py:TestCacheMARCFiles.test_process_lane, I think the culprit is this code, which (if you look at it in context) is creating a link to itself:

```
integration_link = self._external_integration_link(
            integration=integration,
            other_integration=exporter.integration,
            purpose=ExternalIntegrationLink.MARC
        )
```
Can you change the test file so that it sets up a more realistic link from a MARC export integration to a storage integration? Then this code will be properly tested.

tdilauro: This test does not exercise the real `MARCExporter.records` and doesn't really use functionality of the `other_integration`, so the `other_integration` here just needs to be some `ExternalIntegration`. That said, making it something different certainly makes it more clear what _is_ being tested.

tests/test_scripts.py
```diff
@@ -611,7 +620,7 @@ def records(self, lane, annotator, mirror_integration, start_time=None):
 
         eq_(lane, exporter.called_with[0][0])
         assert isinstance(exporter.called_with[0][1], MARCLibraryAnnotator)
-        eq_(exporter.integration, exporter.called_with[0][2])
+        eq_(the_linked_integration, exporter.called_with[0][2])
```
leonardr, line 623: This is exactly what I was looking for -- verifying that `the_linked_integration` is passed in rather than `integration`.

PR #1434, opened 2020-05-14
Allow some OPDS feed controllers to turn off caching by user request

leonardr:
## Description

This branch makes it possible for a client to signal to the circulation manager that it wants a brand-new OPDS feed, not one based on a potentially out-of-date cache. You do this by sticking `max_age=0` onto the URL. This only works if the client is authenticated with the admin interface as an administrator; a random OPDS client can't use this functionality.

There are no links to feeds with this capability -- you have to know to add it -- but once you're in "no cache" mode, OPDS feeds will propagate "no cache mode", whenver possible, to the OPDS feeds they link to.

## Motivation and Context

This is the last part of the backend work for https://jira.nypl.org/browse/SIMPLY-2582. 

## How Has This Been Tested?

Reproduction steps are given in the JIRA ticket. I added test coverage for most of `CirculationManager.cdn_url_for`, which previously fell into the "too difficult and too small to bother testing" category.

## Checklist:

- [] All new and existing tests passed.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -297,9 +297,11 @@ class CirculationManager(object):
         """Generate a URL for a view that (probably) passes through a CDN.
 
         :param view: Name of the view.
-        :param facets: The faceting object used to generate the document that's calling
+        :param _facets: The faceting object used to generate the document that's calling
            this method. This may change which function is actually used to generate the
-           URL; in particular, it may disable a CDN that would otherwise be used.
+           URL; in particular, it may disable a CDN that would otherwise be used. This is
+           called _facets just in case there's ever a view that takes 'facets' as a real
+           keyword argument.
         :param args: Positional arguments to the view function.
         :param kwargs: Keyword arguments to the view function.
         """
@@ -765,7 +767,7 @@ class OPDSFeedController(CirculationManagerController):
 
         url = self.cdn_url_for(
             "acquisition_groups", lane_identifier=lane_identifier,
-            library_short_name=library.short_name, facets=facets
+            library_short_name=library.short_name, _facets=facets
         )
 
         annotator = self.manager.annotator(lane, facets)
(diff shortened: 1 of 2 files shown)


tdilauro:
Any concerns about DOS with this? It might be nice to have a way to configure this per-instance, in case of an attack.

EdwinGuzman:
This seems fine but some unrelated tests are failing.

leonardr:
I thought about DOS, and my initial conclusion was that there are already requests you can make of the circulation manager if you want to waste its time. However those requests require patron authentication. With this in place it's conceivable someone could cause a DOS by accident.

All we really need to meet the requirements is to allow the cache to be disabled by an authenticated _admin_ user. I'd also like to be able to disable the cache from a `curl` command when testing, but that's not as big a deal.

@EdwinGuzman do you have any thoughts on how to restrict certain behavior to admin users without restricting the entire controller method to admin users?

tdilauro:
> I thought about DOS, and my initial conclusion was that there are already requests you can make of the circulation manager if you want to waste its time. However those requests require patron authentication. With this in place it's conceivable someone could cause a DOS by accident.
> 
> All we really need to meet the requirements is to allow the cache to be disabled by an authenticated _admin_ user. I'd also like to be able to disable the cache from a `curl` command when testing, but that's not as big a deal.

Yeah, this seems like a winner. Only allowing an admin to do it seems a lot safer and reduces the chances of both accidental cache disabling and deliberate attacks. I can't think of any good reason for regular clients to trigger feed rebuilds, other than the usual case of missing or expired cached feed.

EdwinGuzman:
I haven't given it too much thought, but is it possible for admin requests to send what time of admin they are for the actions they want to do to?

leonardr:
The admins who will need this functionality are mainly librarians making custom lists, so I think it makes sense to restrict this functionality to any admin at all.

EdwinGuzman:
The update looks good but it seems the PR test build failed to complete.

tdilauro:
> The update looks good but it seems the PR test build failed to complete.

I've been having trouble with this, too. I mentioned it to Leonard last week. I closed and re-opened the PR to get Travis to kick off again.

EdwinGuzman:
@tdilauro I usually restart the build in Travis or add an empty commit `git ci -m "Travis build" --allow-empty` but didn't consider closing/reopening the PR.

leonardr:
Yeah, I always restart in Travis. There are a couple intermittent problems we haven't been able to fix and this branch got hit by both of them.

PR #1436, opened 2020-05-19
Add validation rules for announcement lists

leonardr:
## Description

This branch adds validation rules for announcement lists [per the spec](https://docs.google.com/document/d/1ngvqr7oN-VHMdnygDD9uAeVpa45LYq-JUTT5zY5-gJ4/edit). The validation rules are not yet applied anywhere.

The implementation is modeled after the 'list of geographic locations' validator in api/admin/geographic_validator.

## Motivation and Context

This is work for tickets like [SIMPLY-2736](https://jira.nypl.org/browse/SIMPLY-2736).

We discussed a longer-term project to centralize the validation code, making it easier to understand and reusable, but decided to go ahead with the current system for now, even though 'list of announcements' is a pretty complicated control to validate -- it would be easier to understand as a composite control made out of smaller controls (basically two dates and a string)

## How Has This Been Tested?

Since there's no UI change yet, the unit tests should be enough.

## Checklist:


api/admin/announcement_list_validator.py
```diff
@@ -0,0 +1,155 @@
...
+
+        if len(value) > maximum:
+            return INVALID_INPUT.detailed(
+                _('Value too long (maximum length is %(size)d characters): %(value)s',
+                  size=maximum, value=value)
+            )
+        return value
+
+    @classmethod
+    def validate_date(cls, field, value, minimum=None):
```
adriana-alter, line 128: Don't we also want a maximum, so we can check that the start date isn't later than the end date?

api/admin/announcement_list_validator.py
```diff
@@ -0,0 +1,155 @@
...
+
+        :param value: Proposed value for a field.
+        :param minimum: Minimum length.
+        :param maximum: Maximum length.
+
+        :return: A ProblemDetail if the validation fails; otherwise `value`.
+        """
+        if len(value) < minimum:
+            return INVALID_INPUT.detailed(
+                _('Value too short (minimum length is %(size)d characters): %(value)s',
```
adriana-alter, line 116: I think it would be more helpful for the error message to give the length of the proposed value, rather than the actual value.

api/admin/announcement_list_validator.py
```diff
@@ -0,0 +1,155 @@
...
+    def validate_announcement(self, announcement):
+        validated = dict()
+        if not isinstance(announcement, dict):
+            return INVALID_INPUT.detailed(
+                _("Invalid announcement format: %(announcement)r", announcement=announcement)
+            )
+
+        validated['id'] = announcement.get('id', unicode(uuid.uuid4()))
+
+        for required_field in ('start', 'content'):
```
adriana-alter, line 65: I don't think "start" is required; if the admin leaves the start date field blank, it will automatically be set to today's date.  (If the end date field is blank, it will be set to two months after the start date.  So it doesn't make sense to treat only one of them as required...)

api/admin/announcement_list_validator.py
leonardr: We check that the finish date must be at least a day after the start date, which should take care of this case. I'll add a test for it; currently we only test the case where start and finish are the same.

PR #1437, opened 2020-05-20
Use the AnnouncementListValidator to validate incoming data of the appropriate type

leonardr:
## Description

This branch changes library_configuration_settings to take a dictionary of validator objects instead of just one validator (assumed to be a GeographicValidator). This makes it possible to use different validation rules for fields of different types. Specifically, fields of type `announcements` are validated using an `AnnouncementListValidator`.

There are still no actual fields of type `announcements` in LIBRARY_SETTINGS, so this code won't ever be triggered in real life. I'm assuming the branch that adds the front-end editor for fields of this type will add an appropriate field.

I refactored library_configuration_settings to make it easier to test, and added a lot of unit tests for code that was already covered at a higher level by test_libraries_post_edit and test_libraries_post_create.

## Motivation and Context

This is back-end work towards tickets like https://jira.nypl.org/browse/SIMPLY-2736.

I'm trying to clear a path towards a more general framework for data validation, without going all out and writing that framework when that's not the point of the work we're doing. I'm doing just enough refactoring to be confident that my additions didn't break anything,.

## How Has This Been Tested?

Still not exposed through the UI, so the automated tests should be enough.

## Checklist:

### DIFF
--- a/api/admin/controller/library_settings.py
+++ b/api/admin/controller/library_settings.py
@@ -261,6 +261,9 @@ class LibrarySettingsController(SettingsController):
         #   geographic areas), but not settings that are a single value of that type
         #   (_one_ language code or geographic area). Sometimes there's even an implication
         #   that a certain data type ('geographic') _must_ mean a list.
+        # * A list value is returned as a JSON-encoded string. It
+        #   would be better to keep that as a list for longer in case
+        #   controller code needs to look at it.
         format = setting.get('format')
         type = setting.get('type')
 
--- a/api/admin/controller/library_settings.py
+++ b/api/admin/controller/library_settings.py
@@ -248,41 +248,53 @@ class LibrarySettingsController(SettingsController):
             # Validation succeeded -- set the new value.
             ConfigurationSetting.for_library(setting['key'], library).value = validated_value
 
-    def _validate_setting(self, library, setting, validator):
+    def _validate_setting(self, library, setting, validator=None):
         """Validate the incoming value for a single library setting.
 
         :param library: A Library
         :param setting: Configuration data for one of the library's settings.
-        :param validator: A validation object for data of this type. (Currently there is no standard
-           interface for validators.)
+        :param validator: A validation object for data of this type.
         """
+        # TODO: there are some opportunities for improvement here:
+        # * There's no standard interface for validators.
+        # * We can handle settings that are lists of certain types (language codes,
+        #   geographic areas), but not settings that are a single value of that type
+        #   (_one_ language code or geographic area). Sometimes there's even an implication
+        #   that a certain data type ('geographic') _must_ mean a list.
         format = setting.get('format')
         type = setting.get('type')
+
+        # In some cases, if there is no incoming value we can use a
+        # default value or the current value.
+        #
+        # When the configuration item is a list, we can't do this
+        # because an empty list may be a valid value.
+        current_value = self.current_value(setting, library)
+        default_value = setting.get('default') or current_value
+
         if format == "geographic":
-            locations = validator.validate_geographic_areas(self.list_setting(setting), self._db)
-            if isinstance(locations, ProblemDetail):
-                return locations
-            value = locations or self.current_value(setting, library)
+            value = self.list_setting(setting)
+            value = validator.validate_geographic_areas(value, self._db)
         elif format == "announcements":
-            value = validator.validate(self.list_setting(setting))
-            if isinstance(value, ProblemDetail):
-                return value
-            value = value or self.current_value(setting, library)
+            value = self.list_setting(setting)
+            value = validator.validate(value)
         elif type == "list":
-            value = self.list_setting(setting) or self.current_value(setting, library)
+            value = self.list_setting(setting)
             if format == "language-code":
                 value = json.dumps([LanguageCodes.string_to_alpha_3(language) for language in json.loads(value)])
-        elif type == "image":
-            value = self.image_setting(setting) or self.current_value(setting, library)
         else:
-            default = setting.get('default')
-            value = flask.request.form.get(setting['key'], default)
+            if type == "image":
+                value = self.image_setting(setting) or default_value
+            else:
+                value = self.scalar_setting(setting) or default_value
         return value
 
-    def current_value(self, setting, library):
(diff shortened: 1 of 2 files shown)


api/admin/controller/library_settings.py
```diff
@@ -216,29 +221,83 @@ def get_extra_geographic_information(self, value):
...
 
-    def library_configuration_settings(self, library, validator):
-        for setting in Configuration.LIBRARY_SETTINGS:
-            if setting.get("format") == "geographic":
-                locations = validator.validate_geographic_areas(self.list_setting(setting), self._db)
-                if isinstance(locations, ProblemDetail):
-                    return locations
-                value = locations or self.current_value(setting, library)
-            elif setting.get("type") == "list":
-                value = self.list_setting(setting) or self.current_value(setting, library)
```
leonardr, line 227: This is the only place where I may have changed the old behavior during refactoring. I don't think there's actually a behavior change but I want to point it out and get your ghouthts.

I made the change because I thought I saw a bug: I thought this would prevent you from changing a list value from a non-empty list to an empty list. If you tried to do that, `list_setting` would return an empty list, and the empty list -- which the user had explicitly asked for -- would be replaced by the current value.

But then I saw that `list_setting` doesn't return the list, it returns a string containing a JSON-encoded list. That string will never be empty, so the `or self.current_value` here will never run, even if you do pass in an empty list.

So: `self.list_setting(setting)` will always get you something that evaluates to True, so the default will never be used. But, I don't really like the JSON-encoding thing and would like to eventually change it. If we do change it we'll need to make sure we can still handle the case where someone tries to set a non-empty list to an empty list.

api/admin/controller/library_settings.py
```diff
@@ -216,29 +221,83 @@ def get_extra_geographic_information(self, value):
...
+        # In some cases, if there is no incoming value we can use a
+        # default value or the current value.
+        #
+        # When the configuration item is a list, we can't do this
+        # because an empty list may be a valid value.
+        current_value = self.current_value(setting, library)
+        default_value = setting.get('default') or current_value
+
+        if format == "geographic":
+            value = self.list_setting(setting)
```
leonardr, line 279: All of these `list_setting` calls are where we'd need to be careful about falling back to the current value. But right now we're _always_ ignoring any current value. If we don't get any data from the client we assume the client wants the list to be zeroed out. (I think this is the right behavior, we just need to be careful to preserve it going forward.)

api/admin/controller/library_settings.py
adriana-alter: I don't even remember why we had to JSON-encode it--do you?  Yes, it would be nice to eventually change this (and/or to handle the empty list situation more explicitly)...

api/admin/controller/library_settings.py
```diff
@@ -216,29 +221,83 @@ def get_extra_geographic_information(self, value):
...
+            value = validator.validate_geographic_areas(value, self._db)
+        elif format == "announcements":
+            value = self.list_setting(setting)
```
adriana-alter, line 282: The announcements aren't in quite the same format as the geographic input; I'm not sure we'll want/need to run them through `list_setting`.  I'm envisioning them being a list of dictionaries, each with "content", "start_date", and "end_date" keys.  Whereas the geographic input--if I'm remembering correctly!--is just a list of strings.

api/admin/controller/library_settings.py
```diff
@@ -216,29 +221,83 @@ def get_extra_geographic_information(self, value):
...
-                default = setting.get('default')
-                value = flask.request.form.get(setting['key'], default)
-
-            ConfigurationSetting.for_library(setting['key'], library).value = value
+                value = self.scalar_setting(setting) or default_value
+        return value
 
-    def current_value(self, setting, library):
-        return ConfigurationSetting.for_library(setting['key'], library).value
+    def scalar_setting(self, setting):
```
adriana-alter, line 295: Why "scalar"?

api/admin/controller/library_settings.py
```diff
@@ -68,12 +70,15 @@ def process_get(self):
             )]
         return dict(libraries=libraries, settings=Configuration.LIBRARY_SETTINGS)
 
-    def process_post(self, validator=None):
+    def process_post(self, validators_by_type=None):
```
adriana-alter, line 73: It would be even better if we could eventually fold the calls to the specialized validators into `validate_formats` (in `/api/admin/controller/__init__.py`), i.e. run all the validators from the same place instead of treating these like a whole different category of thing.  (But that may be a ways down the road--just something to think about in the meantime!)

api/admin/controller/library_settings.py
adriana-alter: Anyway, I don't think there's a behavior change either.  But I'll triple-check from the front end once this is being triggered.

leonardr: The only reason I can think of is that a JSON-encoded value is ready to be written right to the database. But I think JSON-encoding makes more sense _right before_ writing to the database, not while the controller might still be validating or working with the value. A task for later!

api/admin/controller/library_settings.py
leonardr: I see. In that case I think list_setting is going to get us a list of JSON-encoded strings, which we'll need to decode. Probably best to pass in a flag and do this inside `list_setting`.

api/admin/controller/library_settings.py
leonardr: It's one of those technical terms where I haven't been able to find a less fancy term  that's equally precise. A scalar value means a single value, as opposed to a more complicated grouping of values such as a list. So this is in opposition to `list_setting`.

api/admin/controller/library_settings.py
leonardr: Absolutely, this is where I want to go as well.

adriana-alter:
This looks good!  Did you still want to make more changes, or shall I go ahead and approve?

leonardr:
Let's merge it if you're comfortable. I was trying to find an easy way of adding test coverage to the new code in  list_setting without covering the whole thing, but it'll be covered when you make an end-to-end test with a real 'announcement list' setting.

PR #1438, opened 2020-05-25
Fix odl expired raises nolicense

tdilauro:
## Description

This branch fixes an issue that led to an unhandled exception when ODL checkout encountered a licensepool in which all available licenses were expired. It checks for this condition and raises `NoLicenses`

## Motivation and Context

During checkout of an ODL edition, if all the licenses in the associated licensepool were expired, checkout would raise `NoAvailableCopies`, rather than `NoLicenses`.

This triggered a hold (`ODLAPI._place_hold`), which in turn triggered an addition check to ensure that there truly were no licenses available. Because there were "available" licenses (albeit expired ones) that check failed, raising an unhandled `CurrentlyAvailable` exception, which Flask handled.

Fixes [SIMPLY-2779](https://jira.nypl.org/browse/SIMPLY-2779).

## How Has This Been Tested?

This was tested end-to-end using two CM instances, one configured as a SharedODL server, the other as a client.

New tests were added to cover the underlying condition on the server and verify the expected response in the client.

## Checklist:

### DIFF
--- a/api/odl.py
+++ b/api/odl.py
@@ -345,8 +345,8 @@ class ODLAPI(BaseCirculationAPI, BaseSharedCollectionAPI):
     def _checkout(self, patron_or_client, licensepool, hold=None):
         _db = Session.object_session(patron_or_client)
 
-        unexpired_licenses_owned = len([license for license in licensepool.licenses if not license.is_expired])
-        if unexpired_licenses_owned < 1:
+        unexpired_licenses = (l for l in licensepool.licenses if not l.is_expired)
+        if not any(unexpired_licenses):
             raise NoLicenses()
 
         # Make sure pool info is updated.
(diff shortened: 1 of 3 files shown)


api/odl.py
```diff
@@ -344,7 +345,8 @@ def checkout(self, patron, pin, licensepool, internal_format):
     def _checkout(self, patron_or_client, licensepool, hold=None):
         _db = Session.object_session(patron_or_client)
 
-        if licensepool.licenses_owned < 1:
+        unexpired_licenses_owned = len([license for license in licensepool.licenses if not license.is_expired])
+        if unexpired_licenses_owned < 1:
```
leonardr, line 349: I think it's possible to have thousands of distinct licenses (for a popular book), which could make this a performance problem. Can you either do this work in the database, or use `any` to bail out of the iteration the first time you find an unexpired title?

tdilauro: Thanks for the insight. I'll look into this.

PR #1440, opened 2020-05-28
Publish a library's announcements through its Authentication For OPDS document

leonardr:
## Description

This branch introduces a per-library setting for an announcements list. An announcement may be active or not, based on when it's scheduled to be run. A library's active announcements are sent to clients via the library's Authentication For OPDS document.

We won't rely on the client knowing when the announcement becomes active or stops being active: instead we have a rule that the A4OPDS document only contains active announcements. This keeps the client-side logic simple: if the announcement is in the A4OPDS document and you haven't displayed it yet, you should display it.

There is still no UI configuration for the `announcements` setting, so it won't show up in the admin interface yet. But if a valid value somehow gets into the database, those announcements will be published to the A4OPDS document.

Announcements in the A4OPDS document look like this:

```
"announcements" : [
 { "id": "announcement1", "content": "This is announcement 1"},
 { "id": "announcement2", "content": "This is announcement 2"}
],
```

If we agree that the announcement format is good, I'll add it to our list of [Authentication For OPDS Extensions](https://github.com/NYPL-Simplified/Simplified/wiki/Authentication-For-OPDS-Extensions)

## Motivation and Context

This fixes https://jira.nypl.org/browse/SIMPLY-2744, but exposes another problem: https://jira.nypl.org/browse/SIMPLY-2790. That problem was easy to fix -- someone else wrote the necessary data structure -- so I put the fix in this branch rather than doing a PR.

## How Has This Been Tested?

There's still no UI element, so automated tests only.

While writing this branch I ran into a problem regarding the validation rules. Under the rules we set up, the start date couldn't be before the current date. This would have made it impossible to edit the announcement list and leave a previously created announcement alone -- you would have to update all the start dates to the current date to get the edit to go through.

## Checklist:

- [x] I have updated [the documentation](https://github.com/NYPL-Simplified/Simplified/wiki/Authentication-For-OPDS-Extensions#sitewide-announcements) accordingly 

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -11,6 +11,7 @@ from time import mktime
 import os
 from collections import defaultdict
 
+from expiringdict import ExpiringDict
 from lxml import etree
 from sqlalchemy.orm import eagerload
 
@@ -299,7 +300,9 @@ class CirculationManager(object):
 
         self.patron_web_domains = patron_web_domains
         self.setup_configuration_dependent_controllers()
-        self.authentication_for_opds_documents = {}
+        self.authentication_for_opds_documents = ExpiringDict(
+            max_len=1000, max_age_seconds=3600
+        )
 
     @property
     def external_search(self):
--- a/api/admin/announcement_list_validator.py
+++ b/api/admin/announcement_list_validator.py
@@ -81,7 +81,7 @@ class AnnouncementListValidator(Validator):
         today = datetime.date.today()
 
         start = self.validate_date(
-            'start', announcement.get('start', today), minimum=today
+            'start', announcement.get('start', today)
         )
         if isinstance(start, ProblemDetail):
             return start
--- a/api/announcements.py
+++ b/api/announcements.py
@@ -1,0 +1,82 @@
+import datetime
+
+from core.util.problem_detail import ProblemDetail
+
+from admin.announcement_list_validator import AnnouncementListValidator
+
+class Announcements(object):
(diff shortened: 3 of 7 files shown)


api/controller.py
```diff
@@ -11,6 +11,7 @@
 import os
 from collections import defaultdict
 
+from expiringdict import ExpiringDict
```
leonardr, line 14: An ExpiringDict is a dictionary that gradually forgets about items that were added to it.

api/controller.py
```diff
@@ -299,7 +300,9 @@ def get_domain(url):
 
         self.patron_web_domains = patron_web_domains
         self.setup_configuration_dependent_controllers()
-        self.authentication_for_opds_documents = {}
+        self.authentication_for_opds_documents = ExpiringDict(
+            max_len=1000, max_age_seconds=3600
+        )
```
leonardr, line 305: To fix [SIMPLY-2790](https://jira.nypl.org/browse/SIMPLY-2790) we want the dictionary of Authentication For OPDS to forget about any documents added to it more than an hour ago.

I don't think there should be a limit on the number of A4OPDS documents in the cache, but ExpiringDict doesn't work unless you specify `max_len`, so I set it to a very large number of libraries. I've asked Jonathan Green what is the largest number of libraries he'd put on a single circulation manager, and if that's more than 500 I'll set this to about twice whatever value he gives.

PR #1441, opened 2020-06-09
Add announcements configuration setting

leonardr:
## Description

This branch adds a per-library configuration setting for a list of announcements.

The branch does _not_ add any special UI control for this list -- it will appear in the admin interface as a normal list of strings. For this reason the branch should not be merged as-is; instead, a UI branch should be built on top of it.

While getting this working, I had to separate out the "validation" part of the validator (which checks that an incoming value is properly formatted and the data makes sense) from the "formatting" part (which converts a potentially complex data structure into a string so it can be stored in `configurationsettings.value`). This is actually really good because it let me change `GeographicValidator` so that the data structure is converted to a string in the (new) formatting code -- previously it was in the validation code and that really bothered me.

## Motivation and Context

This is part of the work on the [anonuncements](https://jira.nypl.org/browse/SIMPLY-2739) epic.

## How Has This Been Tested?

I verified that the "announcements" setting shows up in the admin interface as a list.

## Checklist:


api/admin/controller/library_settings.py
```diff
@@ -319,7 +321,7 @@ def list_setting(self, setting, json_objects=False):
                 if not isinstance(i, list):
                     i = [i]
                 if json_objects:
-                    i = [json.loads(s) for x in i]
+                    i = [json.loads(x) for x in i]
```
leonardr, line 324: This bug went unnoticed because we didn't previously have a setting that was a list of JSON-encoded objects (as opposed to a single JSON-encoded object like the geographic settings, or a list of strings).

api/config.py
```diff
@@ -218,6 +220,14 @@ class Configuration(CoreConfiguration):
...
             "category": "Basic Information",
         },
+        {
+            "key": Announcements.SETTING_NAME,
+            "label": _("Scheduled announcements"),
+            "description": _("Announcements will be displayed to authenticated patrons."),
+            "category": "Basic Information",
+            "type": "list",
+            "format": "announcements"
+        },
```
leonardr, line 230: Feel free to change this however you want when you build the UI.

PR #1442, opened 2020-06-10
Add support for SAML-based authentication for patrons

vbessonov:
## Description

This PR adds a new authentication provider implementing SAML authentication workflow for patrons.

## Motivation and Context

SAML-based authentication for patrons is used by many educational institutions.

## How Has This Been Tested?

I added new unit tests and conducted integration testing using a [testbed environment](https://docs.google.com/document/d/1MZwEgJlEXZjEdG-04AGNLD-z07v8uEK5prMwqk0odww/edit?usp=sharing):
The testbed is a simple test environment demonstrating how SAML-based authentication works. It contains the following components:
- SAML-guarded OPDS feeds:
    - http://opds.hilbertteam.net/feed
    - http://opds.hilbert.team/feed
- Circulation Manager with setup SAML-based patron authentication
- Shibboleth IdP used by the OPDS feed and the Circulation Manager instance
- Test Flask application using Circulation Manager to download books

OPDS feeds and Circulation Manager use a Shibboleth instance as an Identity Provider (IdP) and have their metadata set up with it.

Please refer to [the documentation](https://docs.google.com/document/d/1MZwEgJlEXZjEdG-04AGNLD-z07v8uEK5prMwqk0odww/edit?usp=sharing) for a detailed explanation of how to use the testbed.

## Checklist:


api/authenticator.py
```diff
@@ -2495,3 +2593,95 @@ def _error_uri(self, redirect_uri, pd):
...
+
+    TOKEN_TYPE = "SAML 2.0 token"
+    TOKEN_DATA_SOURCE_NAME = 'SAML 2.0'
+
+    # After successfully authenticating using SAML protocol patrons receive Bearer tokens used
+    # for confirming their authentication state. This configuration setting defines how long these tokens can be valid
+    SAML_TOKEN_EXPIRATION_DAYS = 'token_expiration_days'
+
+    # This is the default value for that configuration setting.
+    DEFAULT_TOKEN_EXPIRATION_DAYS = 42
```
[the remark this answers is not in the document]
tdilauro: For SAML, authentication can be expired early by logging out or by policies implemented in a given IdP, IdP restarts, etc.

api/saml/configuration.py
```diff
@@ -0,0 +1,376 @@
...
+        :param setting_name: Name of the library's configuration setting
+        :type setting_name: string
+
+        :param value: Value to be serialized (should be picklable)
+        :type value: Any
+        """
+        if not value:
+            raise ValueError('Value must be non-empty')
+
+        serialized_value = pickle.dumps(value)
```
leonardr, line 77: Up to this point we've been using `json.dumps` to convert complex data structures to strings for storage in `configurationsettings.value`. I'm concerned about using `pickle.dumps` for a couple reasons. First, any kind of inconsistency will make the system harder to understand. Second, configurationsettings.value is defined as a Unicode value. In the database everything is binary, of course, so it's theoretically possible to store a binary value here. And presumably the code you wrote works. But it's another inconsistency and it could also cause problems down the road -- I remember we had some problems with Unicode database fields containing non-Unicode data when we were initially preparing for a Python 3 migration.

As I go further into the file, I'll understand more about what this value looks like and my opinion may change, but right now I think you should serialize this to/from JSON rather than using pickle. See https://github.com/NYPL-Simplified/circulation/pull/1441/ for some work I did recently with JSON serialization of a complex value stored in `configurationsettings.value`

api/saml/configuration.py
```diff
@@ -0,0 +1,376 @@
...
+        :param setting_name: Name of the library's configuration setting
+        :type setting_name: string
+
+        :return: Any
+        """
+        value = ConfigurationSetting.for_externalintegration(
+            setting_name,
+            self._integration).value
+
+        return value
```
leonardr, line 51: Reading this I was a little confused by your use of "serialize" to describe the act of taking a value of a configuration setting out of the database _and converting it from a string to a usable value_. To the extent we talk about "serializing" we generally use it to refer to the second part. I don't think this is a big deal, but it led to some odd moments for me later on.

api/saml/configuration.py
```diff
@@ -0,0 +1,376 @@
...
+        self._identity_providers = None
+        self._service_provider = None
+
+    def _load_debug(self):
+        """Returns a debug mode indicator
+
+        :return: Debug mode indicator
+        :rtype: bool
+        """
+        debug = bool(self._configuration_serializer.deserialize(self.DEBUG))
```
leonardr, line 139: Here's one of the weird moments I mentioned earlier. We're "deserializing" a value by pulling a string out of the database (not what I would call "deserializing") and then converting it from a string to a bool (what I _would_ call "deserializing"). 

I don't want to make a big deal out of this because it's not the most complicated part of this branch, but in combination with the pickle thing (which is in play in metadata_serializer but not configuration_serializer) it makes this file confusing to read.

api/saml/configuration.py
leonardr: Another problem I've had in the past with long-term storage of data as pickle objects is, pickle objects embed the Python class and module structure that created them. This means a change to the product that does nothing but refactor code can break existing installations. It's also very difficult to migrate data that's in pickle format. (It's not super easy to migrate JSON data stored in a database, but pickle data is even worse.)

So -- still not having gotten all the way through this file -- I now think that if the IdP configuration data comes in as XML, it should be stored in the database as XML, and parsed into Python objects as necessary. Storing it in any other format is introducing an additional layer at which things can go wrong.

If

api/authenticator.py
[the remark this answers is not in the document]
leonardr: Actually, now's a good time to mention that "logging out" of Clever (assuming that were possible) won't invalidate the token that the circulation manager issued to SimplyE. That's because the circulation manager never checks back with Clever after the initial login. If you need to periodically check with the IdP whether the user's session is still valid, that's a new check you'll need to run. I have a vague idea for how to do this using a `Credential` object which I can go into if necessary.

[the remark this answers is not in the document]
vbessonov: Tim is absolutely right, depending on the IdP configuration the session can be very short. SAML assertion contains the following information which can be used to determine how long the SP session be valid:
- AuthnStatement/@SessionNotOnOrAfter
- Conditions/@NotOnOrAfter
- SubjectConfirmationData/@NotOnOrAfter

Please refer to this [page](https://stackoverflow.com/questions/29508906/notonorafter-in-subjectconfirmationdata-and-conditions-and-sessionnotonorafter) for detailed description.

I'm thinking that I could implement the following algorithm by checking these attributes in the following order and taking the first non-empty and setting it as a TTL of a token:
- AuthnStatement/@SessionNotOnOrAfter
- Conditions/@NotOnOrAfter
- SubjectConfirmationData/@NotOnOrAfter

@leonardr, @tdilauro, what do you think?

api/saml/configuration.py
vbessonov: You're absolutely right regarding pickle - it's not safe. 
In the beginning, I was thinking about not introducing new settings and dynamically parsing XML metadata but I then thought that it might affect the performance: IdP configuration might contain multiple IdPs and parsing it might take time.

api/authenticator.py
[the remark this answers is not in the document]
leonardr: In terms of our system, the 'token' you're talking about would be stored as a `Credential` object. You can set `Credential.expires` to set a TTL.

Then I think it would work like this:
* When someone comes in with a bearer token issued by a CM, you look up their Patron record by a Credential -- the bearer token.
* You then look up their SAML assertion -- a second Credential associated with the same Patron -- incidentally checking whether the assertion has expired.
* If the SAML assertion has expired, the bearer token is treated as invalid, and the patron must go through the login process again.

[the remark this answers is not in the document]
leonardr: Again, haven't gotten through this PR so not sure what parts of this you've already implemented. I can tell by searching the diff that you're doing _something_ with Credential.

api/saml/configuration.py
leonardr: Performance shouldn't be an issue as the XML file will only be parsed on startup and when an admin changes the site configuration. (That said, if parsing the XML file takes 30 seconds, that's a long time to wait whenever someone changes the site configuration.)

vbessonov: Actually, those settings are used during the authentication process. We have to specify SP and IdP metadata for the python-saml library to work.
So every time, a client tries to authenticate by calling `saml_authenticate` endpoint and when IdP sends a SAML assertion by calling `saml_callback` we will have to parse them

leonardr: The `SAMLAuthenticationProvider` object will be instantiated on startup and whenever site configuration is reloaded. If you can create the `SAMLConfiguration` object inside the constructor, you should be able to refer to it as needed without parsing the document every time.

leonardr: Just don't store a reference to the database connection or any database objects inside the `SAMLAuthenticationProvider`; that can cause problems in a multithreaded environment.

vbessonov: Sounds great, I'll implement it this way, thank you!

api/authenticator.py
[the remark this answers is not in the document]
tdilauro: @vbessonov Yes, I think it makes sense to use the ...NotOnOrAfter values to determine session `expires` time.

@leonardr The SAML assertion expiration could also be captured in the patron bearer token Credential to avoid the SAML credential lookup (unless something other than the `expires` property is needed from the SAML).

But part of the concern I was trying to get at earlier was that the validity of the assertion can be expired ahead of these times by external events, so the assertion could be invalid even though is hasn't expired.

This statement from the beginning of section 2.5.1.2 of [SAML2 core spec](https://www.oasis-open.org/committees/download.php/35711/sstc-saml-core-errata-2.0-wd-06-diff.pdf) sums it up:

> The NotBefore and NotOnOrAfter attributes specify time limits on the validity of the assertion within the context of its profile(s) of use. They do not guarantee that the statements in the assertion will be correct or accurate throughout the validity period.

I'm not sure if this will be a problem for our profile of use, but just want to make sure that it's in front of both of you now.

api/saml/controller.py
```diff
@@ -0,0 +1,249 @@
...
+
+        :return: Redirection URL
+        :rtype: string
+        """
+        relay_state_parse_result = urlparse.urlparse(relay_state)
+        relay_state_parameters = urlparse.parse_qs(relay_state_parse_result.query)
+
+        del relay_state_parameters[self.PROVIDER_NAME]
+        del relay_state_parameters[self.IDP_ENTITY_ID]
+        del relay_state_parameters[self.LIBRARY_SHORT_NAME]
```
leonardr, line 127: This will raise KeyError if for whatever reason the relay state URL doesn't include these keys. I believe this URL is generated by the circulation manager, so this shouldn't happen in normal usage. But we try to avoid uncaught exceptions even if they could only be triggered by bad input.

api/saml/controller.py
```diff
@@ -0,0 +1,249 @@
...
+        """
+        if self.RELAY_STATE not in request.form:
+            return SAML_INVALID_RESPONSE.detailed('{0} is empty'.format(self.RELAY_STATE))
+
+        relay_state = request.form[self.RELAY_STATE]
+        relay_state_parse_result = urlparse.urlparse(relay_state)
+        relay_state_parameters = urlparse.parse_qs(relay_state_parse_result.query)
+        library_short_name = relay_state_parameters[self.LIBRARY_SHORT_NAME][0]
+        provider_name = relay_state_parameters[self.PROVIDER_NAME][0]
+        idp_entity_id = relay_state_parameters[self.IDP_ENTITY_ID][0]
```
leonardr, line 214: Again, we'll have problems if these keys aren't present in the query string.

api/saml/controller.py
```diff
@@ -0,0 +1,249 @@
...
+
+        provider = self._authenticator.saml_provider_lookup(provider_name)
+        if isinstance(provider, ProblemDetail):
+            return self._redirect_with_error(redirect_uri, provider)
+
+        authentication_manager = self._get_authentication_manager(db, provider)
+        relay_state = self._add_params_to_url(
+            redirect_uri,
+            {
+                self.LIBRARY_SHORT_NAME: provider.library(db).short_name,
```
leonardr, line 184: In addition to the other issues I pointed out with this, it's not necessary to explicitly pass around a library identifier. We have an annotator, `@has_library`, which modifies a route so that it always operates in the context of a particular library. See oauth_callback in api/routes.py for an example.

Within the controller method, you can refer to flask.request.library to get the Library object. As long as you're using `url_for()` to generate links into the controller, the current library will be propagated into those links.

api/saml/configuration.py
```diff
@@ -0,0 +1,376 @@
...
+    def get_identity_provider_settings(self, entity_id):
+        """Returns a dictionary containing identity provider's settings in a OneLogin's SAML Toolkit format
+
+        :return: Dictionary containing identity provider's settings in a OneLogin's SAML Toolkit format
+        :rtype: Dict
+        """
+        if entity_id in self._identity_providers:
+            return self._identity_providers[entity_id]
+
+        identity_providers = filter(lambda idp: idp.entity_id == entity_id, self._configuration.identity_providers)
```
leonardr, line 328: This is cosmetic, but a list comprehension is more readable in situations like this:

`identity_providers = [idp for idp in self._configuration.identity_providers if idp.entity_id == entity_id]`

api/saml/configuration.py
```diff
@@ -0,0 +1,376 @@
...
+
+        if not identity_providers:
+            raise SAMLConfigurationError(
+                message=_('There is no identity provider with entityID = {0}'.format(entity_id)))
+
+        identity_provider = identity_providers[0]
```
leonardr, line 334: What if there's more than one?

api/saml/controller.py
vbessonov: I was thinking about it but, unfortunately, we can't use dynamic routes in the case of SAML callback. It has to be mentioned in the SP metadata which is uploaded to the IdP and it cannot change, otherwise, a SAML request with a SAML callback different from the one configured in the metadata won't pass validation

leonardr: I'm not sure what you mean by dynamic routes here. Is the issue that a given hostname can have one and only one SAML callback? With the library name being "dynamic" information that can go into the query string but not the path portion of the URL?

api/authenticator.py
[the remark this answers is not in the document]
leonardr: We need to handle this case:

* Patron has a book out on loan
* Patron asks CM for a link to the book
* CM provides the link
* Patron follows the link, providing the book vendor whatever information is necessary to the vendor to establish credentials (I remain fuzzy on how exactly this step is supposed to work)
* Book vendor checks back with the IdP to validate the patron's authentication
* IdP says "no, they logged out, that session is no longer valid"
* Book vendor sends a 401 in response to the request for the book

In that case, the client needs to know that a 401 _from the book vendor_ means they should try logging in again _to the circulation manager_. The same as if they had gotten a 401 directly from the circulation manager.

api/saml/controller.py
vbessonov: Yes, the URL of the SP's SAML callback has to be constant. 

For example, if the URL of the SP's SAML callback configured as `https://circulation.manager.org/saml_callback`, we can't use `https://circulation.manager.org/LIBRARY_1/saml_callback` for `LIBRARY 1` and `https://circulation.manager.org/LIBRARY_2/saml_callback` for `LIBRARY 2`, we have to use the same URL configured in the IdP's metadata and we have to pass additional parameters as part of the relay state:
1. Inside `SAMLController.saml_authentication_redirect` we add additional parameters to the redirect URL and save it as the relay state.
2. Inside `SAMLController.saml_callback` we extract parameters from the relay state and delete them from the query part of the URL

leonardr: OK, mention this in a comment.

api/saml/parser.py
```diff
@@ -0,0 +1,357 @@
...
+        if len(name_id_format_nodes) > 0:
+            name_id_format = OneLogin_Saml2_Utils.element_text(name_id_format_nodes[0])
+
+        acs_service = None
+        acs_service_nodes = OneLogin_Saml2_Utils.query(
+            provider_node,
+            "./md:AssertionConsumerService[@Binding='%s']" % required_acs_binding.value
+        )
+        if len(acs_service_nodes) > 0:
+            acs_url = acs_service_nodes[0].get('Location', None)
```
leonardr, line 299: I have a couple more "what if there's more than one?" type questions about this stuff. Does it matter?

api/saml/provider.py
```diff
@@ -0,0 +1,335 @@
...
+
+SAML_INVALID_SUBJECT = pd(
+    'http://librarysimplified.org/terms/problem/saml/invalid-subject',
+    status_code=401,
+    title=_('SAML invalid subject.'),
+    detail=_('SAML invalid subject.')
+)
+
+
+class SAMLAuthenticationProvider(BaseSAMLAuthenticationProvider):
```
leonardr, line 23: Are there going to be other subclasses of BaseSAMLAuthenticationProvider? I think you copied this design from OAuthAuthenticationProvider, which has one subclass -- for Clever -- but could conceivably have other subclasses for other systems that use OAuth. If any authentication provider that uses SAML can use SAMLAuthenticationProvider, you should be able to get rid of BaseSAMLAuthenticationProvider and just have one class.

api/saml/provider.py
```diff
@@ -0,0 +1,335 @@
...
+        :type patron: Patron
+
+        :param token: Token containing SAML subject's metadata
+        :type token: Dict
+
+        :return: Credential object
+        :rtype: Credential
+        """
+        data_source, ignore = self._get_token_data_source(db)
+        duration = datetime.timedelta(days=self.token_expiration_days)
```
leonardr, line 172: Per our discussion above, this is where to put the TTL you get from the IdP, since (I believe) this token contains the actual SAML assertion you got from the IdP.

api/saml/provider.py
vbessonov: This is a very question. `SAMLAuthenticationProvider` implements Web Browser SSO Profile using two specific SAML bindings:
- HTTP-Redirect for SAML requests
- HTTP-Post for SAML responses

There may be other classes inherited from `BaseSAMLAuthenticationProvider` which could implement other SAML profiles or implement the same SSO profile but using other bindings (for example, SOAP).

api/saml/provider.py
```diff
@@ -0,0 +1,335 @@
...
+
+        :param db: Database session
+        :type db: sqlalchemy.orm.session.Session
+
+        :return: Token data source
+        :rtype: DataSource
+        """
+        return get_one_or_create(
+            db, DataSource, name=self.TOKEN_DATA_SOURCE_NAME
+        )
```
leonardr, line 211: I don't think this is going to work in a situation where a library has multiple SAML authentication mechanisms for its patrons. You'll look up a Credential from this data source but you won't be able to tell which IdP it came from.

This would work if the data source identified the IdP -- logically, the IdP is the "source" of this "data".

leonardr: Of course, it's also possible to say that a library can't have multiple SAML authentication mechanisms -- that's what we say now for Basic Auth authentication mechanisms. But this branch isn't there and I don't think that's where you're going.

api/saml/parser.py
```diff
@@ -0,0 +1,357 @@
...
+            ui_info,
+            name_id_format,
+            acs_service,
+            authn_requests_signed,
+            want_assertions_signed,
+            certificates)
+
+        return sp
+
+    def parse(self, xml_metadata):
```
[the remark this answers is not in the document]
vbessonov: `xlmparser.py` is really great. However, OneLogin's `python-saml` relies on [defusedxml](https://pypi.org/project/defusedxml/) library which enhances `lxml` and makes it more secure. `python-saml` uses [defusedxml](https://pypi.org/project/defusedxml/) internally, that's why I decided to use it as well

api/saml/provider.py
vbessonov: I was looking at `OAuthAuthenticationProvider` as a reference and it has the same logic in `authenticator.py`:
```python
    def token_data_source(self, _db):
        return get_one_or_create(
            _db, DataSource, name=self.TOKEN_DATA_SOURCE_NAME
        )
```

What is the best way to fix it?

leonardr: Ugh, that's no good. I guess let's leave it as is for now, add a comment, and we'll figure it out when we need to.

api/saml/controller.py
vbessonov: Added additional checks whether a specific key in the dictionary

api/saml/controller.py
vbessonov: Added additional validation logic to check whether a specific key in the relay's state parameters and return a ProblemDetail instance in the case if it's not

api/saml/configuration.py
vbessonov: Changed it to using a list comprehension

api/saml/configuration.py
vbessonov: Added additional validation logic

api/saml/parser.py
vbessonov: AssertionConsumerService has "isDefault" and "index" attributes, I added additional logic for fetching the default ACS with given attributes and taking the one with the smallest index as a fallback logic

api/authenticator.py
```diff
@@ -597,6 +627,14 @@ def from_config(cls, _db, library, analytics=None, custom_catalog_source=CustomP
...
                 _db
             )
+        if authenticator.saml_providers_by_name:
+            # NOTE: this will immediately commit the database session,
+            # which may not be what you want during a test. To avoid
+            # this, you can create the bearer token signing secret as
+            # a regular site-wide ConfigurationSetting.
+            authenticator.bearer_token_signing_secret = BaseSAMLAuthenticationProvider.bearer_token_signing_secret(
+                _db
+            )
```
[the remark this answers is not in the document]
vbessonov: I moved the logic to a separate class - `BearerTokenSigner`

api/saml/configuration.py
vbessonov: Got rid of `SAMLMetadataSerializer` and didn't even need to use `json.dumps`/`json.loads` because I don't store Python objects anymore XML only which is translated into Python objects dynamically

api/authenticator.py
```diff
@@ -2495,3 +2593,95 @@ def _error_uri(self, redirect_uri, pd):
...
+           'type': 'textarea',
+           'description': _('SAML metadata of Identity Providers in an XML format '
+                            '(may contain multiple IDPSSODescriptor)'),
+           'category': 'IDP',
+           'required': True
+       },
+       {
+           "key": SAML_TOKEN_EXPIRATION_DAYS,
+           "type": "number",
+           "label": _("Days until a Bearer token expires")
```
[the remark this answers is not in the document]
vbessonov: I removed this property and determine credentials expiration based on SAML session expiration and assertion validity

api/saml/provider.py
vbessonov: Set credential.expiration to either AuthnStatement/@SessionNotOnOrAfter or SubjectConfirmationData/@NotOnOrAfter

api/saml/provider.py
vbessonov: Added a comment

api/routes.py
```diff
@@ -543,6 +543,27 @@ def oauth_authenticate():
...
+@library_route('/saml_authenticate')
+@has_library
+@returns_problem_detail
+def saml_authenticate():
+    return app.manager.saml_controller.saml_authentication_redirect(flask.request.args, app.manager._db)
+
+# Redirect URI for SAML providers
+# NOTE: we cannot use @has_library decorator and append a library's name to saml_calback route
+# (e.g. https://cm.org/LIBRARY_NAME/saml_callback).
+# The URL of the SP's assertion consumer service (saml_callback) should be constant:
```
leonardr, line 556: I understand what you're saying because we talked about this earlier, but this will cause confusion in the future because the query string is part of the URL. It would be more accurate to say that the path portion of the URL must be constant, so the library name needs to go into the query string.

requirements.txt
```diff
@@ -28,6 +28,7 @@ uszipcode
 pypostalcode
 watchtower # for cloudwatch logging
 sphinx # for code documentation
+rsa==4.3
```
leonardr, line 31: You should get this change if you merge in the main branch -- Tim merged it earlier.

tests/saml/controller_test.py
```diff
@@ -0,0 +1,18 @@
...
+
+class ControllerTest(BaseControllerTest):
+    def __init__(self):
+        self._library = None
+        self._integration = None
+
+    def setup(self, _db=None, set_up_circulation_manager=True):
+        super(ControllerTest, self).setup(_db, set_up_circulation_manager)
+
+        self._library = self.make_default_library(self._db)
```
leonardr, line 14: It shouldn't be necessary to call make_default_library; it will be created the first time you refer to self._default_library.

tests/saml/database_test.py
```diff
@@ -0,0 +1,20 @@
...
+class DatabaseTest(BaseDatabaseTest):
+    def __init__(self):
+        self._library = None
+        self._integration = None
+        self._authentication_provider = None
+
+    def setup(self, mock_search=True):
+        super(DatabaseTest, self).setup(mock_search)
+
+        self._library = self.make_default_library(self._db)
```
leonardr, line 15: Same comment as above.

api/admin/controller/patron_auth_services.py
```diff
@@ -43,6 +49,30 @@ def __init__(self, manager):
...
+        if protocol == 'api.saml.provider':
+            saml_settings_validator = SAMLSettingsValidator(
+                SAMLMetadataParser(),
+                SAMLMetadataSerializer(integration)
+            )
+
+            return saml_settings_validator
+
+        return None
+
```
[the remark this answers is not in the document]
vbessonov: I was thinking about registering validators inside `LibraryAuthenticator.register_provider`, storing registered inside `LibraryAuthenticator._validators_by_protocol` dictionary and accessing it via `Authenticator.validator_lookup`, `LibraryAuthenticator.validator_lookup`.
For example, `Authenticator.validator_lookup` was supposed to look like:
```python
    def validator_lookup(self, *args, **kwargs):
        return self.invoke_authenticator_method(
            "validator_lookup", *args, **kwargs
        )
```

But it turned out that I can't use it that way because there is no library associated with the current request during settings validation in `PatronAuthServicesController`.

@leonardr, is it the right way to implement it?

tests/saml/test_configuration.py
```diff
@@ -0,0 +1,202 @@
+# FIXME: Required to get rid of the circular import error
+import api.app
```
leonardr, line 2: Can you go into more detail on this? Do you need to import api.app inside the test because putting the import statement where you'd prefer to put it would cause a circular import?

tests/saml/test_controller.py
```diff
@@ -0,0 +1,123 @@
...
+            })
+
+            with self.app.test_request_context('/saml_authenticate?' + query):
+                request.library = self._default_library
+                result = controller.saml_authentication_redirect(request.args, self._db)
+
+                print(result)
+
+                eq_(result.status_code, 302)
+                location_header = result.headers.get('Location')
```
leonardr, line 84: Can you make any assertions about the value of the Location header here?

tests/saml/test_controller.py
```diff
@@ -0,0 +1,123 @@
...
+
+            authenticator.bearer_token_signing_secret = 'test'
+            authenticator.library_authenticators['default'].bearer_token_signing_secret = 'test'
+
+            saml_response = b64encode(SAML_RESPONSE)
+            with self.app.test_request_context('/', data={
+                'SAMLResponse': saml_response,
+                SAMLController.RELAY_STATE: 'http://localhost?' + query
+            }):
+                controller.saml_authentication_callback(request, self._db)
```
leonardr, line 123: This test is also missing some assertions about what happened.

tests/saml/test_parser.py
```diff
@@ -0,0 +1,287 @@
...
+
+        # Act
+        result = metadata_parser.parse(fixtures.CORRECT_ONE_IDP_METADATA)
+
+        # Assert
+        assert isinstance(result, list)
+        eq_(len(result), 1)
+
+        eq_(
+            result[0],
```
leonardr, line 84: It's not necessary to change this, but another cosmetic note: in cases like this I like to assign the list to a variable.

```
[result] = result
eq_(result, ...)
```

This proves that `result` is a list containing a single item without needing an extra explicit assertion.

api/admin/controller/patron_auth_services.py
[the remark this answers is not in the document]
leonardr: `LibraryAuthenticator` is designed for situations where the request is scoped to a specific library, so that won't work. And `Authenticator` is a collection of `LibraryAuthenticator` objects.

What if you defined a function called `validator_factory` in api.saml.provider which returned a `SAMLSettingsValidator` method when called with an ExternalIntegration? This method could then look for a `validator_factory` in the appropriate module and call it if it finds one.

api/saml/provider.py
leonardr: In that case, how about renaming `SAMLAuthenticationProvider` to `WebBrowserSAMLAuthenticationProvider` or something that conveys which part of SAML it implements?

api/saml/parser.py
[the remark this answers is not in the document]
leonardr: This makes sense.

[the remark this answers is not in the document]
leonardr: In fact maybe xmlparser.py should start using defusedxml.

leonardr:
Just to set a baseline, I've just finished going through the entire PR up to this point.

api/saml/provider.py
vbessonov: I renamed `SAMLAuthenticationProvider` to `SAMLWebSSOAuthenticationProvider`. Would you prefer to rename it to `WebBrowserSAMLAuthenticationProvider` instead?

api/routes.py
vbessonov: Actually I meant that both, path and query parts, cannot be changed. We cannot add parameters to the query part of the URL, it has to be passed as a part of the relay state.

We send the relay state to the IdP in a query part using HTTP-Redirect binding and then IdP sends it back in the body of the response using HTTP-Post binding.

api/admin/controller/patron_auth_services.py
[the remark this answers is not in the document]
vbessonov: I created `ValidatorFactory` class inside `api.admin.valiator` and used it in `PatronAuthServicesController`

requirements.txt
vbessonov: I rebased on the recent version of the master and reverted my changes

api/authenticator.py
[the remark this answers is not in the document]
vbessonov: If a book vendor uses SAML, then IdP will just show a login screen and will ask the patron to enter their credentials again

tests/saml/test_parser.py
vbessonov: Sure, changed it in accordance with your comment

tests/saml/controller_test.py
vbessonov: Fixed it, thank you

tests/saml/database_test.py
vbessonov: Started using `self._default_library`

tests/saml/test_controller.py
vbessonov: Added additional tests for `SAMLController`

tests/saml/test_controller.py
vbessonov: Added additional tests for `SAMLController`

api/saml/provider.py
leonardr: I must have missed that; SAMLWebSSOAuthenticationProvider is fine.

api/admin/validator.py
```diff
@@ -163,3 +165,36 @@ def _value(self, field, form):
         elif len(value) == 1:
             return value[0]
         return filter(lambda x: x != None and x != "", value)
+
+
+class ValidatorFactory(object):
```
leonardr, line 170: This is specific to validators for patron authenticator protocols (as opposed to collection protocols or something else) and it should have a name that reflects that; maybe `PatronAuthenticationValidatorFactory`.

api/admin/validator.py
```diff
@@ -163,3 +165,36 @@ def _value(self, field, form):
...
+        :type protocol: string
+
+        :return: Validator object (if there is any for the specific protocol)
+        :rtype: Optional[Validator]
+        """
+        module_name = protocol
+
+        try:
+            provider_module = importlib.import_module(module_name)
+            validator_class = getattr(provider_module, 'Validator', None)
```
leonardr, line 191: (The part that's specific to patron auth protocols is this bit -- we don't do this importlib trick for other types of integrations.)

api/saml/provider.py
```diff
@@ -367,4 +369,9 @@ def saml_callback(self, db, subject):
         return credential, patron, patron_data
 
 
+def validator_factory():
+    return SAMLSettingsValidator(SAMLMetadataParser())
+
+
 AuthenticationProvider = SAMLWebSSOAuthenticationProvider
+Validator = validator_factory
```
leonardr, line 377: Looking at this in context... since this is a function, I'd rather have the ValidatorFactory look for it under the name `validator`. `Validator` makes it look like a class.

leonardr:
At this point I only have one minor request -- lowercasing `Validator` so it's clear from the outside that it's a function. Once that's done I think we're good.

leonardr:
Oh, also update to the new circulation-docker revision, which I just merged.

tests/saml/test_configuration.py
vbessonov: I had some weird errors while trying to run tests in specific files, this import was required to resolve those issues but it's not required anymore. I removed all these redundant imports

api/admin/validator.py
vbessonov: Renamed `ValidatorFactory` to `PatronAuthenticationValidatorFactory`

api/saml/provider.py
vbessonov: Changed logic of `PatronAuthenticationValidatorFactory` a bit:
1. First, it tries to instantiate a validator instance using a class name stored in `Validator`
2. Then, it tries to instantiate a new validator instance using a factory function `validator_factory`

I removed `Validator` from `api.saml.provider` so `validator_factory` will be used by `PatronAuthenticationValidatorFactory`

HadrienGardeur:
@leonardr could you open an issue in https://github.com/opds-community/drafts/issues ?

leonardr:
Done: https://github.com/opds-community/drafts/issues/34

PR #1443, opened 2020-06-11
Fix patron expiration test

### LINKED ISSUE #741: Some tests in test_patron_utility.py fail when run at certain times
I think this is due to `PatronUtility.authorization_is_active` using local time instead of UTC, when the tests and other PatronUtility methods use UTC.

tdilauro:
## Description

This branch fixes a couple of tests that used the wrong time zone for patron expiration date/time values.

## Motivation and Context

Depending on local time zone and time of day when run, the affected tests might or might not fail. The tests in question were using UTC time for patron expiration, but the `PatronUtility.authorization_is_active` uses server local time.

Fixes #741  

## How Has This Been Tested?

Successfully ran the remediated tests late night EDT, when they were originally failing, and mid-morning EDT when they previously were NOT failing.
  ```
  TESTING=true PYTHONWARNINGS=ignore nosetests -w tests \
    test_patron_utility.py:TestPatronUtility.test_has_borrowing_privileges \
    test_patron_utility.py:TestPatronUtility.test_needs_external_sync
  ```

## Checklist:

### DIFF
--- a/api/util/patron.py
+++ b/api/util/patron.py
@@ -59,7 +59,6 @@ class PatronUtility(object):
         :raises OutstandingFines: If the patron has too many outstanding fines.
 
         """
-        now = datetime.datetime.utcnow()
         if not cls.authorization_is_active(patron):
             # The patron's card has expired.
             raise AuthorizationExpired()
--- a/requirements.txt
+++ b/requirements.txt
@@ -38,6 +38,9 @@ feedparser
 # nltk is a textblob dependency, and this is the last release that supports Python 2
 nltk==3.4.5
 
+# rsa is an oauth2client dependency. Version 4.1 silently drops Python 2 support.
+rsa<4.1
+
 # TODO: This is only used for summary evaluation, which I think should
 # only happen in the metadata wrangler, so it should be possible to move
 # it out of core.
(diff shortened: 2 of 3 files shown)


tdilauro:
Any idea why sphinx is erroring in Travis? It's working fine in my local environment.

leonardr:
There was a release of the rsa package 18 hours ago, which silently dropped Python 2 support. This is the same thing that happened recently to [NLTK](https://jira.nypl.org/browse/SIMPLY-2705). You can fix it by pinning rsa to "<4.1" in requirements.txt.

leonardr:
[Here's the bug on the rsa side.](https://github.com/sybrenstuvel/python-rsa/issues/152)

tdilauro:
Thanks, @leonardr. All set.

sybrenstuvel:
I fixed it in the Python-RSA package so that package managers now know which version they can use. Python-RSA 4.3 is now the last one to support Python 2.7, version 4.4 explicitly requires Python 3.5+.

Hope this is now resolved for you, if there are still issues let me know.

leonardr:
Thanks, Sybren, we appreciate it.

PR #1445, opened 2020-06-17
Ebooks can be served in a custom format for use by AxisNow-compatible clients

leonardr:
## Description

This branch adds a new delivery format for ebooks served by Axis 360 -- the "AxisNow" format. This is represented by a new data format that contains the two essential pieces of information: the ISBN Axis 360 has on file for the book, and the "book vault ID" representing (some of) the books the patron has on loan.

When you borrow an book from Axis 360, your loan will be registered in one of two places: the "Book Vault" (if you requested the "AxisNow" format) or the Adobe ACS server (if you requested "ePub" or "PDF"). When you try to fulfill an ebook loan, the Axis 360 API will offer both a link into the ACS server (which gives you an ACSM file) and a "Book Vault UUID" which, when combined with the book's ISBN, can be used to retrieve the ebook through the AxisNow system.

The problem is that only one of the two techniques will work for any given loan. So we need to keep track of whether we originally requested "ePub" or "AxisNow", and provide that information when it's time to fulfill the book. (Fortunately, the general CM code already does this; we just weren't previously using it for Axis 360.) 

When it's time to fulfill a book, we hit the availability endpoint for the loan. At this point there are three possibilities:

1. This is an ebook that was placed into the ACS server. The availability API response will contain an entry point into the ACS server. We create a normal `FulfillmentInfo` object containing the link and return it. This works the same as before and it means we can fulfill an ACS ebook without making any extra HTTP requests.
2. This is an audiobook. The response from the availability endpoint contains a "transaction ID", which we place into an `Axis360FulfillmentInfo` (previously this was called an `AudiobookFulfillmentInfo`). When we need to actually tell the client how to fulfill this audiobook, `Axis360FulfillmentInfo` makes a request to the "getFulfillmentInfo" endpoint, providing the transaction ID. The response contains the information necessary to make a `FindawayManifest`. (This is exactly how audiobooks used to work -- only the class name has changed. I mention it because the next case is new, but very similar.)
3. This is an ebook that was placed into the Book Vault. As with an audiobook, we create an `Axis360FulfillmentInfo` containing the "transaction ID" we got from the availability endpoint. When we need to actually tell the client how to fulfill this audiobook, `Axis360FulfillmentInfo` makes a request to the "getFulfillmentInfo" endpoint, providing the transaction ID. The response contains the information necessary to make an `AxisNowManifest` -- the Book Vault ID and the ISBN.

When a preexisting circulation manager starts up with this code, any Axis 360 ebooks will start out with an "ePub" delivery mechanism but no "AxisNow" delivery mechanism (since that mechanism was originally ignored). A migration script could fix this, but I didn't add one since it will also be fixed gradually by the `axis_reaper` script, which causes the CM to fetch new metadata for each book (including format data) and reevaluate it.

## Motivation and Context

This addresses [SIMPLY-2815](https://jira.nypl.org/browse/SIMPLY-2815), the circulation manager portion of the Open eBooks relaunch.

## How Has This Been Tested?

I ran an end-to-end test where I used the circulation manager to borrow a book from Axis 360 (asking for the "AxisNow" format) and fulfill it as an AxisNowManifest. I then used the data in the AxisNowManifest to acquire a license for the book through AxisNow.

## Checklist:


api/axis.py
```diff
@@ -1247,20 +1277,12 @@ def process_one(self, e, namespaces):
             expiration_date = datetime.strptime(
                 expiration_date, self.FULL_DATE_FORMAT)
 
-        # WTH??? Why is identifier None? Is this ever used?
-        fulfillment = FulfillmentInfo(
-            collection=self.collection, data_source_name=DataSource.AXIS_360,
-            identifier_type=self.id_type,
-            identifier=None, content_link=fulfillment_url,
-            content_type=None, content=None, content_expires=None
-        )
```
leonardr, line 1256: This was not, in fact, ever used. We never fulfill a book immediately after checking it out. I removed this code rather than make it work in a world where an `EbookFulfillmentInfo` is expected to contain a book vault ID. That's because a checkout response document doesn't seem to _have_ a book vault ID -- it's only found in an availability document, which you can obtaine _after_ the checkout..

api/axis.py
```diff
@@ -1358,22 +1380,28 @@ def process_one(self, e, ns):
             download_url = self.text_of_optional_subtag(
                 availability, 'axis:downloadUrl', ns)
             transaction_id = self.text_of_optional_subtag(
```
leonardr, line 1382: Turns out the "transaction ID" is a.k.a. the "book vault ID".

leonardr:
This branch needs to go back in the shop. While testing it in a real situation I discovered that the "transaction ID" it uses is completely different from the "book vault ID". It is possible to get the "book vault ID" but only by hitting the `getfulfillmentInfo` endpoint -- similar to what we do for audiobooks.

api/axis.py
```diff
@@ -334,7 +340,7 @@ def _checkout(self, title_id, patron_id, internal_format):
         response = self.request(url, data=args, method="POST")
         return response
 
-    def fulfill(self, patron, pin, licensepool, **kwargs):
+    def fulfill(self, patron, pin, licensepool, internal_format, **kwargs):
```
leonardr, line 343: `internal_format` was always being passed in, but up to this point it was ignored because the only way to fulfill a book was to return its ACS fulfillment URL (if the Axis 360 API gave you one) or its Findaway fulfillment data (if the API gave you that).

Now we're in a situation where the Axis 360 API may give you _both_ an ACS fulfillment URL _and_ a BookVaultUUID, but only one of the two will actually work. To know which one will work, `fulfill` needs to know which format we told Axis 360 to use when we borrowed the book. That information is kept in `internal_format`.

api/axis.py
```diff
@@ -1321,13 +1338,17 @@ def process_one(self, e, namespaces):
...
-    def __init__(self, api):
+    def __init__(self, api, internal_format=None):
         """Constructor.
 
         :param api: An Axis360API instance, in case the parsing of an
-        availability document triggers additional API requests.
+           availability document triggers additional API requests.
+
+        :param internal_format: The name Axis 360 gave to the format
+           the user requested. Used to distinguish a request for
```
EdwinGuzman, line 1348: This seems unfinished.

api/axis.py
```diff
@@ -1377,14 +1401,22 @@ def process_one(self, e, ns):
...
-                # audio manifest.
-                fulfillment = AudiobookFulfillmentInfo(
+                # If this is an audiobook, we'll need to make two more
+                # requests: the first to get the Findaway content ID,
+                # license ID, and session key; the second to get the
+                # audiobook metadata.
+                #
+                # If this is an ebook being delivered in AxisNow
+                # format, we'll need to make one more request, to get
+                # the Book Vault UUID and ISBN.
```
EdwinGuzman, line 1411: Does this request happen after the first two requests or are two above just for Findaway?

EdwinGuzman: Oh nevermind. It wasn't clear at first that it was audiobook vs ebook.

leonardr: Audiobook means two more requests, AxisNow means one more request. I'll clarify.

PR #1446, opened 2020-06-18
Catalog url pattern

kristojorg:
## Description

I changed the admin interface input validation for the `PATRON_WEB_CLIENT_URL` setting, and updated the setting description as well.

## Motivation and Context

This will allow users to set a regular expression for their allowed CORS domains, which in turn enables them to set up multiple frontends accessing the same circulation-manager. This will be used by Lyrasis to have both a production and a beta instance of `circulation-patron-web`.

## How Has This Been Tested?

`nosetests tests/admin/ tests/test_controller.py`

## Checklist:

I only ran the admin tests.
Is there documentation somewhere I need to update beyond the setting description?

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -183,11 +183,11 @@ class Configuration(CoreConfiguration):
         },
         {
             "key": PATRON_WEB_CLIENT_URL,
-            "label": _("URL of the web catalog for patrons"),
+            "label": _("Hostnames for web application access"),
             "required": True,
             "format": "string",
             "allowed": ["*"],
-            "description": _("You can set this to '*' in development, but you must a real case-sensitive URL or regular expression in production in order to prevent unauthorized CORS requests.")
+            "description": _("Only web applications from these hosts can access this circulation manager. This can be a single hostname ('catalog.library.org'), or a regular expression that matches many hostnames ('.*\.libraryvendor\.com'). You can also set this to '*' to allow access from any host, but you must not do this in a production environment -- only during development.")
         },
         {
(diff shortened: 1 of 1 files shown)


leonardr:
I'm surprised that a regular expression works here; how does that end up in the CORS headers?

kristojorg:
I'm not really sure to be honest. I saw that regex works in the `flask_cors` docs here: https://flask-cors.corydolphin.com/en/latest/api.html

I assume the regex doesn't end up in the header, but that `flask_cors` is simply verifying the request domain against the regex at request time

leonardr:
I tracked down the flask-cors code to https://github.com/corydolphin/flask-cors/blob/master/flask_cors/core.py#L113 and I think you're right.

api/config.py
```diff
@@ -185,9 +185,9 @@ class Configuration(CoreConfiguration):
             "key": PATRON_WEB_CLIENT_URL,
             "label": _("URL of the web catalog for patrons"),
```
leonardr, line 186: This is no longer a good enough description for this setting, if it ever was. Both the `label` and the `description` need to be updated.

What is this now? I think it's something like:

Hostnames for web application access

Only web applications from these hosts can access this circulation manager.

This can be a single hostname ("catalog.library.org"), or a regular expression that matches many hostnames (".*\.libraryvendor\.com"). You can also set this to "*" to allow access from any host, but you must not do this in a production environment -- only during development.

kristojorg: Cool, just updated the label and description to what you suggested

PR #1447, opened 2020-06-23
Patron Web Hostnames

kristojorg:
## Description

This update allows users to pass a list of urls that will be allowed CORS access to the CM. This is a change from the previous single-value setting, and from the attempted regex-based approach. 

To accomplish this properly, I also renamed the setting from `PATRON_WEB_CLIENT` to `PATRON_WEB_HOSTNAMES`, and the DB key from `Patron Web Url` to `patron_web_hostnames`. There is a migration script included in the PR to change the key and update from single string to list of strings.

## Motivation and Context

- To allow multiple web front ends to access the same CM backend. 

## How Has This Been Tested?

`nosetests tests/admin/ tests/test_controller.py`

## Checklist:

### DIFF
--- a/tests/test_controller.py
+++ b/tests/test_controller.py
@@ -445,13 +445,13 @@ class TestCirculationManager(CirculationControllerTest):
         # We also set up some patron web client settings that will
         # be loaded.
         ConfigurationSetting.sitewide(
-            self._db, Configuration.PATRON_WEB_HOSTNAMES).value = json.dumps(list("http://sitewide/1234"))
+            self._db, Configuration.PATRON_WEB_HOSTNAMES).value = json.dumps(["http://sitewide/1234"])
         registry = self._external_integration(
             protocol="some protocol", goal=ExternalIntegration.DISCOVERY_GOAL
         )
         ConfigurationSetting.for_library_and_externalintegration(
             self._db, Registration.LIBRARY_REGISTRATION_WEB_CLIENT,
-            library, registry).value = json.dumps(list("http://registration"))
+            library, registry).value = "http://registration"
 
         # Then reload the CirculationManager...
         self.manager.load_settings()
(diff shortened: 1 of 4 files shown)


kristojorg:
Whoops just forgot to commit that file.

tests/test_controller.py
```diff
@@ -445,7 +445,7 @@ def mock_for_library(cls, incoming_library):
         # We also set up some patron web client settings that will
         # be loaded.
         ConfigurationSetting.sitewide(
-            self._db, Configuration.PATRON_WEB_CLIENT_URL).value = "http://sitewide/1234"
+            self._db, Configuration.PATRON_WEB_HOSTNAMES).value = list("http://sitewide/1234")
```
leonardr, line 448: This is being written to the database, so you can't set it to a list value; use json.dumps() to convert the list into a string first.

A couple lines below this, we set a per-library value for this setting (as we discussed, something that's not currently possible through the UI), and that value is set to a string instead of a list. For reasons I don't understand, this works fine -- probably the code that parses this value into a JSON list falls back to "a list containing a single string" if it can't parse the value. But that bit should also be changed.

I think the failure in `tests.test_custom_index.TestCustomIndexView.test_for_library` is caused by the failure of `test_load_settings`, which mocks some system code and crashes before it can un-mock it. When you fix this test the other failure should go away.

You can run just the failing tests like so:

```
nosetests tests/test_controller.py:TestCirculationManager.test_load_settings tests/test_custom_index.py:TestCustomIndexView.test_for_library
```

kristojorg: Thank you for the details here. Was just digging in to this, v helpful

kristojorg:
> A couple lines below this, we set a per-library value for this setting (as we discussed, something that's not currently possible through the UI), and that value is set to a string instead of a list. For reasons I don't understand, this works fine -- probably the code that parses this value into a JSON list falls back to "a list containing a single string" if it can't parse the value. But that bit should also be changed.

Are you referring to https://github.com/NYPL-Simplified/circulation/blob/d6d370409102df25efb6554a0b7def09e7ff6556/tests/test_controller.py#L454 or something else? I don't believe that should be a list

leonardr:
That's what I was referring to, and I think you're right. It's a different setting that's added to patron_web_domains in controller.py during setup:

```
        from registry import Registration
        for setting in self._db.query(
            ConfigurationSetting).filter(
            ConfigurationSetting.key==Registration.LIBRARY_REGISTRATION_WEB_CLIENT):
            if setting.value:
                patron_web_domains.add(get_domain(setting.value))
```

PR #1448, opened 2020-06-23
Make announcement configuration compatible with frontend code

adriana-alter:
I needed to make a few changes to make this work with the frontend code (I think we were initially expecting the front end to send the announcements in a different format, which turned out not to be compatible with the JavaScript FormData API?)--does this seem okay?

### DIFF
--- a/api/admin/controller/library_settings.py
+++ b/api/admin/controller/library_settings.py
@@ -321,13 +321,14 @@ class LibrarySettingsController(SettingsController):
             # Allow any entered values.
             value = []
             inputs = flask.request.form.getlist(setting.get("key")) if setting.get("type") == "list" else flask.request.form.get(setting.get("key"))
-            if json_objects:
+            if json_objects and inputs:
                 inputs = json.loads(inputs)
-            for i in inputs:
-                if not isinstance(i, list):
-                    i = [i]
-                value.extend(i)
-        
+            if inputs:
+                for i in inputs:
+                    if not isinstance(i, list):
+                        i = [i]
+                    value.extend(i)
+
         return json.dumps(filter(None, value))
 
     def image_setting(self, setting):
--- a/api/admin/controller/library_settings.py
+++ b/api/admin/controller/library_settings.py
@@ -50,12 +50,16 @@ class LibrarySettingsController(SettingsController):
 
             settings = dict()
             for setting in Configuration.LIBRARY_SETTINGS:
+                if setting.get("format") == "announcements":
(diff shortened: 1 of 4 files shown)


api/config.py
```diff
@@ -224,8 +224,7 @@ class Configuration(CoreConfiguration):
             "key": Announcements.SETTING_NAME,
             "label": _("Scheduled announcements"),
             "description": _("Announcements will be displayed to authenticated patrons."),
-            "category": "Basic Information",
-            "type": "list",
+            "category": "Announcements",
             "format": "announcements"
```
leonardr, line 228: Just recording the decision we made in our private conversation: you'll change this from format="announcements" to type="announcements". When we redo the configuration system, we'll use more specific terms like "ui_type" and "data_format".

api/admin/controller/library_settings.py
```diff
@@ -316,13 +320,15 @@ def list_setting(self, setting, json_objects=False):
...
             # Allow any entered values.
             value = []
-            inputs = flask.request.form.getlist(setting.get("key"))
-            for i in inputs:
-                if not isinstance(i, list):
-                    i = [i]
-                if json_objects:
-                    i = [json.loads(x) for x in i]
-                value.extend(i)
+            inputs = flask.request.form.getlist(setting.get("key")) if setting.get("type") == "list" else flask.request.form.get(setting.get("key"))
```
leonardr, line 323: Reformat this as a four-line if/else statement. This is tough to read, and since you're not in a list comprehension it doesn't all have to be one statement.

api/admin/controller/library_settings.py
```diff
@@ -50,12 +50,16 @@ def process_get(self):
...
+                    if value:
+                        value = AnnouncementListValidator().validate_announcements(value)
                 if setting.get("type") == "list":
                     value = ConfigurationSetting.for_library(setting.get("key"), library).json_value
                     if value and setting.get("format") == "geographic":
                         value = self.get_extra_geographic_information(value)
-                    if value and setting.get("format") == "announcements":
-                        value = AnnouncementListValidator().validate_announcements(value)
+                    # if value and setting.get("format") == "announcements":
+                    #     value = AnnouncementListValidator().validate_announcements(value)
```
leonardr, line 62: For the final version, remove the code rather than commenting it out.

leonardr:
Looks good after some minor changes.

adriana-alter:
I'll merge the `add-announcements-configuration-setting` branch into `master` once the front end is done, so I can update the version number.

PR #1449, opened 2020-06-24
Change web color theme settings to primary and secondary

kristojorg:
## Description

The new web interface uses "primary" and "secondary" colors instead of "background" and "foreground". These colors must now both contrast well on white. This PR updates the settings, along with their DB keys. It also configures the default settings to be a teal for primary and a red for secondary.

### Breaking Change

This is a breaking change. It deprecates the `web_foreground_color` and `web_background_color`, which will break any instances of circulation-patron-web currently running that update to the new circulation manager code. I have reached out to Micah at DPLA on slack but have not heard back from him RE if they are using CPW at all. This change will break the NYPL OpenEbooks, but NYPL is hopefully planning to migrate that to the new CPW code soon.

## Motivation and Context

The new version of CPW uses "primary" and "secondary" colors instead of "background" and "foreground". 

## How Has This Been Tested?

Just ran tests via travis.

## To Do:

- [x] Update documentation (can someone point me to where?).
- [x] Check with NYPL and DPLA if this breaking change is acceptable
- [x] Write a DB migration script

### DIFF
--- a/api/admin/controller/library_settings.py
+++ b/api/admin/controller/library_settings.py
@@ -163,8 +163,10 @@ class LibrarySettingsController(SettingsController):
             )
 
     def check_web_color_contrast(self, settings):
-        """Verify that the web primary color and web secondary
-        color go together.
+        """
+        Verify that the web primary and secondary color both contrast
+        well on white, as these colors will serve as button backgrounds with
+        white test, as well as text color on white backgrounds.
         """
         primary = flask.request.form.get(Configuration.WEB_PRIMARY_COLOR, Configuration.DEFAULT_WEB_PRIMARY_COLOR)
         secondary = flask.request.form.get(Configuration.WEB_SECONDARY_COLOR, Configuration.DEFAULT_WEB_SECONDARY_COLOR)
--- a/api/admin/controller/library_settings.py
+++ b/api/admin/controller/library_settings.py
@@ -175,7 +175,7 @@ class LibrarySettingsController(SettingsController):
             return tuple(int(hex[i:i+2], 16)/255.0 for i in (0, 2 ,4))
         primary_passes = wcag_contrast_ratio.passes_AA(wcag_contrast_ratio.rgb(hex_to_rgb(primary), hex_to_rgb("#ffffff")))
         secondary_passes = wcag_contrast_ratio.passes_AA(wcag_contrast_ratio.rgb(hex_to_rgb(secondary), hex_to_rgb("#ffffff")))
-        if not primary_passes and secondary_passes:
+        if not (primary_passes and secondary_passes):
             primary_check_url = "https://contrast-ratio.com/#%23" + secondary[1:] + "-on-%23" + "#ffffff"[1:] 
             secondary_check_url = "https://contrast-ratio.com/#%23" + secondary[1:] + "-on-%23" + "#ffffff"[1:]
             return INVALID_CONFIGURATION_OPTION.detailed(
--- a/api/admin/controller/library_settings.py
+++ b/api/admin/controller/library_settings.py
@@ -163,19 +163,22 @@ class LibrarySettingsController(SettingsController):
             )
 
     def check_web_color_contrast(self, settings):
-        """Verify that the web background color and web foreground
+        """Verify that the web primary color and web secondary
         color go together.
         """
-        background = flask.request.form.get(Configuration.WEB_BACKGROUND_COLOR, Configuration.DEFAULT_WEB_BACKGROUND_COLOR)
-        foreground = flask.request.form.get(Configuration.WEB_FOREGROUND_COLOR, Configuration.DEFAULT_WEB_FOREGROUND_COLOR)
+        primary = flask.request.form.get(Configuration.WEB_PRIMARY_COLOR, Configuration.DEFAULT_WEB_PRIMARY_COLOR)
+        secondary = flask.request.form.get(Configuration.WEB_SECONDARY_COLOR, Configuration.DEFAULT_WEB_SECONDARY_COLOR)
         def hex_to_rgb(hex):
             hex = hex.lstrip("#")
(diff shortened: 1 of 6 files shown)


kristojorg:
@leonardr @jonathangreen can you look this over for me? This is how I propose we update to the new color settings for CPW. I have yet to write the migration script and update documentation, so this is marked as draft for now.

api/authenticator.py
```diff
@@ -1092,12 +1092,12 @@ def create_authentication_document(self):
...
-        foreground = ConfigurationSetting.for_library(
-            Configuration.WEB_FOREGROUND_COLOR, library).value
-        if background or foreground:
-            doc["web_color_scheme"] = dict(background=background, foreground=foreground)
+        primary = ConfigurationSetting.for_library(
+            Configuration.WEB_PRIMARY_COLOR, library).value
+        secondary = ConfigurationSetting.for_library(
+            Configuration.WEB_SECONDARY_COLOR, library).value
+        if primary or secondary:
+            doc["web_color_scheme"] = dict(primary=primary, secondary=secondary)
```
leonardr, line 1100: You could restore backwards compatibility by putting the same colors in here under `background` and `foreground`. We can take it out later. Doing this would make it possible to merge the branch without waiting to hear from everyone involved.

api/admin/controller/library_settings.py
```diff
@@ -163,19 +163,22 @@ def check_for_missing_settings(self, settings):
...
-        foreground = flask.request.form.get(Configuration.WEB_FOREGROUND_COLOR, Configuration.DEFAULT_WEB_FOREGROUND_COLOR)
+        primary = flask.request.form.get(Configuration.WEB_PRIMARY_COLOR, Configuration.DEFAULT_WEB_PRIMARY_COLOR)
+        secondary = flask.request.form.get(Configuration.WEB_SECONDARY_COLOR, Configuration.DEFAULT_WEB_SECONDARY_COLOR)
         def hex_to_rgb(hex):
             hex = hex.lstrip("#")
             return tuple(int(hex[i:i+2], 16)/255.0 for i in (0, 2 ,4))
-        if not wcag_contrast_ratio.passes_AA(wcag_contrast_ratio.rgb(hex_to_rgb(background), hex_to_rgb(foreground))):
-            contrast_check_url = "https://contrast-ratio.com/#%23" + foreground[1:] + "-on-%23" + background[1:]
+        primary_passes = wcag_contrast_ratio.passes_AA(wcag_contrast_ratio.rgb(hex_to_rgb(primary), hex_to_rgb("#ffffff")))
+        secondary_passes = wcag_contrast_ratio.passes_AA(wcag_contrast_ratio.rgb(hex_to_rgb(secondary), hex_to_rgb("#ffffff")))
```
leonardr, line 175: I think the rationale for changing the test so that both colors contrast with white, rather than with each other, is that white is now always the background color. Is this right? If so, please add a comment to this effect.

api/authenticator.py
leonardr: You'd also need a migration script to move the old colors over rather than resetting everything to the default. I will need to be able to deploy a new Open eBooks circulation manager without first upgrading the web catalog, so even though this is a pain for you I'd like to see it in this branch.

leonardr: The migration script might mean that you start out with a primary and secondary color that don't pass the contrast-with-white test, which I think is better than the alternative. If you disagree, then I'm OK with skipping the migration script. So long as you keep `background` and `foreground` in `web_color_scheme` I can hack the database to get Open eBooks web catalog to continue working.

leonardr:
Re: updated documentation, the only relevant documentation here is the description of the settings, which you've updated.

leonardr:
The test failures should go away once you rename the fields in your tests.

api/authenticator.py
kristojorg: Okay I agree with you, we can add the old values at this point in the CM for backwards compat, but this will only work if you don't edit them, because once you edit them, they will need to pass the contrast with white check. We will also need to let people know that when they deploy a new instance of CPW, they need to update these settings in their CM if they had ones set previously.

api/admin/controller/library_settings.py
kristojorg: Yep, will do

kristojorg:
I renamed the fields but am still seeing one test failure that I don't quite know the source of. Also updated comment and added the backwards compatibility bits along with a migration script.

tests/test_authenticator.py
```diff
@@ -1276,8 +1276,8 @@ def annotate_authentication_document(
 
             # The mobile color scheme and web colors are correctly reported.
             eq_("plaid", doc['color_scheme'])
-            eq_("#012345", doc['web_color_scheme']['background'])
-            eq_("#abcdef", doc['web_color_scheme']['foreground'])
+            eq_("#377F8B", doc['web_color_scheme']['primaru'])
```
leonardr, line 1279: Typo here is causing another test failure.

leonardr:
Here's the test that's failing:

```
        # Test a bad contrast ratio between the web foreground and                                                                          
        # web background colors.                                                                                                            
        with self.request_context_with_admin("/", method="POST"):
            flask.request.form = self.library_form(
                library, {Configuration.WEB_BACKGROUND_COLOR: "#000000",
                Configuration.WEB_FOREGROUND_COLOR: "#010101"}
            )
            response = self.manager.admin_library_settings_controller.process_post()
            eq_(response.uri, INVALID_CONFIGURATION_OPTION.uri)
            assert "contrast-ratio.com/#%23010101-on-%23000000" in response.detail
```

Since the rule changed from "colors must contrast with each other" to "each color must contrast with white", a request that used to result in an error no longer does. Change this to verify that both the foreground color must contrast with white, then put it in a loop so that both foreground and background colors are tested.

kristojorg:
@leonardr the checks are finally passing 😓 . Thanks for your assist here

PR #1451, opened 2020-06-25
DB Migration fix

kristojorg:
## Description

This is a new migration to fix a typo in the #1447 migration script. Previously, we updated the value where key is `Patron Web Client` to a list, and then renamed the value of `Patron Web Url` to `patron_web_hostnames`. The first piece of this migration should have said `Patron Web Url` not `Patron Web Client`, so we ended up with `patron_web_hostnames` holding the original non-list value. This migration updates that value to a list.

## How Has This Been Tested?

I have run the migration locally.

## Checklist:


jonathangreen:
I'd like to close this PR in favour of #1453, which incorporates this fix and includes some additional fixed needed after doing some testing.

PR #1453, opened 2020-06-26
Fix cors setting

jonathangreen:
## Description

This fixes some issues introduced in #1447 and supersedes #1451. 

## Motivation and Context

In #1447 there were some issues with how the patron web hostname settings were saved. They are not saved as json, they are saved in text, so there are issues when we try to load them as json.

This PR backs out those changes, and instead saves the hostnames to the DB as a comma seperated list and breaks them out when the settings are loaded.

It also updates the help text to include the netloc, since this needs to be set for the CORS headers to work.

## How Has This Been Tested?

Tested this locally against the DB from one of our production servers that had issues after updating to #1447.

## Checklist:


jonathangreen:
@leonardr @tdilauro @kristojorg can you give this a review?

tests/test_controller.py
```diff
@@ -445,7 +445,7 @@ def mock_for_library(cls, incoming_library):
         # We also set up some patron web client settings that will
         # be loaded.
         ConfigurationSetting.sitewide(
-            self._db, Configuration.PATRON_WEB_HOSTNAMES).value = json.dumps(["http://sitewide/1234"])
+            self._db, Configuration.PATRON_WEB_HOSTNAMES).value = "http://sitewide/1234"
```
kristojorg, line 448: Should we be testing a comma separated string value now?

api/controller.py
```diff
@@ -289,10 +289,10 @@ def get_domain(url):
             return scheme + "://" + netloc
 
         sitewide_patron_web_client_urls = ConfigurationSetting.sitewide(
-            self._db, Configuration.PATRON_WEB_HOSTNAMES).json_value
+            self._db, Configuration.PATRON_WEB_HOSTNAMES).value
         if sitewide_patron_web_client_urls:
-            for url in sitewide_patron_web_client_urls:
-                patron_web_domains.add(get_domain(url))
+            for url in sitewide_patron_web_client_urls.split(','):
+                patron_web_domains.add(get_domain(url.strip()))
```
kristojorg, line 295: Two potential issues here:
1. A comma is a valid url value, though an apostrophe/quotation mark is not. It's unlikely they will need to pass in a url with a comma, but I suppose possible. I guess I assumed that processing it into a json list would avoid that by requiring that only quote-surrounded commas be considered delimiters. 
1. If you pass `'"beta.library.org","catalog.library.org"'`, won't we also need to strip the quotation marks from the string before passing it to `get_domain`?
1. Previously I believe you were able to specify `https://catalog.library.org` and thereby disallow `http://catalog.lobrary.org`, though I'm not sure that's the right mechanism for that protection.

tdilauro: > Two potential issues here:
> 
>     1. A comma is a valid url value, though an apostrophe/quotation mark is not. It's unlikely they will need to pass in a url with a comma, but I suppose possible. I guess I assumed that processing it into a json list would avoid that by requiring that only quote-surrounded commas be considered delimiters.

Yes, comma is a reserved character in the URI spec, so, while it's unlikely to occur in our URLs, I think we should just avoid it. I would suggest the pipe/vertical bar with any number of spaces around it.

>     2. If you pass `'"beta.library.org","catalog.library.org"'`, won't we also need to strip the quotation marks from the string before passing it to `get_domain`?

My own preference would be URLs without quotes (or other paired delimiters) around each of them.

tests/test_controller.py
tdilauro: Yes, though since I'm suggesting a different separator, I would say there should be a test for a list of URLs. :-)

jonathangreen: I added an additional test now that tests the parsing with the separator.

api/controller.py
jonathangreen: My preferences is also not to quote the string. I thought a comma for a separator was okay because its not allowed in the domain or scheme, but I think it makes sense to modify this to be a pipe character to be cautious. I updated the PR.

jonathangreen:
Added an additional commit addressing code review feedback. 

- Use pipes to delimit multiple sites
- Add some tests for values with pipes

I also ran through the migrations again in a test instance to make sure they apply correctly.

jonathangreen:
Thanks @EdwinGuzman!

leonardr:
I'm not arguing we should back this change out and go back to the #1447 system, but we have other configuration settings that are lists, and those are stored in the database correctly. Do you know what the underlying problem was?

jonathangreen:
I don't @leonardr. I can spend some more time looking into it though.

leonardr:
OK, not a big deal. The only reason to make this a priority would be it'll take another migration script if we end up changing it after the next release.

jonathangreen:
Hi @leonardr. Sorry for the delay on getting back to this, I was out the end of last week. 

So the issue here is that the `SitewideSettingsEditForm` component doesn't render lists. It assumes everything is a input, unless it has an options key set, in which case it renders as a select list. 

https://github.com/NYPL-Simplified/circulation-web/blob/master/src/components/SitewideSettingEditForm.tsx#L68-L96

So this setting was just displaying as a text field, and then when saved it was trying to load as JSON. 

Which would result in this exception on any page load: 
```
  File "libsimple/circulation/api/routes.py", line 43, in initialize_circulation_manager
    app.manager = CirculationManager(app._db)
  File "libsimple/circulation/api/controller.py", line 183, in __init__
    self.load_settings()
  File "libsimple/circulation/api/controller.py", line 292, in load_settings
    self._db, Configuration.PATRON_WEB_HOSTNAMES).json_value
  File "libsimple/circulation/core/model/configuration.py", line 720, in json_value
    return json.loads(self.value)
  File ".pyenv/versions/2.7.17/lib/python2.7/json/__init__.py", line 339, in loads
    return _default_decoder.decode(s)
  File ".pyenv/versions/2.7.17/lib/python2.7/json/decoder.py", line 367, in decode
    raise ValueError(errmsg("Extra data", s, end, len(s)))
ValueError: Extra data: line 1 column 4 - line 1 column 8 (char 3 - 7)
```

tdilauro:
There's the [hidden_content_types](https://github.com/NYPL-Simplified/circulation/blob/82d8d2e2a82b5916eb17321b1d7034338bf00a74/api/config.py#L332-L338) property (in Library configuration at the bottom of the `Client Interface Customization` section) that's supposed to capture a list from the [circulation-web](https://github.com/NYPL-Simplified/circulation-web) admin interface. Unfortunately, that field seems to be broken.

PR #1455, opened 2020-06-30
Fix OverDrive credential lookup for multiple collections

tdilauro:
## Description

This branch causes OverDrive patron credentials to be cached (and looked) up by collection.

Associated with [server_core PR #1182](https://github.com/NYPL-Simplified/server_core/pull/1182), which includes a database migration.

## Motivation and Context

I a patron's library has membership in more than one OverDrive collection and an unexpired OverDrive bearer token is cached, `OverdriveAPI.get_patron_credential` will return that credential, even if it is not the right one for the collection being acted upon. Because OverDrive associates that bearer token with the patron in the context of particular collection, any remote action undertaken with that token will be performed in the context of the cached credential, even though our intention was otherwise.

https://jira.nypl.org/browse/SIMPLY-2844

## How Has This Been Tested?

- Debugging/monitoring OverDrive self-tests and monitor scripts.
- Created new test that failed for this issue. It now passes.
- Ran full Circulation Manager test suite. All tests passed.

## Checklist:

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -250,7 +250,9 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI, HasSelfTests):
             return self.refresh_patron_access_token(
                 credential, patron, pin)
         return Credential.lookup(
-            self._db, DataSource.OVERDRIVE, "OAuth Token", patron, refresh)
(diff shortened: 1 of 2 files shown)


tdilauro:
Thanks!!

PR #1456, opened 2020-06-30
Deprecate Header Links and CSS File

kristojorg:
## Description

Circulation Patron Web is about to be launched into production at NYPL and Lyrasis as well as other organizations such as Amigos. At the moment the only other known production version is OpenEbooks at NYPL. Once CPW is in production, it will be very difficult to make breaking changes, so we should consider making them now. This PR is an attempt to clean up the settings that relate to CPW by deprecating two settings that no longer map properly to the current version of CPW.

This PR:
1. Removes support for custom CSS files for the web catalog.
2. Removes support for custom header links in the web catalog.

### CSS file motivation

With the new version of CPW, the CSS file route of style updates will no longer work. For those types of style changes, we will need to expose the them object from CPW in the circulation manager so that theme settings can be overridden. If more style changes are necessary than that allows, we will want to provide a path for libraries to wrap the application and override desired components in the file system. 

### Header links motivation

The Header links setting for circulation-patron-web was implemented for a previous version, seemingly with the intent of letting patrons navigate the main library website from the header of circulation-patron-web. This is no longer a goal of CPW, and instead we provide a link _back_ to the main library website where that navigation can be done natively. Thus, CPW is a separate app, still associated with the main library website, but not intending to mirror its functionality or appear to be the same system.

If we want to keep this functionality and allow libraries to add links to their catalog header, we will need to make significant UX considerations, as this quickly becomes messy. One possibly way to accommodate these extra links would be a dropdown in the header.

## How Has This Been Tested?

Only via Travis

## To Do:

- [ ] Get community approval
- [ ] Write DB migration script
- [ ] Update documentation
- [ ] Make sure tests pass

### DIFF
--- a/api/admin/controller/library_settings.py
+++ b/api/admin/controller/library_settings.py
@@ -137,7 +137,6 @@ class LibrarySettingsController(SettingsController):
         validations = [
             self.check_for_missing_fields,
             self.check_web_color_contrast,
-            self.check_header_links,
             self.validate_formats
         ]
         for validation in validations:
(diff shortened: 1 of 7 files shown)


kristojorg:
After speaking with Micah at DPLA, they are using the old version of CPW in production, so we cannot deprecate these settings currently. We can revisit if and when everybody is migrated to the new version of CPW.

PR #1458, opened 2020-07-01
S3: Add support for S3 content hosting

vbessonov:
## Description

**This is work in progress to illustrate the approach I chose.** 

This PR adds support for using presigned URLs to restrict access to collections stored in S3. Each time an authenticated patron tries to download the book Circulation Manager generates a new presigned expirable URL.

The main changes are:
- I added support for presigned URLs and new `LicensePool.self_hosted` field in [PR 1184](https://github.com/NYPL-Simplified/server_core/pull/1184) and [PR 1189](https://github.com/NYPL-Simplified/server_core/pull/1189)
- I added initialisation of `LicensePool.self_hosted` to `DirectoryImportScript`
- I added URL signing logic to `CirculationAPI.fulfil` method

## Motivation and Context

Please refer to [this document](https://docs.google.com/document/d/1W1QVpkhvaR63FD179PNAzncGNo6Yx2V2mwOhU0vMRv4/edit#) for more details.

## How Has This Been Tested?

Please refer to [this document](https://docs.google.com/document/d/1W1QVpkhvaR63FD179PNAzncGNo6Yx2V2mwOhU0vMRv4/edit#heading=h.aty8e64f1tsg).

## Checklist:


leonardr:
This branch talks about the books in terms of how they got into the collection. But `BaseCirculationAPI` is an API for delivering books to patrons, not for bringing books into the collection. It's not really relevant that the books got in through a manual import; they could have gotten through some other way.

Having a `BaseCirculationAPI` subclass makes sense for sure, but the distinguishing feature here is not that the books were imported manually, but that the books are being delivered through a `MirrorUploader` that can sign URLs. So I would call this something like `SelfHostedCirculationAPI` and instantiate it for any collection that stores its books using a MirrorUploader.

We also had a conversation about how the books you're importing are not being shown in OPDS feeds, probably because `Collection.restrict_to_ready_deliverable_works` filters them out: they are not open-access and their .licenses_owned is set to zero. There are a number of ways to solve this problem, including defining a new field called something like `LicensePool.self_hosted` which is set to true for any LicensePool -- open-access or not -- that was mirrored through a `MirrorUploader`.

jonathangreen:
In this PR we're re-purposing the *Open Access Content Bucket* to both host Open Access Content and Access Controlled Content. Instead of doing this, I think we should be adding a new bucket to the S3 configuration *Protected Access Content Bucket*. This way a CM is able to both host Open Access content that is totally open, and protected access content that will use signed URLs.

Then at the book import stage, we can choose if the content is open access or should be protected with signed URLs.

leonardr:
I agree that creating a new bucket for access-restricted content is a good idea.

PR #1459, opened 2020-07-08
Don't represent loans and holds for books that can't actually be delivered

leonardr:
## Description

For a while now we've given Overdrive audiobooks special treatment, controlled by a sitewide setting. This branch adds another bit of special treatment: loans and holds for Overdrive audiobooks will not be represented in the `loans` and `holds` tables. This will stop them from showing up in patrons' bookshelf feeds, where new versions of SimplyE might incorrectly believe that they are fulfillable through the SimplyE/circulation manager system. (How do they get into bookshelf feeds in the first place? Patrons take out loans for them using the Overdrive web site.)

This solution should work in the short term, while we're certifying that Overdrive audiobooks work with NYPL's collection, but other libraries should continue to treat Overdrive audiobooks as unfulfillable. It should also work in the long term, since the expectation is that every library with an Overdrive collection should have audiobooks enabled -- the existing sitewide setting will go away rather than become a per-library setting.

Since Overdrive is the source of truth regarding these loans, once the sitewide setting is removed, they'll immediately start showing up in the `loans` and `holds` tables, and in bookshelf feeds.

## Motivation and Context

This fixes https://jira.nypl.org/browse/SIMPLY-2869.

## How Has This Been Tested?

I borrowed an Overdrive audiobook with my NYPL card. The audiobook shows up in my bookshelf feed on NYPL's production server, but not on my local installation using NYPL's Overdrive credentials.

## Checklist:

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -1120,7 +1120,6 @@ class CirculationAPI(object):
         )
 
     def sync_bookshelf(self, patron, pin):
-
         # Get the external view of the patron's current state.
         remote_loans, remote_holds, complete = self.patron_activity(patron, pin)
 
@@ -1165,6 +1164,8 @@ class CirculationAPI(object):
             # This is a remote loan. Find or create the corresponding
             # local loan.
             pool = loan.license_pool(self._db)
+            if not self._remote_loans_representable_locally(pool):
+                continue
             start = loan.start_date
             end = loan.end_date
             key = (loan.identifier_type, loan.identifier)
@@ -1201,6 +1202,8 @@ class CirculationAPI(object):
             # This is a remote hold. Find or create the corresponding
             # local hold.
             pool = hold.license_pool(self._db)
+            if not self._remote_loans_representable_locally(pool):
+                continue
             start = hold.start_date
             end = hold.end_date
             position = hold.hold_position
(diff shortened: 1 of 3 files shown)


tdilauro:
Did you want to complete the `Description` above?

leonardr:
Yeah, I was just changing computers.

tdilauro:
Sorry. Jumped the gun.

tdilauro:
How does this compare to the functionality mediated by the per-library `hidden_content_types` setting?

leonardr:
Yeah, that's a good comparison, that's https://jira.nypl.org/browse/SIMPLY-1683. There was a similar problem with PDFs on Open eBooks. Since Overdrive audiobooks have their own media type, we might be able to solve this problem by making that media type a hidden content type.

leonardr:
It's really tempting to solve this with `hidden_content_types`, and I'll give that a try tomorrow. If this wasn't going away in the long term I'd consider this solution a hack, but special treatment for Overdrive audiobooks is also a hack.

leonardr:
I verified this this morning. This SQL will make the Overdrive audiobook format a hidden content type across the CM, which will stop SimplyE from thinking it can fulfill a book it actually can't:

```
update configurationsettings set value='["application/vnd.overdrive.circulation.api+json;profile=audiobook"]' where key='hidden_content_types' and external_integration_id is null and library_id is null;
```

Since we're not using this feature currently, there should be nothing in there now, but you can check with this SQL:

```
select * from configurationsettings where key='hidden_content_types' and external_integration_id is null and library_id is null;
```

Then once your CM is ready for Overdrive audiobooks you can run this to deactivate the feature:

```
update configurationsettings set value=null where key='hidden_content_types' and external_integration_id is null and library_id is null;
```

I'm reluctant to put that SQL into a migration script and force it on everybody, because I don't know everyone's individual situation, but I do think in general everyone should add `application/vnd.overdrive.circulation.api+json;profile=audiobook"` to their hidden_content_types ASAP, and remove it once all the libraries on their CM are set up with Overdrive audiobooks. If we can agree on that, then I don't think it's necessary to change any code.

tdilauro:
I have used this before in a single library scenario when I was having a problem with some Libby DRM content slipping through and it worked fine there. It would be very easy for us to deploy this change at LYRASIS.

Hey @jonathangreen! Any concerns from you on this?

EdwinGuzman:
The code looks good but I'm guessing we're going to wait on a consensus before moving forward with this PR?

leonardr:
Yeah, will hopefully close it without merging.

PR #1460, opened 2020-07-13
Refactor the Overdrive checkout process, unit-test it, and add one new bit of error checking.

leonardr:
## Description

This branch makes a really simple change, made more complicated by the fact that it touches a really important method -- the method that borrows books through Overdrive -- that didn't have test coverage. I refactored the method and added test coverage.

## Motivation and Context

We don't have a complete list of all the error messages Overdrive might send. Recently we discovered a new one which is affecting a lot of our patrons. A problem the patron could fix (they're at their loan limit and need to return something) is reported as a general "Loan failed" error that there's no indication how to fix.

This branch fixes https://jira.nypl.org/browse/SIMPLY-2791.

## How Has This Been Tested?

To make sure that my refactoring didn't break anything, I used a live Overdrive integration in the following test cases:

* Borrow a book
* Hit the same URL just used to borrow the book, trying to borrow it again. (To get this to work I had to comment out some code in controller.py, since we usually catch this case before going into Overdrive code.) This tests the most complicated part of the code: where we get an error from the Overdrive API but we're able to recover from it and act as though everything's normal.
* Try to borrow a book when at the patron loan limit. (I wasn't able to get the exact error message described in 2791, but I got `PatronHasExceededCheckoutLimit`, and they're now handled the same way, by using a dictionary to map the Overdrive message to an exception class.)

## Checklist:

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -153,6 +153,13 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI, HasSelfTests):
     # displayed to a patron, so it doesn't matter much.
     DEFAULT_ERROR_URL = "http://librarysimplified.org/"
 
+    # Map Overdrive's error messages to standard circulation manager
+    # exceptions.
+    ERROR_MESSAGE_TO_EXCEPTION = {
+        "PatronHasExceededCheckoutLimit": PatronLoanLimitReached,
+        "PatronHasExceededCheckoutLimit_ForCPC": PatronLoanLimitReached,
+    }
+
     def __init__(self, _db, collection):
         super(OverdriveAPI, self).__init__(_db, collection)
         self.overdrive_bibliographic_coverage_provider = (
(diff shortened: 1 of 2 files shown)


api/overdrive.py
```diff
@@ -153,6 +153,13 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI, HasSelfTests):
     # displayed to a patron, so it doesn't matter much.
     DEFAULT_ERROR_URL = "http://librarysimplified.org/"
 
+    # Map Overdrive's error messages to standard circulation manager
+    # exceptions.
+    ERROR_MESSAGE_TO_EXCEPTION = {
+        "PatronHasExceededCheckoutLimit": PatronLoanLimitReached,
+        "PatronHasExceededCheckoutLimit_ForCPC": PatronLoanLimitReached,
```
leonardr, line 160: This line is the only new functionality in this branch.

tests/test_overdrive.py
```diff
@@ -276,6 +277,205 @@ def test_scope_string(self):
...
+        assert_raises_regexp(
+            CannotLoan, "Unknown Error",
+            m, patron, pin, pool, dict(errorCodePresent=False)
+        )
+
+        # Some known errors become specific subclasses of CannotLoan.
+        assert_raises(PatronLoanLimitReached, with_error_code,
+                      "PatronHasExceededCheckoutLimit")
+        assert_raises(PatronLoanLimitReached, with_error_code,
+                      "PatronHasExceededCheckoutLimit_ForCPC")
```
leonardr, line 431: Here is the test of the new functionality; everything else is coverage that was previously missing.

PR #1464, opened 2020-07-21
SAML: Add a fallback for the case when IdP metadata does not contain display names

vbessonov:
## Description

This PR extends the algorithm used for generating SAML authentication documents:
1. Check whether `md:Extensions/mdui:UIInfo/mdui:DisplayName` tags exist. If yes, then populate `display_names` in the authentication document and stop.
2. Check whether `md:Organization/mdui:OrganizationDisplayName` tags exist. If yes, then populate `display_names` in the authentication document and stop.
3. Generate an IdP name using `SAMLConfiguration.IDP_DISPLAY_NAME_TEMPLATE` configuration setting

## Motivation and Context

Not all IdP conform with SAML best practices and don't include `mdui:DisplayName` tags inside their metadata. This PR adds several fallbacks for this case.

## Checklist:

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -2681,16 +2681,6 @@ class BaseSAMLAuthenticationProvider(AuthenticationProvider, BearerTokenSigner):
                             'urn:oasis:names:tc:SAML:2.0:bindings:HTTP-Redirect'),
            'category': 'IDP',
            'required': True
-       },
-       {
-           'key': SAMLConfiguration.IDP_DISPLAY_NAME_TEMPLATE,
-           'label': _('Identity Provider\'s display name template'),
-           'type': 'text',
-           'description': _('Template used to generate IdP display name shown in the UI in the case when '
-                            'IdP metadata does not contain neither mduid:DisplayName nor md:OrganizationDisplayName'),
-           'category': 'IDP',
-           'required': False,
-           'default': SAMLConfiguration.IDP_DISPLAY_NAME_DEFAULT_TEMPLATE
        }
     ] + AuthenticationProvider.SETTINGS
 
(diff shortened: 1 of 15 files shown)


vbessonov:
@jonathangreen, @tdilauro, do you think it's required to add additional flexibility or it's enough to hard code IdP display name template as `Identity Provider # {0}`?

vbessonov:
> This makes sense overall. My only concern is with the IDP_DISPLAY_NAME_TEMPLATE configuration setting. I don't think it provides a lot of value over hard-coding 'Identity Provider # {0}' as the template.
> 
> The most you'd be able to do is localize the phrase "Identity Provider". And to get that functionality we'd have to explain to end-users how the `format` syntax works, because we don't currently expose any templated functionality like that. (I was also concerned about security, but after researching `format` it won't be an issue in this case.) If you forgot to include the `{0}`, or got the wrong syntax, all of your IdPs would have the same name.
> 
> Is it likely that someone will realistically want to customize this template?

@leonardr, James asked to add a way to customize the template.
I can extend the description of `IDP_DISPLAY_NAME_TEMPLATE` and add an example:
```
The default template is "Identity Provider # {0}" where {0} will be substituted with an index of this particular IdP
```

leonardr:
I spoke with James about this and his main concern is that we look for an IdP display name in as many places as possible, which this branch does. (I don't know that it looks in _every_  possible place, but it's pretty impressive.) _If all of those attempts fail,_ he's OK with a hard-coded template `"Identity Provider #{0}"`.

Much more minor: while you're removing the configuration setting, also remove the space between `#` and `{0}`. `# 1` looks strange to me in a way that `#1` doesn't. But hopefully this will never show up in real usage.

PR #1465, opened 2020-07-22
Fix RBdigital fulfillment api changes

tdilauro:
## Description

This branch adds credential cacheing support for RBdigital's patron bearer token and uses that credential for fulfillment requests to RBdigital. RBdigital did not previously require the use of the bearer token, but now does. Because current clients don't support this approach, it was also necessary to rewrite client-bound manifests with proxy URLs that direct them to the fulfillment fixes implemented here in Circulation Manager.

In order to improve consistency and clarity for management of both long-term remote patron id and short-term patron bearer token credential types, this branch also refactors RBdigital credential cacheing into a more generalized method.

Because the patron bearer token is such a long-lived (24 hours) credential, if it were to be invalidated for some reason by RBdigital, the unexpired token might prevent the associated patron from being able to fulfill content for up to a day. For that reason, there is now logic to reauthorize and obtain and cache a new token, if a fulfillment request results in a 401. Reauthorization will be attempted only once per fulfillment request.

Finally, as an optimization, the patron bearer token will be captured if present in the create account response when a new RBdigital user is created.

## Motivation and Context

This change is needed because RBdigital changed its `media` fulfillment API such that it now requires a bearer token. If one is not provided, the API returns an authentication error, resulting in a `cannot-fulfill-loan` problem detail and a download error (ebooks) or silent failure (audiobooks) on the iOS client.

https://jira.nypl.org/browse/SIMPLY-2860

## How Has This Been Tested?

Testing was performed using the following methods:
- Tracing code in the debugger.
- Testing manifest and part fulfillment using Postman/curl.
- Running a full client interaction validation script similar to the [self-test-client](https://github.com/NYPL-Simplified/self-test-client).
- Adding some new tests and fixing others.
- Running all test in the test suite.

## Checklist:


api/rbdigital.py
```diff
@@ -142,6 +142,24 @@ class RBDigitalAPI(BaseCirculationAPI, HasSelfTests):
...
+    # - The `lifetime` is used to calculate Credential `expires`. If it
+    #   is None, then the Credential does not expire.
+    CREDENTIAL_TYPES = {
+        CACHED_IDENTIFIER_PROPERTY: dict(
+            label=Credential.IDENTIFIER_FROM_REMOTE_SERVICE,
+            lifetime=None
+        ),
+        BEARER_TOKEN_PROPERTY: dict(
+            label="Patron Bearer Token",
+            lifetime=(23 * 60 + 30) * 60),
```
leonardr, line 159: I figured out that this is 23 hours and 30 minutes but it was hard to read. Can you represent this either as `(24*60*60) * 0.9` or `((24*60) - 30) * 60`, and add a comment saying that it's slightly less than the lifetime RBDigital advertises?

api/rbdigital.py
```diff
@@ -1668,6 +1866,10 @@ def fulfill_part(self, part):
         :return: A FulfillmentInfo if the part could be fulfilled;
             a ProblemDetail otherwise.
         """
+        # TODO: This test is needed until we can convert this kwarg into an arg.
+        # if fulfillment_request is None:
+        #     raise ValueError("fulfillment_request parameter must be specified")
+
```
leonardr, line 1872: This comment seems like a relic of an earlier draft of this branch.

tests/test_rbdigital.py
```diff
@@ -241,6 +241,21 @@ def queue_initial_patron_id_lookup(self):
...
+        This method makes it easier and less error-prone to set up for that.
+        """
+        for filename in (
+            "response_patron_info_found.json",
+            "response_patron_internal_id_found.json",
+            "response_patron_bearer_token_success.json",
+        ):
+            datastr, datadict = self.api.get_data(filename)
+            self.api.queue_response(status_code=200, content=datastr)
+
```
leonardr, line 258: This is fine, but I want to make sure you know about a different solution to a similar problem in core/overdrive.py:MockOverdriveAPI: we override the method that asks for the token (`token_post`), eliminating the need to explicitly queue this stuff on most of the tests.

api/rbdigital.py
tdilauro: Good catch! Thanks.

tests/test_rbdigital.py
tdilauro: Oh, that's handy. I'll keep that in mind.

tdilauro:
Thanks, @leonardr. Addressed your comments. This is ready to merge, assuming tests pass.

PR #1467, opened 2020-07-23
SAML: Fix attribute parsing

vbessonov:
## Description

This PR enhances SAML attribute parsing, every time CM parses attribute's value it checks whether it contains `NameID`.

## Motivation and Context

`eduPersonTargetedID` usually contains `NameID` as a value and CM should parse it correctly.

## Checklist:

### DIFF
--- a/api/saml/parser.py
+++ b/api/saml/parser.py
@@ -8,7 +8,7 @@ from onelogin.saml2.utils import OneLogin_Saml2_Utils
 
 from api.saml.exceptions import SAMLError
 from api.saml.metadata import IdentityProviderMetadata, LocalizableMetadataItem, UIInfo, ServiceProviderMetadata, \
-    Binding, Service, NameIDFormat, Organization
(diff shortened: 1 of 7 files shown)


api/saml/parser.py
```diff
@@ -467,3 +467,89 @@ def parse(self, xml_metadata):
...
+        return name_id
+
+    def _parse_attribute_values(self, attribute_values):
+        """Parses attribute values
+
+        :param attribute_values: List containing SAML attribute values
+        :type attribute_values: List[Union[str, Dict]]
+
+        :return: 2-tupple containing a NameID and a list of SAML attribute values
+        :rtype: Tuple[Optional[NameID], List[str]]
```
leonardr, line 500: This docstring doesn't match the code; you're only returning the list of attribute values.

vbessonov: Updated it, thank you

PR #1468, opened 2020-07-27
SAML: Extract NameID containing in the list of SAML attributes

vbessonov:
## Description

This PR enhances the SAML parser and extracts a NameID containing in the list of SAML attributes.

## Motivation and Context

NYU's unique identifier is contained in eduPersonPrincipalName transmitted as the SAMLv2 NameID.

## Checklist:

### DIFF
--- a/api/saml/parser.py
+++ b/api/saml/parser.py
@@ -491,14 +491,16 @@ class SAMLSubjectParser(object):
         return name_id
 
     def _parse_attribute_values(self, attribute_values):
-        """Parses attribute values
+        """Parses SAML attribute values
 
         :param attribute_values: List containing SAML attribute values
         :type attribute_values: List[Union[str, Dict]]
 
-        :return: List of parsed SAML attribute values
-        :rtype: List[str]
+        :return: 2-tuple containing an optional name ID from the attribute list and
+            a list of parsed SAML attribute values
+        :rtype: Tuple[Optional[NameID, List[str]]
         """
+        name_id = None
         parsed_attribute_values = []
 
         for attribute_value in attribute_values:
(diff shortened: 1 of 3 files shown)


vbessonov:
@tdilauro, @jonathangreen, could you also have a look?

api/saml/parser.py
```diff
@@ -509,24 +511,34 @@ def _parse_attribute_values(self, attribute_values):
...
+        name_id = None
         parsed_attributes = []
         attribute_names = {attribute.value: attribute for attribute in SAMLAttributes}
 
         for name, attribute_values in attributes.iteritems():
             if name in attribute_names:
                 name = attribute_names[name].name
 
-            parsed_attribute_values = self._parse_attribute_values(attribute_values)
+            name_id, parsed_attribute_values = self._parse_attribute_values(attribute_values)
```
leonardr, line 534: This will behave unpredictably if there are multiple items in `attributes`. `name_id` might even end up `None` because an attribute with a name ID shows up in `iteritems` before an attribute that lacks one.

Overall your strategy has been to just find _some_ name ID _somewhere, so I think it would be enough to assign to `name_id` only if the value coming out of `_parse_attribute_values` is None.

vbessonov: I updated the code accordingly and added an additional tests covering the case with multiple SAML attributes

leonardr: I'm OK with this but the name is still unpredictable because the attributes ultimately come out of `attributes.iteritems()`. In Python 3 that order is predictable, in that it won't change over multiple runs, and I don't imagine this scenario happens a lot. My only real concern is that this will cause the unit tests to fail randomly.

tdilauro:
> @tdilauro, @jonathangreen, could you also have a look?

Aside from @leonardr's concern about potential intermittent test failures in Python 2, I don't have any concerns. That said, it may be harder (or more time-consuming) for someone else to figure this out later, so it would be preferable to address it now. Perhaps you could just sort `attributes.iteritems()` like:
```
    for name, attribute_values in sorted(attributes.iteritems()):
        ...
```

If so, would be helpful to add a `# TODO: Python 3 - "sorted" no longer needed` or something similar.

api/saml/parser.py
vbessonov: I can't imagine a case where it could fail. The code now overwrites NameID only in the case when it's not empty so it doesn't matter in what order we iterate attributes, i.e. it doesn't matter whether we get an attribute with NameID first or without

leonardr:
I'm OK with the code as given that there will only ever be one real NameID, we just don't know when it will show up.

PR #1469, opened 2020-07-28
Fix RBdigital audiobook part fulfillment using proxy service

tdilauro:
## Description

This branch adds a lightweight proxy service for RBdigital audiobook chapter/part `access document`s and rewrites audiobook manifest links to use that proxy service. The proxied URLs carry a bearer token and the URL of the access document with the API base URL removed. The proxy reconstructs the URL and fulfills it by authenticating to the RBDigital endpoint with the bearer token. This approach increases  security in two ways:
- There is not enough context in the proxy URLs to fullfill content outside of this system.
- If an arbitrary URL is submitted to the service, it will have the API base URL prepended to it, rendering it useless, in practice.

## Motivation and Context

This change is needed because RBdigital changed its `media` fulfillment API such that it now requires a bearer token. If one is not provided, the API returns an authentication error, resulting in a `cannot-fulfill-loan` problem detail and a download error (ebooks) or silent failure (audiobooks) on the iOS client. This branch addresses issues unique to audiobooks. The eBooks issue was resolved with the earlier PR #1465.

The proxy service is needed because the mobile clients currently expect to be able to dereference a URL without providing any authentication to fulfill this kind of link, so everything needed for fulfillment must be encoded in the URL. Additionally, the lead time to change the clients is relatively long, so a solution that could be implemented entirely in Circulation Manger made sense for the time being.

https://jira.nypl.org/browse/SIMPLY-2860

## How Has This Been Tested?

Testing was performed using the following methods:
- Tracing code in the debugger.
- Testing manifest and part fulfillment using Postman/curl.
- Running a full client interaction validation script similar to the [self-test-client](https://github.com/NYPL-Simplified/self-test-client).
- Fixed existing tests.
- Running all test in the test suite.

## Checklist:


api/rbdigital.py
```diff
@@ -2616,7 +2648,6 @@ def import_spine(self, file_data, proxy_url):
         extra = {}
         extra['proxy_link'] = dict(
             href=proxy_url,
-            type=Representation.guess_media_type(filename)
```
leonardr, line 2619: Why take this out?

tdilauro: We are no longer returning an `alternate` URL, but a proxy link. The clients are expecting to see `vnd.librarysimplified/rbdigital-access-document+json` for an RBdigital access document, which is what the proxy would return. `Representation.guess_media_type(filename)` would return `audio/mpeg`, which would lead the clients to try to download and stash an access document as the content, rather than using it as a pointer to the content.

leonardr: That makes sense, thanks.

api/routes.py
```diff
@@ -394,6 +394,11 @@ def annotations_for_work(identifier_type, identifier):
 def borrow(identifier_type, identifier, mechanism_id=None):
     return app.manager.loans.borrow(identifier_type, identifier, mechanism_id)
 
+@library_route('/works/<license_pool_id>/fulfill/<mechanism_id>/<part>/rbdproxy/<bearer>')
+@has_library
+def proxy_rbdigital_patron_requests(license_pool_id, mechanism_id, part, bearer):
+    return app.manager.rbdproxy.proxy(bearer)
+
```
leonardr, line 401: Since license_pool_id, mechanism_id, and part are not used, can we remove them from the URL?

tdilauro: We could. I left them in so that the URLs would look similar to the fulfillment URLs in any logs, but we don't need them at all to achieve the functionality.

leonardr: Since you're the main audience for this branch I'm OK with whatever makes things easy for you.

tdilauro:
I'm actually working on a few tests for that right now. I just wanted to get the PR into review.

tdilauro:
This is ready to go.

PR #1470, opened 2020-08-05
Patron last activity sync

leonardr:
## Description

This is the circulation portion of the work for https://jira.nypl.org/browse/SIMPLY-2922. (Core is [here](https://github.com/NYPL-Simplified/server_core/pull/1190).) It has four main components:

* When generating a patron's bookshelf feed, we only call out to external vendors if Patron.last_loan_activity_sync is null. (See https://github.com/NYPL-Simplified/server_core/pull/1190 for the rules about when it's null.)

* We clear out Patron.last_loan_activity_sync when revoking a loan or hold.

* The bookshelf feed response includes a Last-Modified header with the value of Patron.last_loan_activity_sync. (This value is always set by the time the bookshelf feed has been generated.)

* The controller that generates the bookshelf feed supports conditional HTTP GET. A client that sends the Last-Modified header back as If-Last-Modified will get a 304 (Not Modified) response if the CM believes there has been no loan activity since then.

## Motivation and Context

The general idea is to allow a client to hit a patron's bookshelf feed whenever it wants, without creating a performance problem on the server side. We address performance with (very limited) server-side caching, and support for conditional HTTP GET.

## Checklist:

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -596,7 +596,7 @@ class CirculationManagerController(BaseCirculationManagerController):
         return search_engine
 
     def handle_conditional_request(self, last_modified=None):
-        """Handle conditional HTTP requests.
+        """Handle a conditional HTTP request.
 
         :param last_modified: A datetime representing the time this
            resource was last modified.
@@ -1182,12 +1182,18 @@ class LoanController(CirculationManagerController):
         """
         patron = flask.request.patron
 
+        # Save some time if we don't believe the patron's loans or holds have
+        # changed since the last time the client requested this feed.
         response = self.handle_conditional_request(
             patron.last_loan_activity_sync
         )
         if isinstance(response, Response):
             return response
 
+        # TODO: SimplyE used to make a HEAD request to the bookshelf feed
+        # as a quick way of checking authentication. Does this still happen?
+        # It shouldn't -- the patron profile feed should be used instead.
+        # If it's not used, we can take this out.
         if flask.request.method=='HEAD':
             return Response()
 
--- a/api/controller.py
+++ b/api/controller.py
@@ -1,4 +1,5 @@
 from nose.tools import set_trace
+import email
 import json
 import logging
 import sys
@@ -607,15 +608,19 @@ class CirculationManagerController(BaseCirculationManagerController):
         if not last_modified:
             return None
 
+        # TODO: This can be cleaned up significantly in Python 3.
         if_modified_since = flask.request.headers.get('If-Modified-Since')
-        parsed_if_modified_since = email.utils.parsedate_to_datetime(
+        if_modified_since_tuple = email.utils.parsedate(
             if_modified_since
+        )
+        parsed_if_modified_since = datetime.datetime.fromtimestamp(
+            mktime(if_modified_since_tuple)
         )
         if not parsed_if_modified_since:
             return None
 
         if parsed_if_modified_since >= last_modified:
-            return Response(status_code=304)
+            return Response(status=304)
         return None
 
     def load_lane(self, lane_identifier):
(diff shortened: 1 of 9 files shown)


tests/saml/test_controller.py
```diff
@@ -238,7 +238,7 @@ def test_saml_authentication_redirect(
                 })
             },
             None,
-            (create_autospec(spec=Credential), create_autospec(spec=Patron), create_patron_data_mock()),
+            (create_autospec(spec=Credential), object(), create_patron_data_mock()),
```
leonardr, line 241: Adding the `last_loan_activity_sync` dynamic property to `Patron` broke this `create_autospec` call. I don't understand the details of how `create_autospec` works but fortunately this test doesn't need a detailed mock of a Patron; it just needs something to fill the space.

api/controller.py
```diff
@@ -1157,11 +1191,23 @@ def sync(self):
...
+        response = self.handle_conditional_request(
+            patron.last_loan_activity_sync
+        )
+        if isinstance(response, Response):
+            return response
+
+        # TODO: SimplyE used to make a HEAD request to the bookshelf feed
+        # as a quick way of checking authentication. Does this still happen?
+        # It shouldn't -- the patron profile feed should be used instead.
+        # If it's not used, we can take this out.
```
tdilauro, line 1207: I think it would be best to leave this in, since we may make such a request in test scripts and the like. In general, I think we should not assume that SimplyE is the only client to a CM.

tests/test_circulationapi.py
```diff
@@ -898,27 +898,33 @@ def yes_we_can(*args, **kwargs):
         result = try_to_fulfill()
         eq_(fulfillment, result)
 
-    def test_revoke_loan_sends_analytics_event(self):
+    def test_revoke_loan(self):
+        self.patron.last_loan_activity_sync = datetime.utcnow()
```
tdilauro, line 902: Using `utcnow()` here could cause the test to fail if, in the future, `loan_activity_max_age` is set to a very small positive value or zero.

tests/test_circulationapi.py
```diff
@@ -898,27 +898,33 @@ def yes_we_can(*args, **kwargs):
...
+        eq_(None, self.patron.last_loan_activity_sync)
+
         # An analytics event was created.
         eq_(1, self.analytics.count)
         eq_(CirculationEvent.CM_CHECKIN,
             self.analytics.event_type)
 
-    def test_release_hold_sends_analytics_event(self):
+    def test_release_hold(self):
+        self.patron.last_loan_activity_sync = datetime.utcnow()
```
tdilauro, line 918: Using `utcnow()` here could cause the test to fail if, in the future, `loan_activity_max_age` is set to a very small positive value or zero.

tests/test_circulationapi.py
tdilauro, line 902: Using `utcnow()` here could cause the test to be ineffective if, in the future, `loan_activity_max_age` is set to a very small positive value or zero. In that case, the cache might expire before `revoke_loan` "invalidates" it, in which case we wouldn't know who was responsible for the invalidation.

tests/test_circulationapi.py
tdilauro, line 918: Similar here: Using `utcnow()` here could cause the test to be ineffective if, in the future, `loan_activity_max_age` is set to a very small positive value or zero. In that case, the cache might expire before `release_hold` "invalidates" it, in which case we wouldn't know who was responsible for the invalidation.

tests/test_circulationapi.py
```diff
@@ -1221,6 +1238,52 @@ def test_sync_bookshelf_applies_locked_delivery_mechanism_to_loan(self):
         self._db.commit()
         assert loan.fulfillment in pool.delivery_mechanisms
 
+    def test_sync_bookshelf_respects_last_loan_activity_sync(self):
```
tdilauro, line 1241: I love this test. It reads like a mystery.

tests/test_circulationapi.py
tdilauro: There are a couple more that are similar.

PR #1472, opened 2020-08-06
Add LCP support

vbessonov:
## Description

This PR adds LCP support to Circulation Manager.

The main changes are:
1. `lcp` module
2. ONIX parser enhancements
3. DirectoryImportScript enhancements

Linked PRs:
- [PR # 1194](https://github.com/NYPL-Simplified/server_core/pull/1194) in `core`
- [PR # 125](https://github.com/NYPL-Simplified/circulation-docker/pull/125) in `circulation-docker`
- [PR # 279](https://github.com/NYPL-Simplified/opds-web-client/pull/279) in `opds-web-client`

## Checklist:


vbessonov:
The import process is implemented in `LCPMirror`, it takes care of encrypting books, sending them to the LCP License Server and then mirroring them in S3 by it

api/lcp/collection.py
```diff
@@ -0,0 +1,311 @@
...
+        LCPServerConfiguration.lcpserver_user.to_settings(),
+        LCPServerConfiguration.lcpserver_password.to_settings(),
+        LCPServerConfiguration.lcpserver_input_directory.to_settings(),
+        LCPServerConfiguration.lcpserver_page_size.to_settings(),
+        LCPServerConfiguration.provider_name.to_settings(),
+        LCPServerConfiguration.passphrase_hint.to_settings(),
+        LCPServerConfiguration.encryption_algorithm.to_settings(),
+        LCPServerConfiguration.max_printable_pages.to_settings(),
+        LCPServerConfiguration.max_copiable_pages.to_settings()
+    ]
```
leonardr, line 101: As I mentioned in the core branch, I would expect this to be more like 

```
SETTINGS = [
 [member.to_settings() for name, member in ConfigurationMetadata.get_configuration_metadata(LCPEncryptionConfiguration)]
+
[member.to_settings() for name, member in ConfigurationMetadata.get_configuration_metadata(LCPServerConfiguration)]
]
```

Or something even simpler with an additional helper function.

api/lcp/collection.py
```diff
@@ -0,0 +1,311 @@
...
+        else:
+            license = self._lcp_server.generate_license(
+                self._db,
+                licensepool.identifier.identifier,
+                patron.id,
+                today,
+                expires
+            )
+
+            patron.external_identifier = license['encryption']['user_key']['key_check']
```
leonardr, line 209: `patron.external_identifier` is used for the unique identifier assigned to the patron by the ILS, so we can track the patron across changes to their library card number. It's not appropriate to use it to store any kind of information from an LCP server. 

I'm not 100% sure what this value is for, but since it differs from one loan to another it would be appropriate to store it as `loan.external_identifier`. But you're using that for `license['id']`, which makes more sense.

Is it necessary to store this value at all?

api/lcp/collection.py
```diff
@@ -0,0 +1,311 @@
...
+        :param patron: A Patron object for the patron who wants to check out the book
+        :type patron: Patron
+
+        :param pin: The patron's alleged password
+        :type pin: string
+
+        :return: List of patron's loans
+        :rtype: List[LoanInfo]
+        """
+        # NOTE: We assume that there can be only ONE LCP collection per library
```
leonardr, line 279: If there were multiple LCP collections per library I'd expect each one to end up with a distinct `LCPAPI`. `patron_activity` would be called on each `LCPAPI` in parallel. So I don't think you're actually making this assumption.

api/lcp/collection.py
```diff
@@ -0,0 +1,311 @@
...
+                start_date=loan.start,
+                end_date=loan.end,
+                fulfillment_info=None,
+                external_identifier=loan.external_identifier
+            ))
+
+        return loan_info_objects
+
+    def place_hold(self, patron, pin, licensepool, notification_email_address):
+        pass
```
leonardr, line 308: Add a comment explaining whether we expect to implement this later.

api/lcp/encrypt.py
```diff
@@ -0,0 +1,499 @@
...
+                self.protected_content_location,
+                self.protected_content_length,
+                self.protected_content_sha256,
+                self.protected_content_disposition,
+                self.protected_content_type
+            )
+
+
+class LCPEncryptorResultJSONEncoder(JSONEncoder):
+    """Subject's JSON encoder"""
```
leonardr, line 239: What does "subject" mean here? We have a data model object called `Subject` but this isn't referring to that.

api/lcp/encrypt.py
```diff
@@ -0,0 +1,499 @@
...
+        :rtype: LCPEncryptionResult
+        """
+        with self._configuration_factory.create(
+                self._configuration_storage, db, LCPEncryptionConfiguration) as configuration:
+            if self._lcpencrypt_exists_locally(configuration):
+                result = self._run_lcpencrypt_locally(file_path, identifier, configuration)
+
+                return result
+            else:
+                raise NotImplementedError()
```
leonardr, line 499: `NotImplementedError` is for cases where an abstract or unfinished method is called. Here the code is implemented but the configuration isn't complete, so a `ValueError` or `core.config.CannotLoadConfiguration` would be a more appropriate exception.

api/lcp/server.py
```diff
@@ -0,0 +1,351 @@
...
+            'URI that identifies the provider in an unambiguous way'
+        ),
+        type=ConfigurationAttributeType.TEXT,
+        required=True
+    )
+
+    passphrase_hint = ConfigurationMetadata(
+        key='passphrase_hint',
+        label=_('Passphrase hint'),
+        description=_('Hint proposed to the user for selecting its passphrase'),
```
leonardr, line 81: "its" should be "their".

api/lcp/server.py
```diff
@@ -0,0 +1,351 @@
...
+        default=DEFAULT_PASSPHRASE_HINT
+    )
+
+    encryption_algorithm = ConfigurationMetadata(
+        key='encryption_algorithm',
+        label=_('Passphrase encryption algorithm'),
+        description=_('Algorithm used for encrypting the passphrase'),
+        type=ConfigurationAttributeType.TEXT,
+        required=False,
+        default=DEFAULT_ENCRYPTION_ALGORITHM
```
leonardr, line 93: How many different encryption algorithms are there? This should be a selection with the valid values prepopulated, not a freeform text field.

api/lcp/server.py
```diff
@@ -0,0 +1,351 @@
...
+
+
+class LCPServerConfiguration(ConfigurationBucket):
+    """Contains LCP License Server's settings"""
+
+    DEFAULT_PAGE_SIZE = 100
+    DEFAULT_PASSPHRASE_HINT = 'If you do not remember your passphrase, please contact your administrator'
+    DEFAULT_ENCRYPTION_ALGORITHM = HashingAlgorithm.SHA256.value
+    DEFAULT_MAX_PRINTABLE_PAGES = 10
+    DEFAULT_MAX_COPIABLE_PAGES = 2048
```
leonardr, line 22: The defaults given in https://readium.org/lcp-specs/releases/lcp/latest.html#36-identifying-rights-and-restrictions-the-rights-object are for both of these values to be unlimited; those should be the defaults here as well.

api/lcp/server.py
```diff
@@ -0,0 +1,351 @@
...
+
+        :param json_encoder: JSON encoder
+        :type json_encoder: JSONEncoder
+
+        :return: Dictionary containing LCP License Server's response
+        :rtype: Dict
+        """
+        json_payload = json.dumps(payload, cls=json_encoder)
+        url = urlparse.urljoin(configuration.lcpserver_url, path)
+        response = requests.request(
```
leonardr, line 228: Usually we make the HTTP request in a method that does nothing else, so we can mock it during tests. But based on your changes to requirements.txt I'm guessing you mock the requests library itself.

api/lcp/server.py
```diff
@@ -0,0 +1,351 @@
...
+        """
+        with self._configuration_factory.create(
+                self._configuration_storage, db, LCPServerConfiguration) as configuration:
+            page_index = 1
+            licenses = []
+
+            while True:
+                path = 'licenses?page={0}&per_page={1}'.format(page_index, configuration.lcpserver_page_size)
+                response = self._send_request(configuration, 'get', path, payload=None)
+                new_licenses = response.json()
```
leonardr, line 340: This may not be under your control, but how does the pagination work here? If licenses are added to the list while you're iterating, you will miss some. Is this a problem? I don't see `get_licenses` used elsewhere in this code.

api/lcp/mirror.py
```diff
@@ -0,0 +1,151 @@
...
+
+    def book_url(self, identifier, extension='.epub', open_access=False, data_source=None, title=None):
+        """Returns the path to the hosted EPUB file for the given identifier."""
+        bucket = self.get_bucket(
+            S3UploaderConfiguration.OA_CONTENT_BUCKET_KEY if open_access
+            else S3UploaderConfiguration.PROTECTED_CONTENT_BUCKET_KEY)
+        root = self.content_root(bucket)
+        book_url = root + self.key_join([identifier.identifier])
+
+        return book_url
```
leonardr, line 107: Is this right? This seems like code copied from the S3 integration. I don't think this type of mirror _has_ a book_url and `NotImplementedError` would be appropriate.

api/lcp/server.py
```diff
@@ -0,0 +1,351 @@
...
+            'rights': {
+                'print': int(configuration.max_printable_pages),
+                'copy': int(configuration.max_copiable_pages)
+            }
+        }
+
+        if patron_id:
+            partial_license['user'] = {
+                'id': str(patron_id)
+            }
```
leonardr, line 199: I don't want to expose our internal database IDs to an external service. It would be more appropriate to create a permanent `Credential` object for a given `Patron` and `Collection`, and pass `Credential.credential` to the LCP server. The `type` of the collection could be "User identiifer passed to LCP server".

See `patron_remote_identifier` in api/rbdigital.py for a similar example: we create an alias for the patron to use when we talk about them to the RBDigital API.

tests/lcp/test_collection.py
```diff
@@ -0,0 +1,516 @@
...
+                configuration.encryption_algorithm = LCPServerConfiguration.DEFAULT_ENCRYPTION_ALGORITHM
+                configuration.max_printable_pages = LCPServerConfiguration.DEFAULT_MAX_PRINTABLE_PAGES
+                configuration.max_copiable_pages = LCPServerConfiguration.DEFAULT_MAX_COPIABLE_PAGES
+
+                # Act
+                license_pool.loan_to(patron, start=today, end=expires, external_identifier=lcp_license['id'])
+                fulfilment_info = lcp_api.fulfill(patron, 'pin', license_pool, 'internal format')
+
+                # Assert
+                eq_(fulfilment_info.collection_id, self._lcp_collection.id)
```
leonardr, line 475: Also verify that `fulfilment_info` is an `LCPFulfilmentInfo` rather than some other kind.

api/lcp/collection.py
```diff
@@ -0,0 +1,311 @@
...
+        """
+        # NOTE: We assume that there can be only ONE LCP collection per library
+        loans = self._db\
+            .query(Loan)\
+            .join(LicensePool)\
+            .join(Collection)\
+            .filter(
+                Loan.patron == patron,
+                Collection.id == self._collection_id
+            )
```
leonardr, line 287: This will find loans that have expired but have not yet been reaped, or which (not likely to happen) have not started yet. Is that OK?

tests/lcp/test_collection.py
```diff
@@ -0,0 +1,516 @@
...
+        external_identifier = '1'
+        license_pool = self._licensepool(
+            edition=None, data_source_name=data_source_name, collection=self._lcp_collection)
+        license_pool.loan_to(patron, start=today, end=expires, external_identifier=external_identifier)
+
+        other_collection = self._collection(protocol=ExternalIntegration.MANUAL)
+        other_external_identifier = '2'
+        other_license_pool = self._licensepool(
+            edition=None, data_source_name=data_source_name, collection=other_collection)
+        other_license_pool.loan_to(patron, start=today, end=expires, external_identifier=other_external_identifier)
```
leonardr, line 500: This test should also create other loans that ought not to show up in the list:

* A loan that's out to some other patron than `patron`.
* A loan that has expired (assuming you don't want to include those -- see https://github.com/NYPL-Simplified/circulation/pull/1472/files#r476572493)
* Optionally, a loan that hasn't started yet (assuming you don't want to include those).

tests/lcp/test_server.py
```diff
@@ -0,0 +1,272 @@
...
+                eq_(json_request['content-encryption-key'], encrypted_content.content_encryption_key)
+                eq_(json_request['protected-content-location'], expected_protected_content_disposition)
+                eq_(json_request['protected-content-disposition'], encrypted_content.protected_content_disposition)
+                eq_(json_request['protected-content-type'], encrypted_content.protected_content_type)
+                eq_(json_request['protected-content-length'], encrypted_content.protected_content_length)
+                eq_(json_request['protected-content-sha256'], encrypted_content.protected_content_sha256)
+
+    def test_generate_license(self):
+        # Arrange
+        patron_id = '1'
```
leonardr, line 85: When you change the code that generates the patron's `Credential` object for identifying them to an LCP server for a given collection, either this test will need to be changed or you'll need a separate test just for the `Credential` generation.

api/lcp/collection.py
vbessonov: `license['encryption']['user_key']['key_check']` contains patron's hashed passphrase, I need to store it somewhere so I can later [add it to OPDS feed](https://github.com/NYPL-Simplified/server_core/pull/1194/files#diff-eb69c9549013acc45b217e06734d0ec9R1522)

vbessonov: `loan.external_identifier` is already used to store license's ID which is needed to be able to fetch an existing license by its ID.
What would be the most appropriate place where I can store patron's hashed passphrase?

api/lcp/encrypt.py
vbessonov: In the beginning I was planning to add support for running `lcpencrypt` in a Docker container, this `else` branch should cover this case

api/lcp/mirror.py
vbessonov: `LCPMirror` has to have `book_url` implemented because it used during the import process. CM first calls `LCPMirror.book_url` and then passes its result to `LCPMirror.mirror_one`

api/lcp/server.py
vbessonov: Yes, it's much easier to mock using `requests-mock` library. For example:
```python
            with requests_mock.Mocker() as request_mock:
                url = urlparse.urljoin(fixtures.LCPSERVER_URL, '/contents/{0}/license'.format(fixtures.CONTENT_ID))
                request_mock.post(url, json=fixtures.LCPSERVER_LICENSE)
```

api/lcp/collection.py
vbessonov: I don't think it's ok. As far as I understand `patron_activity` should return only active loans. What is the best to filter out non-active loans? Should it be something like that?
```python
loans = self._db\
            .query(Loan)\
            .join(LicensePool)\
            .join(Collection)\
            .filter(
                Loan.patron == patron,
                Loan.fulfillment_id != None,
                or_(
                    Loan.start is None,
                    Loan.start >= func.now()
                 ),
                 or_(
                    Loan.end is None,
                    Loan.end < func.now()
                 ),
                Collection.id == self._collection_id
            )
```

api/lcp/collection.py
vbessonov: Changed it to `SETTINGS = LCPServerConfiguration.to_settings() + LCPEncryptionConfiguration.to_settings()`

api/lcp/collection.py
vbessonov: Removed the comment

api/lcp/encrypt.py
vbessonov: Copy & paste error, updated the docstring

api/lcp/server.py
vbessonov: Fixed it

api/lcp/server.py
vbessonov: Changed the defaults to `None`

api/lcp/server.py
```diff
@@ -0,0 +1,351 @@
...
+    )
+
+    max_printable_pages = ConfigurationMetadata(
+        key='max_printable_pages',
+        label=_('Maximum number or printable pages'),
+        description=_(
+            'Maximum number of pages that can be printed over the lifetime of the license. '
+            'The default value is {0}'.format(
+                DEFAULT_MAX_PRINTABLE_PAGES
+            )
```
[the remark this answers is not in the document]
vbessonov: Updated the description

api/lcp/server.py
vbessonov: Add `Credential` generation using `uuid` module

tests/lcp/test_server.py
vbessonov: Updated the test

api/lcp/server.py
vbessonov: I changed type to `ConfigurationAttributeType.SELECT`

api/lcp/server.py
vbessonov: That's a very good question. So far `get_licenses` is not used, I was planning to use in `LCPAPI.patron_activity` but ended up with querying loans from the database

tests/lcp/test_collection.py
vbessonov: Added an assertion

tests/lcp/test_collection.py
vbessonov: Added additional test cases

api/lcp/collection.py
vbessonov: Also, I'm thinking about what should be a passphrase. Currently I'm using `LCPServer.DEFAULT_PATRON_PASSPHRASE` because we rely on [automatic key retrieval feature](https://readium.org/lcp-specs/releases/lcp/latest.html#36-identifying-rights-and-restrictions-the-rights-object) and patrons should never enter their passphrases manually but on the other hand patrons will not be able to use other devices not supporting this feature. Also, it might be not very safe to use the same passphrase for all patrons and books. Does it make sense to generate it dynamically as a UUID for example?

api/lcp/collection.py
vbessonov: Could you tell me what would happen if I don't implement those methods? Circulation Manager will be throwing exceptions when there are no available licenses?

leonardr: Yes, I think you'll get unhandled exceptions, because `CirculationAPI` assumes these methods either raise an exception or return a `HoldInfo`. (This is a problem in enki.py as well, now that I look at it... I'm not sure if Enki books ever lose all their available licenses though.)

If you plan to add this later then I think let these methods inherit the `NotImplementedError` from `BaseCirculationAPI` and file a ticket to finish the work.

api/lcp/collection.py
leonardr: https://readium.org/lcp-specs/releases/lcp/latest.html#44-requirements-for-the-user-key--user-passphrase says: "the Provider should use the same User Key for all licenses issued to the same User." Since the User Key is derived from the User Passphrases I'm OK with having a single User Passphrase.

So, store the User Passphrase for a patron in a permanent `Credential` for that patron. The `Credential` should have data source DataSource.INTERNAL_PROCESSING and a new type, maybe `LCP Passphrase`. The value can be a UUID. 

Another way to do this is to use `Patron.authorization_identifier` as the User Passphrase. Since the user knows this value they'll be able to type it in in another reader.  But patrons do change their authorization_identifiers occasionally and I don't know if patrons who authenticate through SAML will have an authorization_identifier, so I think a UUID is safer.

api/lcp/encrypt.py
leonardr: OK, just make sure there's a ticket for finishing this work.

api/lcp/server.py
leonardr: OK, take out the code and its tests since we're not using it.

api/lcp/collection.py
leonardr: Yes, that looks good.

api/lcp/collection.py
```diff
@@ -78,27 +81,14 @@ def as_response(self):
...
-        LCPServerConfiguration.lcpserver_password.to_settings(),
-        LCPServerConfiguration.lcpserver_input_directory.to_settings(),
-        LCPServerConfiguration.lcpserver_page_size.to_settings(),
-        LCPServerConfiguration.provider_name.to_settings(),
-        LCPServerConfiguration.passphrase_hint.to_settings(),
-        LCPServerConfiguration.encryption_algorithm.to_settings(),
-        LCPServerConfiguration.max_printable_pages.to_settings(),
-        LCPServerConfiguration.max_copiable_pages.to_settings()
-    ]
+    SETTINGS = LCPServerConfiguration.to_settings() + LCPEncryptionConfiguration.to_settings()
```
leonardr, line 91: :smile:

api/lcp/credential.py
```diff
@@ -0,0 +1,47 @@
...
+        def _refresh_credential(credential):
+            credential.expires = None
+            credential.credential = str(uuid.uuid4())
+
+            return credential
+
+        credential = Credential.lookup(
+            db,
+            DataSource.LCP,
+            self.IDENTIFIER_FROM_REMOTE_SERVICE,
```
leonardr, line 39: This makes it look like the credential is coming _from_ the LCP license server rather than being generated here and sent _to_ the license server; see my earlier comment for my suggested strings.

Also, I think you can replace this whole class with `Credential.persistent_token_create`; do you need anything different from that?

api/lcp/server.py
```diff
@@ -188,14 +180,20 @@ def _create_partial_license(self, configuration, patron_id=None, license_start=N
                 }
             },
             'rights': {
-                'print': int(configuration.max_printable_pages),
-                'copy': int(configuration.max_copiable_pages)
+                'print':
+                    int(configuration.max_printable_pages)
+                    if configuration.max_printable_pages is not None
+                    else None,
```
leonardr, line 186: We just ran into a problem (https://jira.nypl.org/browse/SIMPLY-3004) where null values in JSON documents caused a parser to malfunction. For the sake of simplicity I'd prefer to omit `print` and `copy` altogether if the values here are `None`.

There's an additional problem which is the spec doesn't say whether a null value here means "use the default behavior of not limiting this feature at all" or whether it means "disable this feature entirely", the exact opposite. So let's avoid the ambiguity.

api/lcp/credential.py
vbessonov: I added two methods: 
- one for generating patron IDs which is public information and included into LCP licenses 
- the other one for generating passphrases which is private information

api/lcp/server.py
vbessonov: Changed the code accordingly

api/lcp/server.py
vbessonov: Removed the code

api/lcp/collection.py
vbessonov: Removed the methods and added a comment

leonardr:
Merge in changes from the base branch, change the URL of the submodule back, and we'll be good to go once I'm able to release 3.0.11.

api/routes.py
```diff
@@ -569,6 +569,20 @@ def saml_authenticate():
 def saml_callback():
     return app.manager.saml_controller.saml_authentication_callback(request, app.manager._db)
 
+
+@library_route('/lcp/hint')
+@has_library
+@returns_problem_detail
+def lcp_passphrase():
+    return app.manager.lcp_controller.get_lcp_passphrase()
```
leonardr, line 577: I see a couple problems with these new routes:

1. These routes need to use `@requires_auth` since they can't (lcp_passphrase) or shouldn't (lcp_license) be used unauthenticated.
2. How are clients to know about the URLs? I don't see any links to these URLs in this branch; will our license documents include these links? In the LCP spec I see a definition of a link relation `hint` which could link to `lcp_passphrase` but I don't see what would be using `lcp_license` -- maybe a `self` link?

api/lcp/controller.py
```diff
@@ -0,0 +1,111 @@
...
+
+    def get_lcp_passphrase(self):
+        """Returns an LCP passphrase for an authenticated patron"""
+        self._logger.info('Started fetching a patron\'s LCP passphrase')
+
+        patron = self._get_patron()
+        if isinstance(patron, ProblemDetail):
+            return patron
+        if isinstance(patron, Response):
+            return patron
```
leonardr, line 84: I see -- you have the equivalent of `requires_auth` inside here. Unless there's a good reason for duplicating the code I'd rather you use `requires_auth` itself.

api/lcp/controller.py
```diff
@@ -0,0 +1,111 @@
...
+        self._logger.info('Finished fetching a patron\'s LCP passphrase: {0}'.format(response))
+
+        return response
+
+    def get_lcp_license(self, license_id):
+        patron = self._get_patron()
+        if isinstance(patron, ProblemDetail):
+            return patron
+        if isinstance(patron, Response):
+            return patron
```
leonardr, line 101: Same deal here -- is there a reason why `requires_auth` on the route won't do the job?

api/routes.py
vbessonov: 1. Internally they use `BaseCirculationManagerController.authenticated_patron_from_request`. I'll add `@requires_auth` 
2. **The hint link** becomes a part of an LCP license:
```
"links": [{
    "hash": "4b82c412664b4b4b2841f4abb9a0cd1f2bd1a98a815026f83db25f010427487f",
    "title": "9780231543972.epub",
    "length": 798385,
    "href": "http://localhost:9000/encrypted-books/9780231543972",
    "rel": "publication",
    "type": "application/epub+zip"
  }, {
    "href": "http://lsdserver:8990/licenses/15a9b3f9-6ba4-4f11-93cb-5c8ccf3fa336/status",
    "type": "application/vnd.readium.license.status.v1.0+json",
    "rel": "status"
  }, {
    "href": "http://cm.hilbertteam.net:6500/LCP/lcp/hint",
    "rel": "hint"
  }],
```
`The passphrase link` is part of the [LCP Status server's configuration](https://github.com/amigoslibraryservices/lcp-docker/blob/master/base-local/etc/config.yaml.minio#L54).

leonardr: I see -- the LCP server is configured to know about these URLs.

api/lcp/controller.py
vbessonov: There is no specific reason, I just wasn't aware of `requires_auth`. Added `@requires_auth`

api/lcp/controller.py
vbessonov: There is no specific reason, I just wasn't aware of `requires_auth`. Added `@requires_auth`

PR #1478, opened 2020-08-11
The If-Modified-Since value has lower resolution than datetime.datetime, which makes conditional GET not work in real cases.

leonardr:
This branch fixes a bug I found while doing QA for https://jira.nypl.org/browse/SIMPLY-2922. Patron.last_loan_activity_sync has resolution of one microsecond, and If-Last-Modified values have resolution of one second. This means an If-Last-Modified value will almost always evaluate to "before" the last_loan_activity_sync used to generate it (the first microsecond of the second in question, versus let's say the 121301st microsecond). This makes conditional HTTP GET fail in real cases.

## Checklist:

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -608,6 +608,12 @@ class CirculationManagerController(BaseCirculationManagerController):
         if not last_modified:
             return None
 
+        # If-Modified-Since values have resolution of one second. If
+        # last_modified has millisecond resolution, change its
+        # resolution to one second.
+        if last_modified.microsecond:
+            last_modified = last_modified.replace(microsecond=0)
+
         # TODO: This can be cleaned up significantly in Python 3.
         if_modified_since = flask.request.headers.get('If-Modified-Since')
         if not if_modified_since:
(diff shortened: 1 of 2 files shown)


api/controller.py
```diff
@@ -608,6 +608,12 @@ def handle_conditional_request(self, last_modified=None):
         if not last_modified:
             return None
 
+        # If-Modified-Since values have resolution of one second. If
+        # last_modified has millisecond resolution, change its
+        # resolution to one second.
+        if last_modified.microsecond:
+            last_modified = last_modified.replace(microsecond=0)
+
```
tdilauro, line 616: It seems like this would be better captured in the `last_loan_activity_sync` getter and setter in the `patron` model.

leonardr: I think this is more an issue of what's possible with the time format used by HTTP. If I could put microsecond data in the Last-Modified header, I would. If we use last_loan_activity_sync for other purposes, then we might want microsecond resolution there. If the timeout is set to a very short period, the microsecond data will become useful when doing time comparisons between last_loan_activity_sync and the current time--half a second one way or another will make a relatively large difference.

That said, if we use last_loan_activity_sync for other purposes and this keeps being a problem, I'd be OK with moving the resolution change into the model.

PR #1484, opened 2020-09-09
Merge docker submodule into main repository

vbessonov:
## Description

This PR removes `docker` submodule and replaces it with the content of [circulation-docker repository](https://github.com/NYPL-Simplified/circulation-docker) using [git-submodule-rewrite script](https://github.com/jeremysears/scripts/blob/master/bin/git-submodule-rewrite).

## How Has This Been Tested?

`git blame` works correctly:
```bash
git blame docker/Dockerfile.exec
```

## Checklist:


docker/simplified_app.sh
```diff
@@ -0,0 +1,88 @@
...
+  ./configure && make && make install
+)
+rm -rf "/tmp/xmlsec1-${XMLSEC_VERSION}"
+
+# Create a user.
+useradd -ms /bin/bash -U simplified
+
+# Get the proper version of the codebase.
+mkdir /var/www && cd /var/www
+git clone https://github.com/${repo}.git circulation
```
leonardr, line 43: @jonathangreen mentioned that he'd like to see this line removed so that the Docker image is built from the same repo that contains it. I think this can wait for a follow-up branch once we verify that Docker builds still work.

docker/simplified_app.sh
```diff
@@ -0,0 +1,88 @@
...
+chown simplified:simplified circulation
+cd circulation
+git checkout $version
+
+# Use https to access submodules.
+git config submodule.core.url https://github.com/NYPL-Simplified/server_core.git
+git config submodule.docker.url https://github.com/NYPL-Simplified/circulation-docker.git
```
leonardr, line 50: I think this should be removed _before_ doing a follow-up branch to make sure we're not secretly building from the old code.

leonardr (changes requested):
This is great. There's one minor change I'd like to make to make sure Docker gets built from this code and not the submodule, and another change to make in a follow-up branch.

docker/simplified_app.sh
vbessonov: Removed it, thank you

jonathangreen:
Looks like this is failing CI because its pulling in a newer hash of core, maybe we should back that out of this PR?

leonardr:
Yeah, take the `core` change out of this PR, I'll do a separate circulation branch to fix it. Sorry about that.

tdilauro:
@leonard I already opened #1485 which is meant to fix what turned out to be a single call to the changed function.

leonardr:
This is failing due to a datahub.io outage that I hope is temporary. We could merge it anyway; I think the Docker image would build since this file only gets downloaded the first time it's used. Any opinion @jonathangreen @vbessonov ?

leonardr:
See https://github.com/NYPL-Simplified/circulation/pull/1485 for more details on the outage.

tdilauro:
https://old.datahub.io is still up, so hopefully it is, indeed, temporary.

jonathangreen:
Seems like datahub has perhaps been down for a couple days by the comment on this issue: 
https://github.com/MacHu-GWU/uszipcode-project/issues/40

leonardr:
Good catch. I'll merge this and then let's work on a solution to that problem.

jonathangreen:
Agreed @leonardr! I was just about to leave a comment that we shouldn't let that other issue hold this up since they are unrelated.

PR #1485, opened 2020-09-09
Fix site_configuration_has_changed parameter

tdilauro:
## Description

Updates circulation manager to accommodate `site_configuration_has_changed` parameter change introduced in [Server Core PR 1195](https://github.com/NYPL-Simplified/server_core/pull/1195). It turns out the parameter was used in only a single test and was not present elsewhere in the code.

## Motivation and Context

Failing test in this repo.

## How Has This Been Tested?

Reviewed all code in this repo for calls to that function.
Ran all tests.

## Checklist:

### DIFF
--- a/tests/test_controller.py
+++ b/tests/test_controller.py
@@ -1181,9 +1181,9 @@ class TestBaseController(CirculationControllerTest):
         self._default_library.short_name = new_name
         self._db.commit()
 
-        # Bypass the 1-second timeout and make sure the site knows
+        # Bypass the 1-second cooldown and make sure the site knows
         # the configuration has actually changed.
-        model.site_configuration_has_changed(self._db, timeout=0)
+        model.site_configuration_has_changed(self._db, cooldown=0)
 
         # Just making the change and calling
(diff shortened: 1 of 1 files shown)


leonardr (approved):

tdilauro:
@leonardr Can you look at the failing tests? I'm not seeing these in my local environment and it's a part of the codebase with which I'm unfamiliar.

leonardr:
The `uszipcode` module downloads https://datahub.io/machu-gwu/uszipcode-0.2.0-simple_db/r/simple_db.sqlite as part of the installation process, and right now datahub.io is down. I'm OK with merging this branch given that the problem is clearly external and otherwise our circulation module is out of sync with core. But if datahub.io doesn't come back up we'll need to package simple_db.sqlite ourselves.

leonardr:
[zip_code_database.zip](https://github.com/NYPL-Simplified/circulation/files/5196942/zip_code_database.zip)

Here's a copy of simple_db.sqlite from my laptop, just in case it never comes back.

PR #1486, opened 2020-09-09
Fix issue with uszipcode library not loading DB

jonathangreen:
## Description

The [`uszipcode`](https://github.com/MacHu-GWU/uszipcode-project) library downloads a sql file on first request that it uses to look up zip codes. The host of this file is down currently and
appears to have been for several days.

This won't effect running circulation managers that have the file cached, but any new CM or one whose docker containers are restarted will see it failing.

## Motivation and Context

This PR takes the strategy outlined here:
https://github.com/MacHu-GWU/uszipcode-project/issues/40#issuecomment-688323810

It puts the db file in the repo, and passes the location of the DB into the `uszipcode` library so it doesn't have to go out and download it. 

This fix was discussed in slack here: 
https://librarysimplified.slack.com/archives/C3CLGUYGN/p1599680363036000

Issue here: 
https://jira.nypl.org/browse/SIMPLY-3062

## How Has This Been Tested?

Before this change the tests were failing as the library could not download its file. After these changes the tests pass again.

## Checklist:

### DIFF
--- a/api/admin/geographic_validator.py
+++ b/api/admin/geographic_validator.py
@@ -15,12 +15,23 @@ from pypostalcode import PostalCodeDatabase
 import re
 import urllib
 import uszipcode
+import os
 
 class GeographicValidator(Validator):
+
+    @staticmethod
+    def get_us_search():
+        # Use a known path for the uszipcode db_file_dir that already contains the DB that the
+        # library would otherwise download. This is done because the host for this file can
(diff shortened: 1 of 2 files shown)


jonathangreen:
I can't request a review in this repo, but @tdilauro and @leonardr can you take a look?

leonardr (approved):

PR #1487, opened 2020-09-10
Deploy docs from the 'main' branch.

leonardr:
## Description

This branch changes .travis.yml so that generated documentation comes from the `main` branch (previously `master`). As far as I can tell this is the only piece of this change that requires a change to our code base.

### DIFF
--- a/docker/README.md
+++ b/docker/README.md
@@ -59,7 +59,7 @@ For troubleshooting information and installation directions for the entire Circu
 
 This image builds containers that will run a single script and stop. It's useful in conjunction with a tool like Amazon ECS Scheduled Tasks, where you can run script containers on a cron-style schedule.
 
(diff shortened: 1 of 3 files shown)


docker/README.md
```diff
@@ -59,7 +59,7 @@ For troubleshooting information and installation directions for the entire Circu
 
 This image builds containers that will run a single script and stop. It's useful in conjunction with a tool like Amazon ECS Scheduled Tasks, where you can run script containers on a cron-style schedule.
 
-Unlike the `circ-scripts` image, which runs constantly and executes every possible maintenance script--whether or not your configuration requires it--`circ-exec` offers more nuanced control of your Library Simplified Circulation Manager jobs. The most accurate place to look for recommended jobs and their recommended frequencies is [the existing `circ-scripts` crontab](https://github.com/NYPL-Simplified/circulation-docker/blob/master/services/simplified_crontab).
+Unlike the `circ-scripts` image, which runs constantly and executes every possible maintenance script--whether or not your configuration requires it--`circ-exec` offers more nuanced control of your Library Simplified Circulation Manager jobs. The most accurate place to look for recommended jobs and their recommended frequencies is [the existing `circ-scripts` crontab](https://github.com/NYPL-Simplified/circulation/blob/main/docker/services/simplified_crontab).
```
jonathangreen, line 62: Minor thing, but github support relative links to files in the same repo, which solves the problem of which branch / fork, etc to point to. If we're changing this link anyway maybe it should be: 

```md
[the existing `circ-scripts` crontab](docker/services/simplified_crontab)
```

PR #1490, opened 2020-09-15
An OPDS feed designed to help out automated integration tests

EdwinGuzman:
## Description

This branch adds a new grouped OPDS feed at /feed/QA whose sole purpose is to make it easy to run client-side functional tests against a working circulation manager.

Currently the feed contains two "lanes" for each of a library's collections: one containing currently available ebooks from that collection, and one containing currently available audiobooks from that collection. This eliminates the need for functional tests to guess at, search for, or randomly hope to be presented with books from various sources that can be borrowed and read or played.

In the future, we can expand this feed (or add more feeds) to make it easy to write other integration tests like the ones laid out in [SIMPLY-2973](https://jira.nypl.org/browse/SIMPLY-2973).

Note that generating this feed takes a long time on a site such as NYPL that has a lot of different collections. We can do work to optimize this later, but it's always going to be somewhat slow. For this reason adding more QA feeds for specific tests might be better than expanding this feed.

## Motivation and Context

Partially resolves [SIMPLY-2973](https://jira.nypl.org/browse/SIMPLY-2973).

## How Has This Been Tested?

In addition to the unit tests I ran this against NYPL's QA circulation manager. It took about 90 seconds to generate an OPDS feed but the feed contained the data I expected.

## Checklist:

- [ x] All new and existing tests passed.

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -7,6 +7,7 @@ import os
 import sys
 import urllib
 import urlparse
+from nose.tools import set_trace
 from collections import defaultdict
 from time import mktime
 from wsgiref.handlers import format_date_time
@@ -88,6 +89,7 @@ from core.model import (
     CustomList,
     DataSource,
     DeliveryMechanism,
+    Edition,
     ExternalIntegration,
     Hold,
     Identifier,
@@ -798,8 +800,30 @@ class IndexController(CirculationManagerController):
             200, { 'Content-Type' : 'application/opds+json' }
         )
 
-
 class OPDSFeedController(CirculationManagerController):
+    def qa_feed(self, feed_class=AcquisitionFeed):
+        library = flask.request.library
+        filter_list = []
+
+        for collection in self._db.query(Collection):
+            for medium in Edition.FULFILLABLE_MEDIA:
+                for availability in ("now", "all"):
+                    filter_list.append((collection, medium, availability))
+
+        search_engine = self.search_engine
+        if isinstance(search_engine, ProblemDetail):
+            return search_engine
+
+        url = self.cdn_url_for(
+            "qa_feed",
+            library_short_name=library.short_name,
+        )
+
+        annotator = self.manager.annotator(None)
+        return feed_class.qa_feed(
+            _db=self._db, title="QA test feed", url=url, annotator=annotator,
+            search_engine=search_engine, filter_list=filter_list
+        )
 
     def groups(self, lane_identifier, feed_class=AcquisitionFeed):
         """Build or retrieve a grouped acquisition feed.
--- a/api/controller.py
+++ b/api/controller.py
@@ -808,7 +808,7 @@ class OPDSFeedController(CirculationManagerController):
 
         for collection in self._db.query(Collection):
             for medium in Edition.FULFILLABLE_MEDIA:
-                for availability in ("now"):
+                for availability in ['now']:
                     filter_list.append((collection, medium, availability))
 
         search_engine = self.search_engine
@@ -827,6 +827,7 @@ class OPDSFeedController(CirculationManagerController):
         return feed_class.groups(
             _db=self._db, title="QA test feed", url=url, worklist=jwl,
             annotator=annotator, search_engine=search_engine,
+            max_age=CachedFeed.IGNORE_CACHE
         )
 
     def groups(self, lane_identifier, feed_class=AcquisitionFeed):
--- a/api/controller.py
+++ b/api/controller.py
@@ -89,7 +89,6 @@ from core.model import (
     CustomList,
     DataSource,
     DeliveryMechanism,
-    Edition,
     ExternalIntegration,
     Hold,
     Identifier,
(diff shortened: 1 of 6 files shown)


api/controller.py
```diff
@@ -798,8 +800,30 @@ def public_key_document(self):
...
 
-
 class OPDSFeedController(CirculationManagerController):
+    def qa_feed(self, feed_class=AcquisitionFeed):
+        library = flask.request.library
+        filter_list = []
+
+        for collection in self._db.query(Collection):
+            for medium in Edition.FULFILLABLE_MEDIA:
+                for availability in ("now", "all"):
```
leonardr, line 810: Just do "now" for now and add the opposite in a follow-up branch.

api/routes.py
```diff
@@ -239,6 +239,14 @@ def public_key_document():
...
     return app.manager.opds_feeds.groups(lane_identifier)
 
+@library_route('/qa-feed')
+@has_library
+@allows_patron_web
+@returns_problem_detail
+@compressible
+def qa_feed():
+    return app.manager.opds_feeds.qa_feed()
+
```
leonardr, line 249: This should require patron authorization, since its only purpose is to show books to a QA test script that's going to conduct transactions on the books.

leonardr:
I think this is on the right track overall.

EdwinGuzman:
The latest update creates a `JackpotWorkList` instance and passes it to `AcquisitionFeed.groups`.

api/controller.py
```diff
@@ -1075,6 +1076,34 @@ def search(self, lane_identifier, feed_class=AcquisitionFeed):
...
+            library_short_name=library.short_name,
+        )
+
+        jwl = JackpotWorkList(library)
+        annotator = self.manager.annotator(jwl)
+
+        # TODO: Make it possible to pass a pagination object into
+        # groups() to override the standard size of a grouped lane,
+        # improving performance.
+        return feed_class.groups(
```
leonardr, line 1101: @EdwinGuzman By letting `groups` handle this rather than calling `query_works_multi`, as you did, I've simplified the code a lot at the cost of performance. We can optimize this later by creating some kind of interface that lets a child of a `WorkList` signal to its parent that it can participate in a `query_works_multi` call. Then the parent can make a single `query_works_multi` call and mix in the results with the works it had to get by calling `works` on a bunch of individual `WorkList`s.

api/lanes.py
```diff
@@ -1242,3 +1242,82 @@ def initialize(self, library, customlist):
...
+        """Return the faceting object to be used when generating a grouped
+        feed.
+        
+        :param _db: Ignored -- only present for API compatibility.
+        :param facets: Ignored -- only present for API compatibility.
+        """
+        return self.facets
+
+
+class JackpotWorkList(WorkList):
```
leonardr, line 1269: @EdwinGuzman I moved this in from `core` because it's only used by the circulation manager.

api/routes.py
```diff
@@ -267,6 +267,15 @@ def public_key_document():
 def acquisition_groups(lane_identifier):
     return app.manager.opds_feeds.groups(lane_identifier)
 
+@library_route('/feed/qa')
```
leonardr, line 270: This URL format will let us add additional feeds beneath /feed/qa if necessary.

leonardr:
@EdwinGuzman I've moved your core branch into this one simplified it at the cost of performance (I'm running out of time to work on this and I want to get something in place), and added tests.

@vbessonov Since both Edwin and I have worked on this branch, I'd like you to review it and give your feedback. This is a feature we introduced for the benefit of A1QA and their integration tests, which is why I floated the idea of having you add more feeds to this list.

api/lanes.py
EdwinGuzman: Does this mean that https://github.com/NYPL-Simplified/server_core/pull/1198/ can be closed?

leonardr: It sure does!

api/controller.py
```diff
@@ -1075,6 +1076,34 @@ def search(self, lane_identifier, feed_class=AcquisitionFeed):
...
+            _db=self._db, title="QA test feed", url=url, worklist=jwl,
+            annotator=annotator, search_engine=search_engine,
+            max_age=CachedFeed.IGNORE_CACHE
```
vbessonov, line 1104: Would it make sense to cache the generated QA feed by default and just add an additional boolean parameter `reset_cache` to `/feed/qa`?

leonardr: Good question. I talked to @vasilyevi about this and he believes it will only be necessary to fetch this feed once per test run, not once for each individual test. If that's true, then it should be OK to avoid special caching code.

However thinking about this, it's so expensive to generate this feed, that I would like to cache it for a few minutes just to speed up the process of developing the tests. Adding caching here creates an interaction with the `cachedfeeds` table, so it will require a little more work. I'll look into this today and maybe I'll be able to speed up the feed generation as well.

I don't want to cache for too long because the whole point of this feed is all these books are available right now. Running the test suite could put some of these books in the "unavailable" state and invalidate the feed.

leonardr:
The single best way to improve performance here is to pass in a Pagination object that cuts down the number of results we ask for. This speeds up the process by over 3x, making it fast enough that I don't think caching will be necessary.

I'm going to merge this branch and hopefully I can do a follow-up that adds pagination.

PR #1492, opened 2020-09-16
Remove a library name from LCP endpoint URLs

vbessonov:
## Description

It removes a library name from all LCP endpoint URLs which means that LCP Status Server doesn't have to be bind to a particular library anymore.

## Motivation and Context

This PR solves the problem related to inability to reuse the single LCP Status Server with multiple libraries.

## Checklist:

### DIFF
--- a/api/lcp/controller.py
+++ b/api/lcp/controller.py
@@ -3,10 +3,12 @@ import logging
 import flask
 from flask import Response
 
+from api.admin.problem_details import MISSING_COLLECTION
 from api.controller import CirculationManagerController
 from api.lcp.factory import LCPServerFactory
 from core.lcp.credential import LCPCredentialFactory
 from core.model import Session, ExternalIntegration, Collection
+from core.util.problem_detail import ProblemDetail
 
 
 class LCPController(CirculationManagerController):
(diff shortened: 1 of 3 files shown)


api/lcp/controller.py
```diff
@@ -54,21 +54,21 @@ def _get_lcp_passphrase(self, patron):
...
         :return: LCP collection
         :rtype: core.model.collection.Collection
         """
-        lcp_collection = next(iter([
-            collection
-            for collection in library.collections
-            if collection.protocol == ExternalIntegration.LCP
-        ]))
+        db = Session.object_session(patron)
+        lcp_collection, _ = Collection.by_name_and_protocol(db, collection_name, ExternalIntegration.LCP)
```
leonardr, line 71: This allows the possibility that you can authenticate an HTTP request against a collection that's in someone else's library, if you know the name of a collection. If `lcp_collection` is not in `patron.library.collections` this method should return a problem detail document.

leonardr (changes requested):

api/lcp/controller.py
vbessonov: Sorry, missed that. I added a check below as requested

leonardr (approved):

PR #1494, opened 2020-09-21
Explicitly set version numbers for all Python dependencies

leonardr:
## Description

This is the circulation portion of https://jira.nypl.org/browse/SIMPLY-3100. The core portion is https://github.com/NYPL-Simplified/server_core/pull/1199

## Motivation and Context

Releases of Python dependencies can cause our builds to mysteriously break, especially now that Python packages are dropping Python 2 support.

I used a similar technique to the one mentioned in the core branch, but with a couple differences:

* Rather than including dependencies of core, I include the core requirements.txt file by reference.
* A few dependencies that were explicitly added to circulation (such as isodate) turned out to be dependencies-of-dependencies of core, so I didn't specify them again.
* There are a couple dependencies I don't think are actually required; I flagged them but haven't taken them out.

## How Has This Been Tested?

The main test is verifying that the environment set up to run the automated tests works. It's possible that that will work but the docker build will fail.

## Checklist:

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,6 +14,7 @@ docker==4.3.1
 docutils==0.16
 expiringdict==1.2.1
 Flask-Cors==3.0.9
+flask-sqlalchemy-session==1.1
 # freezegun is used for unit tests.
 freezegun==0.3.15
 html-sanitizer==1.9.1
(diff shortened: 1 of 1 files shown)


requirements.txt
```diff
@@ -1,83 +1,61 @@
...
-nose
-python-dateutil
-uwsgi
-loggly-python-handler
-mock
-py-bcrypt
-pyspellchecker
-money
-pymarc
-accept-types
```
leonardr, line 24: I don't think we use accept-types anymore. As I recall the project dropped Python 2 support and it was so small that the simplest solution was to copy it into our code base.

EdwinGuzman: That's right, but I do remember removing it from the file.

EdwinGuzman (approved):

PR #1495, opened 2020-09-21
WIP: Use LCP license ID as a loan's external identifier to fetch the current library

vbessonov:
## Description

This PR updates LCP endpoint URLs. I described the issue in more detail in [SIMPLY-3121](https://jira.nypl.org/browse/SIMPLY-3121).

## Checklist:

### DIFF
--- a/api/base_controller.py
+++ b/api/base_controller.py
@@ -99,16 +99,7 @@ class BaseCirculationManagerController(object):
         """
         self.manager.reload_settings_if_changed()
 
-        loan = None
-
-        try:
-            loan = get_one(self._db, Loan, external_identifier=loan_external_identifier)
-        except:
-            logging.exception(
-                'An unexpected exception occurred during fetching a loan with external identifier {0}'.format(
-                    loan_external_identifier
-                )
-            )
+        loan = get_one(self._db, Loan, external_identifier=loan_external_identifier)
 
         if loan is None:
             return LOAN_NOT_FOUND
--- a/api/base_controller.py
+++ b/api/base_controller.py
@@ -1,10 +1,14 @@
+import logging
+
 import flask
 from flask import Response
-from core.util.problem_detail import ProblemDetail
-from circulation_exceptions import *
-from problem_details import *
 from flask_babel import lazy_gettext as _
-from core.model import Library, get_one
+
+from circulation_exceptions import *
+from core.model import Library, get_one, Loan
+from core.util.problem_detail import ProblemDetail
+from problem_details import *
+
 
 class BaseCirculationManagerController(object):
     """Define minimal standards for a circulation manager controller,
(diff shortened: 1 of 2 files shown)


api/base_controller.py
```diff
@@ -82,6 +86,38 @@ def authenticate(self):
...
+        loan = None
+
+        try:
+            loan = get_one(self._db, Loan, external_identifier=loan_external_identifier)
+        except:
+            logging.exception(
+                'An unexpected exception occurred during fetching a loan with external identifier {0}'.format(
+                    loan_external_identifier
+                )
+            )
```
leonardr, line 111: This method should either re-throw the exception or return a problem detail document; probably the former since an exception here is most likely a problem with the database.

api/routes.py
```diff
@@ -570,14 +598,16 @@ def saml_callback():
     return app.manager.saml_controller.saml_authentication_callback(request, app.manager._db)
 
 
-@app.route('/lcp/hint')
+@library_route('/lcp/hint')
+@has_library
```
leonardr, line 602: Presumably you're going to change this later to use `has_library_through_external_loan_identifier`. If you're confident that your LCP branch will be accepted, I think the change can happen now; we already know the code won't work as is.

leonardr (changes requested):
Looks good overall. The only thing I really need to approve is the code shouldn't swallow an exception in get_one.

api/base_controller.py
vbessonov: [Line # 113](https://github.com/NYPL-Simplified/circulation/pull/1495/files/2ab30d75eb96d2ec2fe9b40f6032263fd5c446e3#diff-f258e5927d0cab67f7bc9611720642a5R113) checks whether `loan` is None and returns `LOAN_NOT_FOUND` if it's empty.
Do you want to return a more specific problem detail document in the case of an exception?

vbessonov:
@leonardr, I updated `lcp/hint` endpoint's URL and added `license_id` parameter to it. It's working locally with the changes I did in [readium-lcp-server](https://github.com/readium/readium-lcp-server/pull/227) but I'll test it one more time once Laurent approves and merges my PR.

api/base_controller.py
leonardr: Normally we wouldn't catch this exception at all -- any database access might raise an unexpected exception. If that happens we would allow it to be treated as an uncaught exception because there's a serious problem with the database. My preferred solution would be to get rid of the try/except block altogether, but if you saw something unusual during development, returning a specific problem detail might be appropriate.

api/routes.py
vbessonov: I updated the `hint` route to `/<collection_name>/lcp/licenses/<license_id>/hint'`

api/base_controller.py
vbessonov: Removed try/except block

vbessonov:
@leonardr, Laurent merged my [PR](https://github.com/readium/readium-lcp-server/pull/227), I tested it locally and it worked as expected

leonardr (approved):
Great, sounds like we're good to go.

PR #1496, opened 2020-09-24
HOTFIX - Fix overwrite of underlying dict in static settings.

tdilauro:
## Description

HOTFIX - Fixes an old regression that removed the ability to leave the `Test Password` credential unset for patron authentication integrations that were subclassed from `BasicAuthenticationProvider`.

In addition, this branch introduces some context-sensitive changes to the explanatory text for the `Test Identifier` and `Test Password` fields. Because of the way `circulation-web` handles the settings, it is currently not possible to group a required field with an optional one. They will end up in different sections. The text changes are a less than ideal, but hopefully workable, approach to giving the user a clue to the relationship between the fields in the case where the test password is optional.

## Motivation and Context

For BasicAuthenticationProvider and its subclasses, at least, the presence of a password when one is not expected causes patron authentication to fail [here](https://github.com/NYPL-Simplified/circulation/blob/develop/api/authenticator.py#L2061-L2063) in BasicAuthenticationProvider.server_side_validation. When the `Test Password` cannot be set/unset appropriately, certain capabilities of the `self-test` functionality are unavailable.

https://jira.nypl.org/browse/SIMPLY-1646

## How Has This Been Tested?

Manually tested UI in my dev environment. Specifically tested the following:
- That the `Test Password` is required when configuring a `Simple Authentication Provider` integration.
- That the `Test Password` is NOT required when configuring a `Millenium` or `SIP2` patron authentication integration.
- That when `Test Password` is required, it appears directly under `Test Identifier` in the `Required Fields` section and both fields have the correct explanatory text.
- That when `Test Password` is NOT required, it appears in the the `Optional Fields` section and both fields have the correct explanatory text (referencing each other from another section).
- That both fields appear in only one section.

## Checklist:

### DIFF
--- a/api/simple_authentication.py
+++ b/api/simple_authentication.py
@@ -31,9 +31,16 @@ class SimpleAuthenticationProvider(BasicAuthenticationProvider):
     TEST_NEIGHBORHOOD = 'neighborhood'
 
     basic_settings = list(BasicAuthenticationProvider.SETTINGS)
-    for setting in basic_settings:
-        if setting['key'] == BasicAuthenticationProvider.TEST_PASSWORD:
-            setting['required'] = True
+    for i, setting in enumerate(basic_settings):
+        if setting['key'] == BasicAuthenticationProvider.TEST_IDENTIFIER:
+            s = dict(**setting)
+            s['description'] = BasicAuthenticationProvider.TEST_IDENTIFIER_DESCRIPTION_FOR_REQUIRED_PASSWORD
+            basic_settings[i] = s
+        elif setting['key'] == BasicAuthenticationProvider.TEST_PASSWORD:
+            s = dict(**setting)
+            s['required'] = True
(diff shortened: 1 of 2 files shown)


api/simple_authentication.py
```diff
@@ -31,9 +31,16 @@ class SimpleAuthenticationProvider(BasicAuthenticationProvider):
     TEST_NEIGHBORHOOD = 'neighborhood'
 
     basic_settings = list(BasicAuthenticationProvider.SETTINGS)
-    for setting in basic_settings:
-        if setting['key'] == BasicAuthenticationProvider.TEST_PASSWORD:
-            setting['required'] = True
```
tdilauro, line 36: This line is changing a value in a dict referenced in `BaseAuthenticationProvider.SETTINGS`, so was have a broader than intended effect.

leonardr (approved):

PR #1497, opened 2020-09-30
Bugfix: Patron auth self-test handles problem detail from test patron lookup

tdilauro:
## Description

This branch fixes a bug in patron authentication self-test by handling the case where the test patron lookup results in a problem detail. It converts the problem detail into a message that will be helpful in the self-test display and throws a reasonable exception so that the associated step in the self-test will be treated as failing.

For completeness, this branch also handles other cases where a value that is not a `Patron` object is returned by the lookup: the existing behavior is preserved when the value is `None` and a more generic error that includes the unexpected value is emitted for other types.

Finally, it adds tests for these new cases

## Motivation and Context

Currently, when a patron authentication integration self-test gets a problem detail while looking up its test patron, it produces an inaccurate display. It appears that the `Authenticating test patron` step is successful – it is not – and that something ultimately went wrong in the `Syncing patron metadata` step (it did, but not for reasons related to the setup of the integration).

https://jira.nypl.org/browse/SIMPLY-3139

## How Has This Been Tested?

Added new tests and ran test suite.

## Checklist:

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -1891,8 +1891,11 @@ class BasicAuthenticationProvider(AuthenticationProvider, HasSelfTests):
         return self.authenticated_patron(_db, header), self.test_password
 
     def testing_patron_or_bust(self, _db):
-        """Look up the Patron object reserved for testing purposes,
-        raising CannotLoadConfiguration if none is configured.
+        """Look up the Patron object reserved for testing purposes.
+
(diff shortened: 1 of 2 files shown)


tdilauro:
Assuming checks pass, this is ready to merge.

leonardr (approved):
Looks great!

PR #1498, opened 2020-09-30
Use "long" web client URLs in MARC 856 fields 

tdilauro:
## Description

This branch does the following:
- Changes the format of the MARC `856|u` links created by `LibraryAnnotator.add_web_client_urls`. 
- Provides a DB migration to perform either `(1a)` or `(1b)` *and* `(2)` below:
  1a. Delete all `generate-marc workcoveragerecords` records, so that the next delta will contain all records (needed only if we decide to retain existing delta files).
  1b. Delete cached MARC delta files (those with a start time), as they will contain invalid links in the `856|u` (needed only if we decide NOT to retain existing delta files).
  2. Null out the end_time for remaining (full dump) cached MARC files, so that the coverage provider will regenerate the records as soon as possible.
- Updates associated tests for the new format and reorganizes the code slightly to make it easier to make future adjustments to the format.

The new format is of the form:
```
<web-client-base-url>/book/<cm-base-url>/<lib-short-name>/works/<qualified-identifier>
```
rather than the previous, _shortened_, form:
```
<web-client-base-url>/book/<qualified-identifier>
```
## Motivation and Context

Handling of so-called shortened URLs has been removed from [Circulation Patron Web PR #135](https://github.com/NYPL-Simplified/circulation-patron-web/pull/135). The links in MARC record `856|u` need to conform to this new format.

https://jira.nypl.org/browse/SIMPLY-3093

## How Has This Been Tested?

Tested various database migration scenarios against the `cache_marc_files` script to ensure that I understood their effects and to arrive at what is articulated above in the description.

## Checklist:

### DIFF
--- a/migration/20200930-re-generate-marc-856-u-urls.sql
+++ b/migration/20200930-re-generate-marc-856-u-urls.sql
@@ -1,0 +1,10 @@
+-- Delete all generate-marc workcoveragerecords records, so that new links will be created.
+delete from workcoveragerecords where operation = 'generate-marc';
+
+-- Delete cached MARC delta files (those with a start time),
+-- as they will contain invalid links in the 856|u.
+delete from cachedmarcfiles where start_time is not null;
+
+-- Null out the end_time for remaining cached MARC files so that the
+-- coverage provider will regenerate the records as soon as possible.
+update cachedmarcfiles set end_time = null;
--- a/tests/test_marc.py
+++ b/tests/test_marc.py
@@ -5,6 +5,7 @@ from nose.tools import (
 from pymarc import Record
 
 from . import DatabaseTest
+from core.config import Configuration
 from core.model import (
     ConfigurationSetting,
     ExternalIntegration,
@@ -109,6 +110,34 @@ class TestLibraryAnnotator(DatabaseTest):
         # Web client URLs can come from either the MARC export integration or
         # a library registry integration.
 
+        # The expected format for a web client url is as follows
+        # - <web-client-base>/book/<cm-base>/<lib-short-name>/works/<qualified-identifier>
+        client_url_template = "{client_base}/book/{cm_base}/{lib}/works/{qid}"
+
+        qualified_identifier = "Gutenberg%20ID%2Fidentifier"
+        lib_short_name = self._default_library.short_name
+
+        cm_base_url = "http://test-circulation-manager"
+        client_base_1 = "http://web_catalog"
+        client_base_2 = "http://another_web_catalog"
+        expected_client_url_1 = client_url_template.format(**dict(
+            client_base=client_base_1, cm_base=cm_base_url,
+            lib=lib_short_name, qid=qualified_identifier
+        ))
+        expected_client_url_2 = client_url_template.format(**dict(
+            client_base=client_base_2, cm_base=cm_base_url,
+            lib=lib_short_name, qid=qualified_identifier
+        ))
+
+        # a few assertions to ensure that our client configurations are useful
+        assert len(lib_short_name) > 0
+        assert client_base_1 != client_base_2
+        assert expected_client_url_1 != expected_client_url_2
+        assert expected_client_url_1.startswith(client_base_1)
+        assert expected_client_url_2.startswith(client_base_2)
+
+        ConfigurationSetting.sitewide(self._db, Configuration.BASE_URL_KEY).value = cm_base_url
+
         annotator = LibraryAnnotator(self._default_library)
 
         # If no web catalog URLs are set for the library, nothing will be changed.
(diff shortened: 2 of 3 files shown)


migration/20200930-re-generate-marc-856-u-urls.sql
```diff
@@ -0,0 +1,10 @@
+-- Delete all generate-marc workcoveragerecords records, so that new links will be created.
+delete from workcoveragerecords where operation = 'generate-marc';
+
+-- Delete cached MARC delta files (those with a start time),
+-- as they will contain invalid links in the 856|u.
+delete from cachedmarcfiles where start_time is not null;
```
tdilauro, line 6: Is there any reason we need to keep the delta files? Or can we just remove them and start fresh with new full dumps? If the former -- keep the deltas -- then we won't use the second statement (line 6). If the latter, we don't need to delete the `generate-marc workcoveragerecords` (line 2) because these records affect only the generation of deltas.

@jonathangreen Do you have an opinion on this?

tests/test_marc.py
```diff
@@ -109,6 +110,34 @@ def test_add_web_client_urls(self):
         # Web client URLs can come from either the MARC export integration or
         # a library registry integration.
 
+        # The expected format for a web client url is as follows
+        # - <web-client-base>/book/<cm-base>/<lib-short-name>/works/<qualified-identifier>
+        client_url_template = "{client_base}/book/{cm_base}/{lib}/works/{qid}"
```
tdilauro, line 115: @kristojorg, @leonardr This is the format for the "long" urls. Can you confirm that this is what is needed?

leonardr: It looks right to me but Kristo's in a position to verify this.

migration/20200930-re-generate-marc-856-u-urls.sql
jonathangreen: I think we can just delete the old delta files. I don't see any reason to keep them around, and they are kind of misleading since they will have bad URLs in them.

tests/test_marc.py
kristojorg: Everything after `/book/` should be encoded as a url parameter. And I'm not sure I've seen the `/lib` section before. This is an example of a url:
```
http://localhost:3000/lyrasis/book/https%3A%2F%2Fdemo.lyrasistechnology.org%2Flyrasis%2Fworks%2FURI%2Fhttps%3A%2F%2Fwww.feedbooks.com%2Fitem%2F2936784
```
If you decode everything after `/book/` you end up with:
```
"https://demo.lyrasistechnology.org/lyrasis/works/URI/https://www.feedbooks.com/item/2936784"
```

kristojorg: Oh, duh, the `{lib}` is the library short name. I think this works

leonardr (approved):

tdilauro:
@leonardr I wasn't ready to merge this yet, but I think it will be okay.

leonardr:
Oops, sorry! Let me know if you need me to back it out.

tests/test_marc.py
tdilauro: Thanks @kristojorg. I'm currently encoding only the qualified identifier (identifier.type + "/" + identifier.identifier), since that is the only part that could have special characters. Here's a kind of fake example for what is produced from one of the test cases:

- http://web_catalog/book/http://test-circulation-manager/default/works/Gutenberg%20ID%2Fidentifier

Does that work for you? Or do you need me to encode the entire CM endpoint again (effectively double encoding the identifier)?

kristojorg: This actually doesn't work in the current CPW. The reason is that this url:
```
http://localhost:3000/lyrasis/book/https://demo.lyrasistechnology.org/lyrasis/works/URI%2Fhttps%3A%2F%2Fwww.feedbooks.com%2Fitem%2F2936784
```
Is telling the application to look for a page at `/lyrasis/book/https://demo.lyrasistechnology.org/lyrasis/works/`, which doesn't exist. The page that does exist is `/lyrasys/book/:bookUrl`. 

Basically, it does need to be double encoded because it is actually a url embedded in a url. This is a valid url: `http://test-circulation-manager/default/works/Gutenberg%20ID%2Fidentifier`, but if you want to embed that in the CPW url, you need to encode it again since in that context, the forward slashes are in fact special characters. Does that make any sense?

kristojorg:
Just a note that this does need to be updated to double encode, so I'm not sure if it should be backed out or a patch applied. Either way, we are embedding the ID in the CM url in the CPW url, so double encoding is necessary here.

tdilauro:
> Oops, sorry! Let me know if you need me to back it out.

You won't be surprised to know that the thing I was most worried about was the database migration, specifically the deletion of the MARC record delta file records. I would have removed the unneeded -- but harmless -- deletion of the `generate-marc workcoveragerecords`, but it's not a big deal.

tdilauro:
@kristojorg Thanks for the feedback. I'll open a new PR to add the needed encoding today.

PR #1500, opened 2020-10-02
Encode work link param for MARC 856|u web client URLs

tdilauro:
## Description

This branch does the following:
- Fixes the format of the MARC `856|u` links created by `LibraryAnnotator.add_web_client_urls` to encode the entire work URL param (which includes the already encoded qualified identifer (`type + '/' + indentifier`).
- Provides a DB migration to:
  1. Delete cached MARC delta files (those with a `start_time`), as they will contain invalid links in the `856|u`.
  2. Null out the `end_time` for remaining (full dump, without a `start_time`) cached MARC files, so that the coverage provider will regenerate the records as soon as possible.
- Updates associated tests for the new format .

The URL for a work is constructed as follows and then encoded:
```
<cm-base>/<lib-short-name>/works/<qualified-identifier>
```

The associated web client URL is constructed as:
```
<web-client-base>/book/<encoded-work-url>
```

## Motivation and Context

PR #1498, the first PR aimed at generating "long" web client URLs, did not encode the work URL param. Circulation Patron Web needs it to be encoded.

A somewhat redundant database migration is included, since it is theoretically possible that new records were generated somewhere based on PR #1498 changes.

https://jira.nypl.org/browse/SIMPLY-3093

## How Has This Been Tested?

Modified one test method for the specific format change and ran all tests.

## Checklist:

### DIFF
--- a/api/marc.py
+++ b/api/marc.py
@@ -66,11 +66,15 @@ class LibraryAnnotator(Annotator):
         qualified_identifier = urllib.quote(identifier.type + "/" + identifier.identifier, safe='')
 
         for web_client_base_url in settings:
-            url = "{}/book/{}/{}/works/{}".format(
-                web_client_base_url,
+            link = "{}/{}/works/{}".format(
                 self.base_url,
                 library.short_name,
                 qualified_identifier,
+            )
+            encoded_link = urllib.quote(link, safe='')
+            url = "{}/book/{}".format(
+                web_client_base_url,
+                encoded_link
             )
             record.add_field(
                 Field(
--- a/migration/20201002-redo-re-generate-marc-856-u-urls.sql
+++ b/migration/20201002-redo-re-generate-marc-856-u-urls.sql
@@ -1,0 +1,7 @@
+-- Delete cached MARC delta files (those with a start time),
+-- as they will contain invalid links in the 856|u.
+delete from cachedmarcfiles where start_time is not null;
+
+-- Null out the end_time for remaining cached MARC files so that the
+-- coverage provider will regenerate the records as soon as possible.
+update cachedmarcfiles set end_time = null where start_time is null;
--- a/tests/test_marc.py
+++ b/tests/test_marc.py
@@ -3,6 +3,7 @@ from nose.tools import (
     set_trace,
 )
 from pymarc import Record
+import urllib
 
 from . import DatabaseTest
 from core.config import Configuration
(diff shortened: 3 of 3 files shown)


leonardr (approved):
This time I'm going to ask if you're ready before I merge the PR! Are you ready?

tdilauro:
I want @kristojorg to have a look to make sure the format is gonna meet his needs.

kristojorg (commented):
RE double encoding: I'm not entirely sure I see where this is being fully encoded, but the tests look like they confirm what we need there.

However, I have a large issue I didn't see earlier. This is the fact that a CPW instance can be run with two different url structures depending on env vars. This is a legacy feature that we've never removed, although I would like to. Basically you can run the app with one or multiple libraries, each with their own catalog root url. The url structure for books (and everything else) is different depending what you do.

**With a single catalog base**
```
http://catalog.library.org/book/:bookUrl
```
This is the simplest form, and is how Open eBooks runs the app today.

**With multiple libraries**
```
http://catalog.library.org/:library/book/:bookUrl
```
This is when a single app instance is serving multiple libraries. The `:library` segment is either the library short name pulled from the config file defined in the env, or it is the library id pulled from a library registry. 

So the question is which case are we building this for? If we want to support both, the CM would have to be configured to know how CPW is in turn configured, which seems... perilous.

I would prefer that CPW has a single stable url structure. If we want to do that, we could either remove support for multiple libraries, and each library would require its own instance (probably not sustainable for lyrasis), or we could remove support for single libraries, and Open eBooks would have to have some shortname in the `:library` section. This could be arbitrarily short, like `catalog.library.org/o/book/:bookUrl`, but we would need something in there. @leonardr has given me the green light on that before, but it was never a priority so I never actually did it. 

A related problem, however, is how the CM knows the library short name? Are we just assuming it is set up the same in the CM as it is in the CPW config file? That also seems error prone, but might be unavoidable. We couldn't move the CPW config into the CM because one CPW can service multiple CMs. 

Curious as to your thoughts on this. I can make the CPW change pretty quickly to remove single library routes, but want to confirm that's the right direction.

api/marc.py
```diff
@@ -66,12 +66,16 @@ def add_web_client_urls(self, record, library, identifier, integration=None):
...
 
         for web_client_base_url in settings:
-            url = "{}/book/{}/{}/works/{}".format(
-                web_client_base_url,
+            link = "{}/{}/works/{}".format(
                 self.base_url,
                 library.short_name,
                 qualified_identifier,
             )
+            encoded_link = urllib.quote(link, safe='')
```
tdilauro, line 74: @kristojorg The work URL is built here. The first encoding is of the `qualified_identifier`. The second encoding is for the `encoded_link`. Only the `qualified identifier` is double encoded.

api/marc.py
```diff
@@ -66,12 +66,16 @@ def add_web_client_urls(self, record, library, identifier, integration=None):
...
+            url = "{}/book/{}".format(
+                web_client_base_url,
+                encoded_link
+            )
```
tdilauro, line 78: The web client URL is built here, using the `encoded_link`.

tdilauro:
> So the question is which case are we building this for? If we want to support both, the CM would have to be configured to know how CPW is in turn configured, which seems... perilous.
> ...
> A related problem, however, is how the CM knows the library short name? Are we just assuming it is set up the same in the CM as it is in the CPW config file? That also seems error prone, but might be unavoidable. We couldn't move the CPW config into the CM because one CPW can service multiple CMs.

If we needed to choose between single- and multi-library, I think generalizing to the multi-library approach would be better. I don't think configuration would be too problematic, so long as the configurations aren't changing too frequently (mostly because of how MARC records are cached on the CMs and because the `MARC Export` integration is incomplete).

The `MARC Export` integration has per-library configuration, including a `base URL for the web catalog for this library`. That field can be configured as `https://<web-catalog-authority>` or `https://<web-catalog-authority>/<arbitrary-path>`. So a per-library setting like `https://<web-catalog-authority>/<library-short-name>` could be manually configured for each library that participates in the integration. 

Currently, that string would be static (i.e., would not be replaced by the current library short name from the database), so could get out of sync with the CM's view of the library. But that might be a good thing, depending on how the associated property in CPW is configured.

api/marc.py
kristojorg: Okay yep this makes sense then, thanks!

kristojorg (approved):
Okay based on the fact that there is a per-library setting for the base url of the web catalog, I think you're right that this is totally okay. That setting can simply be configured to either include or not include the library short name depending on whether that web catalog is being run with a single or multi library setup.

The property is setup in a config file in CPW, so it is a good thing that it is not using the CMs idea of the library's short name... the CPW short name can be different from the CM short name. 

I think this seems okay

leonardr:
I'll merge once I get the okay from @tdilauro .

tdilauro:
This is ready to merge. Thank you both!

PR #1503, opened 2020-10-09
Deny a patron with a root lane access to lanes and books not visible through that lane.

leonardr:
## Description

This branch removes a feature that was had not been used for a long time -- the library's loan policy -- and reimplements it in terms of an existing feature -- the patron's root lane.

A patron may or may not have a root lane. Before this branch, the root lane only controlled which lane a patron saw when they hit the root feed URL. Now, a patron's root lane puts additional restrictions on their activity:

* A patron with a root lane cannot view an OPDS feed for a Lanes that is 'above' their root lane 
* A patron with a root lane cannot view an OPDS feeds for a WorkList that might contain material not age-appropriate for their root lane. (As an example: a `ContributorLane` with `.audiences` set to include an age-inappropriate audience.)
* A patron with a root lane cannot access any controller for a book that is not age-appropriate for their root lane -- the circulation manager will send a problem detail. This includes the "permalink" controller but also the "borrow" and "place hold" controllers.
* A patron with a root lane cannot borrow a book that is not age-appropriate for their root lane. (But it shouldn't come to this because they won't be able to access the "borrow" controller in the first place.)

"Age-appropriate" is a concept defined in the [core branch](https://github.com/NYPL-Simplified/server_core/pull/1201) that goes along with this branch. In fact most of the code is in the core branch; this branch just uses that code to enforce the rules.

## Motivation and Context

https://jira.nypl.org/browse/SIMPLY-3125

## How Has This Been Tested?

`authenticated_patron_from_request` had no specific unit test coverage, so I added some. `load_lane` had test coverage but it was out of date because the class it was in had a name that didn't start with `Test`. I rewrote it.

In addition to the unit tests, I wrote some end-to-end tests illustrating the way people think of this feature: not in terms of certain methods being called, but in terms of certain patrons trying to take certain actions and not being allowed.

## Checklist:

### DIFF
--- a/api/adobe_vendor_id.py
+++ b/api/adobe_vendor_id.py
@@ -739,8 +739,11 @@ class AuthdataUtility(object):
         ):
             raise CannotLoadConfiguration(
                 "Short Client Token configuration is incomplete. "
-                "vendor_id, username, password and "
-                "Library website_url must all be defined.")
+                "vendor_id (%s), username (%s), password (%s) and "
+                "Library website_url (%s) must all be defined." % (
+                    vendor_id, library_uri, library_short_name, secret
+                )
+            )
         if '|' in library_short_name:
             raise CannotLoadConfiguration(
                 "Library short name cannot contain the pipe character."
--- a/api/controller.py
+++ b/api/controller.py
@@ -1882,7 +1882,7 @@ class WorkController(CirculationManagerController):
             lane = RelatedBooksLane(
                 library, work, lane_name, novelist_api=novelist_api
             )
-        except ValueError, e:
+        except ValueError:
             # No related books were found.
             return NO_SUCH_LANE.detailed(e.message)
 
--- a/api/controller.py
+++ b/api/controller.py
@@ -163,13 +163,19 @@ class CirculationManager(object):
         self.load_settings()
 
     def load_facets_from_request(self, *args, **kwargs):
-        """Load a faceting object from the incoming request, but also apply an authentication
-        restriction.
+        """Load a faceting object from the incoming request, but also apply some
+        application-specific access restrictions:
 
-        Specifically, in this application you can't use nonstandard
-        caching rules unless you're an authenticated administrator.
+        * You can't use nonstandard caching rules unless you're an authenticated administrator.
+        * You can't access a WorkList that's not visible to you.
         """
+
         facets = load_facets_from_request(*args, **kwargs)
+
+        worklist = kwargs.get('worklist')
+        if worklist is not None and not worklist.is_visible_to(self.request_patron):
+            return NO_SUCH_LANE.detailed(_("Lane does not exist"))
+
         if isinstance(facets, BaseFacets) and getattr(facets, 'max_cache_age', None) is not None:
             # A faceting object was loaded, and it tried to do something nonstandard
             # with caching.
@@ -612,7 +618,6 @@ class CirculationManagerController(BaseCirculationManagerController):
     def load_lane(self, lane_identifier):
         """Turn user input into a Lane object."""
         library_id = flask.request.library.id
-        has_root_lanes = flask.request.library.has_root_lanes
 
         lane = None
         if lane_identifier is None:
@@ -633,7 +638,7 @@ class CirculationManagerController(BaseCirculationManagerController):
                     self._db, Lane, id=lane_identifier, library_id=library_id
                 )
 
-        if lane and has_root_lanes and not lane.visible_to(self.request_patron):
+        if lane and not lane.visible_to(self.request_patron):
             # The authenticated patron cannot see the lane they
             # requested. Act like the lane does not exist.
             lane = None
--- a/api/controller.py
+++ b/api/controller.py
@@ -167,13 +167,13 @@ class CirculationManager(object):
         application-specific access restrictions:
 
         * You can't use nonstandard caching rules unless you're an authenticated administrator.
-        * You can't access a WorkList that's not visible to you.
+        * You can't access a WorkList that's not accessible to you.
         """
 
         facets = load_facets_from_request(*args, **kwargs)
 
         worklist = kwargs.get('worklist')
-        if worklist is not None and not worklist.is_visible_to(self.request_patron):
+        if worklist is not None and not worklist.accessible_to(self.request_patron):
             return NO_SUCH_LANE.detailed(_("Lane does not exist"))
 
         if isinstance(facets, BaseFacets) and getattr(facets, 'max_cache_age', None) is not None:
@@ -638,8 +638,8 @@ class CirculationManagerController(BaseCirculationManagerController):
                     self._db, Lane, id=lane_identifier, library_id=library_id
                 )
 
-        if lane and not lane.visible_to(self.request_patron):
-            # The authenticated patron cannot see the lane they
+        if lane and not lane.accessible_to(self.request_patron):
+            # The authenticated patron cannot access the lane they
             # requested. Act like the lane does not exist.
             lane = None
 
--- a/api/controller.py
+++ b/api/controller.py
@@ -1882,7 +1882,7 @@ class WorkController(CirculationManagerController):
             lane = RelatedBooksLane(
                 library, work, lane_name, novelist_api=novelist_api
             )
-        except ValueError:
+        except ValueError as e:
             # No related books were found.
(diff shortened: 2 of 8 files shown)


api/adobe_vendor_id.py
```diff
@@ -739,8 +739,11 @@ def from_config(cls, library, _db=None):
         ):
             raise CannotLoadConfiguration(
                 "Short Client Token configuration is incomplete. "
-                "vendor_id, username, password and "
-                "Library website_url must all be defined.")
+                "vendor_id (%s), username (%s), password (%s) and "
+                "Library website_url (%s) must all be defined." % (
+                    vendor_id, library_uri, library_short_name, secret
+                )
+            )
```
leonardr, line 746: This is an unrelated change that helps us recover more quickly from SIMPLY-2983 incidents -- with the current error message it's not clear which piece of configuration is missing.

api/controller.py
```diff
@@ -625,9 +626,20 @@ def load_lane(self, lane_identifier):
             elif isinstance(lane, WorkList):
                 lane.children = [self._db.merge(child) for child in lane.children]
         else:
-            lane = get_one(
-                self._db, Lane, id=lane_identifier, library_id=library_id
-            )
+            try:
+                lane_identifier = int(lane_identifier)
+            except ValueError, e:
+                pass
```
leonardr, line 632: Previously if you asked for a lane with a string name (rather than a numeric ID) you would get a database error.

api/controller.py
```diff
@@ -625,9 +626,20 @@ def load_lane(self, lane_identifier):
...
+
+            if isinstance(lane_identifier, int):
+                lane = get_one(
+                    self._db, Lane, id=lane_identifier, library_id=library_id
+                )
+
+        if lane and not lane.visible_to(self.request_patron):
```
leonardr, line 639: This is why I had to add `request_patron`. Authentication is not required for any controller that uses `load_lane`, so `authenticated_patron_for_request` probably hasn't been called yet. I need a yes-or-no answer as to whether there is an authenticated patron or not, and `authenticated_patron_for_request` can return all sorts of things.

api/controller.py
```diff
@@ -694,9 +707,11 @@ def load_licensepooldelivery(self, pool, mechanism_id):
     def apply_borrowing_policy(self, patron, license_pool):
         if isinstance(patron, ProblemDetail):
             return patron
-        if not patron.can_borrow(license_pool.work, self.manager.lending_policy):
```
leonardr, line 697: The `can_borrow` method and `lending_policy` have been removed.

api/controller.py
```diff
@@ -1958,6 +1959,8 @@ class ProfileController(CirculationManagerController):
     def _controller(self):
         """Instantiate a CoreProfileController that actually does the work.
         """
+        # TODO: Probably better to use request_patron and check for
+        # None here.
```
leonardr, line 1963: It's not a good idea to call `authenticated_patron_from_request` and assume it will return a `Patron`, but that's what's happening here. I don't want to spend time fixing it in this branch though.

api/routes.py
```diff
@@ -81,6 +81,9 @@ def decorated(*args, **kwargs):
 def allows_auth(f):
     """Decorator function for a controller method that supports both
     authenticated and unauthenticated requests.
+
+    NOTE: This decorator might not be necessary; you can probably call
+    BaseCirculationManagerController.request_patron instead.
```
leonardr, line 86: I experimented with marking the controllers that use `load_lane` as `@allow_auth` but I think it's actually better to get rid of the decorator.

api/controller.py
```diff
@@ -255,9 +260,6 @@ def load_settings(self):
         self.circulation_apis = new_circulation_apis
         self.custom_index_views = new_custom_index_views
         self.shared_collection_api = self.setup_shared_collection()
-        self.lending_policy = load_lending_policy(
-            Configuration.policy('lending', {})
-        )
```
leonardr, line 260: The lending policy was not tested or supported, and now has been removed.

tests/test_controller.py
```diff
@@ -1199,52 +1297,72 @@ def test_library_for_request_reloads_settings_if_necessary(self):
...
             assert new_name in self.manager.auth.library_authenticators
 
+    def test_load_lane(self):
+        # Verify that requests for specific lanes are mapped to
+        # the appropriate lane.
 
-class FullLaneSetupTest(CirculationControllerTest):
-    """Most lane-based tests don't need the full multi-tier setup of lanes
-    that we would see in a real site. We use a smaller setup to save time
-    when running the test.
```
leonardr, line 1206: The test in this class was long out of date and not being run because its name didn't start with `Test`.

api/controller.py
```diff
@@ -744,29 +761,20 @@ def authentication_document(self):
...
-        ).filter(
-            Lane.root_for_patron_type!=None
-        )
-        return root_lanes.count() > 0
+        """Does the active library feature root lanes for patrons of
+        certain types?
+
+        :return: A boolean
+        """
+        return flask.request.library.has_root_lanes
```
leonardr, line 769: Implementation was moved to the `Library` class in core.

api/controller.py
```diff
@@ -164,13 +163,26 @@ def __init__(self, _db, testing=False):
...
     def load_facets_from_request(self, *args, **kwargs):
-        """Load a faceting object from the incoming request, but also apply an authentication
-        restriction.
+        """Load a faceting object from the incoming request, but also apply some
+        application-specific access restrictions:
 
-        Specifically, in this application you can't use nonstandard
-        caching rules unless you're an authenticated administrator.
+        * You can't use nonstandard caching rules unless you're an authenticated administrator.
+        * You can't access a WorkList that's not accessible to you.
```
leonardr, line 170: `load_facets_from_request` is called every time we're building an OPDS feed, and it already might short-circuit the request by returning a ProblemDetail (if the facets are incorrectly specified) so it's a convenient place to put other logic that might short-circuit the request by returning a ProblemDetail.

tests/test_controller.py
```diff
@@ -1010,6 +1118,40 @@ def test_load_licensepools(self):
         )
         eq_(NO_LICENSES.uri, problem_detail.uri)
 
+    def test_load_work(self):
```
leonardr, line 1121: This method previously had no standalone tests.

tests/test_controller.py
```diff
@@ -796,6 +829,80 @@ def test_unscoped_session(self):
...
+        o2 = object()
+        def set_patron():
+            flask.request.patron = o2
+        mock = MagicMock(side_effect = set_patron,
+                         return_value = "return value will be ignored")
+        self.controller.authenticated_patron_from_request = mock
+        with self.app.test_request_context("/"):
+            eq_(o2, self.controller.request_patron)
+
+    def test_authenticated_patron_from_request(self):
```
leonardr, line 855: A lot of this test coverage was missing.

api/controller.py
```diff
@@ -647,7 +667,13 @@ def load_work(self, library, identifier_type, identifier):
 
         # We know there is at least one LicensePool, and all LicensePools
         # for an Identifier have the same Work.
-        return pools[0].work
+        work = pools[0].work
+
+        if work and not work.age_appropriate_for_patron(self.request_patron):
+            # This work is not age-appropriate for the authenticated
+            # patron. Don't show it.
+            work = NOT_AGE_APPROPRIATE
```
vbessonov, line 675: Would it make sense to change the current approach and instead of returning `ProblemDetail` objects from functions and handling them with `if isinstance(result, ProblemDetail)` statements just raise a `ProblemDetail` exception handling it at the bottom of the callstack where we could log them appropriately and return a `ProblemDetail` object back to the caller if needed? Of course, I don't imply that it's needed to be done in this PR but maybe we can consider it for the future? I think it could simplify the codebase, we could get rid of a lot of `if isinstance(result, ProblemDetail)` statements

vbessonov (changes requested):
This PR also looks good, added a couple of suggestions

api/controller.py
leonardr: I have had similar thoughts in the past and I think it could work. The main thing we'd have to change is situations where one piece of code calls `ProblemDetail.detailed` to add more information to a `ProblemDetail` object. I think on balance the code would be easier to understand if we made this change.

PR #1504, opened 2020-10-13
Add support for RWPM .audiobook files

vbessonov:
## Description

This PR adds media types for supporting RWPM `.audiobook` files. You can find more information in [SIMPLY-3183](https://jira.nypl.org/browse/SIMPLY-3183). 

This PR depends on [PR # 1203 in server_core](https://github.com/NYPL-Simplified/server_core/pull/1203).

## Checklist:

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -1312,7 +1312,7 @@ class DirectoryImportScript(TimestampScript):
         parser.add_argument(
             '--default-medium-type',
             help=u'Default medium type used in the case when it\'s not explicitly specified in a metadata file. '
-                 u'Valid values are: EditionConstants.FULFILLABLE_MEDIA.',
+                 u'Valid values are: {0}.'.format(', '.join(EditionConstants.FULFILLABLE_MEDIA)),
             type=str,
             choices=EditionConstants.FULFILLABLE_MEDIA
         )
--- a/scripts.py
+++ b/scripts.py
@@ -90,7 +90,7 @@ from core.model import (
     Subject,
     Timestamp,
     Work,
-)
+    EditionConstants)
 from core.model.configuration import ExternalIntegrationLink
 from core.opds import (
     AcquisitionFeed,
@@ -1309,6 +1309,14 @@ class DirectoryImportScript(TimestampScript):
             help=u"Show what would be imported, but don't actually do the import.",
             action='store_true',
         )
+        parser.add_argument(
+            '--default-medium-type',
+            help=u'Default medium type used in the case when it\'s not explicitly specified in a metadata file. '
+                 u'Valid values are: Book, Audio.',
+            type=str,
+            choices=(EditionConstants.BOOK_MEDIUM, EditionConstants.AUDIO_MEDIUM)
+        )
+
         return parser
 
     def do_run(self, cmd_args=None):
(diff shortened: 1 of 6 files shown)


scripts.py
```diff
@@ -1309,6 +1309,14 @@ def arg_parser(cls, _db):
             help=u"Show what would be imported, but don't actually do the import.",
             action='store_true',
         )
+        parser.add_argument(
+            '--default-medium-type',
+            help=u'Default medium type used in the case when it\'s not explicitly specified in a metadata file. '
+                 u'Valid values are: Book, Audio.',
```
leonardr, line 1315: Substitute EditionConstants.FULFILLABLE_MEDIA both here and in `choices` below.

scripts.py
```diff
@@ -1323,16 +1331,33 @@ def do_run(self, cmd_args=None):
...
+            collection_name,
+            collection_type,
+            data_source_name,
+            metadata_file,
+            metadata_format,
+            cover_directory,
+            ebook_directory,
+            rights_uri,
+            dry_run,
+            default_medium_type
```
leonardr, line 1346: It's up to you, but at this point when I'm adding yet another argument to a method that takes a lot of arguments, I usually switch over to passing in keyword arguments. This reduces the risk that I'll mix up the order of two arguments and create a difficult-to-find bug.

tests/test_scripts.py
```diff
@@ -877,7 +877,8 @@ def run_with_arguments(self, *args):
                 "--cover-directory=covers",
                 "--ebook-directory=ebooks",
                 "--rights-uri=rights",
-                "--dry-run"
+                "--dry-run",
+                "--default-medium-type=Audiobook"
```
leonardr, line 881: This should be `Audio`, not `Audiobook`. This is what's causing the test failure -- the script is exiting early due to invalid command-line arguments.

leonardr (changes requested):
This looks good once the test failure is fixed and the list of allowable media is changed to use the list defined in EditionConstants.

scripts.py
vbessonov: I changed the code accordingly. You also wanted to replace `u'Valid values are: Book, Audio.'` with `u'Valid values are: EditionConstants.FULFILLABLE_MEDIA.'`, right?

scripts.py
vbessonov: It's a good advice, changed the code accordingly

tests/test_scripts.py
vbessonov: Sorry, didn't notice it at first, strangely enough the tests worked locally (probably had some other changes). I replaced `Audiobook` with `Audio`

scripts.py
leonardr: Sorry, I should have been more clear. I meant something like this:
```
"Valid values are: {}".format(", ".join(EditionConstants.FULFILLABLE_MEDIA))
```

vbessonov: My bad, fixed it

scripts.py
```diff
@@ -1312,9 +1312,9 @@ def arg_parser(cls, _db):
         parser.add_argument(
             '--default-medium-type',
             help=u'Default medium type used in the case when it\'s not explicitly specified in a metadata file. '
-                 u'Valid values are: Book, Audio.',
+                 u'Valid values are: EditionConstants.FULFILLABLE_MEDIA.',
```
leonardr, line 1315: I think you changed the default_medium_type but not this piece, which is the confusing part.

vbessonov: That's weird, I must've forgotten to push

leonardr (approved):

PR #1512, opened 2020-10-27
Significantly reduce the size of the jackpot feed, improving performance

leonardr:
## Description

Make it possible to pass a Pagination object into WorkList.groups(), and use a Pagination object with size=1 when creating  the jackpot feed.

This is the circulation portion of https://github.com/NYPL-Simplified/server_core/pull/1206. It improves performance of the jackpot feed significantly, though it may be necessary to do more work if integration tests are being run against libraries with a lot of collections.

We call groups() in a few cases, but the jackpot feed is the only place where we need to overwrite the default behavior, so none of the other tests needed to change.

## Motivation and Context

https://jira.nypl.org/browse/SIMPLY-3234

## Checklist:

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -1095,12 +1095,13 @@ class OPDSFeedController(CirculationManagerController):
         jwl = JackpotWorkList(library)
         annotator = self.manager.annotator(jwl)
 
-        # TODO: Make it possible to pass a pagination object into
-        # groups() to override the standard size of a grouped lane,
-        # improving performance.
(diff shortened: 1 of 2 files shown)


leonardr:
@vbessonov I'm officially laying down this branch (and the corresponding core branch) so that you can finish it if you get time. If I come back to pick it up, I'll let you know so we don't duplicate work.

EdwinGuzman (approved):

PR #1521, opened 2020-11-09
Add faceting capability to the jackpot feed

leonardr:
## Description

This branch makes it possible to request a jackpot feed with different facets, similar to the way other feed controllers work. 

## Motivation and Context

In combination with https://github.com/NYPL-Simplified/server_core/pull/1211, which adds a "not_now" availability facet, this completes https://jira.nypl.org/browse/SIMPLY-3236.

## How Has This Been Tested?

To test the 'not_now' facet specifically:

`curl http://user:pass@localhost:6500/feed/qa?available=not_now`

## Checklist:

### DIFF
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -1288,7 +1288,8 @@ class JackpotFacets(Facets):
                 config, facet_group_name
             )
 
-        return [Facets.AVAILABLE_NOW, Facets.AVAILABLE_NOT_NOW]
+        return [Facets.AVAILABLE_NOW, Facets.AVAILABLE_NOT_NOW,
+                Facets.AVAILABLE_OPEN_ACCESS]
 
 
 class JackpotWorkList(WorkList):
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -1280,6 +1280,11 @@ class KnownOverviewFacetsWorkList(WorkList):
 
 
 class JackpotFacets(Facets):
+    """A faceting object for a jackpot feed.
+
+    Unlike other faceting objects, AVAILABLE_NOT_NOW is an acceptable
+    option for the availability facet.
+    """
 
     @classmethod
     def default_facet(cls, config, facet_group_name):
--- a/api/lanes.py
+++ b/api/lanes.py
@@ -1336,7 +1336,7 @@ class JackpotWorkList(WorkList):
                     data_source_name = collection.data_source.name
                 else:
                     data_source_name = "[Unknown]"
-                display_name = "[Collection test] - License source {%s} - Medium {%s} - Collection name {%s}" % (data_source_name, medium, collection.name)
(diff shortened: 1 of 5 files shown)


api/lanes.py
```diff
@@ -1303,30 +1325,19 @@ def __init__(self, library):
...
-            #   that are _not_ available.
-            #  A lane where all books belong to some series
-            #  A lane where all books use a particular DRM scheme/format
+                # Give each Worklist a name that is distinctive
+                # and easy for a client to parse.
+                if collection.data_source:
+                    data_source_name = collection.data_source.name
+                else:
+                    data_source_name = "[Unknown]"
+                display_name = "[Collection test] - License source {%s} - Medium {%s} - Collection name {%s}" % (data_source_name, medium, collection.name)
```
leonardr, line 1334: The availability status is no longer mentioned in the lane name, because whoever requested the feed had a chance to specify a value for the `availability` facet. They ought to know what they requested, so there's no reason to spell t out in the feed itself.

tests/test_lanes.py
```diff
@@ -936,6 +937,54 @@ def test_overview_facets(self):
...
+
+        # For other facet groups, the class defers to the Facets
+        # superclass. (But this doesn't matter because it's not relevant
+        # to the creation of jackpot feeds.)
+        for group in (Facets.COLLECTION_FACET_GROUP_NAME,
+                      Facets.ORDER_FACET_GROUP_NAME):
+            eq_(m(self._default_library, group),
+                Facets.default_facet(self._default_library, group))
+
+        eq_(Facets.ORDER_SERIES_POSITION, SeriesFacets.DEFAULT_SORT_ORDER)
```
EdwinGuzman, line 959: I don't get this test. These two are constants but how are they related to the JackpotFacets?

leonardr: They're not; this is a copy-and-paste error. I copied the code from `TestSeriesFacets.test_default_sort_order` and I'm not sure why. Maybe I intended to modify it but then I wrote new code instead.

tests/test_lanes.py
```diff
@@ -978,34 +1021,35 @@ def test_constructor(self):
...
+            eq_(facets, internal_facets)
 
         # These worklists show ebooks and audiobooks from the two
         # collections associated with the default library.
         [default_ebooks, default_audio, overdrive_ebooks, overdrive_audio] = (
             available_now
         )
 
-        eq_("[Collection test] - License source {[Unknown]} - Medium {Book} - Availability {now} - Collection name {%s}" % self._default_collection.name, 
+        eq_("License source {[Unknown]} - Medium {Book} - Collection name {%s}" % self._default_collection.name,
```
leonardr, line 1036: I realized that the logic that applied to removing "Availability {now}" also applied to removing "[Collection test]" -- you asked for the collection test so you know that's what you're getting.

EdwinGuzman (approved):

PR #1523, opened 2020-11-16
SAML enhancements

vbessonov:
## Description

This PR implements the following SAML enhancements:
- ability to select federated IdPs from InCommon instead of manually specifying their XML metadata
- ability to filter patrons by their attributes (for example, we can filter out all the patrons who don't have `eresources` entitlement attribute)

This PR uses enhancements in `server_core` added in [PR # 1217](https://github.com/NYPL-Simplified/server_core/pull/1217).

## Motivation and Context

[SIMPLY-3272](https://jira.nypl.org/browse/SIMPLY-3272)

## Checklist:


leonardr:
What does it mean to filter patrons by their attributes? Are we using the admin interface to filter a big list of patrons to make the list a manageable size, or are we "filtering" them  to deny them access or something?

(I have not looked at the branch yet--I want to know what I'm looking at.)

api/saml/configuration/model.py
```diff
@@ -0,0 +1,604 @@
...
+        key="saml_filter_expression",
+        label=_("Filter Expression"),
+        description=_(
+            "Python expression used for filtering out patrons by their SAML attributes."
+            "<br>"
+            "<br>"
+            'For example, if you want to authenticate using SAML only patrons having "eresources" '
+            'as the value of their "eduPersonEntitlement" then you need to use the following expression:'
+            "<br>"
+            '<pre>subject.attribute_statement.attributes["eduPersonEntitlement"].values[0] == "eresources"</pre>'
```
leonardr, line 89: So this answers the question I asked initially -- we're denying patrons access based on the results of the 'filter' -- but now I have another question. Is this a DSL or are we asking people to write Python code here? That seems really bad, both in terms of usability and security.

api/saml/configuration/model.py
```diff
@@ -0,0 +1,604 @@
...
+            'If "eduPersonEntitlement" can have multiple values, you can use the following expression:'
+            "<br>"
+            "<pre>"
+            """
+any([
+        value 
+        for value in subject.attribute_statement.attributes["eduPersonEntitlement"].values 
+        if value == "eresources"
+    ]
+)
```
leonardr, line 100: This is definitely Python. I'll check to see how you're actually executing this code, but this is pretty close to being a deal breaker for me. Even without the security risk it seems easy to bring the whole site down accidentally.

api/saml/metadata/filter.py
```diff
@@ -0,0 +1,131 @@
...
+            raise ValueError("Argument 'expression' must be a non-empty string")
+        if not isinstance(subject, SAMLSubject):
+            raise ValueError("Argument 'subject' must an instance of Subject class")
+
+        self._logger.info(
+            "Started applying expression '{0}' to {1}".format(expression, subject)
+        )
+
+        try:
+            result = eval(expression)
```
leonardr, line 97: This is what I was afraid of. Is possible to represent these filters as XPath expressions instead of Python code?

vbessonov: The problem is that we don't know how complex can these filter expression can be: depending on the university they might want to filter based on several attributes or their combinations using plain mathematical operators or regular expressions.
I see the following alternatives:
- Add additional UI controls to `circulation-web` which would allow to create complex expression using disjunction, conjunction, and composition
- Change `eval` to [`asteval`](https://newville.github.io/asteval/motivation.html#how-safe-is-asteval) (haven't checked whether it really has all the features)
- [Minimise the security issues of `eval`](https://realpython.com/python-eval-function/#minimizing-the-security-issues-of-eval)

leonardr: How do other SAML clients implement features like this?

`asteval` seems mainly concerned with preventing segfaults; my big concern is more access to the database.

Of the three options, the first one is the one that appeals to me, though we could achieve the same result by writing a DSL for matching SAML attributes -- in fact I hope someone else has written one and we can just copy it. e.g. would [Shibboleth's AttributeFilterPolicyConfiguration](https://wiki.shibboleth.net/confluence/display/IDP4/AttributeFilterPolicyConfiguration) do the job? We could implement a subset of that and gradually improve it as necessary.

vbessonov: Attribute Filtering policies are used by IdPs:
> It is about constraining the set of attributes and/or attribute values prior to them being used for some purpose, typically either for passing along to a relying party or less often to limit data passing into the IdP itself from another source. In this way, you can tailor the attributes seen by a specific relying party for a specific subject. 

I think in this case it's better to go with some business rule engines to create a DSL. For example, [this one](https://github.com/venmo/business-rules) looks promising

vbessonov: Actually, I found a [question](https://stackoverflow.com/questions/50275634/python-saml-onelogin-equivalent-to-simplesamlphp-filters) on StackOverflow asking about a similar thing. This feature is implemented in [SimpleSAMLphp](https://simplesamlphp.org/docs/stable/simplesamlphp-authproc) and partially in [PySAML2](https://github.com/IdentityPython/pysaml2) but it's not present in [python-saml](https://github.com/onelogin/python-saml) which we're using.

vbessonov: Actually, I was wrong, the libraries I mentioned above doesn't allow to filter by attribute values. It seems that only Shibboleth SP allows to it using [AttributeFilter](https://wiki.shibboleth.net/confluence/display/SP3/AttributeFilter).

leonardr: Do you think a library is likely to have an AttributeFilter already configured that they could copy-and-paste into their CM configuration?

leonardr:
If you can provide a branch that only includes the first bit of functionality you mentioned -- "ability to select federated IdPs from InCommon instead of manually specifying their XML metadata" -- I think we can get that merged without a lot of additional work.

api/saml/metadata/filter.py
vbessonov: Right now I'm not even sure that AttributeFilter has the functionality that we need. [The documentation](https://wiki.shibboleth.net/confluence/display/SP3/AttributeFilter) describes the following cases:
- limiting the values of an attribute whose values are required to be from an enumeration (e.g., the eduPersonScopedAffiliation attribute)
- applying automated rules for the acceptance of attribute "scopes" based on SAML metadata
- blocking self asserted personal identification data from known "open" IdPs
- limiting custom attributes intended to be used by only a single IdP

api/saml/configuration/model.py
```diff
@@ -0,0 +1,611 @@
...
+        """
+        identity_providers = []
+
+        if self.non_federated_identity_provider_xml_metadata:
+            parsing_results = self._metadata_parser.parse(
+                self.non_federated_identity_provider_xml_metadata
+            )
+            identity_providers = [
+                parsing_result.provider for parsing_result in parsing_results
+            ]
```
leonardr, line 193: Note to myself: need to make sure that this is only parsed during LibraryAuthenticator construction.

migration/20201112-add-incommon-saml-federation-metadata.py
```diff
@@ -0,0 +1,31 @@
...
+
+    if not incommon_federation:
+        incommon_federation = SAMLFederation(
+            incommon.FEDERATION_TYPE,
+            incommon.IDP_METADATA_SERVICE_URL,
+            incommon.CERTIFICATE
+        )
+
+        db.add(incommon_federation)
+        db.commit()
```
leonardr, line 31: This adds a `SAMLFederation` for sites that are being migrated from an earlier release. What about sites that are started up after this release happens? I don't think they'll have this.

I recommend a follow-up branch where you look at `SessionManager.initialize_data` and change it to use a registry, so different data model classes can register data initialization code.

Since data initialization code runs every time the server starts up, this will let you get rid of the migration script -- write the code correctly and a site will be guaranteed to have an appropriate SAMLFederation object on startup.

tests/saml/metadata/federations/test_validator.py
```diff
@@ -0,0 +1,134 @@
...
+                tests.saml.fixtures.FEDERATED_METADATA_WITH_VALID_UNTIL_ATTRIBUTE,
+                None,
+            ),
+            (
+                "with_real_incommon_metadata",
+                datetime.datetime(2020, 11, 26, 14, 32, 42),
+                open(
+                    os.path.join(
+                        os.path.dirname(os.path.abspath(__file__)),
+                        "../../../files/saml/incommon-metadata-idp-only.xml",
```
leonardr, line 70: Is it necessary to include this entire 38M document to do a good test? If the answer is yes then that's fine; I ran the test to check on timing and it only added about 1 second to the test time. But it's the biggest file in the repo so I want to see whether there's a way to run this test with a subset of the metadata, or whether it's better to make sure we handle every real case.

api/saml/metadata/federations/incommon.py
```diff
@@ -0,0 +1,33 @@
+FEDERATION_TYPE = "InCommon"
+
+IDP_METADATA_SERVICE_URL = "https://mdq.incommon.org/entities/idps/all"
+
+CERTIFICATE = """-----BEGIN CERTIFICATE-----
```
matthewhelt, line 5: Ideally a certificate should not be in the code base but instead stored somewhere in either a configuration file or database. This is ok for now but we'll want to move things like certs & secret keys to configuration in the future.

tests/saml/metadata/federations/test_validator.py
vbessonov: Unfortunately, we can't just rip a snippet out of this document. The document contains a signed signature and the snippet won't pass validation of the signature because snippet's hash would be different.
However, there is a workaround: we can create a new pair of keys, sign the snippet using the private keys and use the public one to validate it.
But I just wanted to make sure that the real data will be processed correctly.

leonardr:
I know you've just made a bunch of branches, but this PR should have been even more branches:

* One just for cosmetic changes like changing single quotes to double quotes, class renames, and import moves
* One adding localization support.
* One adding InCommon support.
* One adding filter support for patrons by attributes.

That said, the only issues I have are the use of `exec`, which you solved with the DSL, and the issue about new installations not being populated with InCommon data, which you can resolve in a separate branch.

I'll approve merge once I hear from Matthew, who knows much more about SAML than I do.

api/saml/metadata/federations/incommon.py
leonardr: This does get put into the database as part of a migration script. This also tripped me up--putting this data in  its own module makes it look like it's going to be imported and used as a source of constants, rather than inserted into the database and pulled out later.

I don't see a way to make this stuff end up in the database upon initialization without including it in the source code, but it does need to be revisited so it doesn't rely on a specific migration script having run.

tests/saml/metadata/federations/test_validator.py
leonardr: This makes sense.

leonardr (approved):

leonardr:
I filed https://jira.nypl.org/browse/SIMPLY-3386 to deal with the issue of populating a brand new database with InCommon data, so this is ready to merge.

PR #1526, opened 2020-11-19
Remove NYPL URL in docker build

jonathangreen:
## Description

The docker build currently assumes that the core sub-module has to come from an NYPL repo. This can be annoying when building a version of the docker containers for testing that has core commits from a different repo.

## Motivation and Context

This replaces the hard-coded NYPL url with a perl regex that replaces the `git` url with a `https` one. This does the same thing as the previous replacement for a NYPL URL, but allows other URLs to be used as well.

## How Has This Been Tested?

A container was build and pushed with this change an its [available on dockerhub](https://hub.docker.com/layers/lyrasis/circ-webapp/remove_hardcoded_nypl_url/images/sha256-10fcb78a9171d7f6943847241ea4aadc73b91d57702c87ea25c49978c50f9750?context=repo&tab=layers).

### DIFF
--- a/docker/simplified_app.sh
+++ b/docker/simplified_app.sh
@@ -46,7 +46,7 @@ cd circulation
 git checkout $version
 
 # Use https to access submodules.
-git config submodule.core.url https://github.com/NYPL-Simplified/server_core.git
+set +x && git config submodule.core.url $(git config submodule.core.url | perl -p -e 's|git@(.*?):|https://\1/|g') && set -x
 git submodule update --init --recursive
 
 # Add a .version file to the directory. This file
--- a/docker/simplified_app.sh
+++ b/docker/simplified_app.sh
@@ -46,6 +46,7 @@ cd circulation
 git checkout $version
 
 # Use https to access submodules.
+git submodule init
 git config submodule.core.url $(git config submodule.core.url | perl -p -e 's|git@(.*?):|https://\1/|g')
 git submodule update --init --recursive
 
--- a/docker/simplified_app.sh
+++ b/docker/simplified_app.sh
@@ -46,7 +46,7 @@ cd circulation
 git checkout $version
 
 # Use https to access submodules.
-set +x && git config submodule.core.url $(git config submodule.core.url | perl -p -e 's|git@(.*?):|https://\1/|g') && set -x
+git config submodule.core.url $(git config submodule.core.url | perl -p -e 's|git@(.*?):|https://\1/|g')
 git submodule update --init --recursive
 
 # Add a .version file to the directory. This file


docker/simplified_app.sh
```diff
@@ -46,7 +46,8 @@ cd circulation
 git checkout $version
 
 # Use https to access submodules.
-git config submodule.core.url https://github.com/NYPL-Simplified/server_core.git
+git submodule init
```
jonathangreen, line 49: This has to be called in order to initialize the config that is changed below. It initializes the config, but doesn't actually checkout the module. From the [git manual](https://git-scm.com/docs/git-submodule#Documentation/git-submodule.txt-init--ltpathgt82308203).

> Initialize the submodules recorded in the index (which were added and committed elsewhere) by setting submodule.$name.url in .git/config. ...
> You can then customize the submodule clone URLs in .git/config for your local setup and proceed to git submodule update; you can also just use git submodule update --init without the explicit init step if you do not intend to customize any submodule locations.

leonardr (commented):
This makes sense to me but I'd like to get an opinion from someone like @alsrlw who would also be using this feature.

alsrlw:
I like the idea of allowing a variable value for this, which would be handy especially in development (essential in customizations not in the community Dockerfile). But I must admit to being lost in regular expressions. I'm looking to see how a custom URL gets passed in to that expression. It looks like, maybe, the 1 parameter references the ${repo} value in the Dockerfile (or maybe I've proved I get lost in it!). If so, is there a missing `github.com/` bit (in the Dockerfile `ARG repo=` value) to fully reconstitute the original NYPL URL? 

Pardon the ignorance if this is all off-track! In short, yes, I'm good with the idea of making it a variable value.

jonathangreen:
The `repo` argument is used in the [git clone](https://github.com/NYPL-Simplified/circulation/blob/develop/docker/simplified_app.sh#L43). But then once we have the git repo cloned down it has the submodule URL embedded in it. 

What we used to do is just clobber that embedded URL with the NYPL git URL. 

What this PR does is:
- Run `git config submodule.core.url` this dumps out the URL embedded in the repo
  - In the case of the main repo this is:
    ```
    $ git config submodule.core.url
    git@github.com:NYPL-Simplified/server_core.git
    ```
- Pipe that to a perl regex:
  ```
  $ git config submodule.core.url | perl -p -e 's|git@(.*?):|https://\1/|g'
  https://github.com/NYPL-Simplified/server_core.git
  ```
- Set that as the new `submodule.core.url`

So even though the `repo` arg doesn't take the full URL the PR still works, since by the time it gets here we are pulling the full URL out of the git config. 

Hopefully that makes sense! Regex are always a bit difficult to sort out.

alsrlw:
Ah, that makes sense. Thanks. I knew I was in the weeds. :-)

leonardr:
Sounds like we're good to merge this?

PR #1527, opened 2020-11-19
Add logrotation for UWSGI logs

jonathangreen:
## Description

UWSGI puts logs in `/var/log/uwsgi` and these are not currently being rotated by the logrotate config.

## Motivation and Context

The logs aren't huge, so they don't quickly fill the disk, but they do slowly creep up in size. This makes sure they are properly rotated, just as the other logs are.

## How Has This Been Tested?

This config has been running on the LYRASIS servers for a few months and has been properly rotating the UWSGI logs for us. Thought I would contribute the change upstream if there is interest in it.

### DIFF
--- a/docker/services/simplified_logrotate.conf
+++ b/docker/services/simplified_logrotate.conf
@@ -9,3 +9,15 @@
     notifempty
     dateext
 }
+
+/var/log/uwsgi/*.log {
+    missingok
+    daily
+    create 0660 simplified adm
+    rotate 13
+    copytruncate
+    compress
+    delaycompress
+    notifempty
+    dateext
+}


leonardr (approved):

docker/services/simplified_logrotate.conf
```diff
@@ -9,3 +9,15 @@
     notifempty
     dateext
 }
+
+/var/log/uwsgi/*.log {
+    missingok
+    daily
+    create 0660 simplified adm
+    rotate 13
```
alsrlw, line 17: Since the uwsgi log isn't terribly large, I think I'd prefer a longer default retention time. Every once in a great while that may come in handy. Maybe 30 days? Or, perhaps the value could be "variablized" and an ENV made available for run time? Otherwise, all good.

leonardr: Apologies for merging this before you had a chance to weigh in. I'm OK with either 13 or 30 days as the retention time.

jonathangreen: I'm fine with 30 days. I only chose 13 so it was the same as the other logrotate configs in the container. 

If everyone is happy with 30, I can make another PR to change it.

jonathangreen: PR here to change the logs kept to 30: 
https://github.com/NYPL-Simplified/circulation/pull/1529

PR #1528, opened 2020-11-19
Upgrade Nodejs Package in Docker Image

jonathangreen:
## Description

This PR does two things: 
- Updates the base image that we use
- Uses a version of Node 10 from package management

### Change in base image

We are currently using the [`latest-amd64`](https://registry.hub.docker.com/layers/phusion/baseimage/latest-amd64/images/sha256-6c28526342cc588732aff25fc97e2c7054899cee6ff03f8f0f34ec4c0c66bb4e?context=explore) tag for our base image. It was last pushed more then a year ago. The [phusion docs](https://github.com/phusion/baseimage-docker#getting-started) say we should be using one or their [released versions](https://github.com/phusion/baseimage-docker/releases). The latest stable release is 'bionic-1.0.0' which was pushed 5 months ago. Since its not architecture specific and has been released more recently this seems like the tag we should be using.

### Node 10 from Nodesource Packages

Currently we are compiling our own version of `libxmlsec1` because the verison of OpenSSL it requires conflicts with the version of OpenSSL needed for NodeJS 8. 

NodeJS 8 is out of support and several node modules we are using complain about the old version we are using when doing 'npm install' inside the container.
```
simplified-circulation-web@0.4.30: wanted: {"node":">=8.0.0","npm":">=5.0.0"} (current: {"node":"8.10.0","npm":"3.5.2"})
jsdom@16.4.0: wanted: {"node":">=10"} (current: {"node":"8.10.0","npm":"3.5.2"})
data-urls@2.0.0: wanted: {"node":">=10"} (current: {"node":"8.10.0","npm":"3.5.2"})
w3c-xmlserializer@2.0.0: wanted: {"node":">=10"} (current: {"node":"8.10.0","npm":"3.5.2"})
webidl-conversions@6.1.0: wanted: {"node":">=10.4"} (current: {"node":"8.10.0","npm":"3.5.2"})
whatwg-url@8.4.0: wanted: {"node":">=10"} (current: {"node":"8.10.0","npm":"3.5.2"})
saxes@5.0.1: wanted: {"node":">=10"} (current: {"node":"8.10.0","npm":"3.5.2"})
html-encoding-sniffer@2.0.1: wanted: {"node":">=10"} (current: {"node":"8.10.0","npm":"3.5.2"})
```

This PR switches from using the version for Node from Ubuntu package management to the version from the [nodesource  repo](https://github.com/nodesource/distributions). This allows us to just install `libxmlsec` from Ubuntu package management, which should reduce container build time and make sure we have any security updates to that package provided by Ubuntu.

@EdwinGuzman @adriana-alter @kristojorg Is it fine to be building `simplified-circulation-web` using Node 10? It appears the travis tests on that repo are being run on Node 10 and the changelog seems to confirm that building with Node 10 is supported, but I wanted to check to be sure.

## How Has This Been Tested?

There is a build of this branch [on dockerhub](https://registry.hub.docker.com/layers/lyrasis/circ-webapp/upgrade-node/images/sha256-f3d5a001f4676e8dbbe87caf6d69f9653f52274fea68c24ae9061a5d409979ba?context=explore). I've run the circulation manager and core tests inside this container and tested out the container in our development environment and everything seems to be working fine.

### DIFF
--- a/docker/Dockerfile.exec
+++ b/docker/Dockerfile.exec
@@ -12,7 +12,7 @@ RUN go get -v github.com/readium/readium-lcp-server/lcpencrypt
 ## Final image
 ###############################################################################
 
-FROM phusion/baseimage:latest-amd64
+FROM phusion/baseimage:bionic-1.0.0
 
 COPY --from=builder /go/bin/lcpencrypt /go/bin/lcpencrypt
 
--- a/docker/Dockerfile.scripts
+++ b/docker/Dockerfile.scripts
@@ -12,7 +12,7 @@ RUN go get -v github.com/readium/readium-lcp-server/lcpencrypt
 ## Final image
 ###############################################################################
 
-FROM phusion/baseimage:latest-amd64
+FROM phusion/baseimage:bionic-1.0.0
 
 COPY --from=builder /go/bin/lcpencrypt /go/bin/lcpencrypt
 
--- a/docker/Dockerfile.webapp
+++ b/docker/Dockerfile.webapp
@@ -1,4 +1,4 @@
-FROM phusion/baseimage:latest-amd64
+FROM phusion/baseimage:bionic-1.0.0
 MAINTAINER Library Simplified <info@librarysimplified.org>
 
 ARG version
--- a/docker/simplified_app.sh
+++ b/docker/simplified_app.sh
@@ -6,7 +6,17 @@ set -x
 repo="$1"
 version="$2"
 
-apt-get update && $minimal_apt_get_install python-dev \
+# Install the nodesource nodejs package
+# This lets us use node 10 and avoids dependency conflict between node and libxmlsec1 over the
+# version of the ssl library that we find from package managemnet
+curl -sSL https://deb.nodesource.com/gpgkey/nodesource.gpg.key | apt-key add -
+echo "deb https://deb.nodesource.com/node_10.x bionic main" >> /etc/apt/sources.list.d/nodesource.list
+echo "deb-src https://deb.nodesource.com/node_10.x bionic main" >> /etc/apt/sources.list.d/nodesource.list
+
+# Add packages we need to build the app and its dependancies
+apt-get update
+$minimal_apt_get_install --no-upgrade \
+  python-dev \
   python2.7 \
   python-nose \
   python-setuptools \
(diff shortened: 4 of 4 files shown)


jonathangreen:
Travis failed on this, but I think it just needs to be restarted. Looks like it was a hiccup rather then an actual fail.

alsrlw:
I like both the updates: to the base image and to Node. Without having tested, the major factor would seem to be ongoing support for Python 2.7. If the newer base images have no problem with our current staple of frozen Python 2.7 packages, then it should be a better option. As for Node, it looks like the dev team for `circulation-patron-web` is working from Node v12. At least the Dockerfile in that project begins with a Node 12 base image (https://github.com/NYPL-Simplified/circulation-patron-web/blob/dev/Dockerfile). Anyway, I like the idea.

tdilauro:
Looks good. Definitely a step forward. Very happy to have that custom xmlsec1 build out of there.

EdwinGuzman:
The node update looks good but I can't comment on the rest. `circulation-web` works on node 10 which is what Adriana and I use. I haven't tested it on node 12 though. The `circulation-patron-web` app is a different application, but I don't think there'll be any issues if `circulation-web` is also on node 12. To be safe, node 10 would be the best choice right now until we test node 12 capability.

vbessonov:
@jonathangreen, change of `libxmlsec` version can also affect SAML, it's used internally by OneLogin library

jonathangreen:
@vbessonov is there a particular version of `libxmlsec` you would recommend or know we need for SAML? Or is there additional testing that should be done here? If its possible, I think it is preferable to use the version from package management. All of the SAML tests pass with this version of the library.

vbessonov:
@jonathangreen, I just checked `.travis.yml` and it also has the following dependencies installed using `apt`:
```
      - libxmlsec1-dev
      - libxml2-dev
```
[The new PR](https://github.com/NYPL-Simplified/circulation/pull/1523) containing SAML enhancements where I use `xmlsec` under the hood to verify the validity of the XML's signature also passed using those dependencies.
It seems it should be enough to tell that SAML authentication will correctly after the changes done in this branch.

kristojorg:
Yep edwin is right this shouldn't impact circulation-patron-web at all. We build on Node 12 in docker though

jonathangreen:
@leonardr bumping this one to see if there is anything else I should change here.

leonardr:
I don't know enough about this part of the system to have an informed opinion, but if Kristo and Edwin are happy with this then let's go for it.

leonardr (approved):

leonardr:
I filed https://jira.nypl.org/browse/SIMPLY-3351 to track this work.

PR #1529, opened 2020-11-19
Increase the logs we keep for UWSGI.

jonathangreen:
## Description

Followup to https://github.com/NYPL-Simplified/circulation/pull/1527 based on @alsrlw feedback that perhaps keeping these logs for longer would be a good idea. 

## Motivation and Context

I modified this value to be 30 days, based on the feedback on #1527. But I'm open to other values as well. I was mostly concerned that this log doesn't grow unbounded more then the retention time.


leonardr:
30 sounds good to me.

PR #1530, opened 2020-12-01
Update Overdrive availability API to v2

leonardr:
## Description

This branch goes along with https://github.com/NYPL-Simplified/server_core/pull/1221 to get us using v2 of the Overdrive availability API without making any other changes.

## Motivation and Context

https://jira.nypl.org/browse/SIMPLY-3302

## How Has This Been Tested?

I have tested this using real Overdrive credentials, but they were for a regular Overdrive account, not an Overdrive Advantage account.

## Checklist:

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -1037,13 +1037,18 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI, HasSelfTests):
                 exc_info=e
             )
 
-        if status_code != 200:
+        # TODO: If you ask for a book that you know about, and
+        # Overdrive says the book doesn't exist in the collection,
+        # then it's appropriate to update an existing
+        # LicensePool. However we shouldn't be creating a *brand new*
+        # LicensePool for a book Overdrive says isn't in the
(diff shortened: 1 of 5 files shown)


api/overdrive.py
```diff
@@ -1037,13 +1037,18 @@ def update_licensepool(self, book_id):
...
             )
 
-        if status_code != 200:
+        # TODO: If you ask for a book that you know about, and
+        # Overdrive says the book doesn't exist in the collection,
+        # then it's appropriate to update an existing
+        # LicensePool. However we shouldn't be creating a *brand new*
+        # LicensePool for a book Overdrive says isn't in the
+        # collection.
+        if status_code not in (200, 404):
```
leonardr, line 1046: This is the actual fix for the problem we're seeing: the Overdrive API is telling us that a book is no longer in the collection, but because they're sending a 404 error (rather than a 200 that says "there are no licenses") we're ignoring that information and acting like the book is still present in the collection.

tdilauro (approved):
Looks good to me.

PR #1531, opened 2020-12-10
Fix the issue related to ProQuest book covers not loading in SimplyE

vbessonov:
## Description

This PR changes overrides the way `ProQuestOPDS2Importer` processes cover links: for each cover links `ProQuestOPDS2Importer` must create two `LinkData` objects, one for the full cover and the other for the thumbnail.

**NOTE:** This PR depends on [PR # 1523](https://github.com/NYPL-Simplified/circulation/pull/1523).

## Motivation and Context

[SIMPLY-3364](https://jira.nypl.org/browse/SIMPLY-3364)

## Checklist:

### DIFF
--- a/api/proquest/importer.py
+++ b/api/proquest/importer.py
@@ -3,19 +3,20 @@ import logging
 from contextlib import contextmanager
 
 import six
+from flask_babel import lazy_gettext as _
+from requests import HTTPError
+from sqlalchemy import or_
+from webpub_manifest_parser.utils import encode
+
 from api.circulation import BaseCirculationAPI, FulfillmentInfo, LoanInfo
 from api.circulation_exceptions import CannotFulfill, CannotLoan
 from api.proquest.client import ProQuestAPIClientConfiguration, ProQuestAPIClientFactory
 from api.proquest.credential import ProQuestCredentialManager
 from api.proquest.identifier import ProQuestIdentifierParser
 from api.saml.metadata.model import SAMLAttributeType
-from flask_babel import lazy_gettext as _
-from requests import HTTPError
-from sqlalchemy import or_
-
 from core.classifier import Classifier
 from core.exceptions import BaseError
-from core.model import Collection, Identifier, LicensePool, Loan, get_one
+from core.model import Collection, Hyperlink, Identifier, LicensePool, Loan, get_one
 from core.model.configuration import (
     ConfigurationAttributeType,
     ConfigurationFactory,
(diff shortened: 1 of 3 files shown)


api/proquest/importer.py
```diff
@@ -390,6 +389,49 @@ def _get_book(self, patron, configuration, document_id):
...
+                default_link_rel=Hyperlink.IMAGE,
+            )
+            image_links.append(cover_link)
+
+            thumbnail_link = self._extract_link(
+                image_link,
+                feed_self_url,
+                default_link_rel=Hyperlink.THUMBNAIL_IMAGE,
+            )
+            image_links.append(thumbnail_link)
```
leonardr, line 425: cover_link will be a `LinkData`, and `LinkData` has a `.thumbnail` attribute, so you should set `cover_link.thumbnail = thumbnail_link` rather than adding `thumbnail_link` to `image_links`. Otherwise it's not explicit that `thumbnail_link` is a thumbnail _of_ `cover_link` -- it could be a thumbnail of some other image.

metadata_layer might be smart enough to figure it out, since there's one image and one thumbnail, but it's better to be explicit.

leonardr: I see how it's working now, because the Proquest feed doesn't specify a link relation for its cover image, but  `default_link_rel` won't do anything if a link relation is specified in the OPDS. In particular, if Proquest starts specifying rel="image", you'll end up with the thumbnail having a rel="image", which metadata_layer will reject.

Since you know these are images (they're in the `images` section of the OPDS entry), i would explicitly set `cover_link.rel` and `thumbnail_link.rel` rather than (or in addition to) expecting `default_link_rel` to do it.

tests/proquest/test_importer.py
```diff
@@ -839,64 +842,39 @@ def test_fulfil_refreshes_expired_token(self):
...
+
+        full_cover_link = publication_metadata.links[0]
+        eq_(True, isinstance(full_cover_link, LinkData))
+        eq_(fixtures.PROQUEST_RAW_PUBLICATION_COVER_HREF, full_cover_link.href)
+        eq_(Hyperlink.IMAGE, full_cover_link.rel)
+
+        thumbnail_cover_link = publication_metadata.links[1]
+        eq_(True, isinstance(thumbnail_cover_link, LinkData))
+        eq_(fixtures.PROQUEST_RAW_PUBLICATION_COVER_HREF, thumbnail_cover_link.href)
+        eq_(Hyperlink.THUMBNAIL_IMAGE, thumbnail_cover_link.rel)
```
leonardr, line 877: Yes, this only tests that the LinkData objects were created. I think if you call `metadata.apply` in this situation you won't get the results you want, because the image and the thumbnail aren't connected.

leonardr (changes requested):

tests/proquest/test_importer.py
leonardr: I see, the end result is still valid, even though the work's record in the database is in an inconsistent state. I'm glad that works but yes, it would be better to have the image associated with its thumbnail.

leonardr (approved):
This looks good now, i'll merge once the test suite completes.

PR #1533, opened 2020-12-10
Update simplified-circulation-web

vbessonov:
## Description

This PR updates the version of `simplified-circulation-web` package to the latest - `v0.4.30`, containing different bug fixes, the most important of which is described in [SIMPLY-3367](https://jira.nypl.org/browse/SIMPLY-3367).

## Motivation and Context

[SIMPLY-3367](https://jira.nypl.org/browse/SIMPLY-3367)

## Checklist:

### DIFF
--- a/api/admin/package.json
+++ b/api/admin/package.json
@@ -13,6 +13,6 @@
     "npm": ">=5.0.0"
   },
   "dependencies": {
-    "simplified-circulation-web": "0.4.30"
+    "simplified-circulation-web": "^0.4.30"
   }
 }
--- a/api/admin/package-lock.json
+++ b/api/admin/package-lock.json
@@ -5,20 +5,20 @@
   "requires": true,
   "dependencies": {
     "@babel/runtime": {
-      "version": "7.8.4",
-      "resolved": "https://registry.npmjs.org/@babel/runtime/-/runtime-7.8.4.tgz",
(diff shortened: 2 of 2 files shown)


api/admin/package.json
```diff
@@ -13,6 +13,6 @@
     "npm": ">=5.0.0"
   },
   "dependencies": {
-    "simplified-circulation-web": "0.4.30"
+    "simplified-circulation-web": "^0.4.30"
```
leonardr, line 16: Do you know why this was 0.4.30 before when package-lock.json had it at 0.4.11? Is it just that the dependencies hadn't been updated to reflect a change in package.json?

vbessonov: Yes, that's what I see in the git log. @adriana-alter, am I right that we have to update `package-lock.json` because `npm` gets package versions out of it, not from `package.json`?

adriana-alter: Absolutely need to update `package-lock.json` (it's just a description of your currently installed package versions; it should never be out of sync with `package.json`!), but not manually; delete your local `package-lock.json` and run `npm install` to generate a new, up-to-date one.

vbessonov: That's what I basically did. However, I didn't delete `package-lock.json` before but just run `npm install`.

leonardr:
OK, based on what Adriana said I think we're good to merge.

leonardr (approved):

PR #1535, opened 2020-12-10
Do not save transient Name Ids in credentials

vbessonov:
## Description

This PR adds additional logic which doesn't allow Circulation Manager to save transient Name IDs.
**NOTE:** This PR depends on [1534](https://github.com/NYPL-Simplified/circulation/pull/1534).

## Motivation and Context

[SIMPLY-3368](https://jira.nypl.org/browse/SIMPLY-3368)

## Checklist:

### DIFF
--- a/api/saml/provider.py
+++ b/api/saml/provider.py
@@ -1,6 +1,7 @@
 import datetime
 import json
 import logging
+from copy import deepcopy
 
 from contextlib2 import contextmanager
 from flask import url_for
@@ -14,6 +15,7 @@ from api.saml.configuration.validator import SAMLSettingsValidator
 from api.saml.metadata.filter import SAMLSubjectFilter
 from api.saml.metadata.model import (
     SAMLLocalizedMetadataItem,
+    SAMLNameIDFormat,
     SAMLSubject,
     SAMLSubjectJSONEncoder,
     SAMLSubjectUIDExtractor,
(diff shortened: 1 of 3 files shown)


api/saml/provider.py
```diff
@@ -209,9 +211,32 @@ def _get_idp_display_names(identity_provider_index, identity_provider):
...
+        :rtype: str
+        """
+        subject = deepcopy(subject)
+
+        # We should not save a transient Name ID because it changes each time
+        if (
+            subject.name_id
+            and subject.name_id.name_format == SAMLNameIDFormat.TRANSIENT.value
+        ):
+            subject.name_id = None
```
leonardr, line 231: I assume the reason this code isn't in `SAMLSubjectJSONEncoder` is that not only do we not want to put a transient name into the token, we don't even want to store it in the `Subject` object.  Is that right

leonardr (approved):
This will work but it seems a little unusual that the transient names are wiped from the Subject when it's time to create the token, rather than this happening when the Subject is created,  or happening _only_ inside the token, with the Subject being unaffected.

Am I missing something or is there a better way to do this?

api/saml/provider.py
vbessonov: Actually, I wanted to keep the `SAMLSubject` object untouched, I didn't want to mutate the state during saving process because I thought it might lead to some unwanted side-effects.

leonardr: That makes sense, but doesn't `subject.name_id = None` change the state of the `SAMLSubject`?

vbessonov:
This is a really good question. I think we have to create a `SAMLSubject` object a transient name ID in the begging because it could be used during filtration. I can't provide with any real world examples but still think it's better to have "full" SAML subject. Regarding making a copy rather than directly modifying the object, I just wanted to get rid of any possible side effects. Right now the logic is pretty simple and the token creation is the last stage but if it changes in the future and we decide to add some post-processing I would rather prefer to explicitly specify that I want to strip out a transient name ID rather than it being it removed during token creation.

api/saml/provider.py
vbessonov: Yes, it does but it changes the deep copy of it so the original object left unchanged.

leonardr: Got it, thanks.

PR #1540, opened 2020-12-21
Use v2 of the availability API even if we get a v1 link from another API request.

leonardr:
## Description

This branch fixes https://jira.nypl.org/browse/SIMPLY-3370, in conjunction with [core#1228](https://github.com/NYPL-Simplified/server_core/pull/1228).

## Motivation and Context

When we generate a URL to Overdrive's availability endpoint, we use v2 of the API. But sometimes Overdrive gives us links that we follow, and those links may use v1 of the API.

## Checklist:

### DIFF
--- a/api/overdrive.py
+++ b/api/overdrive.py
@@ -991,8 +991,10 @@ class OverdriveAPI(BaseOverdriveAPI, BaseCirculationAPI, HasSelfTests):
             )
             book = dict(id=book_id)
         else:
-            book_id = book['id']
             circulation_link = book['availability_link']
+            # Make sure we use v2 of the availability API,
+            # even if Overdrive gave us a link to v1.
+            circulation_link = self.make_link_safe(circulation_link)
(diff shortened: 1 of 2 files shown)


api/overdrive.py
```diff
@@ -991,8 +991,10 @@ def circulation_lookup(self, book):
             )
             book = dict(id=book_id)
         else:
-            book_id = book['id']
             circulation_link = book['availability_link']
+            # Make sure we use v2 of the availability API,
+            # even if Overdrive gave us a link to v1.
+            circulation_link = self.make_link_safe(circulation_link)
```
leonardr, line 997: There may be other places in overdrive.py where we need to make this change, but this is the only place where we reference 'availability_link'.

This bit of the API client is actually the first code I wrote for this project, and it's a mess that needs refactoring... but not today.

vbessonov (approved):

PR #1543, opened 2021-01-07
Solo OPDS catalog entries must contain self links

vbessonov:
## Description

This PR adds an ability to override `rel` of `application/atom+xml;type=entry;profile=opds-catalog` links: depending on whether the OPDS entry is complete or incomplete `rel` will be set to `self` or `alternate` accordingly.

This PR depends on [PR # 1230](https://github.com/NYPL-Simplified/server_core/pull/1230) in server_core.

## Motivation and Context

[SIMPLY-3355](https://jira.nypl.org/browse/SIMPLY-3355)

## Checklist:

### DIFF
--- a/api/opds.py
+++ b/api/opds.py
@@ -81,6 +81,21 @@ class CirculationManagerAnnotator(Annotator):
         self.hidden_content_types = hidden_content_types
         self.test_mode = test_mode
 
+    def is_work_entry_solo(self, work):
+        """Return a boolean value indicating whether the work's OPDS catalog entry is served by itself,
+            rather than as a part of the feed.
+
+        :param work: Work object
+        :type work: core.model.work.Work
+
+        :return: Boolean value indicating whether the work's OPDS catalog entry is served by itself,
+            rather than as a part of the feed
+        :rtype: bool
+        """
+        return any(
+            work in x for x in (self.active_loans_by_work, self.active_holds_by_work, self.active_fulfillments_by_work)
+        )
+
     def _lane_identifier(self, lane):
         if isinstance(lane, Lane):
             return lane.id
--- a/tests/test_opds.py
+++ b/tests/test_opds.py
@@ -6,6 +6,7 @@ import os
 import re
 import json
 from lxml import etree
+from mock import create_autospec
 from nose.tools import (
     set_trace,
     eq_,
@@ -191,7 +192,6 @@ class TestCirculationManagerAnnotator(DatabaseTest):
         # The EPUB is hidden, and this license pool has no delivery
         # mechanisms.
         eq_([], list(no_epub.visible_delivery_mechanisms(pool)))
-
 
     def test_rights_attributes(self):
         m = self.annotator.rights_attributes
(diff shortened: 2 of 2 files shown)


api/opds.py
```diff
@@ -81,6 +81,14 @@ def __init__(self, lane,
         self.hidden_content_types = hidden_content_types
         self.test_mode = test_mode
 
+    def is_work_entry_complete(self, work):
+        """Return a boolean value indicating whether the work's OPDS catalog entry is complete.
```
leonardr, line 85: I can see a couple of problems with this. Most importantly, this will return true if this Annotator is prepared to generate a complete OPDS entry for _any_ work. It should only return true if the Annotator is prepared to generate a complete OPDS entry for _this_ work. So the code should look like:

```
return any(work in x for x in self.active_loans_by_work, self.active_holds_by_work, self.active_fulfillments_by_work)
```

Second, I don't think "complete catalog entry" is the right concept here. Section 5.1.2 of the OPDS spec says a complete catalog entry must "include all metadata elements".  I don't understand how an active loan/hold/fulfillment makes the difference here. An acquisition link isn't a metadata element.

I think it would be better to change the link from rel="alternate" to rel="self" across the board -- all of these OPDS entries are complete, and the use of "self" instead of "alternate" to indicate a complete entry makes sense as a convention.

Assuming it doesn't cause problems with existing clients, is there any reason not to _always_ use rel="self" here?

vbessonov: This is a very good question, indeed. This issue was raised on behalf of Saad, who was working on integrating CM OPDS feeds into Thorium. Since all the catalog entries can be considered "complete" according to the spec, I think it should be sufficient and absolutely right. I'll follow up with Saad and return with feedback.

leonardr: Thinking it out, the distinction I'm seeing is between an OPDS entry which is generic, not customized with special acquisition links based on the authenticated patron (e.g. because it's in a big feed that came from a cache), and an OPDS entry that _is_ customized (because of your recent change to the `permalink` endpoint).

This is a good distinction to make but it doesn't have a basis in the spec. I can see the appeal of using rel="alternate" in the big feed and rel="self" in the permalink for this reason.

leonardr (changes requested):
I see a path towards a change that will get Thorium working, but based on my conversation with Ettore we need to coordinate with the mobile developers so that the change doesn't break SimplyE or circulation-patron-web. These changes are taking place within the wide latitude allowed by the spec. It's good to get some conventions in place but all the clients need to be modified to understand the conventions.

leonardr (changes requested):
On a code level, both of these branches look fine. My only requested changes are around terminology:

1. All of these entires are "complete" as per the OPDS spec because they "include all metadata elements". Let's use a different term like "solo" to describe the entries that are served by themselves, rather than as part of a feed. So this would mean renaming "is_work_entry_complete" to "is_work_entry_solo".

2. Backwards compatibility isn't the reason the rel="alternate" links are being left in. SimplyE, circulation-patron-web, and Thorium are different products that make different assumptions. We're leaving the links in to support a variety of clients. At some point we may standardize the behavior, and at that point one of the links would be in there for backwards compatibility, but at this point it's not even clear which link we'd remove.

Sorry to be so pedantic but I want to be really careful when we are dealing with terms defined (or not defined) by the OPDS specs.

vbessonov:
Sure, I understand completely. I renamed the method to `is_work_entry_solo` and updated the comments in the tests accordingly.

leonardr (approved):

PR #1544, opened 2021-01-07
Add a jackpot feed for books that belong to _some_ series.

leonardr:
## Description

This branch adds a "jackpot" feed guaranteed to give you a book that belongs to a series.

This resolves https://jira.nypl.org/browse/SIMPLY-3235.

## Motivation and Context

Up to this point I've been adding jackpot feeds by giving more functionality to the `qa_feed` controller, because the goal of the jackpot feeds was to show the same sort of thing -- one book and audiobook from every supported vendor -- in different contexts. (Books that are available now, books that are not available now, etc.)

This type of feed is different -- we want a single book with a certain property that is not related to vendor or availability status. I refactored the `qa_feed` controller so that most of the code could be shared by this new endpoint, `qa_series_feed`. This should make it easier to write additional jackpot feeds in the future.

## How Has This Been Tested?

I created a local library using NYPL's Overdrive credentials, partially imported the collection, and tested the old qa_feed endpoint as well as the new qa_series_feed endpoint.

## Checklist:


api/lanes.py
```diff
@@ -1304,6 +1304,13 @@ def available_facets(cls, config, facet_group_name):
         return [cls.AVAILABLE_NOW, cls.AVAILABLE_NOT_NOW,
                 cls.AVAILABLE_ALL, cls.AVAILABLE_OPEN_ACCESS]
 
+class HasSeriesFacets(Facets):
+    """A faceting object for a feed containg books guaranteed
+    to belong to _some_ series.
+    """
+    def modify_search_filter(self, filter):
```
matthewhelt, line 1311: This might be the intended effect but the method name implies one could set this to true or false. Would direct modification of the filter.series attribute be more readable?

leonardr: We discussed this in Slack. `modify_search_filter` is the name of a hook method defined by `Facets` that can basically do anything it wants to a `Filter` object; it's not specific to this class.

matthewhelt (approved):
Class/method naming conventions were clarified by Leonard so my singular comment on this PR has been explained.

As this only change really applies to facilitating QA via "jackpot" feeds for testing I believe it to be low risk and ready for merge and validation.

PR #1545, opened 2021-01-08
Permalink must correctly process works with fulfillments

vbessonov:
## Description

This PR fixes a regression introduced in [SIMPLY-3354](https://jira.nypl.org/browse/SIMPLY-3354) when `permalink` route fails for works with fulfillments.

## Motivation and Context

[SIMPLY-3419](https://jira.nypl.org/browse/SIMPLY-3419)

## Checklist:

### DIFF
--- a/tests/test_controller.py
+++ b/tests/test_controller.py
@@ -1,49 +1,44 @@
 # encoding=utf8
-from nose.tools import (
-    assert_raises,
-    eq_,
-    set_trace,
-)
+import base64
 import calendar
-from contextlib import contextmanager
-import email
-import os
 import datetime
+import email
+import json
+import os
+import random
 import time
-import urlparse
-from wsgiref.handlers import format_date_time
-from time import mktime
+import urllib
+from contextlib import contextmanager
 from decimal import Decimal
-from mock import (
-    MagicMock,
-    patch
-)
+from time import mktime
+from wsgiref.handlers import format_date_time
 
+import feedparser
 import flask
-from flask import (
-    url_for,
-    Response as FlaskResponse,
-)
-from flask_sqlalchemy_session import (
-    current_session,
-    flask_scoped_session,
-)
+import urlparse
+from flask import Response as FlaskResponse
+from flask import url_for
+from flask_sqlalchemy_session import current_session
+from mock import MagicMock, patch
+from nose.tools import assert_raises, eq_
 from werkzeug.datastructures import ImmutableMultiDict
 from werkzeug.exceptions import NotFound
 
-from . import DatabaseTest
+from api.adobe_vendor_id import AuthdataUtility, DeviceManagementProtocolController
+from api.annotations import AnnotationWriter
 from api.app import app, initialize_database
-from api.config import (
-    Configuration,
-    temp_config,
+from api.authenticator import (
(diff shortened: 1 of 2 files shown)


leonardr (commented):
We definitely shouldn't be treating a LicensePoolDeliveryMechanism as though it were a FulfillmentInfo. But I'm not convinced this is the right solution. I suspect what we want inside `permalink` is to always pass in `fulfillment=None`.

I say this because "converting a LicensePoolDeliveryMechanism into a FulfillmentInfo" can be an extremely complicated vendor-specific task that might result in multiple calls to API endpoints. You generally only want to do that when you are calling `CirculationAPI.fulfill`, because `fulfill` handles all the edge cases and lets vendor-specific code run.

I think the `fulfillment` in `single_item_feed` is designed for a situation where _you already called CirculationAPI.fulfill()_ and the thing you got back can't be sent directly to the user -- it needs to be turned into an OPDS entry first.

`single_item_feed` is called in a couple places apart from `permalink` (which calls `single_item_feed` as of your recent PR).

1. When someone asks to fulfill a book in a streaming HTML format (within CirculationAPI.fulfill itself). We need to make a separate request to the API to find the time-limited URL of the web page that we can give to the patron. Then we generate an <entry> tag that includes that link. This is the only situation I could find where `fulfill` was not None, and it happened _inside_ a call to `fulfill`.
2. In LoanController.revoke and LoanController.detail. In this case `fulfill` is None. (LoanController.detail actually seems very similar to the changes you made to permalink, but I'm not sure where it's used.)
3. In SharedCollectionLoanController.loan_info, hold_info, and borrow. In these case `fulfill` is always None.

Basically, since you are not actually fulfilling a loan here, only giving information about a loan, it should not be necessary to provide `fulfill` at all.

vbessonov:
Thank you for noticing this. I just realised that sending `fulfillment` to `single_item_feed` in `permalink` could lead to really bad consequences. For example, in the case of self-hosted content protected by presigned S3 URLs it would lead to having unsigned URLs in an OPDS catalog entry.

tests/test_controller.py
```diff
@@ -3109,6 +3061,96 @@ def test_permalink_returns_fulfillment_links_for_authenticated_patrons_with_loan
...
+                pool.identifier.identifier,
+                datetime.datetime.utcnow(),
+                datetime.datetime.utcnow() + datetime.timedelta(seconds=3600),
+            )
+            self.manager.d_circulation.queue_checkout(
+                pool,
+                loan_info
+            )
+
+            fulfillment = FulfillmentInfo(
```
leonardr, line 3105: It should no longer be necessary to create a FulfillmentInfo or a Resource as part of test setup, since it's no longer considered when generating the OPDS.

leonardr (approved):
I think there's some test setup that can be removed but otherwise this looks good.

tests/test_controller.py
vbessonov: The following code requires a `FulfillmentInfo` object:
```python
self.manager.loans.fulfill(
    pool.id, delivery_mechanism.delivery_mechanism.id,
)
```
It might make sense to remove the whole test case but I would like to still keep it to make sure that the code correctly processes works with fulfillments (by just ignoring them).

PR #1546, opened 2021-01-11
Implement removals handling in the ProQuest integration

vbessonov:
## Description

The PR includes the following changes:
- it optimizes the import process and reduces the memory footprint by dumping all feed pages to a local disk
- it adds an ability to handle removals in the ProQuest feed and hide items that are no longer present there from the CM's catalog.

## Motivation and Context

[SIMPLY-3421](https://jira.nypl.org/browse/SIMPLY-3421)

## Checklist:

### DIFF
--- a/api/proquest/importer.py
+++ b/api/proquest/importer.py
@@ -1,10 +1,16 @@
 import datetime
 import json
 import logging
+import tempfile
 from contextlib import contextmanager
 
 import six
 import webpub_manifest_parser.opds2.ast as opds2_ast
+from flask_babel import lazy_gettext as _
+from requests import HTTPError
+from sqlalchemy import or_
+from webpub_manifest_parser.utils import encode
+
 from api.circulation import BaseCirculationAPI, FulfillmentInfo, LoanInfo
 from api.circulation_exceptions import CannotFulfill, CannotLoan
 from api.proquest.client import ProQuestAPIClientConfiguration, ProQuestAPIClientFactory
@@ -13,6 +19,7 @@ from api.proquest.identifier import ProQuestIdentifierParser
 from api.saml.metadata.model import SAMLAttributeType
 from core.classifier import Classifier
 from core.exceptions import BaseError
+from core.metadata_layer import TimestampData
 from core.model import (
     Collection,
     DeliveryMechanism,
@@ -36,10 +43,6 @@ from core.model.configuration import (
 from core.opds2_import import OPDS2Importer, OPDS2ImportMonitor, parse_feed
 from core.opds_import import OPDSImporter
 from core.util.string_helpers import is_string
-from flask_babel import lazy_gettext as _
-from requests import HTTPError
-from sqlalchemy import or_
-from webpub_manifest_parser.utils import encode
 
 MISSING_AFFILIATION_ID = BaseError(
     _(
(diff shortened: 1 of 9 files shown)


api/proquest/importer.py
```diff
@@ -789,61 +811,199 @@ def _parse_identifier(self, identifier):
...
+                        else 0
+                    )
 
-            # FIXME: We cannot short-circuit the feed import process
-            #  because ProQuest feed is not ordered by the publication's modified date.
-            #  This issue will be addressed in https://jira.nypl.org/browse/SIMPLY-3343
-            # if not self.feed_contains_new_data(feed):
-            #     break
+                processed_number_of_items += (
+                    feed.metadata.items_per_page if feed.metadata.items_per_page else 0
```
leonardr, line 954: This might indicate a problem with the OPDS2 parser rather than this code, but does this imply that every page of a paginated OPDS2 feed has the same number of items? That's not necessarily true.

leonardr: Or does `items_per_page` contain the number of items that were actually found on that particular page?

api/proquest/importer.py
```diff
@@ -789,61 +811,199 @@ def _parse_identifier(self, identifier):
...
+        :type feed: opds2_ast.OPDS2Feed
 
-            self._logger.info("Page # {0}. Finished parsing the feed".format(page))
+        :param feed_identifiers: List of identifiers in the ProQuest feed
+        :type feed_identifiers: List[str]
+        """
+        for publication in self._get_publications(feed):
+            identifier = parse_identifier(self._db, publication.metadata.identifier)
+
+            feed_identifiers.append(identifier.identifier)
```
leonardr, line 894: Don't bother changing this, but if this method returned its own list of identifiers, rather than modifying a list in place, it wouldn't need to take `feed_identifiers` as an argument.

api/proquest/scripts.py
```diff
@@ -14,6 +13,18 @@ def __init__(self, *args, **kwargs):
...
         self._logger = logging.getLogger(__name__)
 
+    @classmethod
+    def arg_parser(cls):
+        parser = OPDSImportScript.arg_parser()
+        parser.add_argument(
+            "--process-removals",
+            help="Remove from the Circulation Manager's catalog items that are no longer present in the ProQuest feed",
+            dest="process_removals",
+            action="store_true",
```
leonardr, line 23: The default value here will be false. If this is functionality you only want to activate once a week or so, then that's the right decision--you'll just have to add the script to the crontab twice.

leonardr (commented):
Overall this looks good, I have a couple minor questions and would like to hear what your plan is for invoking the import script with process_removals=True.

api/proquest/importer.py
vbessonov: Yes, it must contain the number of the items on a specific page.

vbessonov:
We agreed that we start from running the import script twice a week. I think that we can schedule to run it for the second time with `--process-removals` flag turned on.
I want to test it first and then update the cron schedule.

api/proquest/importer.py
leonardr: OK, I think the name could be improved but that's an issue for the parser library.

leonardr (approved):

PR #1548, opened 2021-01-20
Make SAML authentication provider return 401 error in the case when a bearer token is invalid

vbessonov:
## Description

This PR changes the error the SAML authentication provider raises when a bearer token sent by a client app to Circulation Manager is invalid (has an invalid format).

I'm not 100% sure that this is a right solution. If yes, then it would make sense to update the status code in `INVALID_OAUTH_BEARER_TOKEN` as well.

## Motivation and Context

[LSSE-85](https://jira.lyrasis.org/browse/LSSE-85)

## Checklist:

### DIFF
--- a/api/problem_details.py
+++ b/api/problem_details.py
@@ -251,7 +251,7 @@ UNKNOWN_SAML_PROVIDER = pd(
 
 INVALID_SAML_BEARER_TOKEN = pd(
     "http://librarysimplified.org/terms/problem/credentials-invalid",
-    status_code=400,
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
Looks good.

I think 401 is the right code, according to the spec.

leonardr:
Agreed, 401 is appropriate -- this is similar to the INVALID_CREDENTIALS problem detail.

leonardr (approved):

PR #1549, opened 2021-01-27
Fix docker builds by pinning PIP version

jonathangreen:
Looks like PIP 21.0 is out and it drops Python 2 support which breaks our docker builds: 
https://github.com/pypa/pip/issues/6148

This should pin PIP in the docker install by pinning pip to a version that works with Python 2.

This should fix the docker containers that didn't build on this merge: https://github.com/NYPL-Simplified/circulation/commit/a22e205812ecd9a66f9229c537f97dfd40e4a9c6

### DIFF
--- a/docker/simplified_app.sh
+++ b/docker/simplified_app.sh
@@ -53,7 +53,7 @@ git submodule update --init --recursive
 printf "$(git describe --tags)" > .version
 
 # Use the latest version of pip to install a virtual environment for the app.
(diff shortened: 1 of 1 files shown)


tdilauro (approved):
Don't want to assume that you're done, but  this looks like it should do the trick.

jonathangreen:
Thanks @tdilauro! Yes, should be good to go from my perspective.

leonardr (approved):

leonardr:
Thanks for taking care of this.

PR #1550, opened 2021-01-27
Bump submodule commit up to the latest.

jonathangreen:
## Description

Another attempt to fix the docker build. It is closer after fixing pip, but the docker build is still failing. The sub-module commit seems like it is a problem now. This one bumps it up to the most recent. I believe this should sort it out.

Tagging: @tdilauro @leonardr


jonathangreen:
Closing this since https://github.com/NYPL-Simplified/circulation/pull/1551 was merged.

PR #1551, opened 2021-01-28
Pin urllib3 to the last 1.24 version.

leonardr:
## Description

This branch pins urllib3 to the final 1.24 version as a way of avoiding https://jira.nypl.org/browse/SIMPLY-3477.

This is a temporary measure until either Axis 360 changes their ACS setup or we figure out a better long-term solution.

## How Has This Been Tested?

Axis 360 fulfillment starts succeeding once urllib3 is downgraded to a 1.24 version.

## Checklist:


tdilauro (approved):
This is basically a submodule commit at this point. Looks good, assuming tests pass.

PR #1554, opened 2021-02-05
Add an ability to override the default loan length for ProQuest collections

vbessonov:
## Description

This PR adds standard `Default loan length` setting to the list of available ProQuest configuration settings.

## Motivation and Context

At the moment there is no way to override the default loan length for ProQuest collection. Global default value, 21, is used all the time.

## Checklist:

### DIFF
--- a/api/proquest/importer.py
+++ b/api/proquest/importer.py
@@ -174,6 +174,9 @@ class ProQuestOPDS2Importer(OPDS2Importer, BaseCirculationAPI, HasExternalIntegr
         ProQuestOPDS2ImporterConfiguration.to_settings()
         + ProQuestAPIClientConfiguration.to_settings()
     )
+    LIBRARY_SETTINGS = BaseCirculationAPI.LIBRARY_SETTINGS + [
+        BaseCirculationAPI.DEFAULT_LOAN_DURATION_SETTING
+    ]
 
     def __init__(
         self,
@@ -645,15 +648,16 @@ class ProQuestOPDS2Importer(OPDS2Importer, BaseCirculationAPI, HasExternalIntegr
             with self._get_configuration(self._db) as configuration:
                 self._get_or_create_proquest_token(patron, configuration)
 
-                today = datetime.datetime.utcnow()
-
+                loan_period = self.collection.default_loan_period(patron.library)
+                start_time = datetime.datetime.utcnow()
+                end_time = start_time + datetime.timedelta(days=loan_period)
                 loan = LoanInfo(
                     licensepool.collection,
                     licensepool.data_source.name,
                     identifier_type=licensepool.identifier.type,
                     identifier=licensepool.identifier.identifier,
-                    start_date=today,
-                    end_date=None,
+                    start_date=start_time,
+                    end_date=end_time,
                     fulfillment_info=None,
                     external_identifier=None,
                 )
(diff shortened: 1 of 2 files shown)


tdilauro:
The new library editable field presents as expected and this change looks like it should work, but I can't get it to. I tried to do it in my dev environment with an instance of the library in which we're testing ProQuest, but get what seems to be an unrelated error for other properties in the configuration:
```
Error: The configuration value for List of SAML attributes containing an affiliation ID is invalid.
```
so I am unable to save the collection configuration.

leonardr:
This is fine as far as it goes -- I expect the problem Tim encountered is not related -- but I don't understand how this gets used. The loan is created on the CM side, and it currently doesn't have an end date, so it should be good forever, not 21 days.

vbessonov:
This feature is required for Adobe DRM protected publications. The default loan length for all of them set on ProQuest's side is 14 days, so we have to use the same value on the CM's side. The other option would be to parse ACSM files and extract this information from them but it would require changing the way the ProQuest integration currently works and moving loan creation logic to the fulfilment stage.

leonardr:
That makes sense, but the only way this will be used right now is if it's in behavior inherited from a subclass, and I don't see that happening. Right now I only see this setting used in vendor-specific APIs such as rbdigital.py.

vbessonov:
I think it's used by default when we set `LoanLength.end_date` to `None. 
I tested my changes by setting `Default loan length` to `14` days and checking out a book. 
The resulting loans feed contained the following:
```
<link href="http://cm.hilbertteam.net:6500/TEST/works/518/fulfill/4" rel="http://opds-spec.org/acquisition" type="application/vnd.librarysimplified.bearer-token+json">
<opds:indirectAcquisition type="application/epub+zip"/>
<opds:availability since="2021-02-05T16:58:37Z" status="available" until="2021-02-19T16:58:37Z"/>
</link>
```
The loan was valid only for `14` days.

leonardr:
I see... the code is in core/model/collection.py, not in circulation.

Now I understand how these dates show up in the OPDS feed -- `core/opds.py:AcquisitionFeed.license_tags` uses the default value, but I don't understand how `loans.end` gets set to this date when you actually take out a loan.

That's because when CirculationAPI.borrow actually calls LicensePool.loan_to, it passes in `end=loan_info.end_date`. And in proquest/importer.py:ProquestImporter.checkout, `LoanInfo.end_date` is set to None when creating a new loan. So I'd expect the default date to show up in the OPDS feed but not actually to be set in `loans.end`.

Are you sure that `loans.end` is set? If so, then I'm OK to merge this, but I want to make sure you're seeing the date set once the loan is actually taken out.

vbessonov:
@leonardr, you're absolutely right, `Loan.end_date` wasn't set and remained empty. I added a couple of new lines to load `BaseCirculationAPI.DEFAULT_LOAN_DURATION_SETTING` explicitly by using `Collection.default_loan_period` method.

vbessonov:
@leonardr, unfortunately, I can restart the failed pipeline but I hope you would be able to do so. If not, I can push recreate the PR or just push force some changes.

tdilauro (approved):
Tested and verified that this is working properly:

- Accepted default loan period of 21 days.
  - Bookshelf feed contains correct availability information:
    - `<opds:availability since="2021-02-16T19:13:27Z" status="available" until="2021-03-09T19:13:27Z"/>`
  - `loans` table `start`/`end` values (truncated to the second) match those in the feed

- Reset loan period to 1 day.
  - Bookshelf feed contains correct availability information:
    - `<opds:availability since="2021-02-16T19:20:21Z" status="available" until="2021-02-17T19:20:21Z"/>`
  - `loans` table `start`/`end` values (truncated to the second) match those in the feed

PR #1555, opened 2021-02-11
Bugfix/simply 2913

EdwinGuzman:
## Description

This uses `timestamp.counter` to consider whether not getting events from Bibliotheca during a time period is an error or not. Initially, not getting any events is not considered to be an error. If a no events error is raised, the next time the script runs it'll continue past the time period the no events error occurred (assuming that no events really occurred during that time period).

## Motivation and Context

Resolves [SIMPLY-2913](https://jira.nypl.org/browse/SIMPLY-2913)

## How Has This Been Tested?

Locally. Running `./bin/bibliotheca_monitor` fetches events from 2/01/2018. Eventually, this breaks when no events were returned between 5/17/18 - 5/18/18:

Screen Shot 2021-02-11 at 10 33 24 AM

The next time the script runs, we don't consider no events during the time period as an error and it continues to fetch events for later time periods:

Screen Shot 2021-02-11 at 10 33 11 AM

## Checklist:

### DIFF
--- a/api/bibliotheca.py
+++ b/api/bibliotheca.py
@@ -1350,8 +1350,7 @@ class BibliothecaEventMonitor(CollectionMonitor, TimelineMonitor):
     def catch_up_from(self, start, cutoff, progress):
         added_books = 0
         i = 0
-        init_minute_span = 5
-        timespan_to_check = timedelta(minutes=init_minute_span)
+        timespan_to_check = timedelta(minutes=5)
 
         # Not having events in a time period is not considered to be
         # an error.
@@ -1362,18 +1361,17 @@ class BibliothecaEventMonitor(CollectionMonitor, TimelineMonitor):
         # reach a 70-hour timespan to check for events.
         if (progress.counter > 0):
             no_events_error = True
-            minutes = init_minute_span * progress.counter
-            timespan_to_check = timedelta(minutes=minutes)
-            # We ran through the import and now we can consider not getting
-            # events as an error, but if the start date is more than a week
-            # ago, then we don't consider not getting events as an error.
-            one_week_ago = datetime.utcnow() - timedelta(days=7)
-            if (start < one_week_ago):
+            timespan_to_check = timedelta(hours=70)
+
+            import_time_delta = cutoff - start
+            # If we want to check for events in a time period that's
+            # longer than 70 hours, then revert back to the initial
+            # import configuration.
+            if (import_time_delta > timespan_to_check):
                 no_events_error = False
                 # Start by checking for events in a 5 minute timespan.
-                progress.counter = 1
-                minutes = init_minute_span * progress.counter
-                timespan_to_check = timedelta(minutes=minutes)
+                progress.counter = 0
+                timespan_to_check = timedelta(minutes=5)
 
         for slice_start, slice_cutoff, full_slice in self.slice_timespan(
             start, cutoff, timespan_to_check
@@ -1382,19 +1380,9 @@ class BibliothecaEventMonitor(CollectionMonitor, TimelineMonitor):
                 "Asking for events between %r and %r", slice_start,
                 slice_cutoff
             )
-            try:
-                events = self.api.get_events_between(
-                    slice_start, slice_cutoff, full_slice, no_events_error
-                )
-            except Exception, e:
-                # If we are considering no events as an error, we increase the
-                # timespan we check for events until we get to 840, which
-                # is the multiple that will get use to the 70-hour
-                # timespan check.
-                progress.counter += 1
-                if progress.counter == 840:
-                    progress.counter = 1
-                raise e
+            events = self.api.get_events_between(
+                slice_start, slice_cutoff, full_slice, no_events_error
+            )
             for event in events:
                 event_timestamp = self.handle_event(*event)
                 i += 1
--- a/api/bibliotheca.py
+++ b/api/bibliotheca.py
@@ -1359,7 +1359,7 @@ class BibliothecaEventMonitor(CollectionMonitor, TimelineMonitor):
         # an error. Each subsequent time this is run after the first time,
         # the timespan to check for event increases by 5 minutes until we
         # reach a 70-hour timespan to check for events.
-        if (progress.counter > 0):
+        if (progress.counter == 1):
             no_events_error = True
             timespan_to_check = timedelta(hours=70)
 
@@ -1390,7 +1390,13 @@ class BibliothecaEventMonitor(CollectionMonitor, TimelineMonitor):
                     self._db.commit()
             self._db.commit()
         progress.achievements = "Events handled: %d." % i
-        progress.counter = 1
+        # If we are in "catch up" mode and we encounter some events, we can
+        # reset the counter back to 0. Otherwise, keep the counter at 1. This
+        # will also set it to 1 after the initial run.
+        if progress.counter == 1:
+            progress.counter = 0
+        else:
+            progress.counter = 1
 
     def handle_event(self, bibliotheca_id, isbn, foreign_patron_id,
                      start_time, end_time, internal_event_type):
--- a/api/bibliotheca.py
+++ b/api/bibliotheca.py
@@ -326,7 +326,7 @@ class BibliothecaAPI(BaseCirculationAPI, HasSelfTests):
                 _count_activity
             )
 
-    def get_events_between(self, start, end, cache_result=False, no_events_error=False):
+    def get_events_between(self, start, end, cache_result=False, no_events_error=False, timestamp=None):
         """Return event objects for events between the given times."""
         start = start.strftime(self.ARGUMENT_TIME_FORMAT)
         end = end.strftime(self.ARGUMENT_TIME_FORMAT)
@@ -340,7 +340,10 @@ class BibliothecaAPI(BaseCirculationAPI, HasSelfTests):
             self._db.commit()
         try:
             events = EventParser().process_all(response.content, no_events_error)
+            timestamp.counter = 0
         except Exception, e:
+            if timestamp:
+                timestamp.counter = 1
             self.log.error(
                 "Error parsing Bibliotheca response content: %s", response.content,
                 exc_info=e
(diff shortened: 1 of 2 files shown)


EdwinGuzman:
@leonardr The `TestBibliothecaEventMonitor.test_run_once` test breaks because I updated the slice timespan from 1 day to 5 minutes in `catch_up_from`. If I revert back to the 5 minutes time span, does that mean the timestamps in mocked files need to be updated, or is it in the test itself where it's simulating that the script last ran 24 hours ago? I tried both but didn't get far.

leonardr:
The shape of this is right, but I think the crucial logic is backwards. Looking at the issue, I see that I didn't state the problem precisely and I kept parts of it in my head. I'm going to try to fix that now:

Libraries are starting from some period long in the past (say 2017) and trying to find all the events between the distant past and the present day. They're going to encounter many days like 05/17/2018 when there are no events. Right now, they encounter a day like 05/17/2018 and they get stuck. We need to push past those days and keep going up to the present day.  Once we reach the present day, we should start treating a span of time when there are no events as an error.

To take the example you tested with, we should blow right past 5/17/2018 and go on to 5/18. Once we get to 02/11/2021 -- once we successfully get to the current time -- _then_ we should set `timestamp.counter=1`. And after that point, an empty event list should be considered an error.

You can set `progress.counter` at the end of `catch_up_from`, where we currently set `progress.achievements`  -- that's the code that runs whenever we "catch up" with the current time. 

Also, while we're here, I think you can work around https://jira.nypl.org/browse/SIMPLY-3417 by making the timespan slice  much smaller--like, five minutes instead of one day. Bibliotheca started limiting the size of these lists to 100-200 items, so whenever we ask about a whole day we're going to miss out on a lot of events. It's still possible, even when we ask about five minutes! That's why we'll need to move to a different API once Bibliotheca releases it.

api/bibliotheca.py
```diff
@@ -1347,16 +1355,29 @@ def catch_up_from(self, start, cutoff, progress):
         added_books = 0
         i = 0
         one_day = timedelta(days=1)
+        timestamp = self.timestamp()
+
+        # If the start date is more than a month ago, then we don't consider
+        # not getting events as an error.
+        one_month_ago = datetime.utcnow() - timedelta(days=30)
```
leonardr, line 1362: This is a good idea, but you don't use this variable. It should override `no_events_error = True` below.

leonardr:
> @leonardr The `TestBibliothecaEventMonitor.test_run_once` test breaks because I updated the slice timespan from 1 day to 5 minutes in `catch_up_from`. If I revert back to the 5 minutes time span, does that mean the timestamps in mocked files need to be updated, or is it in the test itself where it's simulating that the script last ran 24 hours ago? I tried both but didn't get far.

I don't think the timestamps need to be updated, since the only thing we're checking is whether there were any events at all. I could be wrong though. Let's see what it looks like once you respond to my first round of feedback.

leonardr:
Ugh, this isn't going to work. Once we get to the current day, if we ever have a 5-minute period with no events, we'll start throwing errors and we won't recover for a month.

If we start expanding the time period when we're caught up but we don't get any results, we'll run into the problem Robert Williams mentioned in https://jira.nypl.org/browse/SIMPLY-2913: if you ask for a span of time longer than 72 hours, you'll never get any results. And the one time https://jira.nypl.org/browse/SIMPLY-2750 happened (that incident being the reason we put the original workaround into place), it lasted for much longer than 72 hours.

Can you set the time slice to five minutes if `no_events_error` is False (i.e. when we're scouring the timeline trying to catch up with the present), and one day if `no_events_error` is True (where we'll spend most of our time)?

If this doesn't work, I think we have to get rid of the idea that a period with no events is _ever_ considered an error. If https://jira.nypl.org/browse/SIMPLY-2750 happens again, it'll be bad, but there's a variety of overlapping problems with this API and we just won't be able to handle all the cases.

leonardr:
After sleeping on it, one more idea: change the date at which a start date is considered "old" from a month to a week. Then I think we'll have a good balance:

* If a library legitimately goes 70 hours without any events, you'll catch up again in a week, rather than waiting for a month.
* The API can go down for up to a week without causing any problems.

alsrlw:
@leonardr I think the approach discussed here addresses my comment on JIRA 2930 about a 72-hour limit for a single request timespan. (And, honestly, now that I reflect on the comment, it's based on my tests back last August. I'm not sure if Bibliotheca addressed any of the concerns I raised in the report I filed with them. Since they've begun an API revision, and they haven't mentioned addressing any of the concerns I noted, I'm assuming not--but we could test if we want to verify.)

I like the process of treating an initial ingest (which will encounter LOTS of no activity days) treating zero results as a good result and moving on (stepping the start datestamp). When you mention switching to a 5-minute span when "caught up," does that mean making 24 * 12 5-minute requests (and stepping the start datestamp) to get a days' events (with zero events assumed to be a good response)? While a lot, it seems like that should minimize bad effects of the undocumented 100-event limit being imposed.

I'm not sure how to balance the need to detect an error condition against the possibility of misinterpreting an actual zero-event period as an error. If a revised API wasn't in the near future, I'd suggest looking at some other way to evaluate a zero-results condition--perhaps falling back to querying on of the other endpoints to see if the results are different/the same. But given the API revision, I think it may not be worth the effort.

leonardr:
My proposed solution uses a 5-minute span while catching up, to minimize the effect of the 100-event limit. This means the initial crawl will be very slow, since it'll take 288 API requests to get through a day with no events, rather than one.

Once we're caught up, we will _start_ by asking for a 5-minute span. If there are no recent events, rather than moving on, we will ask next time about a 10-minute span, then a 15-minute span, etc., until there are either some events or we find no events in a 70-hour span.

Once we find no events in a 70-hour span, we will keep asking about that 70-hour span over and over again, assuming that the API is broken and once it's fixed we'll find out about some events.

If at some point we hear about some events from the 70-hour span, we assume the API has been fixed and start from the current time with a 5-minute span. (n.b. we probably lose most of the events from those 70 hours because of the 100-event limit).

If at the end of a week, we're still waiting to hear about that 70-hour span, then we assume it was just a slow couple of days and actually nothing happened during that time. We resume going through the timeline 5 minutes at a time.

I think we all agree that this is inefficient, but it should get Bibliotheca libraries more or less going until we can start using the new API.

leonardr:
A much simpler alternative is to switch to a 5-minute span across the board and remove the SIMPLY-2750 workaround. Assume that if the API says "no events", it really means "no events", and that the problem we saw earlier won't recur. I can't say I feel comfortable with that, though.

EdwinGuzman:
I did a minor update where the timespan now increases by 5 minutes if not getting events is considered an error. Why is switching to a 5-minute span all the time a bad idea?

alsrlw:
I understand better now. At this point (hoping the revised API drops soon), I think your solution is as good as it's going to get, balancing ease with protection. Given the nature of the different problems with the current API, there's just not an effective overall solution. This will allow hosters to get new collections ingested.

api/bibliotheca.py
```diff
@@ -1346,25 +1350,59 @@ def create_default_start_time(self, _db, cli_date):
...
+        # an error.
+        no_events_error = False
+        # But in this case, we do want to consider not getting events as
+        # an error. Each subsequent time this is run after the first time,
+        # the timespan to check for event increases by 5 minutes until we
+        # reach a 70-hour timespan to check for events.
+        if (progress.counter > 0):
+            no_events_error = True
+            minutes = init_minute_span * progress.counter
+            timespan_to_check = timedelta(minutes=minutes)
```
leonardr, line 1366: The timespan to check is the entire time between `start` and `cutoff` -- unless that time is more than 70 hours. If it's more than 70 ours, timespan_to_check should be bumped back down to 5 minutes, no_events_error should be set to False, and progress.counter should be set to zero.

Basically, timespan_to_check grows until it reaches 70 hours, at which point we go back into "initial import" mode.

I was imprecise when I said that the time slice would get five minutes bigger every time--that's because the script is run every five minutes. It shouldn't be necessary to explicitly do that.

Does this make sense?

leonardr:
Grouping these together into two different modes might make it easier to write the code:

**Initial import mode**

`catch_up_from` calls `slice_timespan(start, cutoff)` to divide up the time between `start` and the present time into 5-minute increments. If there is no event in a given 5-minute period, it's not a problem--just keep moving and process the next slice.

Once the `slice_timespan` iterator is exhausted, switch to real-time mode by setting `progress.state` to one. On the next run, we'll start in real-time mode.

*Real-time mode* 

First off, `catch_up_from` should calculate the delta between `start` and `cutoff`. If it's more than 70 hours, we must exit real-time mode. Set progress.state to zero to go back to initial import mode, and call the method again so the real-time code runs.

If the delta is less than 70 hours, we make a single call to `get_events_between` covering the entire timespan. If there are no events, it's treated as an error and we will not update `progress.start`.

api/bibliotheca.py
EdwinGuzman: Yes, it does.

EdwinGuzman:
@leonardr I updated the `catch_up_from` function and the related test.

tests/test_bibliotheca.py
```diff
@@ -1053,31 +1055,27 @@ def test_run_once(self):
...
-
         # The timestamp's achivements have been updated.
-        eq_("Events handled: 2.", after_timestamp.achievements)
+        eq_("Events handled: 13.", after_timestamp.achievements)
 
-        # If we tell run_once() to work through an amount of time
-        # where the are no events, it errors out, because we can't distinguish
-        # between a slow day and a problem with the API.
+        # After the initial run, the progress timestamp's `counter` property
+        # is set to `1`. This means we are now in "catch up" mode where we
```
leonardr, line 1064: Add an assertion that the `counter` is set to 1.

tests/test_bibliotheca.py
```diff
@@ -1053,31 +1055,27 @@ def test_run_once(self):
...
-        self.time_eq(now, final_timestamp.finish)
+        
+        assert_raises_regexp(
+            RemoteInitiatedServerError,
+            "No events returned from server. This may not be an error, but treating it as one to be safe.",
+            monitor.run_once, after_timestamp
+        )
+
+        # One request was made but no events were found.
+        eq_(16, len(api.requests))
```
leonardr, line 1078: Verify that once events _are_ found, `counter` is set to zero.

leonardr (changes requested):
This is looking really good, it just needs some additional tests around when we change between modes.

EdwinGuzman:
I added a bit more to the function because I initially did not revert back the counter to 0.

leonardr (approved):

PR #1556, opened 2021-02-11
Adjust permissions

adriana-alter:
Resolves https://jira.nypl.org/browse/SIMPLY-3486 (it's the project for extending non-system-admins' access to various system configuration stuff).  Goes with [this core branch](https://github.com/NYPL-Simplified/server_core/pull/1239) and [this front-end branch](https://github.com/NYPL-Simplified/circulation-web/pull/334).

Two changes:

1) I gave each setting in `LIBRARY_SETTINGS` a `level` key; this is how the front end is going to know whether the admin has high enough permissions to edit the setting.  (If the `level` key gets left out, that won't break anything; the front end will just assume that the setting can be edited by all admins.)

2) Everyone, not just system admins, is supposed to be able to access the analytics settings now; I changed the controller code (and tests) to reflect that.  Librarians are allowed to create/edit Google analytics settings, but not local analytics ones.  The front-end code should make it impossible for them to mess with local analytics stuff, but I've put in a fail-safe check just in case, and made it so that all existing and new local analytics services will be set to level 2 (i.e. higher than librarians can access).

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -211,19 +211,25 @@ class Configuration(CoreConfiguration):
         }
     ]
 
+    # The "level" property determines which admins will be able to modify the setting.  Level 1 settings can be modified by anyone.
+    # Level 2 settings can be modified only by library managers and system admins (i.e. not by librarians).  Level 3 settings can be changed only by system admins.
+    # If no level is specified, the setting will be treated as Level 1 by default.
+
     LIBRARY_SETTINGS = CoreConfiguration.LIBRARY_SETTINGS + [
         {
             "key": LIBRARY_DESCRIPTION,
             "label": _("A short description of this library"),
             "description": _("This will be shown to people who aren't sure they've chosen the right library."),
             "category": "Basic Information",
+            "level": 1
         },
         {
             "key": Announcements.SETTING_NAME,
             "label": _("Scheduled announcements"),
             "description": _("Announcements will be displayed to authenticated patrons."),
             "category": "Announcements",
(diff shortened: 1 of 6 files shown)


leonardr (approved):
Both this and the core branch make sense to me. I'm a little concerned about using numbers for the permission levels. At the least, I'd like to see us use constants for the numbers; even as I write this I've forgotten whether "1" is the lowest level of permission or the highest. But that's OK to put in later.

adriana-alter:
It turns out that, after a few days away, I have also forgotten which level is which.  I can't switch away from using numbers (that would entail completely reworking all of the front end logic), but I'm putting in constants.

api/config.py
```diff
@@ -211,53 +211,67 @@ class Configuration(CoreConfiguration):
         }
     ]
 
+    # The "level" property determines which admins will be able to modify the setting.  Level 1 settings can be modified by anyone.
+    # Level 2 settings can be modified only by library managers and system admins (i.e. not by librarians).  Level 3 settings can be changed only by system admins.
+    # If no level is specified, the setting will be treated as Level 1 by default.
+    ALL_ACCESS = 1
+    SYS_ADMIN_OR_MANAGER = 2
+    SYS_ADMIN_ONLY = 3
+
```
leonardr, line 220: You should be able to import these from core/config.py rather than defining them again -- that way there's no risk they'll get out of sync.

leonardr (approved):

PR #1557, opened 2021-02-12
Fix rendering of menu controls in the admin UI

vbessonov:
## Description

This PR fixes the rendering of menu controls in the admin UI. Menu controls with multiple selected values are currently rendered like that:
image

@tdilauro, this PR should allow you to test [PR # 1554](https://github.com/NYPL-Simplified/circulation/pull/1554).

## Checklist:

### DIFF
--- a/api/admin/controller/collection_settings.py
+++ b/api/admin/controller/collection_settings.py
@@ -109,7 +109,7 @@ class CollectionSettingsController(SettingsController):
                         value = str(storage_integration.other_integration_id)
                     else:
                         value = self.NO_MIRROR_INTEGRATION
-                elif protocol_setting.get("type") == "list":
+                elif protocol_setting.get("type") in ("list", "menu"):
                     value = collection_object.external_integration.setting(key).json_value
                 else:
                     value = collection_object.external_integration.setting(key).value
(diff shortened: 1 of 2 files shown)


leonardr:
@adriana-alter Does this make sense to you?

adriana-alter:
It definitely makes sense, but is the list of SAML attributes the only use case?  If so, I'd rather see it called something other than `menu`, just to be sure that nothing is conflicting with the existing `menu` functionality.

leonardr:
@vbessonov What's the difference between the list of SAML attributes and some other `menu` type configuration setting? Or is the new behavior an improvement for all `menu` type settings?

tdilauro (approved):
This change does resolve the issues that I was experiencing.

I'm approving now, pending feedback from @leonardr and @adriana-alter.

vbessonov:
@leonardr, yes, my changes should suit any `menu` controls. I used SAML configuration settings as an example but I think I can make a more generic example.

leonardr:
I'm OK with merging on the basis that this will improve things for all `menu` type attributes -- do you want to complete the more generic example in this branch or leave it til later?

vbessonov:
Since it's blocking [PR # 1554](https://github.com/NYPL-Simplified/circulation/pull/1554) which is pretty critical I would prefer to merge it as is and later I'll add a more generic example.

leonardr:
OK, sounds good.

PR #1560, opened 2021-02-19
Let non-system-admins submit the library form

adriana-alter:
After all that, I forgot to make it so that non-system-admins can actually submit the create/edit library form without getting an error message...

### DIFF
--- a/api/admin/controller/library_settings.py
+++ b/api/admin/controller/library_settings.py
@@ -73,8 +73,6 @@ class LibrarySettingsController(SettingsController):
         return dict(libraries=libraries, settings=Configuration.LIBRARY_SETTINGS)
 
     def process_post(self, validators_by_type=None):
-        self.require_system_admin()
-
         library = None
         is_new = False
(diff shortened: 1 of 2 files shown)


api/admin/controller/library_settings.py
```diff
@@ -73,8 +73,6 @@ def process_get(self):
         return dict(libraries=libraries, settings=Configuration.LIBRARY_SETTINGS)
 
     def process_post(self, validators_by_type=None):
-        self.require_system_admin()
-
```
leonardr, line 77: Should you call require_library_manager instead, or is this taken care of by the @requires_admin that guards the route?

leonardr (approved):
Looks good assuming @requires_admin is all the permission control necessary here.

api/admin/controller/library_settings.py
adriana-alter: No, librarians can edit libraries too now, not just system admins and library managers.  @requires_admin should be plenty.

PR #1563, opened 2021-02-23
Define and use a setting that controls how long authentication documents are cached.

leonardr:
## Description

This branch gives us the power to fine-tune or disable the caching of authentication documents. This was introduced for performance reasons but it makes certain deployment problems difficult to test, so it should be possible to disable it.

## Motivation and Context

https://jira.nypl.org/browse/SIMPLY-3548

## How Has This Been Tested?

In addition to the automated tests, I tested this with a local database by setting the timeout to 10 seconds, or 0 seconds, and adding a debug statement to the code to signal whenever a fresh auth document was created.

## Checklist:

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -302,8 +302,13 @@ class CirculationManager(object):
 
         self.patron_web_domains = patron_web_domains
         self.setup_configuration_dependent_controllers()
+        authentication_document_cache_time = int(
+            ConfigurationSetting.sitewide(
+                self._db, Configuration.AUTHENTICATION_DOCUMENT_CACHE_TIME
+            ).value_or_default(3600)
+        )
         self.authentication_for_opds_documents = ExpiringDict(
-            max_len=1000, max_age_seconds=3600
+            max_len=1000, max_age_seconds=authentication_document_cache_time
         )
 
     @property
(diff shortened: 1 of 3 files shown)


api/controller.py
```diff
@@ -538,9 +543,15 @@ def authentication_for_opds_document(self):
         OPDS document in the cache, then return the cached version.
         """
         name = flask.request.library.short_name
-        if name not in self.authentication_for_opds_documents:
-            self.authentication_for_opds_documents[name] = self.auth.create_authentication_document()
-        return self.authentication_for_opds_documents[name]
+        value = self.authentication_for_opds_documents.get(name, None)
+        if value is None:
```
EdwinGuzman, line 547: This is similar to how it was before except that now it's checking if it's `None`. Why the change to this syntax?

EdwinGuzman (commented):
This looks good but the PR branch Travis test is failing.

api/controller.py
leonardr: The old code won't work if the caching time is set to zero. I create the authentication document and stick it into `self.authentication_for_opds_documents`, where it is _immediately deleted_. So `self.authentication_for_opds_documents[name]` fails even though I just put the document into there.

Instead I make sure the document (whether retrieved from cache or created fresh) is stored in `value`. I stick `value` in the cache for next time, but it doesn't matter if it immediately gets deleted; I already have it.

leonardr:
Yeah, I don't know what to do about that. I can't use Travis to restart the build; it fails with an unspecified error.

leonardr:
Made a cosmetic change to force the tests to run again and this time it worked.

PR #1564, opened 2021-02-24
Fix the ProQuest ulimits issue

vbessonov:
## Description

This PR limits a number of open temporary files used during the ProQuest import to address the ulimits issue.

## Motivation and Context

The ProQuest import script (`proquest_import_monitor`) fails when running through cron with `Too many open files` error.

## Checklist:

### DIFF
--- a/api/proquest/importer.py
+++ b/api/proquest/importer.py
@@ -1,6 +1,8 @@
+import codecs
 import datetime
 import json
 import logging
+import shutil
 import tempfile
 from contextlib import contextmanager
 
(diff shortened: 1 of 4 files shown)


leonardr (commented):
This general approach looks good. My guess is, it works because the temporary files are closed after being written, and reopened as necessary, reducing the number of filehandles open at once.

api/proquest/importer.py
```diff
@@ -835,48 +837,75 @@ def _get_publications(feed):
...
+                feed_page_file.write(feed_page_content)
+                feed_page_file.flush()
+
+                feed_page_files.append(feed_page_file.name)
+            except:
+                self._logger.exception(
+                    "An unexpected error occurred during saving a feed page"
+                )
+
+                feed_page_file.delete()
```
tdilauro, line 881: If we're on the second or subsequent feed page, will all previous pages get deleted?

Might be easier to let `ProQuestOPDS2ImportMonitor._parse_feed_page` and handle clean up of the entire `feed_pages_directory`.

api/proquest/importer.py
```diff
@@ -930,16 +959,18 @@ def _get_feeds(self):
         """
         self._logger.info("Started fetching ProQuest paged OPDS 2.0 feeds")
 
-        feed_temporary_files = self._download_feed_pages()
+        feed_pages_directory = tempfile.mkdtemp()
+        feed_page_files = self._download_feed_pages(feed_pages_directory)
```
tdilauro, line 963: Moving the `._download_feed_pages` call into the `try` block below catch download errors and allow the `finally` block to clean up the `feed_pages_directory`.

vbessonov: Totally makes sense, thank you for noticing this.

api/proquest/importer.py
leonardr: I agree; the code that makes the temporary directory should be in charge of clearing out anything in that directory before deleting it.

vbessonov: This part is actually not required, since there is `raise` statement the stack will be unwind and got to the `finally` block in `_get_feeds` method where `shutil.rmtree` will take care of recursively removing the temporary directory with all its content.

tdilauro: Yes.

leonardr: I'm OK merging this either with `feed_page_file.delete` or without it, though I'd prefer it be removed. LMK what you'd prefer.

tdilauro: I would prefer that `feed_page_file.delete` be removed, since it might confuse someone later. That someone might be me. 🙃

vbessonov: I removed `feed_page_file.delete` and extended the log message a bit but anyways it should contain all the required information about the exception because we use `exception` method.

leonardr (approved):

PR #1566, opened 2021-02-25
Make basic info settings sys-admin-only

adriana-alter:
Along with https://github.com/NYPL-Simplified/server_core/pull/1241, resolves https://jira.nypl.org/browse/SIMPLY-3556 and https://jira.nypl.org/browse/SIMPLY-3557; it turns out that librarians and library managers weren't supposed to have access to the fields in the "Basic Information" section after all.  (The "Name" and "Short Name" fields from the "Basic Information" section are taken care of on the front end.)

### DIFF
--- a/api/config.py
+++ b/api/config.py
@@ -232,7 +232,7 @@ class Configuration(CoreConfiguration):
             "label": _("A short description of this library"),
             "description": _("This will be shown to people who aren't sure they've chosen the right library."),
             "category": "Basic Information",
-            "level": CoreConfiguration.ALL_ACCESS
+            "level": CoreConfiguration.SYS_ADMIN_ONLY
         },
         {
             "key": Announcements.SETTING_NAME,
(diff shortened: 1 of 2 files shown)


api/config.py
```diff
@@ -232,7 +232,7 @@ class Configuration(CoreConfiguration):
             "label": _("A short description of this library"),
             "description": _("This will be shown to people who aren't sure they've chosen the right library."),
             "category": "Basic Information",
-            "level": CoreConfiguration.ALL_ACCESS
+            "level": CoreConfiguration.SYS_ADMIN_ONLY
```
tdilauro, line 235: @adriana-alter Since you're now using the `CoreConfiguration` level constants, it's probably useful to get rid of the comments above (ll. 225-227), since the constants are pretty self-explanatory and the numbers there might get out of sync with the constants.

adriana-alter: Hi Tim!  The comments are still relevant (to me, at least!), because the front end uses the numbers.  (If the constants got out of sync with the numbers in the comments, that would probably mean they were also out of sync with the front end, i.e. that the functionality was broken from the user's perspective; I can't see it happening accidentally.)

tdilauro: I see. Thanks for clarifying!

leonardr (approved):

PR #1567, opened 2021-03-01
Pick the right licensepool for our collection.

tdilauro:
## Description

Fixes the case in which there is already a license pool for the same edition, which would lead to `DirectoryImportScript.work_from_metadata` crashing after we add another with our collection and try to pick just one.

## Motivation and Context

Details on https://jira.nypl.org/browse/SIMPLY-3567.

## How Has This Been Tested?

Added tests which failed before change and pass after change.

## Checklist:

- N/A I have updated the documentation accordingly.

### DIFF
--- a/scripts.py
+++ b/scripts.py
@@ -1493,9 +1493,8 @@ class DirectoryImportScript(TimestampScript):
 
         edition, new = metadata.edition(self._db)
         metadata.apply(edition, collection, replace=policy)
-        data_source = metadata.data_source(self._db)
         [pool] = [x for x in edition.license_pools
-                  if x.data_source == data_source]
+                  if x.collection == collection]
         if new:
             self.log.info("Created new edition for %s", edition.title)
         else:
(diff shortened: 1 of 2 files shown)


scripts.py
```diff
@@ -1493,9 +1493,8 @@ def work_from_metadata(self, collection, collection_type, metadata, policy, *arg
 
         edition, new = metadata.edition(self._db)
         metadata.apply(edition, collection, replace=policy)
-        data_source = metadata.data_source(self._db)
         [pool] = [x for x in edition.license_pools
-                  if x.data_source == data_source]
+                  if x.collection == collection]
```
tdilauro, line 1497: `Edition.license_pools` already filters license pools with the edition's data source (and primary identifier), but not by collection. So, we need to ensure that we have the right one for our collection.

vbessonov (approved):

tdilauro:
@leonardr This is ready to merge, if you approve.

leonardr (approved):

PR #1568, opened 2021-03-02
Add saml_monitor to the crontab

vbessonov:
## Description

This PR adds [saml_monitor](https://github.com/NYPL-Simplified/circulation/blob/develop/bin/saml_monitor), the script refreshing federated SAML metadata (at the moment, only the InCommon Federation is supported), to the crontab.

This script must update metadata of federated IdPs available in Circulation Manager at leasts once a day.

## Checklist:

### DIFF
--- a/docker/services/simplified_crontab
+++ b/docker/services/simplified_crontab
@@ -121,3 +121,7 @@ HOME=/var/www/circulation
 # ProQuest
 # Run proquest_import_monitor once a week on Sunday at 00:00
 0 0 * * 0 root core/bin/run proquest_import_monitor >> /var/log/cron.log 2>&1
+
+# SAML
+#
+0 5 * * * root core/bin/run saml_monitor >> /var/log/cron.log 2>&1


docker/services/simplified_crontab
```diff
@@ -121,3 +121,7 @@ HOME=/var/www/circulation
 # ProQuest
 # Run proquest_import_monitor once a week on Sunday at 00:00
 0 0 * * 0 root core/bin/run proquest_import_monitor >> /var/log/cron.log 2>&1
+
+# SAML
+#
+0 5 * * * root core/bin/run saml_monitor >> /var/log/cron.log 2>&1
```
tdilauro, line 127: No newline at the end of the file, but otherwise looks good.

tdilauro (approved):

leonardr:
This looks fine on its own, but will this do anything if SAML is not configured for a given circulation manager? While reviewing the SAML code I seem to recall that the Incommon federation is set up in every circulation manager, even if it doesn't use SAML as an authentication mechanism. In a case like that I'd rather not spend the effort to keep the Incommon metadata updated.

leonardr (approved):
Since this branch is just adding an existing script to the cron, I'm fine with merging it even if it does unnecessary work on a CM that doesn't have SAML authentication enabled. Let me know if that's the case and I'll file a ticket.

docker/services/simplified_crontab
vbessonov: Fixed it.

vbessonov:
@leonardr, yes, [this migration](https://github.com/NYPL-Simplified/circulation/blob/develop/migration/20201112-add-incommon-saml-federation-metadata.py) adds the InCommon metadata by default, so the script will always be running, even if there is no SAML auth provide set up.

leonardr:
OK, I filed https://jira.nypl.org/browse/SIMPLY-3577 to deal with that problem.

PR #1572, opened 2021-03-08
SAML: add an ability to override the patron ID extraction logic

vbessonov:
## Description

This PR adds an ability to override the patron ID extraction logic by adding 3 additional configuration settings:
- `Patron ID: Use SAML NameID`:
image
- `Patron ID: SAML Attributes`:
image
- `Patron ID: Regular expression`:
image

## Motivation and Context

At the moment the SAML authentication provider uses a [hard-coded algorithm](https://github.com/NYPL-Simplified/circulation/blob/develop/api/saml/metadata/model.py#L1241) for extracting patron IDs from SAML assertions. It iterates through all the SAML attributes and takes the first available attribute from the following list: 

- eduPersonUniqueId
- eduPersonTargetedID
- uid
- Name ID

This algorithm does work in general but not for Columbia and NYU. They store their patron IDs in **eduPersonPrincipalName** attribute using the following format:
```
<Patron ID>@domain
```
It means we have to extract a patron ID from **eduPersonPrincipalName** attribute.

## Checklist:


tdilauro (commented):
There are a couple of issues related to coding conventions and DRY that it would be helpful to update and one issue where the example will not render as you expect.

An issue that concerns me more is the use of the named group. If you know that you will always be able to construct a relatively simple regex that needs only one, this won't be a problem; however, if you need to do something more complex that required an additional group, it would be a problem. I'm not sure how likely it is that these would appear in the wild, though.

api/saml/configuration/model.py
```diff
@@ -64,6 +64,54 @@ class SAMLConfiguration(ConfigurationGrouping):
         format="narrow",
     )
 
+    patron_id_use_name_id = ConfigurationMetadata(
```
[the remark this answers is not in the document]
vbessonov: Would you like to have all the attributes of `SAMLConfiguration` to be capitalized? If yes, then probably, I would also need to do the same for ProQuest configuration classes: [ProQuestAPIClientConfiguration](https://github.com/NYPL-Simplified/circulation/blob/develop/api/proquest/client.py#L21), and [ProQuestOPDS2ImporterConfiguration](https://github.com/NYPL-Simplified/circulation/blob/develop/api/proquest/importer.py#L85).

tests/saml/metadata/test_model.py
```diff
@@ -116,14 +117,190 @@ class TestSubjectUIDExtractor(object):
...
+                        ]
+                    ),
+                ),
+                "patron",
+                False,
+                [
+                    SAMLAttributeType.eduPersonPrincipalName,
+                    SAMLAttributeType.mail,
+                ],
+                r"(?P<patron_id>.+)@university\.org",
```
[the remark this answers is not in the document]
vbessonov: I added two constants:
- `PATRON_ID_REGULAR_EXPRESSION_ORG`
- `PATRON_ID_REGULAR_EXPRESSION_COM`

api/saml/configuration/model.py
```diff
@@ -64,6 +64,54 @@ class SAMLConfiguration(ConfigurationGrouping):
...
+            SAMLAttributeType.uid.name,
+        ],
+        format="narrow",
+    )
+
+    patron_id_regular_expression = ConfigurationMetadata(
+        key="saml_patron_id_regular_expression",
+        label=_("Patron ID: Regular expression"),
+        description=_(
+            "Regular expression used to extract a unique patron ID from SAML attributes. "
```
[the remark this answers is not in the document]
vbessonov: I added clarification.

api/saml/configuration/model.py
```diff
@@ -64,6 +64,54 @@ class SAMLConfiguration(ConfigurationGrouping):
         format="narrow",
     )
 
+    patron_id_use_name_id = ConfigurationMetadata(
+        key="saml_patron_id_use_name_id",
+        label=_("Patron ID: Use SAML NameID"),
+        description=_(
+            "Boolean value indicating whether SAML NameID should be searched for a unique patron ID: "
+            "<b>0</b> means that NameID won't be used, "
+            "<b>1</b> means that Circulation Manager will scan NameID in search for a patron ID."
```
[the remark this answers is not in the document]
vbessonov: I specified that `1` is the default value.

api/saml/configuration/validator.py
```diff
@@ -266,12 +275,67 @@ def _process_filtration_expression(
...
                 return SAML_INCORRECT_FILTRATION_EXPRESSION.detailed(
                     _(
-                        "SAML filtration expression has incorrect format: {0}".format(
+                        u"SAML filtration expression has an incorrect format: {0}".format(
                             six.ensure_text(str(exception))
                         )
                     )
                 )
 
+    def _process_patron_id_regular_expression(
```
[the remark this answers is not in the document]
vbessonov: Changed the name of the method accordingly.

api/saml/configuration/model.py
```diff
@@ -64,6 +64,54 @@ class SAMLConfiguration(ConfigurationGrouping):
...
+        ),
+        type=ConfigurationAttributeType.NUMBER,
+        required=False,
+        default=1,
+    )
+
+    patron_id_attributes = ConfigurationMetadata(
+        key="saml_patron_id_attributes",
+        label=_("Patron ID: SAML Attributes"),
+        description=_("List of SAML attributes that MAY contain a unique patron ID."),
```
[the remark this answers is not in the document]
vbessonov: I specified that the attributes will be searched through in the order they were chosen by the admin.

api/saml/configuration/model.py
```diff
@@ -64,6 +64,54 @@ class SAMLConfiguration(ConfigurationGrouping):
...
+    patron_id_regular_expression = ConfigurationMetadata(
+        key="saml_patron_id_regular_expression",
+        label=_("Patron ID: Regular expression"),
+        description=_(
+            "Regular expression used to extract a unique patron ID from SAML attributes. "
+            "The expression MUST contain a named group <b>patron_id</b> used to match the patron ID. "
+            "For example:"
+            "<br>"
+            "<pre>"
+            r"(?P<patron_id>.+)@university\.org"
```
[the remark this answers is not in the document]
vbessonov: Thank you for noticing this! I updated the code accordingly.

api/saml/metadata/model.py
```diff
@@ -1235,74 +1238,146 @@ def decode(self, raw_subject, _w=WHITESPACE.match):
...
+    PATRON_ID_REGULAR_EXPRESSION_NAMED_GROUP = "patron_id"
 
-        3. uid
-           http://oid-info.com/get/0.9.2342.19200300.100.1.1
+    def __init__(self, use_name_id=True, attributes=None, regular_expression=None):
+        """Initialize a new instance of SAMLSubjectPatronIDExtractor class.
 
-           See IETF RFC 4519.
-           IETF RFC 1274 uses the identifier "userid".
+        :param use_name_id: Boolean value indicating whether NameID is also should be searched for a unique patron ID
```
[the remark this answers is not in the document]
vbessonov: I wiped `is also` out.

api/saml/metadata/model.py
```diff
@@ -1235,74 +1238,146 @@ def decode(self, raw_subject, _w=WHITESPACE.match):
...
+       with a simpler syntax and safer comparison rules.
+       Existing use of this attribute in SAML 1.1 or SAML 2.0 should be phased out
+       in favor of the new Subject Identifier attributes."
 
-           A persistent, non-reassigned, opaque identifier for a principal.
-           eduPersonTargetedID is an abstracted version of the SAML V2.0 Name Identifier format of
-           "urn:oasis:names:tc:SAML:2.0:nameid-format:persistent"
-           (see http://www.oasis-open.org/committees/download.php/35711).
+    3. uid
+       http://oid-info.com/get/0.9.2342.19200300.100.1.1
```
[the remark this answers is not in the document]
vbessonov: Sure, totally makes sense, I updated the docstring.

vbessonov:
@tdilauro, I believe it should be enough to use the single `patron_id` because the examples we have so far are relatively simple and depicted in the example in `Patron ID: Regular expression` configuration setting's description:
- `(?P<patron_id>.+)@university\.org`

api/saml/configuration/model.py
```diff
@@ -64,6 +65,64 @@ class SAMLConfiguration(ConfigurationGrouping):
...
+
+    patron_id_attributes = ConfigurationMetadata(
+        key="saml_patron_id_attributes",
+        label=_("Patron ID: SAML Attributes"),
+        description=_(
+            "List of SAML attributes that MAY contain a unique patron ID. "
+            "The attributes will be scanned sequentially in the order you chose them, "
+            "and the first existing attribute will be used to extract a unique patron ID."
+            "<br>"
+            "NOTE: If a SAML attribute contains several values, only the first be used."
```
tdilauro, line 91: "...only the first [will] be used."

api/saml/configuration/model.py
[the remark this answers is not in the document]
tdilauro: @jonathangreen @leonardr Do you have thoughts on this?

api/saml/configuration/model.py
```diff
@@ -64,6 +65,64 @@ class SAMLConfiguration(ConfigurationGrouping):
...
 
+    patron_id_use_name_id = ConfigurationMetadata(
+        key="saml_patron_id_use_name_id",
+        label=_("Patron ID: Use SAML NameID"),
+        description=_(
+            "Boolean value indicating whether SAML NameID should be searched for a unique patron ID: "
+            "<br>"
+            "- <b>0</b> means that NameID won't be used, "
+            "<br>"
+            "- <b>1</b> (default) means that Circulation Manager will scan NameID in search for a patron ID."
```
tdilauro, line 76: Suggest appending the following to value 1 for clarity: "If found, it will supersede any SAML Attributes selected in the next section."

api/saml/configuration/model.py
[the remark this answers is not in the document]
tdilauro: The screenshot image in the PR text body above doesn't reflect the change, but not a big deal to me.

api/saml/configuration/model.py
```diff
@@ -64,6 +65,64 @@ class SAMLConfiguration(ConfigurationGrouping):
...
+        ],
+        format="narrow",
+    )
+
+    patron_id_regular_expression = ConfigurationMetadata(
+        key="saml_patron_id_regular_expression",
+        label=_("Patron ID: Regular expression"),
+        description=_(
+            "Regular expression used to extract a unique patron ID from the attributes "
+            "specified in <b>Patron ID: SAML Attributes</b>. "
```
tdilauro, line 113: The regular expression would also be applied to NameID, if found (in addition to the Patron ID: SAML Attributes), right?

api/saml/metadata/model.py
```diff
@@ -1235,74 +1238,154 @@ def decode(self, raw_subject, _w=WHITESPACE.match):
...
-                        unique_id_attribute.name
+            for patron_id_attribute in self._patron_id_attributes:
+                if patron_id_attribute in subject.attribute_statement.attributes:
+                    patron_id_attribute = subject.attribute_statement.attributes[
+                        patron_id_attribute
                     ]
 
-                    return unique_id_attribute.values[0]
+                    # NOTE: It takes the first value.
+                    patron_id_candidate = patron_id_attribute.values[0]
```
tdilauro, line 1373: Just wondering here: Is there a reason why we only look at the first one? Or is another case in which it is unlikely to get more than one?

vbessonov: I think it's quite unlikely to get more than one. Also, we simply don't know how to process more than one values. For example, if we don't have a regular expression set up, what should we do with multiple values?

tdilauro: Funny. The case of [0] not matching the regex, but [1] matching, is exactly what got me thinking about this. :-)

api/saml/configuration/model.py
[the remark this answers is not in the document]
vbessonov: I updated the screenshot as well.

api/saml/configuration/model.py
vbessonov: Added the suggested line.

vbessonov: However, it might be confusing. NameID used only when there are _NO_ SAML attributes satisfying the conditions specified by the admin because it has the lowest priority.

api/saml/configuration/model.py
vbessonov: Yes, that's true, I updated the description.

api/saml/configuration/model.py
```diff
@@ -64,6 +65,67 @@ class SAMLConfiguration(ConfigurationGrouping):
         format="narrow",
     )
 
+    patron_id_use_name_id = ConfigurationMetadata(
+        key="saml_patron_id_use_name_id",
+        label=_("Patron ID: Use SAML NameID"),
+        description=_(
+            "Boolean value indicating whether SAML NameID should be searched for a unique patron ID: "
+            "<br>"
+            "- <b>0</b> means that NameID won't be used, "
```
vbessonov, line 74: @leonardr, @tdilauro, @jonathangreen, how do you think we need to represent boolean configuration settings in the admin UI and in the code?
- Is it enough to use an integer text input and treat them as binary zeros and ones? 
- Or maybe it would be better to represent them using a dropdown box containing "Yes", "No" choices?

P.S. Radio buttons would work nicely here, but, unfortunately, we don't have them in the admin UI yet.

api/saml/metadata/model.py
```diff
@@ -1235,74 +1238,154 @@ def decode(self, raw_subject, _w=WHITESPACE.match):
...
+                    patron_id = self._extract_patron_id(patron_id_candidate)
+
+                    if patron_id is not None:
+                        break
+
+        if self._use_name_id:
```
tdilauro, line 1379: Here it looks like the presence of NameID could override a previously set patron_id.

api/saml/provider.py
```diff
@@ -74,6 +74,21 @@ def __init__(self, library, integration, analytics=None):
         self._configuration_factory = SAMLConfigurationFactory(SAMLMetadataParser())
         self._authentication_manager_factory = SAMLAuthenticationManagerFactory()
 
+        db = Session.object_session(library)
+
+        with self.get_configuration(db) as configuration:
+            self._patron_id_use_name_id = (
+                str(configuration.patron_id_use_name_id) == "1"
```
vbessonov, line 81: This is super ugly and can be fixed in two different ways:
1. Adding a new method to `ConfigurationMetadata` class:
```python
class ConfigurationMetadata(object):
    @property
    def bool_value(self):
         return str(self) == "1"
```
And replacing the code above with the following:
```python
self._patron_id_use_name_id = configuration.patron_id_use_name_id.bool_value
```

2. Changing the type of `patron_id_use_name_id` from the integer text input to a dropdown, adding the following method to `ConfigurationMetadata` class:
```python
class ConfigurationMetadata(object):
    @property
    def bool_value(self):
         return str(self).lower() in ConfigurationSetting.MEANS_YES
```
And replacing the code above with the following:
```python
self._patron_id_use_name_id = configuration.patron_id_use_name_id.bool_value
```

tests/saml/test_provider.py
```diff
@@ -350,10 +360,126 @@ def test_authentication_document(self, _, identity_providers, expected_result):
                     complete=True,
                 ),
             ),
+            (
+                "subject_has_unique_name_id_but_use_of_name_id_is_switched_off_using_integer_literal",
```
vbessonov, line 364: Again, representing boolean configuration settings as binary zeros and ones makes the code more complex since we don't whether we're going to end up with an integer zero or one or their string version.

api/saml/configuration/model.py
vbessonov: I updated the description.

api/saml/metadata/model.py
vbessonov: You're absolutely right, fixed the issue and added additional tests.

api/saml/configuration/model.py
tdilauro: I like Yes/No, since they make more sense to humans. That would prevent nonsensical values and you could just set the current/default value, allowing us to omit text explaining the default.

leonardr: Take a look at core/config.py to see how we currently handle this.

```
        {
            "key": ALLOW_HOLDS,
            "label": _("Allow books to be put on hold"),
            "type": "select",
            "options": [
                { "key": "true", "label": _("Allow holds") },
                { "key": "false", "label": _("Disable holds") },
            ],
            "default": "true",
            "category": "Loans, Holds, & Fines",
            "level": SYS_ADMIN_ONLY
        },
```

The user interface looks good and it doesn't rely on people remembering constants. The dropdown format (rather than a checkbox) lets you explain each option individually. You can use `ConfigurationSetting.bool_value` to coerce the string value stored in the database (whatever it is) into a boolean.

api/saml/configuration/model.py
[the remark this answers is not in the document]
leonardr: I don't have a strong opinion. Those are instances of application classes, so not your typical constants. But they are being used as groupings of constants.

Since we don't currently use `ConfigurationMetadata` everywhere I think it makes sense to use lowercase for these configuration objects at least so long as we're using all caps for the names of configuration _settings_. That way it's clear that `patron_id_use_name_id` is a configuration object and `ALLOW_HOLDS` is just a string.

api/saml/provider.py
leonardr: I prefer the second solution due to its similarity to ConfigurationSetting.bool_value.

leonardr:
Needs some work around how the boolean configuration is stored, presented, and retrieved; everything else LGTM.

api/saml/configuration/model.py
vbessonov: I updated the PR and turned `Patron ID: SAML NameID` setting into a boolean configuration setting, dropdown allowing admins to select one of the two values:
- `Use SAML NameID`
- `Do NOT use SAML NameID`

api/saml/provider.py
vbessonov: As I described in [server_core/PR # 1253](https://github.com/NYPL-Simplified/server_core/pull/1253) I couldn't make it a property because `ConfigurationMetadata` has `__get__` method.

PR #1574, opened 2021-03-12
 Replace nosetests with pytest

jonathangreen:
## Description

In anticipation of the work to move from Python 2 to Python 3 I want to make sure that the CI system in place can smoothly handle this transition. During the initial testing with Python 3 we had some issues with the nose log capture handler.

This is a follow up to the work to move core to pytest https://github.com/NYPL-Simplified/server_core/pull/1245

It build on that work to get the circulation tests running on pytest. 

## Motivation and Context

Looking into nosetests, I noticed it had its last update in 2016 and has recommended that it not be used for new projects since 2015.

> Nose has been in maintenance mode for the past several years and will likely cease without a new person/team to take over maintainership. New projects should consider using Nose2, py.test, or just plain unittest/unittest2.

## Work

This is a large diff, but there are fewer logic changes here then there were in the core PR, since most of the fixture changes for pytest happened there.

## How Has This Been Tested?

Should be able to test this using docker with the following commands:
```
# Start DB
docker run -d -p 9005:5432/tcp --name db -e POSTGRES_USER=simplified_test -e POSTGRES_PASSWORD=test -e POSTGRES_DB=simplified_circulation_test postgres:9.6

# Set environment variables
export SIMPLIFIED_TEST_DATABASE="postgres://simplified_test:test@localhost:9005/simplified_circulation_test"

# Run tox
tox
```

As always any review / feedback is appreciated. @leonardr @tdilauro @vbessonov


.travis.yml
```diff
@@ -1,56 +0,0 @@
...
-  - ./verbose-test
-
-deploy:
-  provider: pages
-  skip_cleanup: true
-  keep-history: true
-  github_token: $GITHUB_TOKEN
-  on:
-    branch: main
-  local_dir: docs/build/html/
```
jonathangreen, line 55: I didn't move this deployment to github actions, since I'm not sure all the details here. Someone from NYPL will probably have to write the action for this.

integration_tests/test_borrow.py
```diff
@@ -1,7 +1,4 @@
-from nose.tools import (
```
jonathangreen, line 1: The tests in the integration_tests probably warrant some discussion. I haven't been able to run them for testing. There are a few changes from the automated runs, and find replace, but nothing in the `integration_tests` folder has been tested.

.travis.yml
leonardr: That will be @EdwinGuzman, though I can figure it out if necessary. I'm okay with doing this in between merging this branch and doing the Python 3 conversion.

integration_tests/test_borrow.py
leonardr: Most of the files in integration_tests could be removed. I think the combination of self-test-client (for diagnosing problems) and our mobile-based integration test suites (for everyday testing) negate the need for most of these scripts.

The one exception is test_search.py, which has a lot of knowledge that isn't contained anywhere else -- knowledge about what search requests ought to return what sorts of books.

I'm OK with keeping test_search.py without modifying it; we can fix it up later the next time we need to run search benchmarks. And I'm OK with removing the rest of the integration tests since we have other means of testing that stuff.

Some automatically-run integration tests that verify behavior on various CMs without relying on the mobile app would be useful, but I don't think the existing code (which is quite old) is a good starting point for that project.

leonardr: For now, leave integration_tests alone, and we'll decide to either fix or remove later.

leonardr:
This looks good overall. Next week Edwin and/or I will figure out what to do about publishing the docs, and we can get this and the core branch merged.

.travis.yml
leonardr: It looks pretty simple: https://github.com/marketplace/actions/deploy-to-github-pages

leonardr:
I filed https://jira.nypl.org/browse/SIMPLY-3615 to cover the work of porting to Github Actions the Sphinx doc build and push to Github Pages.

PR #1575, opened 2021-03-15
[WIP] SAML: migrate patrons with "old" IDs

vbessonov:
## Description

NOTE: This PR depends on [circulation # 1572](https://github.com/NYPL-Simplified/circulation/pull/1572).

This PR adds a new configuration setting `Patron ID: Migrate Patrons with "Old" IDs`:
image

This configuration setting allows to turn on the patron migration process and update patron IDs according to new **Patron ID** configuration setting values.

## Motivation and Context

Please consider the following scenario as an example:
1. "Old" patron IDs were extracted from NameID.
2. The requirements have changed and now patron IDs have to be extracted from eduPersonalPrincipalName attribute using `(?P<patron_id>.+)@university\.org` regular expression.

If we don't want to loose patron's holds and loans we have to migrate them. This process includes the following steps:
1. Find a patron object with an "old" ID.
2. Find or create a patron object with a "new" ID.
3. Transfer all the holds and loans from the "old" patron object to the "new" patron object.
4. Delete the "old" patron object.

## Checklist:

### DIFF
--- a/api/saml/configuration/model.py
+++ b/api/saml/configuration/model.py
@@ -126,6 +126,33 @@ class SAMLConfiguration(ConfigurationGrouping):
         required=False,
     )
 
+    patron_id_migrate_patrons_with_old_ids = ConfigurationMetadata(
+        key="saml_patron_id_migrate_patrons_with_old_ids",
+        label=_("Patron ID: Migrate Patrons with \"Old\" IDs"),
+        description=_(
+            "Boolean value indicating whether Circulation Manager should migrate patrons with \"old\" IDs, "
+            "i.e. IDs that were extracted using the default <b>Patron ID</b> settings from either "
+            "eduPersonUniqueId, eduPersonTargetedID, uid attributes, or NameID without using the regular expression:"
+            "<br>"
+            "- <b>0</b> (default) means that Circulation Manager won't migrate patrons with \"old\" IDs and "
+            "after changing the <b>Patron ID</b> configuration settings some patrons "
+            "MIGHT loose their holds and loans, "
+            "<br>"
+            "- <b>1</b> means that Circulation Manager will migrate patrons with \"old\" IDs by assigning them "
+            "a new ID extracted from a SAML assertion according to new <b>Patron ID</b> configuration setting values."
+            "<br>"
+            "Please consider the following scenario as an example:"
+            "<br>"
+            "1. \"Old\" patron IDs were extracted from NameID."
+            "<br>"
+            "2. The requirements changed and now IDs have to be extracted from eduPersonalPrincipalName attribute "
+            "using a regular expression."
+        ),
+        type=ConfigurationAttributeType.NUMBER,
+        required=False,
+        default=0,
+    )
+
     non_federated_identity_provider_xml_metadata = ConfigurationMetadata(
         key="idp_xml_metadata",
         label=_("Identity Provider's XML metadata"),
--- a/api/saml/configuration/model.py
+++ b/api/saml/configuration/model.py
@@ -1,3 +1,4 @@
+import cgi
 import json
 from threading import Lock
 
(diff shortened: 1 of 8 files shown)


tdilauro:
Is updating the old patron in place an option, too? Or is it necessary for some reason to create a new patron for this?

vbessonov:
@leonardr, what would be the right choice here: updating the ID of the existing `Patron` object or creating a new one and transferring all the holds and loans to it?

leonardr:
A couple questions:

1. It's unusual in this system to implement a migration process as part of the product rather than using the migration system. Is this something that needs to happen on an ongoing basis? That is, you might be serving a patron for years, and then one day the rules for extracting their NameID  have changed, and now they need to be migrated in this way? Or is this a one-time operation based on an improvement made by PR #1572?

2. It's also unusual in this system to offer a choice about whether to "migrate" or not. Why would someone want to change the rules for extracting NameID and _not_ migrate over the old accounts? Usually in situations like this the "migration" happens automatically.

3. Do the "old" patron and the "new" patron have the same permanent_id? If so, the right thing to do is to change the authorization_identifier. This is the same situation as a patron of a public library who loses their plastic card and gets a new one. The number on the plastic card (their authorization_identifier) is different but they have the same account number (permanent_id) as before.

If the old patron and new patron have different permanent_id, are you really sure they are the same person?

If you are sure, I think a new Patron object would be an acceptable solution. But it should be done as a _merge_ of two accounts--it's possible that the "new" permanent_id is already in the database!

vbessonov:
@leonardr, thank you for those questions above.
This functionality covers the following case: universities started using Circulation Manager with the default patron ID algorithm, meaning that both patron's `authorization_identifier` and `permanent_id` are set to a random GUID returned by the university's IdP in NameID. 
This really works, but universities already have their own patron IDs returned in eduPersonPrincipleName and usually have the following format: `NetID@university.org`. So we decided to start using `NetID` as a unique patron ID because it would allow us to share analytics with universities (they know only about NetIDs and don't know about GUIDs coming from SAML NameIDs).
The functionality in [PR #1572](https://github.com/NYPL-Simplified/circulation/pull/1572) is intended to allow us to override patron ID extraction logic and start using NetIDs. However, we don't want existing patrons to lose their holds and loans.
1. In this specific case, it really looks like a one-time operation: once we migrate all the university's patron to NetIDs, we're done.
2. However, functionality in [PR #1572](https://github.com/NYPL-Simplified/circulation/pull/1572) allows to change the extraction algorithm dynamically, and other universities may face the same problem in the future: they started using the default extraction algorithm and then after they decided to switch to change it to use some different SAML attribute as a unique patron ID.
The only concern I have is that the migration logic is currently designed so that it allows to migrate from the default settings to overridden settings. But it doesn't allow to migrate from one set of overriden settings to another. For example, we can migrate patrons which IDs were extracted using the default algorithm to IDs extracted from `eduPersonPrincipleName`. But we won't be able to migrate patrons with IDs from `eduPersonPrincipleName` to IDs from `uid`.
3. As I mentioned above SAML auth provider sets both `authorization_identifier` and `permanent_id` and treats them equally.

leonardr:
This is a lot clearer now. I see three options:

1. Treat this as a one-time change. Don't put this migration in the product at all, but keep track of this discussion. As similar "one-time" operations need to happen in the future, we can build up a picture of what a general solution would look like.
2. Put this in the product as a Script, triggered by a script in the bin/repair directory, so that it can be manually triggered as necessary. 
3. Go for the general solution right now. It would probably look like this: write a Monitor that can migrate from one field to another. Whenever the ID extraction setting is changed, set some state in the system so that a) the Monitor knows it needs to run on a particular SAML integration b) the Monitor knows what the old field was. Do this automatically whenever the setting is changed and add a notification that it will take some time to migrate users over to the new ID.

Right now I think the first solution sounds best. We've only seen a single type of migration (from `eduPersonPrincipleName` to `uid`), and it sounds like that particular migration won't be repeated, because future libraries will use `eduPersonPrincipleName` from the start. If that's right, the second solution will put code in the product that won't see use, and the third solution is a lot of engineering for a problem we don't have yet.

leonardr:
It looks like finding the patron and modifying their permanent ID (and authorization_identifier) will work in all real cases, and I think that's more reliable than creating a new patron and transferring things over. For example, this branch doesn't transfer over a patron's credentials or annotations.

PR #1583, opened 2021-03-26
Python 3 Conversion

EdwinGuzman:
## Description

This PR updates the repo to use python 3, first by running it through `2to3` and then doing manual updates that `2to3` didn't update as well as changes based on updated packages.
This also includes the python 3 PR branch of [server_core](https://github.com/NYPL-Simplified/server_core/pull/1244)

Todos:
* update docker to use this update.
* ~`test_individual_admins_post_create_on_setup` stalls while running and need a closer look why.~ - not adding `flask.request.files = {}` caused a python stack overflow.
* There seems to be an issue with the `xmlsec` package in the python 3.9 test Github Action job [ERROR: Failed building wheel for dm.xmlsec.binding](https://github.com/NYPL-Simplified/circulation/runs/2202892474#step:6:455). Not sure yet how to resolve this.
* Cleaning up some byte to string changes.

List of changes:
* Switched from StringIO to BytesIO for the buffer used for images and for zip files.
* Getting the data from a Flask Response is now `.get_data(as_text=True)` instead of `.data` or `.response[0]`.
* Removed `encodestring` and `decodestring` for `encodebytes` and `decodebytes`, respectively.
* HMACAlgorithm `.sign` function now takes in a byte string.
* `jwt.encode` now returns a byte string and must be decoded.
* `exportKey()` from `RSA.generate` now returns a byte string and must be decoded.
* Exceptions no longer have a `.message` property and must use `str(e)` to get the exception's error message.
* `hashlib.sha256` now takes in a byte string.
* `os.urandom(...).encode("hex")`  no longer works because `urandom` returns a byte string. I switched this to use `binascii.hexlify(os.urandom(...)).decode("utf-8")`.
* `OneLogin_Saml2_Utils` no longer has the functions `element_text`, `query`, or `validate_xml`. `OneLogin_Saml2_XML` now replaces `OneLogin_Saml2_Utils`.
* Added a `__hash__` function to the `BaseError` class. There were multiple `TypeError: unhashable type: X` errors where X was one of many subclasses of `BaseError`.
* Switched form using `iteritems()` to `items()`.
* Some if statements that compared string/int with `None` were updated (can't compare with None).
* `has_key` was removed and now `in` is recommended.

## Motivation and Context

Resolves [SIMPLY-2214](https://jira.nypl.org/browse/SIMPLY-2214).

## How Has This Been Tested?

Through `tox` with python 3.6, 3.7, 3.8, 3.9.

## Checklist:


api/enki.py
```diff
@@ -557,16 +557,16 @@ class BibliographicParser(object):
...
-        "French" : u"fre",
-        "Spanish": u"spa",
+        "English": "eng",
+        "French" : "fre",
+        "Spanish": "spa",
     }
 
     def process_all(self, json_data):
-        if isinstance(json_data, basestring):
+        if isinstance(json_data, (bytes, str)):
```
leonardr, line 566: I'm fine with this one going forward because json.loads works on both bytes and strings.

api/local_analytics_exporter.py
```diff
@@ -43,8 +43,7 @@ def export(self, _db, start, end, locations=None, library=None):
         writer = csv.writer(output, encoding="utf-8")
         writer.writerow(header)
         writer.writerows(results)
-
-        return output.getvalue()
+        return output.getvalue().decode("utf-8")
```
leonardr, line 46: I was hoping we could avoid this encode/decode step but it looks like we can't.

api/registry.py
```diff
@@ -246,7 +246,7 @@ def _decode_data_url(cls, url):
             raise ValueError(
                 "Unsupported media type in data: URL: %s" % media_type
             )
-        html = base64.b64decode(encoded)
+        html = base64.b64decode(encoded.encode("utf-8")).decode("utf-8")
```
leonardr, line 249: This seems excessive?

api/local_analytics_exporter.py
EdwinGuzman: Yeah, not for BytesIO buffer.

api/registry.py
EdwinGuzman: `b64decode` now takes a byte string. I was trying to use the base64 from the util class but then other encoding/decoding in this file break. I think I got fewer errors using the normal base64 package and doing the above than using the unicode aware base64 class and changing other strings in this file.

tests/lcp/test_server.py
```diff
@@ -89,7 +89,7 @@ def test_add_content(self, _, input_directory):
         ('none_rights', None, None, None, None),
         (
                 'license_start',
-                datetime.datetime(2020, 01, 01, 00, 00, 00),
+                datetime.datetime(2020, 0o1, 0o1, 00, 00, 00),
```
leonardr, line 92: This works, but it's defining 1 as an octal number, which doesn't make sense and could cause problems in the future. I'm guessing this happened automatically as an attempt to make sense of `01`, which is not valid in Python 3. I think just change these to `1`.

tests/proquest/test_client.py
```diff
@@ -310,8 +310,8 @@ def test_get_book_correctly_fails(
     def test_get_book_correctly_extracts_open_access_books(self):
         # Arrange
         book_content = "PDF Book12345"
-        response_arguments = {"content": book_content}
-        expected_open_access_book = ProQuestBook(content=bytes(book_content))
+        response_arguments = {"content": book_content.encode("utf-8")}
+        expected_open_access_book = ProQuestBook(content=book_content.encode("utf-8"))
```
leonardr, line 314: Not a big deal but it would make more sense conceptually to have the mock book contents start out as bytestrings, since they represent files on disk.

tests/saml/configuration/test_model.py
```diff
@@ -503,7 +504,6 @@ def test_get_settings_returns_correct_result(self):
                 "logoutResponseSigned": False,
                 "wantMessagesSigned": False,
                 "metadataCacheDuration": None,
-                "rejectUnsolicitedResponsesWithInResponseTo": False,
```
leonardr, line 506: I don't understand why Python 3 conversion by itself would make these changes.

tests/test_controller.py
```diff
@@ -1,22 +1,20 @@
 # encoding=utf8
-import base64
+# import base64
```
leonardr, line 2: Remove this rather than commenting it out.

leonardr:
I've been through this and see no major issues. We can address the Python 3.9 issue later; for now I'm fine with removing that test from tox.

tdilauro:
@EdwinGuzman I have seen the `ERROR: Failed building wheel for dm.xmlsec.binding` issue in containers based on ubuntu 20.04, where it builds fine under `ubuntu 18.04`. I have no idea, though, why this would fail only with Python 3.9.

@leonardr may recall that we also have had issues with libxmlsec in circ containers in the past. If I remember correctly, there was a dependency conflict caused by packages needed by an older `node` and I had to build xmlsec manually, at least for a time, to get the containers to build.

tdilauro:
I just thought of one other issue: I recall hearing that there was an issue with `datetime.utcnow()` between Python 2 and 3. I'll see what I can find tomorrow.

tdilauro:
Here's some commentary about potential issues with `datetime.utcnow()` between Python 2 and 3:
- https://bugs.python.org/issue37488
- https://blog.ganssle.io/articles/2019/11/utcnow.html

leonardr:
I filed https://jira.nypl.org/browse/SIMPLY-3663 for the utcnow thing.

leonardr:
Viacheslav found some nondeterministic code in tests/admin/test_geographic_validator.py.TestGeographicValidator.test_ask_registry; see https://github.com/NYPL-Simplified/circulation/pull/1586/checks?check_run_id=2243831603 and https://github.com/NYPL-Simplified/circulation/pull/1586

Rather than fix this in Py2 and then again in Py3, can you see if it's easy to add to this branch? Otherwise I'll open a ticket and we'll address it separately after the Py3 conversion.

tests/saml/configuration/test_model.py
EdwinGuzman: It didn't. This property is no longer in the object.

tests/lcp/test_server.py
EdwinGuzman: Yeah, this was an odd update from the converter.

EdwinGuzman:
> Viacheslav found some nondeterministic code in tests/admin/test_geographic_validator.py.TestGeographicValidator.test_ask_registry; see https://github.com/NYPL-Simplified/circulation/pull/1586/checks?check_run_id=2243831603 and #1586
Rather than fix this in Py2 and then again in Py3, can you see if it's easy to add to this branch? Otherwise I'll open a ticket and we'll address it separately after the Py3 conversion.

Looking into this now.

EdwinGuzman:
Fixed the nondeterministic test in https://github.com/NYPL-Simplified/circulation/pull/1583/commits/9ce79a549009bb9bdaa243108fef5c863bf6914a

tests/saml/configuration/test_validator.py
```diff
@@ -77,31 +77,31 @@ def setup_class(cls):
...
-                        u"SAML patron ID regular expression '[' has an incorrect format: "
-                        u"unexpected end of regular expression"
-                    ),
+                "incorrect_patron_id_regular_expression",
+                fixtures.CORRECT_XML_WITH_ONE_SP,
+                fixtures.CORRECT_XML_WITH_IDP_1,
+                r"[",
+                SAML_INCORRECT_PATRON_ID_REGULAR_EXPRESSION.detailed(
+                    "SAML patron ID regular expression '[' has an incorrect format: "
+                    "unterminated character set at position 0"
```
EdwinGuzman, line 103: The regex error message changed in python 3.

leonardr:
I've started using this branch as a basis for development work and I pretty quickly ran into a problem in novelist.py:

```
        if content.startswith('"Missing'):
            raise Exception("Invalid NoveList parameters: %s" % content)
```

`content` is a bytestring at this point because it came out of an HTTP request, so `'Missing` also needs to be a bytestring. Tests didn't catch this because test_review_response passes in a Unicode string to simulate an error message.

It won't help this particular case, but we could probably catch some more errors like this by making sure that whenever you queue a response in a mock API, the content of the response is converted into a bytestring.

EdwinGuzman:
That makes sense. I can go ahead and update all the instances where we are mocking API responses so that content messages are byte strings. This should be the same situation with reading/writing files which I covered in tests but haven't actually tested in real life. Are there any other instances where byte strings are read besides from API responses and files?

leonardr:
The only ones I can think of are when we run data through an encryption, decryption, or Base64 encoding process

PR #1584, opened 2021-03-30
Feature/bibliotheca event monitor multi collection enhancements

tdilauro:
## Description

This branch:
- Adds a new script runner for Bibliotheca Event Monitor to support the following command line options.
  - `--default-date` option to override a monitor's intrinsic default date with an ISO 8601 date of arbitrary resolution.
  - `--override-timestamp` flag to allow a command line default date to override start time from timestamp.
- Support the new `CollectionArgumentsMonitor` support in `server_core.
- Improves readability of log messages by displaying dates in ISO 8601 date format.
- Renames some variables and concepts to improve clarity (I hope).
- Refactor `catch_up_from` to clarify intent.
- Add warning of potentially missed events, if 100 or more. (See [SIMPLY-3417](https://jira.nypl.org/browse/SIMPLY-3417)).
- Enhances some docstrings.

## Motivation and Context

Needed to make it easier to: 
- Run `bibliotheca_monitor` with a default start time.
- Override an existing timestamp without manually modify the datebase.
- Run the monitor for specific collection(s) and understand which log messages apply to which collection.

https://jira.nypl.org/browse/SIMPLY-3416.

## How Has This Been Tested?

Manual testing and new tests.

## Checklist:

### DIFF
--- a/api/bibliotheca.py
+++ b/api/bibliotheca.py
@@ -1401,19 +1401,14 @@ class BibliothecaEventMonitor(CollectionMonitor, TimelineMonitor):
 
     @staticmethod
     def _impose_bibliotheca_timespan_constraint(timespan):
+        # Due to the workings of the Bibliotheca API, no mode
+        # should have a timespan longer than 72 hours.
         if timespan <= timedelta(hours=72):
             return
         raise ValueError('A timespan greater than 72 hours can result in spurious'
                          ' results from the Bibliotheca events API.')
 
     def catch_up_from(self, start, cutoff, progress):
-        recent_mode_counter_value = 1
-        backlog_mode_counter_value = 0
-        # Due to the workings of the Bibliotheca API, no mode
-        # should have a timespan longer than 72 hours.
-        recent_mode_timespan = timedelta(hours=70)
-        backlog_mode_timespan = timedelta(minutes=5)
-
         # This method operates in three modes: Uninitialized, Backlog,
         # and Recent. Uninitialized Mode happens when the monitor's
         # `progress.counter` is null, but is functionally identical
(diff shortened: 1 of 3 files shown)


api/bibliotheca.py
```diff
@@ -1288,9 +1282,35 @@ class BibliothecaEventMonitor(CollectionMonitor, TimelineMonitor):
...
+                 override_timestamp=False,
+                 ):
+        """Initializer.
+
+        :param _db: Database session object.
+
+        :param collection: Collection for which this monitor operates.
+
+        :param api_class: API class or an instance thereof for this monitor.
+        :param api-class: Union[Type[BibliothecaAPI], BibliothecaAPI]
```
EdwinGuzman, line 1298: Is `api-class` meant to have the hyphen rather than the underscore?

api/bibliotheca.py
```diff
@@ -1447,9 +1501,44 @@ def handle_event(self, bibliotheca_id, isbn, foreign_patron_id,
...
         return start_time
 
+
+class RunBibliothecaMonitorScript(RunCollectionMonitorScript):
+
+    def __init__(self, monitor_class, _db=None, cmd_args=None, **kwargs):
+        super(RunBibliothecaMonitorScript, self).__init__(monitor_class, _db, cmd_args, **kwargs)
+
+    # TODO: Both '--default-start-date' and '--force-default' (or equivalents) should probably end up
+    #   in a more generalized script runner superclass that is focused on Monitor or TimelineMonitor.
```
leonardr, line 1515: I'm not sure about this; the Bibliotheca monitor is the only one that has needed a default start date. The other ones use the time of the first run as the start date, since they don't need to ask about the past. And once the new Bibliotheca API comes out I think even Bibliotheca won't need this.

bin/bibliotheca_monitor
```diff
@@ -6,5 +6,8 @@ bin_dir = os.path.split(__file__)[0]
 package_dir = os.path.join(bin_dir, "..")
 sys.path.append(os.path.abspath(package_dir))
 from core.scripts import RunCollectionMonitorScript
-from api.bibliotheca import BibliothecaEventMonitor
-RunCollectionMonitorScript(BibliothecaEventMonitor, cli_date=sys.argv[1:2]).run()
+from api.bibliotheca import BibliothecaEventMonitor, RunBibliothecaMonitorScript
+try:
+    RunBibliothecaMonitorScript(BibliothecaEventMonitor).run()
+except KeyboardInterrupt:
+    pass
```
leonardr, line 13: Why ignore this exception, especially since the script is going to end anyway?

EdwinGuzman (approved):
Looks good to me. Btw, it's still in "draft" mode.

bin/bibliotheca_monitor
tdilauro: Mainly because it's annoying to someone running it on the command line, but I can take it out, so that we can see a traceback in case there's an infinite loop or similar.

api/bibliotheca.py
tdilauro: Typo. Thanks for catching that.

bin/bibliotheca_monitor
leonardr: Yeah, I would take it out.

api/bibliotheca.py
tdilauro: There RBdigital, which I know is gone, is another case where I would have found that useful and I suspect that there will be cases in the future where it would be nice to get a do-over without having to manually change the database. But I certainly don't mind removing this.

leonardr: It's just a TODO, and who knows what the future will hold.

bin/bibliotheca_monitor
tdilauro: Done.

leonardr (approved):

EdwinGuzman (approved):
The update looks good 👍

tdilauro:
I'm done and this is ready to merge. Thanks!

PR #1586, opened 2021-04-01
Update core submodule to the latest version of server_core

vbessonov:
## Description

This PR updates `core` submodule to the latest version of `server_core` needed by recently merged [PR # 1572](https://github.com/NYPL-Simplified/circulation/pull/1572) and [PR # 1584](https://github.com/NYPL-Simplified/circulation/pull/1584).

## Checklist:


vbessonov:
@leonardr, @tdilauro, the second test run failed because [test_ask_registry](https://github.com/NYPL-Simplified/circulation/blob/ee0ba53ac1070ffc320795e4984da251ab99ea03/tests/admin/test_geographic_validator.py#L155) is not deterministic:
1. It saves a list of registries in the database.
2. Then using `RemoteRegistry.for_protocol_and_goal` inside `GeographicValidator.ask_registry` it tries to fetch available registries. The problem is that the order we get them is not guaranteed since we don't explicitly sort them.
This is what [the PostgreSQL documentation](https://www.postgresql.org/docs/9.1/queries-order.html) says regarding the default order:
> The actual order in that case will depend on the scan and join plan types and the order on disk, but it must not be relied on. A particular output ordering can only be guaranteed if the sort step is explicitly chosen.

The simplest solution would be to mock `RemoteRegistry.for_protocol_and_goal` and make it return registries in a specific order.

leonardr:
I asked Edwin to look into making this change in the Python 3 branch, so we don't have to address it separately in Py2 and Py3.

https://github.com/NYPL-Simplified/circulation/pull/1583#issuecomment-811914596

tdilauro (approved):
Looks good, aside from the other issue you tracked down.

PR #1587, opened 2021-04-06
Disable the authentication document cache by default, and make it the location of optional UWSGI debugging information.

leonardr:
## Description

This branch productizes two things I've learned while trying to fix SIMPLY-3346:

1. At the moment, the authentication document cache is too much trouble to justify the relatively small performance improvement it brings. I've disabled it on every site NYPL hosts and it should be disabled by default.
2. Some deployment configurations seem to send bad UWSGI data to the app server, but there's no easy way to look at that data on a running site.

I've put item #2 into QA servers as a hotfix; this productizes the hotfix so it can be turned on as necessary, to solve this problem or other problems in the future. If UWSGI debug is enabled (this is a secret site-wide setting called `uwsgi_debug`) you can get UWSGI debug information by making a request to `authentication_document` and putting `debug` in the query string, e.g. `https://qa-circulation.openebooks.us/USOEI/authentication_document?debug`.

## How Has This Been Tested?

In addition to the unit tests I tested this against a local development database.

## Checklist:

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -304,11 +304,14 @@ class CirculationManager(object):
         authentication_document_cache_time = int(
             ConfigurationSetting.sitewide(
                 self._db, Configuration.AUTHENTICATION_DOCUMENT_CACHE_TIME
-            ).value_or_default(3600)
+            ).value_or_default(0)
         )
         self.authentication_for_opds_documents = ExpiringDict(
             max_len=1000, max_age_seconds=authentication_document_cache_time
         )
+        self.uwsgi_debug = ConfigurationSetting.sitewide(
+            self._db, Configuration.UWSGI_DEBUG_KEY
+        ).bool_value or False
 
     @property
     def external_search(self):
@@ -550,6 +553,20 @@ class CirculationManager(object):
             # time.
             value = self.auth.create_authentication_document()
             self.authentication_for_opds_documents[name] = value
+
+        if self.uwsgi_debug and 'debug' in flask.request.args:
+            # Annotate with debugging information about the UWSGI
+            # environment and the authentication document cache
+            # itself.
+            value = json.loads(value)
+            value['_debug'] = dict(
+                url=self.url_for(
+                    'authentication_document', library_short_name=name
+                ),
+                environ=str(dict(flask.request.environ)),
+                cache=str(self.authentication_for_opds_documents),
+            )
+            value = json.dumps(value)
         return value
 
     @property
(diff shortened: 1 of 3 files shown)


api/controller.py
```diff
@@ -304,11 +304,14 @@ def get_domain(url):
         authentication_document_cache_time = int(
             ConfigurationSetting.sitewide(
                 self._db, Configuration.AUTHENTICATION_DOCUMENT_CACHE_TIME
-            ).value_or_default(3600)
+            ).value_or_default(0)
```
matthewhelt, line 307: Is there a way to completely skip the cache instead of setting the retention time span to 0? There might be a performance impact to skip it altogether (although I have no profiled the run time for this). This question is moot if the cache is an intended/required function of the architecture.

tests/test_controller.py
```diff
@@ -422,7 +425,11 @@ def mock_for_library(cls, incoming_library):
 
         ConfigurationSetting.sitewide(
             self._db, Configuration.AUTHENTICATION_DOCUMENT_CACHE_TIME
-        ).value = "0"
+        ).value = "60"
```
matthewhelt, line 428: I'd add a comment that these settings are seconds where the Configuration.AUTHENTICATION_DOCUMENT_CACHE_TIME value is in MS - just for clarity when reviewing test output.

api/controller.py
leonardr: I could remove the cache altogether without changing much, but conceptually it makes a lot of sense, so I'd like to keep the code in place and fix it (or decide to remove it) later.

tests/test_controller.py
leonardr: Where are you getting the milliseconds idea? The default value for AUTHENTICATION_DOCUMENT_CACHE_TIME is 3600 (which I'm about to change in a follow-up).

matthewhelt (commented):
Would it not be better to have a separate endpoint for the WSGI data to be populated so one could still evaluate the auth document and also see application state? I think overloading the endpoint with 2 functions may not be ideal if one needed to debug the rendered state of a non-cached auth doc and still be able to grab the state of a given request, this would require a change to log the wsgi request object in the same thread and require more telemetry, so this might be viable for immediate needs regardless.

matthewhelt:
> Would it not be better to have a separate endpoint for the WSGI

Its been clarified that this change adds a "_debug" key to the auth document and does not override that document or any content that would be rendered otherwise. As it's not a default option and would be very helpful this presents only the issues of inadvertent use which could disclose information and thus should be documented.

matthewhelt (approved):

leonardr:
I added the docs you requested to the docstring of the controller method. I also changed the term used throughout from "uwsgi" (which is a specific package) to "wsgi" (the name of the spec).

matthewhelt (approved):
Per approach to adjust timeout in the given architecture.

PR #1589, opened 2021-04-14
Datetime update

leonardr:
## Description

This branch changes all circulation code to use timezone-aware objects, generally with the timezone being set to UTC. It's the circulation equivalent to https://github.com/NYPL-Simplified/server_core/pull/1256

## Motivation and Context

https://jira.nypl.org/browse/SIMPLY-3663.

## Checklist:

### DIFF
--- a/tests/test_nyt.py
+++ b/tests/test_nyt.py
@@ -227,11 +227,13 @@ class TestNYTBestSellerList(NYTBestSellerAPITest):
 
         jan_17 = datetime.datetime(2015, 1, 17).replace(tzinfo=self.api.TIME_ZONE)
         assert (True ==
-            all([x.first_appearance == jan_17 for x in custom.entries]))
+            all([x.first_appearance.replace(tzinfo=self.api.TIME_ZONE) ==
+                jan_17 for x in custom.entries]))
 
         feb_1 = datetime.datetime(2015, 2, 1).replace(tzinfo=self.api.TIME_ZONE)
         assert (True ==
-            all([x.most_recent_appearance == feb_1 for x in custom.entries]))
+            all([x.most_recent_appearance.replace(tzinfo=self.api.TIME_ZONE) ==
+                feb_1 for x in custom.entries]))
 
         # Now replace this list's entries with the entries from a
         # different list. We wouldn't do this in real life, but it's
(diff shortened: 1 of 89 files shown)


EdwinGuzman:
Sorry, for I got confused and thought this was my working branch and merged in the `python3` branch. I noticed there's a new file which you also pointed out. I can't find a trace of it outside this branch, however.

leonardr:
It looks like I wrote test_announcement_list_validator when I was working with Adriana on the announcement list feature, and just didn't put it in the repo. I think it's just been living on my laptop.

tests/test_nyt.py
```diff
@@ -227,11 +227,13 @@ def test_to_customlist(self):
 
         jan_17 = datetime.datetime(2015, 1, 17).replace(tzinfo=self.api.TIME_ZONE)
         assert (True ==
-            all([x.first_appearance == jan_17 for x in custom.entries]))
+            all([x.first_appearance.replace(tzinfo=self.api.TIME_ZONE) ==
```
EdwinGuzman, line 230: `x.first_appearance` gets its timezone information from the database since the list entries are queried from the database. It'll appear as `datetime.datetime(2015, 1, 17, 0, 0, tzinfo=psycopg2.tz.FixedOffsetTimezone(offset=0, name=None))` so that's why this test fails. But not sure if this is the _best_ way to check this.

leonardr: Yeah, I don't get it -- first_appearance should be set at a date that represents midnight Eastern time, not midnight UTC.

EdwinGuzman: I updated `jan_17` to `datetime.datetime(2015, 1, 17, tzinfo=dateutil.tz.tzoffset("", 0))` which gives a better representation of what's coming from the database.

EdwinGuzman:
I think I covered the last datetime objects that needed to be converted from naive to aware objects. There is one test that fails but it's because in that one instance it just took slightly longer to run https://github.com/NYPL-Simplified/circulation/pull/1589/checks?check_run_id=2394598448#step:6:332

tests/test_nyt.py
leonardr: See my later comment -- first_appearance was set to midnight UTC because the timezone info got wiped when we converted datetime to date.

PR #1597, opened 2021-05-14
Implement BibliothecaPurchaseMonitor to take the burden of learning about new titles off of BibliothecaEventMonitor

leonardr:
## Description

This branch creates a new client method for the Bibliotheca API, `marc_request`. This method uses the `/data/marc` endpoint to iterate over a paginated list of book purchases for a given date. The API returns a MARC XML record for each book purchased, and `marc_request` yields a sequence of parsed MARC record objects.

This branch introduces a new monitor, the BibliothecaPurchaseMonitor, which retrieves all book purchase data for each Bibliotheca collection, starting in 2014 (or a user-specified date). It uses `marc_request` to retrieve book purchases for one day at a time.

`/data/marc` is much faster and more reliable than `/data/cloudevents`, and it gives us all the events we absolutely have to hear about. This lets me remove a lot of the complexity we built up around the `BibliothecaEventMonitor` while trying to make it more reliable.

I also moved the code that lets you specify a start date from `BibliothecaEventMonitor` to `BibliothecaPurchaseMonitor`, for collections that go back earlier than 2014. Basically, we are no longer trying to learn everything that happened in a Bibliotheca collection for the past several years. We don't really need that information and we know we'll miss a bunch of it anyway. What we absolutely have to have are the purchase records, which we can get quite reliably from `/data/marc`.

## Motivation and Context

https://jira.nypl.org/browse/SIMPLY-3739

Despite the changes we've made over the past few months to BibliothecaEventMonitor, it's simply impossible to get it to reliably find all of the "new book" events (since librarians frequently purchase titles in batches of 100 or more). This branch reduces the role of BibliothecaEventMonitor to "makes things run more smoothly but is not essential."

## How Has This Been Tested?

I set up a local environment with NYPL's Bibliotheca credentials and ran the BibliothecaPurchaseMonitor. It has been running all day and so far it's made it from 1-1-2014 to 9-15-2019, with about 42000 MARC records processed and about 21000 works created.

## Checklist:


api/bibliotheca.py
```diff
@@ -1392,7 +1440,7 @@ def _intrinsic_start_time(self, _db):
         # We don't use Monitor.timestamp() because that will create
         # the timestamp if it doesn't exist -- we want to see whether
         # or not it exists.
-        default_start_time = utc_now() - self.DEFAULT_START_TIME
+        default_start_time = self.DEFAULT_START_TIME
```
leonardr, line 1443: Previously the default start time was a certain number of years in the past. Now it's a specific date: January 1, 2014.

api/bibliotheca.py
```diff
@@ -1413,96 +1461,218 @@ def timestamp(self):
...
-        # and Recent. Uninitialized Mode happens when the monitor's
-        # `progress.counter` is null, but is functionally identical
-        # to Backlog Mode (0, backlog_mode_counter_value).
-        # In both of these modes, the timespan for which we fetch
-        # events is `backlog_mode_timespan` and it is NOT considered
-        # an error when the response contains no events.
-        # In Recent Mode, the timespan (`backlog_mode_timespan`) is
-        # longer and a response contains no events DOES raise an error.
-        recent_mode_counter_value = 1
-        backlog_mode_counter_value = 0
```
leonardr, line 1443: I was able to remove the code that switched from one mode to another, since it's not an issue for the BibliothecaPurchaseMonitor, and the BibliothecaEventMonitor no longer needs to be super reliable.

api/bibliotheca.py
```diff
@@ -1413,96 +1461,218 @@ def timestamp(self):
...
                 events_handled += 1
-                if not events_handled % 1000:
-                    self._db.commit()
             self._db.commit()
-            _events_this_request = events_handled - _events_pre_request
-            if _events_this_request >= reliable_max_cloudevents_count:
-                self.log.warning(
-                    'Response contained %d events, which equals or exceeds the number '
-                    'reliably returned by this API (%d). Some events may have been lost.',
-                    _events_this_request, reliable_max_cloudevents_count)
```
leonardr, line 1495: We could keep this warning but the docstring makes it  understood that this isn't super reliable, there's nothing we can do about this, and we'll recover from the problem automatically when the circulation sweep runs. So I don't think it does much now to have this warning.

tests/test_bibliotheca.py
```diff
@@ -932,19 +977,19 @@ def test_exception(self):
         error = parser.process_all(self.TRIED_TO_HOLD_BOOK_ON_LOAN)
         assert isinstance(error, CannotHold)
 
-
-class TestBibliothecaEventMonitor(BibliothecaAPITest):
+class TestBibliothecaPurchaseMonitor(BibliothecaAPITest):
```
leonardr, line 980: Most of the EventMonitor tests end up in the PurchaseMonitor test class because they focused on the ability to set a custom start date, something that now lives in PurchaseMonitor.

api/bibliotheca.py
```diff
@@ -1283,32 +1311,62 @@ def _process_metadata(
...
+    collection, even if that date was years in the past. That's
+    because this monitor may be the only time we hear about a
+    particular book.
+
+    Because of this, this monitor has a very old DEFAULT_START_TIME
+    and special capabilities for customizing the start_time to go back
+    even further.
+    """
+    SERVICE_NAME = "Bibliotheca Purchase Monitor"
+    DEFAULT_START_TIME = datetime_utc(2014, 1, 1)
```
leonardr, line 1364: DEFAULT_START_TIME has moved from the event monitor, and changed from a timedelta (3 years in the past) to a constant date (1/1/2014).

api/bibliotheca.py
```diff
@@ -1413,96 +1461,221 @@ def timestamp(self):
...
+            cutoff.strftime(self.LOG_DATE_FORMAT)
+        )
+        events_handled = 0
+
+        # Since we'll never get more than about 100 events from a
+        # single API call, slice the timespan into relatively small
+        # chunks.
         for slice_start, slice_cutoff, full_slice in self.slice_timespan(
-            start, cutoff, timespan_to_check
+            start, cutoff, timedelta(minutes=5)
```
leonardr, line 1671: Since it's no longer essential that we get all the events from this monitor, we can go back to a hard-coded delta that will work pretty well most of the time.

api/bibliotheca.py
```diff
@@ -1545,25 +1721,22 @@ def handle_event(self, bibliotheca_id, isbn, foreign_patron_id,
...
         )
 
-        if is_new:
-            # This is our first time seeing this LicensePool. Log its
-            # occurance as a separate event.
-            license_pool.collect_analytics_event(
-                self.analytics, CirculationEvent.DISTRIBUTOR_TITLE_ADD,
-                license_pool.last_checked or start_time,
-                0, 1
-            )
```
leonardr, line 1555: This went to the purchase monitor.

tests/test_bibliotheca.py
```diff
@@ -1200,16 +1510,19 @@ def test_handle_event(self):
...
-        # license available, one for the first appearance of a new
-        # LicensePool, and a redundant 'license add' event
-        # which was registered with analytics but which did not
-        # affect the counts.
-        assert 4 == analytics.count
+        # Three analytics events were collected: one for the license
+        # add event itself, one for the 'checkin' that made the new
+        # license available, and a redundant 'license add' event which
+        # was registered with analytics but which did not affect the
+        # counts.
```
leonardr, line 1517: I still don't understand this redundant event but it's not new in this branch so I didn't investigate too closely.

api/bibliotheca.py
```diff
@@ -1283,32 +1311,62 @@ def _process_metadata(
...
-    the BibliothecaCirculationSweep.
-    """
+        :param collection: Collection for which this monitor operates.
 
-    SERVICE_NAME = "Bibliotheca Event Monitor"
-    DEFAULT_START_TIME = timedelta(days=365*3)
-    PROTOCOL = ExternalIntegration.BIBLIOTHECA
-    LOG_DATE_FORMAT = '%Y-%m-%dT%H:%M:%S'
+        :param api_class: API class or an instance thereof for this monitor.
+        :param api_class: Union[Type[BibliothecaAPI], BibliothecaAPI]
```
EdwinGuzman, line 1329: I think you meant `:type` instead of another `:param`.

api/bibliotheca.py
```diff
@@ -1413,96 +1461,221 @@ def timestamp(self):
...
+        progress.apply(self._db)
+        self._db.commit()
+
+    def purchases(self, start, end):
+        """Ask Bibliotheca for a MARC record for each book purchased
+        between `start` and `end`.
+
+        :yield: A sequence of pymarc Record objects
+        """
+        offset = 1     # Smallest allowed offset
```
EdwinGuzman, line 1551: What's the offset for in this case?

EdwinGuzman: Never mind, I saw it afterwards.

EdwinGuzman (approved):

leonardr:
I added a note to https://jira.nypl.org/browse/SIMPLY-3739 about a blip you're likely to see in Google Analytics the first time you run this monitor.

PR #1599, opened 2021-05-20
Py3 and flake8 updates to informational scripts

nballenger:
## Description

Two things:

* Updates the scripts in `bin/informational` to be Python3  compatible, and use `python3` in their `#!` line
* Changes the `adobe-id-for-short-client` script to be `click` based and accept a `--registry-url` CLI option

## Motivation and Context

Part of the Py3 QA process.

## How Has This Been Tested?

I've only actually tested the `adobe-id-for-short-client` script, and that was done manually. However the other changes I made were almost entirely formatting related, so shouldn't impact script functionality.

## Checklist:

### DIFF
--- a/requirements-base.txt
+++ b/requirements-base.txt
@@ -3,7 +3,7 @@ backports.functools-lru-cache==1.6.1
 backports.ssl-match-hostname==3.7.0.1
 beautifulsoup4==4.9.1
 cffi==1.14.3
-click==8.0.1
+click==7.1.2
 contextlib2==0.6.0.post1
 cryptography==3.3.2
 # defusedxml is required for SAML authentication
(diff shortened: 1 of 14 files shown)


requirements-base.txt
```diff
@@ -3,6 +3,7 @@ backports.functools-lru-cache==1.6.1
 backports.ssl-match-hostname==3.7.0.1
 beautifulsoup4==4.9.1
 cffi==1.14.3
+click==8.0.1
```
leonardr, line 6: Core has a requirement of click 7.1.2, which is causing the conflict that breaks the tests. We don't use click directly, so it must have been introduced as a dependency. Try upgrading it in core and if that doesn't work, try removing this requirement and see if we can use 7.1.2 here.

leonardr (approved):
This is a necessary first step; the information scripts are rarely used so we can fix them as problems come up. Also, _since_ they are rarely used any problems that do show up are now more likely to be related to general bit rot than Python 3 in particular.

PR #1600, opened 2021-05-21
Switching db task from auto to ignore

nballenger:
## Description

Sets the env var `SIMPLIFIED_DB_TASK` to `ignore` in both the webapp and exec dockerfiles, to keep those instances from running database migrations, which should only be run by the script runner.

## Motivation and Context

Should resolve https://jira.nypl.org/browse/SIMPLY-3745, once a new image is built and pushed.

### DIFF
--- a/docker/Dockerfile.exec
+++ b/docker/Dockerfile.exec
@@ -19,7 +19,7 @@ COPY --from=builder /go/bin/lcpencrypt /go/bin/lcpencrypt
 ARG version
 ARG repo="NYPL-Simplified/circulation"
 
-ENV SIMPLIFIED_DB_TASK "auto"
+ENV SIMPLIFIED_DB_TASK "ignore"
 ENV SIMPLIFIED_SCRIPT_NAME ""
 
 # Copy over all Library Simplified build files for this image
--- a/docker/Dockerfile.webapp
+++ b/docker/Dockerfile.webapp
@@ -4,7 +4,7 @@ LABEL maintainer="Library Simplified <info@librarysimplified.org>"
 ARG version
 ARG repo="NYPL-Simplified/circulation"
 
-ENV SIMPLIFIED_DB_TASK "auto"
+ENV SIMPLIFIED_DB_TASK "ignore"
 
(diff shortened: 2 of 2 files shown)


leonardr (approved):

alsrlw:
Technically, I would consider this a deployment decision rather than an app decision. I've deployed multiple scripts containers (with separate, non-redundant cron task files) for a single CM at times. Conversely, I've deployed multiple webapp containers set to "migrate" without issue. But that's just anecdotal. In my way of thinking, webapp is the primary because one can begin setting up a CM with just the webapp container and a database. Scripts can come into play after some integrations are configured. 

It seems better as a deployer's choice, which container gets "turned up first." As is how to make certain the database is ready to accept connections before deploying the scripts/webapp/exec container(s). Does changing this setting have any negative potential to mess with someone's deployment code who chooses to deploy differently than the "single script container first" paradigm?

In one sense it doesn't matter what the defaults are. Maybe we need to include guidance that the scripts container is set to run "first" and perform database initialization/migrations by default. And choosing to run/deploy another container first would require setting the proper run-time environment variable to init/migrate to accommodate that deployment scenario.

leonardr:
That's a good point, Robert. I think this works as a default, since "one script runner, N app servers" works out of the box and "N script runners, N app servers" won't. But we could instead shoot for an invariant that works no matter how . I think this would always work:

Immediately after gaining a database connection:

* If the database needs migration, and no migration is active, set a migration active and run the migration.
* If the database needs migration, but a migration is nothing, do... what? Wait? Fail out?
* If the database doesn't need migration, proceed immediately.

Nick, should I file a ticket for this? You've mentioned moving to a third-party migration framework rather than rolling our own; do you know if they can deal with this situation?

leonardr:
If we're addressing this issue by redoing the whole migration system then the "right" place for it is as a requirement in a BRD.  I don't want to write one tiny part of a BRD and fill in the rest later, so the next best place is is an epic. https://jira.nypl.org/browse/SIMPLY-3875 is that epic.

PR #1603, opened 2021-05-27
Change sample_data to return a bytestrings by default, to flush out discrepancies between mocked and real APIs

leonardr:
## Description

This branch changes the `sample_data` test function to return a bytestring by default. This makes tests of mocked APIs more like tests of the real APIs, since the content of an HTTP response is always a bytestring.

The problems I saw were limited to APIs that serve HTML-like markup (Millenium Patron and Bibliotheca). This is because JSON-like markup from APIs always gets run through `json.loads`, which always returns Unicode.

## Motivation and Context

This came out of our Python 3 QA process. The Millenium Patron API bug was found on the QA server; the Bibliotheca bug was found as a result of changing the test code.

## Checklist:

### DIFF
--- a/api/bibliotheca.py
+++ b/api/bibliotheca.py
@@ -1136,6 +1136,10 @@ class DateResponseParser(BibliothecaParser):
 
     def process_all(self, string):
         parser = etree.XMLParser()
+        # If the data is an HTTP response, it is a bytestring and
+        # must be converted before it is parsed.
+        if isinstance(string, bytes):
+            string = string.decode("utf-8")
         root = etree.parse(StringIO(string), parser)
         m = root.xpath("/%s/%s" % (self.RESULT_TAG_NAME, self.DATE_TAG_NAME))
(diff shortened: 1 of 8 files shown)


api/bibliotheca.py
```diff
@@ -1136,6 +1136,10 @@ class DateResponseParser(BibliothecaParser):
 
     def process_all(self, string):
         parser = etree.XMLParser()
+        # If the data is an HTTP response, it is a bytestring and
+        # must be converted before it is parsed.
+        if isinstance(string, bytes):
+            string = string.decode("utf-8")
         root = etree.parse(StringIO(string), parser)
```
leonardr, line 1143: Presumably we could also use BytesIO instead; I copied this code from elsewhere in the file.

nballenger (approved):
Looks good to me! (Though of course my knowledge of the circ manager is still limited.)

PR #1606, opened 2021-06-02
Convert the decrypted shared secret to Unicode before storing it in the database

leonardr:
## Description

This branch fixes bugs found during in-person testing of qa-circulation.librarysimplified.org using the SimplyE mobile app. The corresponding core PR  is [#1262](https://github.com/NYPL-Simplified/server_core/pull/1262).

## Motivation and Context

* The shared secret between circulation manager and library registry is decrypted as a bytestring, but it's stored in ConfigurationSettings.value, which takes a Unicode string. I fixed this here, and fixed the general problem in core. 

Encryption/decryption is an area I hadn't considered as a source of bytes/Unicode confusion, so I searched through the code for other places we use the `Crypto` library or the `cipher` wrapper class. There are a couple such places but I didn't see any other errors of this type.

## Checklist:

### DIFF
--- a/tests/test_registry.py
+++ b/tests/test_registry.py
@@ -715,6 +715,8 @@ class TestRegistration(DatabaseTest):
         key2 = RSA.generate(2048)
         encryptor2 = PKCS1_OAEP.new(key2)
 
+        # NOTE: In a real case, shared_secret must be UTF-8 string,
+        # but this particular method will work even if it's not.
         shared_secret = os.urandom(24)
         encrypted_secret = base64.b64encode(encryptor.encrypt(shared_secret))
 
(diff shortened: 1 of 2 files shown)


tests/test_registry.py
```diff
@@ -754,7 +756,7 @@ def test__process_registration_result(self):
         class Mock(Registration):
             def _decrypt_shared_secret(self, encryptor, shared_secret):
                 self._decrypt_shared_secret_called_with = (encryptor, shared_secret)
-                return "cleartext"
+                return "👉 cleartext 👈".encode("utf8")
```
leonardr, line 759: We didn't catch this problem earlier because the mock _decrypt_shared_secret returned Unicode, whereas the real method returns a bytestring.

EdwinGuzman (approved):

PR #1608, opened 2021-06-04
Fixed mock tests for the Firstbook API(s) to be more realistic.

leonardr:
## Description

This branch fixes the mock FirstBook APIs so `response.content` is always a bytestring, as it would be in real life. Two points in the code assumed that `response.content` would be a Unicode string; I fixed those points after fixing the mock and verifying it caused the existing tests to fail.

I changed both firstbook2.py and firstbook.py, even though firstbook.py is no longer used.


nballenger (approved):
Looks good to me!

PR #1610, opened 2021-06-09
Merge to OpenEbooks prod deploy branch

nballenger:
Need to do this to deploy Py3 version to OpenEbooks prod, per https://jira.nypl.org/browse/OE-138


leonardr (approved):

nballenger:
@leonardr pointed out we should actually merge `openebooks-deploy-qa` into prod, not `develop`.

PR #1612, opened 2021-06-15
Port the SIP2 client to Python 3

leonardr:
## Description

This branch makes the SIP2 client compatible with Python 3. This required revisiting an underlying assumption in the SIP2 client code that SIP2 messages are ASCII. In point of fact, SIP2 messages are by default encoded using an old MS-DOS encoding called Code Page 850. This is now the default encoding used by our SIP2 client, but I've added a configuration setting for ILSes that use something more modern, like UTF-8.

As often happens with Python 3 work, the problems were revealed when I changed our test mocks to make sure that simulated SIP2 responses came in over the wire as bytestrings rather than Unicode strings. The problems were largely fixed by the usual trick of decoding the bytestring to Unicode immediately upon receipt. But what encoding to use? That's where Code Page 850 comes in.

## Motivation and Context

https://jira.nypl.org/browse/SIMPLY-3804

## How Has This Been Tested?

This was initially based on a reading of the SIP2 spec, but I've gotten access to a test Koha server hosted by Amigos and done basic end-to-end tests against it. Going forward we can hook qa-circulation.librarysimplified.org up to that server and use that to verify basic SIP2 support.

## Checklist:

### DIFF
--- a/api/sip/__init__.py
+++ b/api/sip/__init__.py
@@ -21,6 +21,7 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
     PORT = "port"
     LOCATION_CODE = "location code"
     FIELD_SEPARATOR = "field separator"
+    ENCODING = "encoding"
     USE_SSL = "use_ssl"
     SSL_CERTIFICATE = "ssl_certificate"
     SSL_KEY = "ssl_key"
@@ -33,6 +34,12 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
         { "key": ExternalIntegration.USERNAME, "label": _("Login User ID") },
         { "key": ExternalIntegration.PASSWORD, "label": _("Login Password") },
         { "key": LOCATION_CODE, "label": _("Location Code") },
+        {
+            "key": ENCODING,
+            "label": _("Data encoding"),
+            "default": "cp850",
+            "description": _("By default, SIP2 servers encode outgoing data using the Code Page 850 encoding, but some ILSes allow some other encoding to be used, usually UTF-8."),
+        },
         { "key": USE_SSL, "label": _("Connect over SSL?"),
           "description": _("Some SIP2 servers require or allow clients to connect securely over SSL. Other servers don't support SSL, and require clients to use an ordinary socket connection."),
           "type": "select",
@@ -148,18 +155,33 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
         else:
             self.fields_that_deny_borrowing = []
 
+    @property
+    def _default_client(self):
+        """Initialize a SIPClient object using the default settings.
+
+        :return: A SIPClient
+        """
+        return SIPClient(
+            target_server=self.server,
+            target_port=self.port,
+            login_user_id=self.login_user_id,
+            login_password=self.login_password,
+            location_code=self.location_code,
+            institution_id=self.institution_id,
+            separator=self.field_separator,
+            use_ssl=self.use_ssl,
+            ssl_cert=self.ssl_cert,
+            ssl_key=self.ssl_key,
+            encoding=self.encoding.lower(),
+            dialect=self.dialect
+        )
+
     def patron_information(self, username, password):
         try:
             if self.client:
                 sip = self.client
             else:
-                sip = SIPClient(
-                    target_server=self.server, target_port=self.port,
-                    login_user_id=self.login_user_id, login_password=self.login_password,
-                    location_code=self.location_code, institution_id=self.institution_id, separator=self.field_separator,
-                    use_ssl=self.use_ssl, ssl_cert=self.ssl_cert, ssl_key=self.ssl_key,
-                    dialect=self.dialect
-                )
+                sip = self._default_client
             sip.connect()
             sip.login()
             info = sip.patron_information(username, password)
@@ -200,13 +222,7 @@ class SIP2AuthenticationProvider(BasicAuthenticationProvider):
         if self.client:
             sip = self.client
         else:
-            sip = SIPClient(
-                target_server=self.server, target_port=self.port,
-                login_user_id=self.login_user_id, login_password=self.login_password,
-                location_code=self.location_code, institution_id=self.institution_id, separator=self.field_separator,
-                use_ssl=self.use_ssl, ssl_cert=self.ssl_cert, ssl_key=self.ssl_key,
-                dialect=self.dialect
-            )
+            sip = self._default_client
 
         connection = self.run_test(
             ("Test Connection"),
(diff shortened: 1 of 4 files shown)


api/sip/client.py
```diff
@@ -852,6 +858,10 @@ def __init__(self, login_user_id=None, login_password=None, separator="|",
         self.status = []
 
     def queue_response(self, response):
+        if isinstance(response, str):
+            # Make sure responses come in as bytestrings, as they would
+            # in real life.
+            response = response.encode("cp850")
```
leonardr, line 864: This is the change that exposed all the problems fixed by this branch.

api/sip/__init__.py
```diff
@@ -148,18 +163,35 @@ def __init__(self, library, integration, analytics=None, client=None, connect=Tr
         else:
             self.fields_that_deny_borrowing = []
 
+    @property
+    def _client(self):
+        """Initialize a SIPClient object using the default settings.
```
leonardr, line 168: This code was duplicated within this class so I moved it to a helper.

api/sip/client.py
```diff
@@ -622,6 +626,8 @@ def parse_response(self, data, expect_status_code, *fields):
 
         :param return: A dictionary containing the parsed-out information.
         """
+        data = data.decode(self.encoding)
+
```
leonardr, line 630: Here's the essential Python 3 fix.

api/sip/client.py
```diff
@@ -764,7 +770,7 @@ def summary(self, hold_items=False, overdue_items=False,
     def send(self, data):
         """Send a message over the socket and update the sequence index."""
         data = data + '\r'
-        return self.do_send(data)
+        return self.do_send(data.encode(self.encoding))
```
leonardr, line 773: And the corresponding fix on the outbound side.

api/sip/client.py
```diff
@@ -838,12 +844,12 @@ class MockSIPClient(SIPClient):
...
-        super(MockSIPClient, self).__init__(
-            None, None, login_user_id=login_user_id,
-            login_password=login_password, separator=separator, institution_id=institution_id
-        )
+    def __init__(self, **kwargs):
+        # Override any settings that might cause us to actually
+        # make requests.
+        kwargs['target_server'] = None
+        kwargs['target_port'] = None
+        super(MockSIPClient, self).__init__(**kwargs)
```
leonardr, line 852: This is doing the same thing as it was before but the reasoning is spelled out.

api/sip/client.py
```diff
@@ -874,3 +884,20 @@ def read_message(self, max_size=1024*1024):
...
+    every simulated server interaction, making it impossible to queue
+    responses or look at the results.
+    """
+    def __init__(self):
+        self.client = None
+
+    def __call__(self, **kwargs):
+        if self.client is None:
+            self.client = MockSIPClient(**kwargs)
+        return self.client
```
leonardr, line 903: What we had before worked, but it's one of those situations where I'm not really sure how. This makes it really easy to write a test where you set up an `ExternalIntegration` with a certain configuration and then pass it into a `SIP2AuthenticationProvider`. With `MockSIPClientFactory` you can then run tests on a mocked-up `SIPClient` that is a) persistent across requests (something we wouldn't want for a real client) and b) is instantiated with the `ExternalIntegration` settings you configured at the start of the test.

tests/sip/test_authentication_provider.py
```diff
@@ -17,30 +21,32 @@ class TestSIP2AuthenticationProvider(DatabaseTest):
     # starting point the actual (albeit redacted) SIP2 messages we
     # receive from servers.
 
-    sierra_valid_login = "64              000201610210000142637000000000000000000000000AOnypl |AA12345|AESHELDON, ALICE|BZ0030|CA0050|CB0050|BLY|CQY|BV0|CC15.00|BEfoo@example.com|AY1AZD1B7"
-    sierra_excessive_fines = "64              000201610210000142637000000000000000000000000AOnypl |AA12345|AESHELDON, ALICE|BZ0030|CA0050|CB0050|BLY|CQY|BV20.00|CC15.00|BEfoo@example.com|AY1AZD1B7"
-    sierra_invalid_login = "64Y  YYYYYYYYYYY000201610210000142725000000000000000000000000AOnypl |AA12345|AESHELDON, ALICE|BZ0030|CA0050|CB0050|BLY|CQN|BV0|CC15.00|BEfoo@example.com|AFInvalid PIN entered.  Please try again or see a staff member for assistance.|AFThere are unresolved issues with your account.  Please see a staff member for assistance.|AY1AZ91A8"
+    sierra_valid_login_unicode = "64              000201610210000142637000000000000000000000000AOnypl |AA12345|AELE CARRÉ, JOHN|BZ0030|CA0050|CB0050|BLY|CQY|BV0|CC15.00|BEfoo@example.com|AY1AZD1B7"
+    sierra_valid_login = sierra_valid_login_unicode.encode("cp850")
+    sierra_valid_login_utf8 = sierra_valid_login_unicode.encode("utf-8")
```
leonardr, line 26: The other changes are just turning ASCII Unicode strings to ASCII bytestrings. For this one, I changed the patron name to include a non-ASCII character, and encoded it both as Code Page 850 and UTF-8. This is used to test server setups with different encodings.

leonardr:
Attn @alsrlw

api/sip/client.py
```diff
@@ -779,13 +785,13 @@ def read_message(self, max_size=1024*1024):
         A SIP2 message ends with a \\r character.
         """
         done = False
-        data = ""
+        data = b""
```
leonardr, line 788: This method contains the other essential Py3 fixes. This method wasn't previously tested at all so I added coverage.

tests/sip/test_client.py
```diff
@@ -14,17 +14,29 @@
...
 
     def settimeout(self, value):
         self.timeout = value
 
+    def recv(self, size):
+        block = self.data[:size]
+        self.data = self.data[size:]
+        return block
+
+
```
leonardr, line 39: This makes `MockSocket` more like a real socket, allowing us to add test coverage to `read_message`.

nballenger (approved):
Approving because I don't see anything that's a blocker, but added a few inline comments that may be worth exploring.

leonardr:
Using more constants is a good idea but I'm not comfortable guessing at the encoding. It's true that failure will be hard to diagnose because it will only manifest for patrons with certain names -- all pre-4.0.0 versions of the CM are in that failure mode now and no one has complained -- but with no 100% reliable way of guessing I'd rather not guess.

I went looking to see if anyone even uses an encoding other than CP850, and according to [this page](https://wiki.koha-community.org/wiki/SIP2_configuration), UTF-8 is the default SIP2 encoding for the Koha ILS (although I tested on a Koha ILS and it uses CP850).

PR #1613, opened 2021-06-16
If a borrow attempt falls afoul of a vendor-specific loan limit, try placing the book on hold before giving up.

leonardr:
## Description

It's possible for `CirculationAPI.borrow` to fail at the "take out a loan" stage for a book that can't be loaned out at all. When it looks like that might be happening, we store the exception we got at the "borrow" stage and proceed to the "place a hold" stage.

If "place a hold" succeeds, there's no problem. If "place a hold" fails, we re-raise the original exception.

The specific case we've seen this is when a vendor enforces a loan limit before checking whether or not the requested book is actually loanable.

## Motivation and Context

https://jira.nypl.org/browse/SIMPLY-3750

## How Has This Been Tested?

I used SimplyE to borrow up to my limit of Axis 360 titles. Then I ran this code against NYPL's QA database, and attempted to borrow a fourth Axis 360 title that I knew was unavailable. Without this code, the fourth borrow attempt failed with a problem detail. With this code, the fourth borrow attempts manages to put the title on hold.

## Checklist:

### DIFF
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -627,15 +627,30 @@ class CirculationAPI(object):
 
         new_loan = False
 
+        # Some exceptions may be raised during the borrow process even
+        # if the book is not actually available for loan.  In those
+        # cases, we will store the exception here and try to place the
+        # book on hold. If the hold placement succeeds, there's no
+        # problem. If the hold placement fails because the book is
+        # actually available, it's better to raise this exception than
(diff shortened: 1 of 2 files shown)


leonardr:
@EdwinGuzman Can you take a look at this once you're done with the latest round of card creator work, maybe Friday?

EdwinGuzman (approved):

PR #1616, opened 2021-06-24
Fix a timezone conversion problem in the ODL (DPLA Exchange) client

leonardr:
## Description

I found this while doing a live-integration test of the ODL integration, using a tiny catalog on qa-circulation.librarysimplified.org that DPLA Exchange gave to NYPL so we could implement a player for their audiobooks.

When we create a loan, we tell the ODL server what expiration date we would like for the loan. Previously this expiration date was a timestamp with no explicit timezone, which we indicated by sticking a 'Z' on to the end. Now it's explicitly a timestamp in the UTC timezone and sticking a 'Z' on to the end makes no sense. Sending a timestamp that ends with "+00:00Z" to the ODL server results in an error.

## Motivation and Context

Part of https://jira.nypl.org/browse/SIMPLY-2214

## How Has This Been Tested?

Right now, this curl command will result in a problem detail document:

```
curl https://[user]:[pass]@qa-circulation.librarysimplified.org/DEX/works/URI/https://www.feedbooks.com/item/3264308/borrow
```

I set up this branch to go against the QA server and was able to duplicate the problem and fix it locally. I was then able to check out an audiobook, fulfill its manifest, and return it with these commands:

```
curl http://[user]:[pass]@localhost:6500/DEX/works/URI/https://www.feedbooks.com/item/3264308/borrow
curl http://[user]:[pass]@localhost:6500/DEX/works/403595/fulfill/25
curl http://[user]:[pass]@localhost:6500/DEX/loans/403595/revoke
```

## Checklist:

### DIFF
--- a/tests/test_odl.py
+++ b/tests/test_odl.py
@@ -96,8 +96,15 @@ class TestODLAPI(DatabaseTest, BaseODLTest):
         # Loans expire in 21 days by default.
         now = utc_now()
         after_expiration = now + datetime.timedelta(days=23)
-        expires = urllib.parse.unquote_plus(params.get("expires")[0])
+        expires = urllib.parse.unquote(params.get("expires")[0])
+
+        # The expiration time passed to the server is associated with
+        # the UTC time zone.
+        assert expires.endswith('+00:00')
         expires = dateutil.parser.parse(expires)
+        assert expires.tzinfo == dateutil.tz.tz.tzutc()
+
+        # It's a time in the future, but not _too far_ in the future.
         assert expires > now
         assert expires < after_expiration
 
--- a/api/odl.py
+++ b/api/odl.py
@@ -253,7 +253,7 @@ class ODLAPI(BaseCirculationAPI, BaseSharedCollectionAPI):
                 id=id,
                 checkout_id=checkout_id,
                 patron_id=patron_id,
-                expires=(expires.isoformat() + 'Z'),
+                expires=expires.isoformat(),
                 notification_url=notification_url,
             )
         response = self._get(url)


tests/test_odl.py
```diff
@@ -96,8 +96,15 @@ def test_get_license_status_document_success(self):
         # Loans expire in 21 days by default.
         now = utc_now()
         after_expiration = now + datetime.timedelta(days=23)
-        expires = urllib.parse.unquote_plus(params.get("expires")[0])
+        expires = urllib.parse.unquote(params.get("expires")[0])
+
+        # The expiration time passed to the server is associated with
+        # the UTC time zone.
+        assert expires.endswith('+00:00')
         expires = dateutil.parser.parse(expires)
```
leonardr, line 104: This wasn't catching the error because dateutil.parser.parse doesn't have a problem parsing the technically nonsensical "+00:00Z" syntax; it makes a relatively obvious guess. But the actual DPLA Exchange server refuses to make that guess.

tests/test_odl.py
```diff
@@ -96,8 +96,15 @@ def test_get_license_status_document_success(self):
         # Loans expire in 21 days by default.
         now = utc_now()
         after_expiration = now + datetime.timedelta(days=23)
-        expires = urllib.parse.unquote_plus(params.get("expires")[0])
+        expires = urllib.parse.unquote(params.get("expires")[0])
```
leonardr, line 99: Using unquote_plus destroys the plus sign in "+00:00".

nballenger (approved):
Looks good to me!

PR #1621, opened 2021-07-22
Simply 449/add useful message

oliviawongnyc:
## Description

- Added an error message to be displayed in the browser and in the console if the CSS or JavaScript files can't be located.
- Updated the README with some debugging instructions.

## Motivation and Context

Without this change, if this error occurs, users see a blank page with no feedback.

https://jira.nypl.org/browse/SIMPLY-449

## How Has This Been Tested?

I removed node_modules from the circulation repo and the circulation-web repo, and the dist folder from the latter, as well, to test the change.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,5 @@
 # Library Simplified Circulation Manager
+
 ![Build Status](https://github.com/nypl-simplified/circulation/actions/workflows/test.yml/badge.svg?branch=develop) [![GitHub Release](https://img.shields.io/github/release/nypl-simplified/circulation.svg?style=flat)]()
 
 ## Overview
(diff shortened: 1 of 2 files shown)


api/admin/templates.py
```diff
@@ -18,11 +21,21 @@
...
     });
+    const elementsToRemove = document.getElementsByClassName("error");
+    while(elementsToRemove.length > 0){
+        elementsToRemove[0].parentNode.removeChild(elementsToRemove[0]);
+    }
+  } catch (e) {
+    document.getElementById("error1").innerHTML = "We're having trouble displaying this page."
+    document.getElementById("error2").innerHTML = "Contact your administrator, and ask them to check the console for more information."
+    console.warn("The CSS and/or JavaScript files for this page could not be found. Try running `npm install` in the api/admin directory of the circulation repo.")
+  }
```
leonardr, line 32: Is the exception `e` itself useful here, or are you pretty sure this will only ever happen due to missing files? If it could be a different problem, it would be good to log `e` to the console as well.

README.md
```diff
@@ -252,10 +259,14 @@ tox -e py36-docker -- tests/test_google_analytics_provider.py
 
 Check out the [Docker README](/docker/README.md) in the `/docker` directory for in-depth information on optionally running and developing the Circulation Manager locally with Docker, or for deploying the Circulation Manager with Docker.
 
+## Debugging
+
+If you are served an error message on the admin home screen, you may need to run `npm install` in the api/admin directory of the circulation repo. If you're running the application locally, you can also try running the same command in the root directory of the circulation-web repo.
```
EdwinGuzman, line 264: You should add at the end that the latter command will work if the circulation-web and circulation repos are linked through `npm link`.

leonardr (approved):
This looks fine, I just want to make sure we're not hiding useful information.

api/admin/templates.py
```diff
@@ -4,11 +4,14 @@
 <head>
 <title>Circulation Manager</title>
 <meta name=\"viewport\" content=\"width=device-width, initial-scale=1\">
-<link href=\"/admin/static/circulation-web.css\" rel="stylesheet" />
+<link href=\"/admin/static/circulation-web.css\" rel="stylesheet"/>
 </head>
 <body>
+  <p class="error" id="error1" style="color:#1B7FA7;font-family:sans-serif;font-weight:bold;font-size:x-large;margin-top:50px;margin-left:30px"></p>
+  <p class="error" id="error2" style="color:#1B7FA7;font-family:sans-serif;font-size:medium;margin-top:10px;margin-left:30px"></p>
```
EdwinGuzman, line 11: I mentioned the inline style for this thinking it would be one element. Since there are two elements and most of the rules are the same, you can break it out into a style element:

```
<style>
  .error {
    color: #1B7FA7;
    font-family: sans-serif;
    etc...
  }
</style>
```

api/admin/templates.py
oliviawongnyc: I added a console.error for `e`!

README.md
oliviawongnyc: Done!

api/admin/templates.py
oliviawongnyc: Done!

EdwinGuzman (approved):

PR #1622, opened 2021-07-23
An attempt to borrow a book that's already on loan is treated as success without even going to the remote API.

leonardr:
## Description

This branch makes the "checkout" operation idempotent. If you try to borrow a book and we know you already have the book on loan, we send an OPDS entry that explains how to fulfill the book, rather than sending a problem detail document.

## Motivation and Context

https://jira.nypl.org/browse/SIMPLY-3831

If looks like SimplyE sometimes loses track of a book's fulfillment links, or possibly all information about its current state. When that happens, it may attempt to borrow the book again, either to get a copy of the fulfillment links or because it doesn't know there's already a loan.

This may or may not represent a problem on the client side, but there's no reason we can't accommodate a client that forgets whether a book is on loan.

## How Has This Been Tested?

I set up the new branch against the QA database and tried to "borrow" a book I already had on loan.

```
curl http://leonardr:xxxx@localhost:6500/QANYPL/works/Overdrive%20ID/b95e6c49-0d9d-4670-92b8-1d788f72f730/borrow
```

This got me an OPDS entry with fulfillment links.

Then I tried the same request on the real QA server running the old code:

```
curl https://leonardr:xxxx@qa-circulation.librarysimplified.org/QANYPL/works/Overdrive%20ID/b95e6c49-0d9d-4670-92b8-1d788f72f730/borrow
```

This got me a problem detail document saying that I've already checked out the book.

## Checklist:

### DIFF
--- a/api/controller.py
+++ b/api/controller.py
@@ -1406,6 +1406,9 @@ class LoanController(CirculationManagerController):
         patron = flask.request.patron
         library = flask.request.library
 
+        header = self.authorization_header()
+        credential = self.manager.auth.get_credential_from_header(header)
+
         result = self.best_lendable_pool(
             library, patron, identifier_type, identifier, mechanism_id
         )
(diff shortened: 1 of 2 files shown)


api/controller.py
```diff
@@ -1406,6 +1406,9 @@ def borrow(self, identifier_type, identifier, mechanism_id=None):
         patron = flask.request.patron
         library = flask.request.library
 
+        header = self.authorization_header()
+        credential = self.manager.auth.get_credential_from_header(header)
+
```
leonardr, line 1411: This changes behavior slightly in that if you send an invalid request without authentication, you'll now get an authentication failure rather than an "invalid request" error, but you get an error in either case.

api/controller.py
```diff
@@ -1416,74 +1419,104 @@ def borrow(self, identifier_type, identifier, mechanism_id=None):
...
-        problem_doc = None
+        if isinstance(result, Loan):
+            # We already have a Loan, so there's no need to go to the API.
+            loan_or_hold = result
+            is_new = False
+        else:
+            # We need to actually go out to the API
+            # and try to take out a loan.
+            pool, mechanism = result
+            loan_or_hold, is_new = self._borrow(patron, credential, pool, mechanism)
```
leonardr, line 1434: Code that used to be inline in this method, I moved to a helper method.

api/controller.py
```diff
@@ -1416,74 +1419,104 @@ def borrow(self, identifier_type, identifier, mechanism_id=None):
...
+        # a feed that tells the patron how to fulfill the loan. If a hold,
+        # serve a feed that talks about the hold.
+        response_kwargs = {}
+        if is_new:
+            response_kwargs['status'] = 201
+        else:
+            response_kwargs['status'] = 200
+        return LibraryLoanAndHoldAnnotator.single_item_feed(
+            self.circulation, loan_or_hold, **response_kwargs
+        )
```
leonardr, line 1449: This is the same code as before, but the diff makes it look new because it now comes before the code in the helper method, where it used to come afterwards.

api/controller.py
```diff
@@ -1502,7 +1535,10 @@ def best_lendable_pool(self, library, patron, identifier_type, identifier,
             Loan.patron==patron
         ).all()
         if existing_loans:
-            return ALREADY_CHECKED_OUT
+            # The patron already has at least one loan on this book already.
+            # To make the "borrow" operation idempotent, return one of
+            # those loans instead of an error.
+            return existing_loans[0]
```
leonardr, line 1541: Here's the core of the change. A preexisting loan is no longer an error.

nballenger (approved):
This all makes sense to me, though I'm still rusty on the conceptual pieces around license pools, so could easily have missed something WRT that.

PR #1626, opened 2021-08-17
Simply 3412/make-https-with-local-host-possible

oliviawongnyc:
## Description

- Updated api/app.py run function so that app can be run over https when given the proper scheme.
- Updated the README to include this information.
- Updated the README to make it more clear that the app only works with certain Python versions (per feedback from Mike).

## Motivation and Context

- Originally configured to investigate a redirect issue.
- The above issue will be resolved outside of the app code (in AWS), but it will still be useful to have the ability to run localhost over https in future.

## How Has This Been Tested?

- Running `python app.py` will start the app locally on http.
- Running `python app.py https://localhost:6500/` will start the app locally on https.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,10 +57,6 @@ While you're at it, go ahead and install the following required dependencies:
 - `$ brew install libxmlsec1`
 - `$ brew install libjpeg`
 
-Please note: only certain versions of Python 3 will work with this application. One such version is Python 3.6.5. Check to see which version you currently have installed by running `$ python -V`.
-
-If you're using a version of Python that doesn't work, install [pyenv](https://github.com/pyenv/pyenv-installer) using command `$ curl https://pyenv.run | bash`, then install the version of Python you want to work with, ie `$ pyenv install python3.6.5`, and then run `$ pyenv global 3.6.5`. Check the current version again with `$ python -V` to make sure it's correct before proceeding.
-
 You will need to set up a local virtual environment to install packages and run the project. If you haven't done so before, use pip to install virtualenv – `$ pip install virtualenv` – before creating the virtual environment in the root of the circulation repository:
 
 ```sh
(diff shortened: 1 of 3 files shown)


oliviawongnyc:
@nballenger, can you please take a look at the small adjustment I made in api/app.py and confirm that it won't interfere with the way Docker runs the app? Thank you!

README.md
```diff
@@ -129,13 +131,19 @@ And install the dependencies:
...
+Generally, you can run the application with:
 
 ```sh
 $ python app.py
 ```
 
-And visit `http://localhost:6500/`.
+Then, navigate to `http://localhost:6500/`.
+
+But, if you need to run the application over https, explicitly state that in the command:
```
nballenger, line 142: I think it's worth expanding on this a bit, because I want people to understand that there's a pretty specific context where you would do this, and why it works since it's reliant on the Werkzeug implementation (and `ssl_context` doesn't show up in the Flask docs). Something like:

> **Note on HTTP/S in development:** When deployed, the application should be run behind a secure proxy responsible for SSL termination. While developing locally if it becomes necessary to observe app code serving HTTP/S requests, it is possible to start the Flask/Werkzeug development server with an ad-hoc SSL context (see [`werkzeug.serving.run_simple()`](https://werkzeug.palletsprojects.com/en/2.0.x/serving/#werkzeug.serving.run_simple) for more details).
>
> To have the server listen for HTTP/S requests, supply an `https://` URL on start:
>
> ```sh
> $ python app.py https://localhost:6500/
> ```
>
> Also note that this does not fully replicate secure requests as they would appear on a deployed app instance. In particular the `X-Forwarded-*` headers may be different, since you are hitting the application server directly rather than through one or more proxy layers.

nballenger (changes requested):
Code bit should be fine, but I think the docs need expanding. I left an in-line note.

README.md
EdwinGuzman: This should also mention that the adhoc certs also work because of the `pyopenssl` package.

I was originally concerned about this approach because I thought the docker scripts ran `python app.py ...` internally and that nginx's SSL configs would conflict with this local SSL config. It doesn't look like that is the case and there are no issues with the CLI command, right?

nballenger: Right--the actual app runs in Docker via uWSGI knowing where to look for the app object, not via that `run()` command.

oliviawongnyc: Thanks for taking a look, both! I have updated the README.

nballenger (approved):
Looks good, thanks for making the additions.

PR #1632, opened 2021-09-01
Scrub leading/trailing whitespace from provided credentials.

leonardr:
## Description

This branch fixes https://jira.nypl.org/browse/SIMPLY-3872 for all authentication techniques that use HTTP Basic Auth.

## Checklist:

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -1940,6 +1940,12 @@ class BasicAuthenticationProvider(AuthenticationProvider, HasSelfTests):
             patron
         )
 
+    def scrub_credential(self, value):
+        """Scrub an incoming value that is part of a patron's set of credentials."""
+        if not isinstance(value, (str, bytes)):
+            return value
+        return value.strip()
+
     def authenticate(self, _db, credentials):
         """Turn a set of credentials into a Patron object.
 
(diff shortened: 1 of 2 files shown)


api/authenticator.py
```diff
@@ -1940,6 +1940,12 @@ def _run_self_tests(self, _db):
             patron
         )
 
+    def scrub_credential(self, value):
+        """Scrub an incoming value that is part of a patron's set of credentials."""
+        if not isinstance(value, (str, bytes)):
```
nballenger, line 1945: I'll preface this by saying I don't grok the class inheritance here, so ignore this if it doesn't make sense. That said--if `value` can be something other than a string/bytestring, does it make sense to have this trigger based on whether the passed in object responds to a `.strip()` call, instead of by type? That would let you pass a custom object as the value if necessary. Something like:

```Python
if callable(getattr(value, 'strip', None)):
    return value.strip()
else:
    return value
```

nballenger (approved):
Looks good to me, one question/suggestion in-line that may or may not be useful.

api/authenticator.py
leonardr: Since the data is coming in from the Authorization header, I think the only acceptable values are string, bytestring, and None; but I don't want a crash if I'm wrong, and this case does handle None.

PR #1636, opened 2021-09-03
Odl palace update

leonardr:
## Description

This branch brings in recent improvements to ODL collection management originated by @vbessonov in https://github.com/ThePalaceProject/circulation/pull/25. It still needs some adaptation but the basic code should be in place now.

## Motivation and Context

This is per https://jira.nypl.org/browse/SIMPLY-3868

## Checklist:


api/odl.py
```diff
@@ -596,7 +586,7 @@ def update_hold_queue(self, licensepool):
                 Loan.end>utc_now()
             )
         ).count()
-        remaining_licenses = licensepool.licenses_owned - loans_count
+        remaining_licenses = max(licensepool.licenses_owned - loans_count, 0)
```
leonardr, line 589: This stops the number of licenses from going below zero.

api/odl.py
```diff
@@ -887,8 +878,13 @@ def _detail_for_elementtree_entry(cls, parser, entry_tag, feed_url=None, do_get=
...
                 concurrent_checkouts = subtag(terms[0], "odl:concurrent_checkouts")
                 expires = subtag(terms[0], "odl:expires")
+
                 if expires:
-                    expires = dateutil.parser.parse(expires)
+                    expires = util.datetime_helpers.to_utc(dateutil.parser.parse(expires))
+                    now = util.datetime_helpers.utc_now()
+
+                    if expires <= now:
+                            continue
```
leonardr, line 887: First, this makes sure `expires` is a timezone-aware datetime. Second, this skips any books that were already expired the first time we saw them.

api/odl.py
```diff
@@ -959,11 +956,12 @@ def run_once(self, progress):
         progress = TimestampData(achievements=message)
         return progress
 
+
 class MockODLAPI(ODLAPI):
     """Mock API for tests that overrides _get and _url_for and tracks requests."""
 
     @classmethod
-    def mock_collection(self, _db):
+    def mock_collection(cls, _db, protocol=ODLAPI.NAME):
```
leonardr, line 964: We don't need this right now -- it's for the ODL-via-OPDS-2 code that I didn't bring in -- but it doesn't hurt to have it.

api/odl.py
```diff
@@ -1040,6 +1038,25 @@ def __init__(self, _db, collection):
...
+        """
+        response_content = response.content
+
+        if not isinstance(response_content, (str, bytes)):
+            raise ValueError("Response content must be a string or byte-encoded value")
+
+        feed = feedparser.parse(response_content)
+
+        return feed
+
```
leonardr, line 1059: This just adds a little bit of error checking to the feed parsing process.

api/odl.py
```diff
@@ -1518,3 +1539,38 @@ def _get(self, url, patron=None, headers=None, allowed_response_codes=None):
...
+    def process_item(self, identifier):
+        for licensepool in identifier.licensed_through:
+            licenses_owned = licensepool.licenses_owned
+            licenses_available = licensepool.licenses_available
+
+            for license in licensepool.licenses:
+                if license.is_expired:
+                    licenses_owned -= 1
+                    licenses_available -= 1
```
leonardr, line 1561: Setting this to -1 instead of 0 ensures that if licenses_owned drops to zero and _then_ the license expires, the expiration is treated as a distinct event. However, I'm not sure there will actually be a real difference in terms of which circulation events are issued.

leonardr:
I ran this in the context of NYPL's QA server (where we happen to have a test ODL collection with some expired licenses), and watching it run made me concerned that the reaper algorithm only works in simple cases.
https://github.com/ThePalaceProject/circulation/pull/25/files#r703719081

I'm waiting to hear back before proceeding with this PR.

PR #1640, opened 2021-09-24
Simply 3907 reindex lists and lanes

nballenger:
## Description

We need to reindex works that caused a membership change in custom lists, so that the upstream counts of those lists and the lanes they populate (which are pulled from the search index) are correct when recalculated.

The relevant changes to review are in 3d6b707--the changeset in d3dea3d is formatting related.

### DIFF
--- a/api/admin/controller/__init__.py
+++ b/api/admin/controller/__init__.py
@@ -9,32 +9,36 @@ from datetime import date, datetime, timedelta
 
 import flask
 import jwt
-from flask import (
-    Response,
-    redirect,
-)
-from flask_babel import lazy_gettext as _
+from flask import Response, redirect
+from flask_babel import lazy_gettext as lgt
 from sqlalchemy.sql import func
-from sqlalchemy.sql.expression import desc, nullslast, and_, distinct, select, join
-from api.admin.exceptions import *
+from sqlalchemy.sql.expression import (and_, desc, distinct, join, nullslast, select)
+
+from api.admin.exceptions import (
+    ADMIN_AUTH_MECHANISM_NOT_CONFIGURED, ADMIN_AUTH_NOT_CONFIGURED,
+    CANNOT_CHANGE_LIBRARY_FOR_CUSTOM_LIST, CANNOT_CHANGE_PROTOCOL,
+    CANNOT_EDIT_DEFAULT_LANE, CANNOT_SHOW_LANE_WITH_HIDDEN_PARENT,
+    COLLECTION_NOT_ASSOCIATED_WITH_LIBRARY, CUSTOM_LIST_NAME_ALREADY_IN_USE,
+    DUPLICATE_INTEGRATION, INCOMPLETE_CONFIGURATION,
+    INTEGRATION_NAME_ALREADY_IN_USE, INTEGRATION_URL_ALREADY_IN_USE,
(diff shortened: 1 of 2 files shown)


leonardr (changes requested):
It might be possible to fix your failed test  by accepting the new value and changing the comments. I'm not sure, since the search index is mocked to always say there are 2 items in a lane. "4" might be the actual right answer but I don't see how your change modifies the mock data.

The failure of that test also shows that it's pretty easy to write a unit test for this code: add some works to a list, remove some works, and verify that each of those works was immediately reindexed.

api/admin/controller/__init__.py
```diff
@@ -937,13 +937,16 @@ def _create_or_update_list(self, library, name, entries, collections, deletedEnt
         list.name = name
         membership_change = False
 
+        works_to_update_in_search = []
```
leonardr, line 940: It doesn't matter much, but this should be a set.

nballenger: Good call.

nballenger:
The test turned out to be failing because the records in the mock search engine were different from the ones added, so they were adding to the count. I set it to start with an empty list of documents and now the count is correct again. Also switched over to a set for the iterable of works to reindex.

leonardr:
I don't think it's sufficient anymore to simply check how many items are in the search index. You need to look inside the index and verify that the specific books that were added and deleted got reindexed. But I think putting that check in the test you're already changing should be fine.

nballenger:
@leonardr - I've added assertions in 4058de5 to determine which items in the search engine are associated with the list, and those are working. Unfortunately now the test fails here:

https://github.com/NYPL-Simplified/circulation/blob/4058de570ae1f3d5d675918090a6cab6ce66322f/tests/admin/controller/test_controller.py#L1223-L1225

After the works are updated in the index, there's a call to update the size of lanes affected by that custom list:

https://github.com/NYPL-Simplified/circulation/blob/4058de570ae1f3d5d675918090a6cab6ce66322f/api/admin/controller/__init__.py#L962-L969

But right now it's returning 3. I looked at `Lane.update_size()`, but I don't understand the facet filter it's using well enough to know why it's returning all three works, instead of just those associated with the custom list:

https://github.com/NYPL-Simplified/server_core/blob/1d39c85f98cd8d75a0357937b44318dace52f535/lane.py#L2825-L2842

I dropped into the debugger there, and it definitely does pull a count of 3. Do you see any obvious answer?

leonardr:
You're running this against a mocked ExternalSearchIndex, not a real Elasticsearch index (which would make the test very slow). I bet at some point you're calling a mocked method that just returns the total number of items in the index instead of simulating a real Elasticsearch query.

If I'm right, there are a couple solutions. Instead of comparing lane.size to a specific value you can verify that Lane.update_size was called. You can also change the mock to make it return a specific value when asked for the size of a given lane, instead of returning the total number of items in the index.

nballenger:
Updated so that the lane size check shows that `update_size()` got called.

nballenger:
Tests passed for 58ec1a8, and bc5e533 just removes `, monkeypatch` from the test method signature. I stopped the tests manually--should still be fine from the previous run.

leonardr (approved):

PR #1642, opened 2021-09-24
SIMPLY-3919 Infrastructure Enabler - Improve Instructions on Local Development with Docker

mattkoch614:
## Description
Re-organizes the project README to introduce a table of contents and include steps for developing locally with Docker and Docker Compose. Almost no changes were made to the actual Docker / Docker Compose configuration.

## Motivation and Context

The hypothesized benefits are that these instructions will make it easier to get started and that onboarding new developers will take less time.

## How Has This Been Tested?

1. Ran the following commands to spin up Docker containers:
```bash
cd circulation/docker
docker compose build
docker compose up -d
```
2. Ran `docker ps -a` to confirm all containers were up and running.
3. Navigated to http://localhost:80/admin to confirm the site loads.
4. Followed [these](https://docs.google.com/document/d/1a0QNWTvt9NKChr8TcLJ3G1PfpBCKV6hC2JuB5Yk_CYY/edit?usp=sharing) instructions for adding test data such as a test library, search service, and books to a catalog.

## Checklist:
- [X ] I have updated the documentation accordingly.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,3 @@
+We welcome your contributions to new features, fixes, or updates, large or small; we are always thrilled to receive pull requests, and do our best to process them as fast as we can.
+
+Before you start to code, we recommend discussing your plans through a [GitHub issue](https://github.com/NYPL-Simplified/circulation/issues/new), especially for more ambitious contributions. This gives other contributors a chance to point you in the right direction, give you feedback on your design, and help you find out if someone else is working on the same thing.
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,3 +1,3 @@
-We welcome your contributions to new features, fixes, or updates, large or small; we are always thrilled to receive pull requests, and do our best to process them as fast as we can.
+## Contribution Guidelines
 
-Before you start to code, we recommend discussing your plans through a [GitHub issue](https://github.com/NYPL-Simplified/circulation/issues/new), especially for more ambitious contributions. This gives other contributors a chance to point you in the right direction, give you feedback on your design, and help you find out if someone else is working on the same thing.
+Our contribution guidelines are currently under construction. Please see https://librarysimplified.org/community/ for more information.
--- a/README.md
+++ b/README.md
@@ -95,7 +95,7 @@ Finally, open the administration panel at http://localhost:80/admin to view the 
 ---
 #### **Python Set Up**
 
-If you do not have Python 3 installed, you can use [Homebrew](https://brew.sh/)\* to install it by running the command `$ brew install python3`.
+If you do not have Python 3 installed, and you are running MacOS or Linux, you can use [Homebrew](https://brew.sh/)\* to install it by running the command `$ brew install python3`.
 
 \*If you do not yet have Homebrew, you can install it by running the following:
 
@@ -335,7 +335,7 @@ The latter command will only work if the circulation-web and circulation repos a
 
 ## Contributing
 
-Contributions are always welcome! See [here](CONTRIBUTING.md) for ways to get started.
+See [here](CONTRIBUTING.md) for more information.
 
 ## License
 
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -10,12 +26,6 @@ This is the Circulation Manager for [Library Simplified](https://www.librarysimp
 
 It depends on [Library Simplified Server Core](https://github.com/NYPL-Simplified/server_core) as a git submodule.
 
-## Installation
```
mattkoch614, line 13: This section was removed in favor of linking directly to the NYPL Simplified Wiki in the Appendix.

oliviawongnyc (commented):
Looks great to me! Thanks, Matt. Will let Leonard/Nick give the approval.

CONTRIBUTING.md
```diff
@@ -0,0 +1,3 @@
+We welcome your contributions to new features, fixes, or updates, large or small; we are always thrilled to receive pull requests, and do our best to process them as fast as we can.
+
+Before you start to code, we recommend discussing your plans through a [GitHub issue](https://github.com/NYPL-Simplified/circulation/issues/new), especially for more ambitious contributions. This gives other contributors a chance to point you in the right direction, give you feedback on your design, and help you find out if someone else is working on the same thing.
```
leonardr, line 3: It would be better to remove this than move it, since it doesn't accurately represent how we do development right now.

docker/docker-compose.yml
```diff
@@ -12,7 +12,17 @@ services:
...
+    ports:
+      - "9200:9200"
+      - "9300:9300"
+    environment:
+      - discovery.type=single-node
+      - bootstrap.memory_lock=true
+      - "ES_JAVA_OPTS=-Xms512m -Xmx512m"
+    command: >
+      /bin/sh -c "./bin/elasticsearch-plugin list | grep -q analysis-icu 
+      || ./bin/elasticsearch-plugin install analysis-icu; tail -f /dev/null"
```
leonardr, line 25: Attn @nballenger, this was a necessary step for making real use of the `es` image.

CONTRIBUTING.md
mattkoch614: If we want to make changes to the language for accepting open source contributions, that seems appropriate to better reflect our expectations. Since this is an open source repo, anyone can still submit a PR, even though the NYPL has its own roadmap - maybe even indicating that and a few extra notes about how we plan to triage contributions would be enough.

docker/docker-compose.yml
nballenger: It's possible this command is enough by itself, if we just add a volume or two to persist the install. I pulled that ES image, and it looks like the entire install is contained under `/usr/share/elasticsearch`. If you set that as a volume it would keep the entire install across container runs. Alternatively you could add one or two volumes for `/usr/share/elasticsearch/plugins` and `/usr/share/elasticsearch/data`, if you wanted to just persist those bits.

nballenger (approved):
Looks good to me, a few suggestions in-line to take or leave.

README.md
```diff
@@ -39,9 +49,51 @@ There are additional protected branches that are used for _NYPL-specific_ deploy
...
+  docker compose up -d
+```
+
+Confirm all of the containers are running successfully:
+
+```bash
+  docker ps -a
+```
+
+Finally, open the administration panel at http://localhost:80/admin to view the Circulation Manager Administration app.
```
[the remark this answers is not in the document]
mattkoch614: You're right, but I really only added it to distinguish between the "manual" instructions which have the user use port 6500.

README.md
```diff
@@ -39,9 +49,51 @@ There are additional protected branches that are used for _NYPL-specific_ deploy
...
+  docker ps -a
+```
+
+Finally, open the administration panel at http://localhost:80/admin to view the Circulation Manager Administration app. 
+
+### **Manual Setup**
+---
+#### **Python Set Up**
 
 If you do not have Python 3 installed, you can use [Homebrew](https://brew.sh/)\* to install it by running the command `$ brew install python3`.
```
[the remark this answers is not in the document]
mattkoch614: Good idea.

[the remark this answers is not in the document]
leonardr: Our default background assumption is that developers are on a Mac. I certainly wouldn't worry about adding doc support for Windows right now.

PR #1644, opened 2021-09-29
Add volumes for es service in Docker Compose setup

mattkoch614:
## Description

Adds two volumes to the docker compose setup for Elastic Search. 

## Motivation and Context

Based on a [suggestion ](https://github.com/NYPL-Simplified/circulation/pull/1642#discussion_r717693597) by @nballenger


nballenger (approved):
Looks good to me. I was slightly worried about having the entrypoint script as the last bit of the command, since in some cases the pattern is to have the contents of the command get passed to the value of `ENTRYPOINT` as args to execute. I pulled the ES image and ran it though, all seems fine--installs the plugin on first run, has it already on subsequent runs, responds normally on 9200, etc.

PR #1645, opened 2021-09-30
Make use of the new Axis 360 API endpoint for returning titles.

leonardr:
## Motivation and Context

This fixes https://jira.nypl.org/browse/SIMPLY-1742.

## Checklist:

### DIFF
--- a/api/axis.py
+++ b/api/axis.py
@@ -2,6 +2,7 @@ import json
 import logging
 import os
 import re
+import urllib
 
 from collections import defaultdict
 from datetime import datetime, timedelta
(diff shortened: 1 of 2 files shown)


api/axis.py
```diff
@@ -1210,9 +1241,17 @@ def __init__(self, collection):
...
+                                 ignore_error_codes=None):
         """Raise an error if the given lxml node represents an Axis 360 error
         condition.
+
+        :param e: An lxml Element
+        :param ns: A dictionary of namespaces
+        :param custom_error_classes: A dictionary of errors to map to custom
+           classes rather than the defaults.
+        :param ignore_error_codes: A list of error codes to treat as success
+           rather than as cause to raise an exception.
```
leonardr, line 1253: ignore_error_codes is new.

api/axis.py
```diff
@@ -1183,20 +1209,24 @@ class ResponseParser(Axis360Parser):
         (3113, "Title ID is not available for checkout") : NoAvailableCopies,
         3114 : PatronLoanLimitReached,
         3115 : LibraryInvalidInputException, # Missing DRM format
+        3116 : LibraryInvalidInputException, # No patron session ID provided -- we don't use this
```
leonardr, line 1212: These are previously unseen error codes defined in the documentation for the new endpoint.

api/axis.py
```diff
@@ -1253,6 +1295,22 @@ def _raise_exception_on_error(cls, code, message, custom_error_classes={}):
...
+    def process_all(self, string):
+        for i in super(CheckinResponseParser, self).process_all(
+                string, "//axis:EarlyCheckinRestResult", self.NS):
+            return i
+
+    def process_one(self, e, namespaces):
+        """Either raise an appropriate exception, or do nothing.
+        """
+        self.raise_exception_on_error(
+            e, namespaces, ignore_error_codes = [4058]
```
leonardr, line 1309: Most of the time, "No checkout is associated with patron for the title" is an error. But when returning a book, we can just claim that the return has succeeded. This makes the return operation idempotent.

leonardr:
It's going to be at least 6 weeks until B&T makes their API changes, so let's get this reviewed and deployed now, and do an upgrade later.

nballenger (approved):
This all looks good to me, though I admittedly don't have a ton of context with the Axis API. Only thing that jumped out at me was that sometimes `code_to_exception` has a numeric key and sometimes a tuple key, but it works in context in the method doing the lookup.

PR #1648, opened 2021-10-15
Update from Palace for ODL reaper improvements

leonardr:
## Description

This branch copies in code from the Palace project -- specifically these two PRs [1](https://github.com/ThePalaceProject/circulation/pull/39) [2](https://github.com/ThePalaceProject/circulation/pull/25) -- to address https://jira.nypl.org/browse/SIMPLY-3868.

There are two main goals:
1. Add a reaper for ODL licenses, so that books disappear from the collection once all loans have been used up.
2. Change the ODL importer so that books aren't imported in the first place if there aren't actually any licenses.

## Motivation and Context

In my review of the original branch I noticed that the algorithm for determining whether an ODL license _had_ expired only handled the most simple use cases, so Tim and Viacheslav from Lyrasis added some more complex test cases and make them work.

To make this particular PR I just copied the api/odl.py and tests/test_odl.py files over from the second Palace PR. We haven't done any development on those files for a long time, so there should have been no divergence. The Palace repo has a lot of new code in odl2.py and test_odl2.py, but that's for the JSON flavor of ODL based on OPDS 2, which is a new flavor of ODL that we don't currently support.

## Checklist:

### DIFF
--- a/api/odl.py
+++ b/api/odl.py
@@ -1,82 +1,51 @@
 import datetime
-import dateutil
 import json
+import logging
 import uuid
-from flask_babel import lazy_gettext as _
-import urllib.parse
-from collections import defaultdict
-import flask
-from flask import Response
-import feedparser
-from lxml import etree
-from .problem_details import NO_LICENSES
 from io import StringIO
-import re
+from typing import Callable, Optional
+
+import dateutil
+import feedparser
+import flask
+from flask import url_for
+from flask_babel import lazy_gettext as _
+from lxml import etree
+from sqlalchemy.sql.expression import or_
 from uritemplate import URITemplate
 
-from sqlalchemy.sql.expression import or_
-
-from core.opds_import import (
-    OPDSXMLParser,
-    OPDSImporter,
-    OPDSImportMonitor,
-)
-from core.monitor import (
-    CollectionMonitor,
-    TimelineMonitor,
-)
+from core import util
+from core.analytics import Analytics
+from core.metadata_layer import CirculationData, FormatData, LicenseData, TimestampData
 from core.model import (
     Collection,
     ConfigurationSetting,
-    Credential,
     DataSource,
     DeliveryMechanism,
     Edition,
     ExternalIntegration,
     Hold,
     Hyperlink,
-    Identifier,
-    IntegrationClient,
     LicensePool,
     Loan,
     MediaTypes,
+    Representation,
     RightsStatus,
     Session,
-    create,
     get_one,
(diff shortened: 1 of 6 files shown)


api/odl.py
```diff
@@ -1518,3 +1608,41 @@ def _get(self, url, patron=None, headers=None, allowed_response_codes=None):
...
+        super(ODLExpiredItemsReaper, self).__init__(_db, collection)
+
+    def process_item(self, identifier):
+        for licensepool in identifier.licensed_through:
+            remaining_checkouts = 0   # total number of checkouts across all the licenses in the pool
+            concurrent_checkouts = 0  # number of concurrent checkouts allowed across all the licenses in the pool
+
+            # 0 is a starting point,
+            # we're going through all the valid licenses in the pool and count up available checkouts.
+            for license_pool_license in licensepool.licenses:
```
leonardr, line 1629: This is the main addition to the second PR. Previously we were assuming that a LicensePool had only a single license.

tests/test_odl.py
```diff
@@ -1925,7 +1932,493 @@ def canonicalize_author_name(self, identifier, working_display_name):
...
 
+        :return: Test ODL feed
+        """
+        env = Environment(
+            loader=FileSystemLoader(self.ODL_TEMPLATE_DIR), autoescape=select_autoescape()
+        )
+        template = env.get_template(self.ODL_TEMPLATE_FILENAME)
+        feed = template.render(licenses=licenses)
+
+        return feed
```
leonardr, line 2079: This uses a Jinja2 template to generate an ODL feed for whatever licenses you want to test. This way you can create the License objects for your use case and generate the feed, without having to pregenerate a separate feed for each use case.

This isn't how we've done this kind of testing up to this point, but it does make it easier to test a large number of  slight variant files.

tests/test_odl.py
```diff
@@ -1925,7 +1932,493 @@ def canonicalize_author_name(self, identifier, working_display_name):
...
+        ),
+        (
+            "total_checkouts_is_zero",
+            TestLicense(total_checkouts=0)
+        ),
+        (
+            "remaining_checkouts_is_zero",
+            TestLicense(total_checkouts=10, concurrent_checkouts=5),
+            TestLicenseInfo(remaining_checkouts=0, available_concurrent_checkouts=0)
+        )
```
leonardr, line 2136: So at this point we can just list all the ways a license might need to be reaped, and test that the reaper takes care of them.

PR #1649, opened 2021-10-18
SIMPLY-3873 Hotfix for production script runner

nballenger:
## Description

The script runner in production is having scripts crash with the error `EOFError: marshal data too short`, which is detailed in [SIMPLY-3873](https://jira.nypl.org/browse/SIMPLY-3873). This changeset will update the script runner's reference to `core` to point at a revision which includes `PYTHONDONTWRITEBYTECODE=1` in the `core/bin/run` script, which will force the script runner not to cache compiled Python files.

## Motivation and Context

See above.

## How Has This Been Tested?

Been running in QA a while, seems to have fixed the script crashes there.

## Checklist:


leonardr:
Looks good. This core change will also fix the "multiple items in heaviest_child" issue which I'd forgotten about and which also causes scripts to crash.

PR #1653, opened 2021-11-01
SIMPLY-3923: Make it possible to do circulation-web development without installing circulation manager locally

samanthaandrews:
## Description

- Creates a new site-wide setting for listing hostnames for admin web application access
- Refactors `allows_patron_web` decorator to check a customizable list of domains
- Make all admin routes allow CORS access from customizable domains

## Motivation and Context

Currently, circulation-web is designed under the assumption that it's making requests against a circulation manager installed on the same hostname. This means any circulation-web development requires installing a local circulation manager. But most circulation-web development doesn't require any development  work on the circulation manager, especially as we focus on refactoring.

It should be possible to set up a local development deployment of circulation-web to talk to an externally hosted circulation manager – most commonly NYPL's QA circulation manager at qa-circulation.librarysimplified.org.

This requires changes to both circulation-web (to make the target URL customizable) and circulation (to set the appropriate CORS headers that make cross-domain requests possible).

https://jira.nypl.org/browse/SIMPLY-3925

## How Has This Been Tested?

I wrote a few unit tests similar to the ones we have for the patron web domains.

Manual testing while running locally:

Screen Shot 2021-11-04 at 2 42 26 PM

Screen Shot 2021-11-04 at 2 42 35 PM

## Checklist:

I have yet to update the documentation, but I believe that when I do I should add it to the [google doc](https://docs.google.com/document/d/1a0QNWTvt9NKChr8TcLJ3G1PfpBCKV6hC2JuB5Yk_CYY/edit) where setting up your circulation manager is outlined.

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -102,80 +102,46 @@ def allows_auth(f):
 # This is mostly taken from the cross_origin decorator in flask_cors, but we
 # can't use that decorator because we aren't able to look up the patron web
 # client url configuration setting at the time we create the decorator.
-# def allows_cors(f, web_domains):
-#     # Override Flask's default behavior and intercept the OPTIONS method for
-#     # every request so CORS headers can be added.
-#     f.required_methods = getattr(f, 'required_methods', set())
-#     f.required_methods.add("OPTIONS")
-#     f.provide_automatic_options = False
-
-#     def wrapped_function(*args, **kwargs):
-#         if request.method == "OPTIONS":
-#             resp = app.make_default_options_response()
-#         else:
-#             resp = make_response(f(*args, **kwargs))
-
-#         if web_domains:
-#             options = get_cors_options(
-#                 app, dict(origins=web_domains,
-#                           supports_credentials=True)
-#             )
-#             set_cors_headers(resp, options)
-
-#         return resp
-#     return update_wrapper(wrapped_function, f)
-
-def allows_admin_web(f):
+def allows_cors(allowed_domain_type):
     # Override Flask's default behavior and intercept the OPTIONS method for
     # every request so CORS headers can be added.
-    f.required_methods = getattr(f, 'required_methods', set())
-    f.required_methods.add("OPTIONS")
-    f.provide_automatic_options = False
+    def decorator(func):
+        @wraps(func)
+        def wrapper(*args, **kwargs):
+            func.required_methods = getattr(func, 'required_methods', set())
+            func.required_methods.add("OPTIONS")
+            func.provide_automatic_options = False
 
-    def wrapped_function(*args, **kwargs):
-        if request.method == "OPTIONS":
-            resp = app.make_default_options_response()
-        else:
-            resp = make_response(f(*args, **kwargs))
-        
-        web_domains = app.manager.admin_web_domains
-        if web_domains:
-            options = get_cors_options(
-                app, dict(origins=web_domains,
-                          supports_credentials=True)
-            )
-            set_cors_headers(resp, options)
+            if request.method == "OPTIONS":
+                resp = app.make_default_options_response()
+            else:
+                resp = make_response(func(*args, **kwargs))
+            
+            patron_web_domains = app.manager.patron_web_domains
+            admin_web_domains = app.manager.admin_web_domains
+            
+            if "patron" in allowed_domain_type:
+                web_domains = " ,".join(admin_web_domains.union(patron_web_domains))
+            else:
+                web_domains = admin_web_domains
 
-        return resp
-    return update_wrapper(wrapped_function, f)
+            if web_domains:
+                options = get_cors_options(
+                    app, dict(origins=web_domains,
+                            supports_credentials=True)
+                )
+                set_cors_headers(resp, options)
 
-def allows_patron_and_admin_web(f):
-    # Override Flask's default behavior and intercept the OPTIONS method for
-    # every request so CORS headers can be added.
-    f.required_methods = getattr(f, 'required_methods', set())
-    f.required_methods.add("OPTIONS")
-    f.provide_automatic_options = False
+            return resp
 
-    def wrapped_function(*args, **kwargs):
-        if request.method == "OPTIONS":
-            resp = app.make_default_options_response()
-        else:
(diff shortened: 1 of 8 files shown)


api/routes.py
```diff
@@ -126,9 +126,34 @@ def wrapped_function(*args, **kwargs):
...
+            )
+            set_cors_headers(resp, options)
+
+        return resp
+    return update_wrapper(wrapped_function, f)
+
 h = ErrorHandler(app, app.config['DEBUG'])
 @app.errorhandler(Exception)
 @allows_patron_web
+@allows_admin_web
```
samanthaandrews, line 156: Are my decorators in the right place? I understand that they run in inverse order... generally I put them underneath the `@allows_patron_web` decorator or below the `@returns_problem_detail` decorator.

api/routes.py
```diff
@@ -126,9 +126,34 @@ def wrapped_function(*args, **kwargs):
...
+        admin_web_domains = app.manager.admin_web_domains
+        if admin_web_domains:
+            options = get_cors_options(
+                app, dict(origins=admin_web_domains,
+                          supports_credentials=True)
+            )
+            set_cors_headers(resp, options)
+
+        return resp
+    return update_wrapper(wrapped_function, f)
```
samanthaandrews, line 151: This function is a copy + paste of `allows_patron_web` function right above it... is there a way to not duplicate this code? I tried doing as Leonard suggested, where I rename the function to `allows_cors()` and declare two variables as so:

```
allows_patron_web = allows_cors(f, app.manager.patron_web_domains)
allows_admin_web = allows_cors(f, app.manager.admin_web_domains)
```
but that didn't seem to work when I used `allows_patron_web` or `allows_admin_web` as a decorator. i received errors that `f` was not defined.

leonardr: Yes, what I wrote was wrong. `f` is the function you're decorating and it's not available until later. Try this slight variation:

```
def allows_patron_web(f):
    return allows_cors(f, app.manager.patron_web_domains)

def allows_admin_web(f):
    return allows_cors(f, app.manager.admin_web_domains)
````

File: api/routes.py (hunk 1 above)
Comment (leonardr) [reply api/routes.py:156]: I was about to say that for this particular decorator the order doesn't matter much since the decorator either does nothing, or it sets an HTTP header.

But this question made me realize that @allows_patron_web sets the same header, potentially overriding it! That's a problem. I looked at [set_cors_headers](https://github.com/corydolphin/flask-cors/blob/master/flask_cors/core.py#L233) and it looks like you can only call it once per request. I think the second call will do nothing.

You should be able to verify this by using `curl` to request a URL that uses both decorators, like `/`. It should also be possible to make the tests in test_routes.py make assertions about what the decorators did to a request. Try writing a test that calls one of these methods and looks at the response headers -- you should get a real Response object back and it should have been run through the decorators.

I think the solution is to define a third decorator, @allows_patron_and_admin_web, which combines both sets of domains and uses them to call set_cors_headers once. Whenever you would need to use both decorators, use that third one instead.

File: api/routes.py (hunk 2 above)
Comment (samanthaandrews) [reply api/routes.py:151]: Hmm this solved the `f is not defined` issue but now it is saying that it app doesn't have a manager property... do i need to wrap the function in order to get access to the app.manager?

File: api/routes.py (hunk 1 above)
Comment (samanthaandrews) [reply api/routes.py:156]: You're right. I was able to write a test showing that my cors headers were being overwritten. I have a fix that I'll push up soon.

File: api/routes.py (hunk 2 above)
Comment (leonardr) [reply api/routes.py:151]: I don't know if this will get you all the way to a solution but I think it should help:

app.manager is set in api/routes.py in a function decorated with @app.before_first_request. For some reason, the allows_cors call is running before that function runs. It looks like your decorators are being called much earlier than I thought they would be.

Looking at the other decorator definitions in api/routes.py, I think the secret might be the @wraps decorator used _inside_ the definition of the main decorator. I think @wraps will make your decorators return a function which gets called much later, during an actual request -- by that time app.manager will be defined.

Comment (samanthaandrews) [reply api/routes.py]: Ok I pushed up a change to this decorator so that it can be re-used as `allows_cors(allowed_domain_type=set({"patron", "admin"}))`. @mwbenowitz let me know if this is what you had in mind. I think it is an improvement.

**Comment by leonardr** [comment]
I really like the idea of keeping it at one decorator but adding an argument.

**Comment by samanthaandrews** [comment]
Giving this a little nudge. Should I schedule time to review together with @nballenger and @leonardr and Keith?

**Comment by leonardr** [comment]
Thanks for the nudge. I'll go through this again just to make sure but I'm pretty confident it's ready to merge.

File: api/routes.py [hunk 3]
```diff
@@ -95,40 +95,53 @@ def decorated(*args, **kwargs):
...
+            else:
+                resp = make_response(func(*args, **kwargs))
+            
+            patron_web_domains = app.manager.patron_web_domains
+            admin_web_domains = app.manager.admin_web_domains
+            
+            if "patron" in allowed_domain_type:
+                web_domains = " ,".join(admin_web_domains.union(patron_web_domains))
+            else:
+                web_domains = admin_web_domains
```
leonardr, line 126: Since you treat `allowed_domain_type` as a collection, it looks as though calling this code will set the patron web domains but not the admin web domains:

```
@allows_cors(allowed_domain_type=set({"patron"}))
```

But actually, both sets of domains will be allowed.

Since your calls to @allows_cors are already explicit about which domains are allowed, I think it's best to make this method explicit as well. Add the patron_web_domains to a set if 'patron' is one of the allowed types, and then add the admin_web_domains to that set if 'admin' is one of the allowed types.

Apart from this minor issue I think this looks good and is ready to merge.

samanthaandrews: Ah, yes, this should be made more clear. I pushed up a change that I think is better. Let me know if this is what you were thinking.

leonardr: This looks great but your commit also introduced some Dockerfile changes which should go into another branch (from Nick?). Take those out and this is good to merge.

samanthaandrews: Good catch. I have unintentionally committed those files about 5 times on this PR 😵‍💫 . Reverted!

leonardr (approved):

PR #1656, opened 2021-11-23
Single dockerfile

nballenger:
## Description

Updates the containerization of the Circulation Manager to make local development simpler.


nballenger:
If folks are reviewing and want to give it a go locally, you should be able to check out the `single-dockerfile` branch, then use the instructions here:

https://github.com/NYPL-Simplified/circulation/blob/single-dockerfile/docs/Development.md

I'm looking for feedback on whether everything works on your system, but also whether the document was clear, whether there are other Makefile commands you'd find useful, etc. Thanks!

keithbauer (approved):
I was able to run this locally with no issues. The instructions were clear and easy to follow. Looks great!

oliviawongnyc (approved):
Works for me!!

nballenger:
@ettore was kind enough to test on his arm64 Mac, and found that the elasticsearch image fails during build. I wiped out my build cache and existing images, and I can replicate, though it fails with a different error. In my case Gradle is unable to resolve dependencies, I _think_ because Elastic Co. has removed some relevant .pom files from their CDN.

I suspect that building a snapshot that is not officially supported will only get more brittle as time goes on, particularly since Elastic has changed their licensing and profit models. The build still works fine for people who have amd64 chips, but is currently broken from arm64 users.

I'm going to merge this code anyway, because I want to get the tests and deployment squared away, and start being able to base new feature branch work off of what's in this branch. I'll make a new ticket for getting an Elasticsearch server working again for arm64 users.

PR #1659, opened 2021-12-09
OE-162: Implement OAuth Client Credential Password flow as an alternative to HTTP Basic Auth

keithbauer:
## Description

Check that `bearer_token_signing_secret` is set at app start.

## Motivation and Context

[OE-293](https://jira.nypl.org/browse/OE-293) is the first task towards implementing OAuth Client Credential Password flow for HTTP Basic Auth.

## Checklist:


api/routes.py
```diff
@@ -49,6 +50,20 @@ def initialize_circulation_manager():
...
 
+@app.before_first_request
+def initialize_app_settings():
+    _db = app.manager._db
+
+    bearer_token_signing_secrets = [
+        token.value for token in
+        _db.query(ConfigurationSetting).filter(
+            ConfigurationSetting.key==Configuration.BEARER_TOKEN_SIGNING_SECRET
+        ).all()
```
keithbauer, line 61: I wasn't sure if this should be a list, `first()`, or `one_or_none()`.

nballenger: Couple things, some of which I just figured out:

1. I think there can only be one bearer token signing secret at any given time, because rows in `ConfigurationSetting` have to be [unique for a given combination of key name, external integration id, and library id](https://github.com/NYPL-Simplified/server_core/blob/6c54a9105f530c29389feeba3beb0d9313314355/model/configuration.py#L568-L570), and since this is a sitewide secret, the only operative value is the key name. So you don't have to worry about getting multiple ones by query.
2. The normal way to get the signing secret value is to call [`api.authenticator.BearerTokenSigner.bearer_token_signing_secret()`](https://github.com/NYPL-Simplified/circulation/blob/develop/api/authenticator.py#L2247). However, that method relies on a call to `ConfigurationSetting.sitewide_secret()`, which will [create a randomly generated value](https://github.com/NYPL-Simplified/server_core/blob/6c54a9105f530c29389feeba3beb0d9313314355/model/configuration.py#L625-L628) if the named setting doesn't already exist. That would mean you couldn't actually set the value explicitly if it was unset, since checking for it would also set it.

Given the way it's set up, I think the simplest thing may be to drop the requirement that you be able to set the value from an environment variable, and just make the single call to `BearerTokenSigner.bearer_token_signing_secret(_db)` that you've got below. That'll make sure a value is in there, creating one if necessary.

nballenger: Honestly it's possible this whole thing isn't necessary, since the first time any piece of code tries to retrieve the signing secret it'll get set if it doesn't already exist. That said, I think having an `initialize_app_settings()` function is probably worthwhile, and it's reasonable to be explicit in that function that we want a signing secret to exist.

api/authenticator.py
```diff
@@ -1949,11 +1887,45 @@ def scrub_credential(self, value):
...
+            if an error occurs.
+        """
+        credential = Credential.lookup_by_token(
+            _db, None, BasicAuthenticationProvider.TOKEN_TYPE, credentials
+        )
+
+        if isinstance(credential, Credential):
+            return credential.patron
+        else:
+            return EXPIRED_HTTP_BASIC_BEARER_TOKEN
```
leonardr, line 1918: Do you need to distinguish between the case where the token has expired and the case where it's just invalid? If we're only handling one case I think it's better to just say that the token is invalid.

api/authenticator.py
```diff
@@ -2661,3 +2633,37 @@ class BaseSAMLAuthenticationProvider(AuthenticationProvider, BearerTokenSigner,
...
+            # into a valid patron.
+            return patron
+
+        if isinstance(patron, Patron):
+            # Create a temporary inner token with a lifetime of 1 hour
+            duration = datetime.timedelta(seconds=3600)
+            data_source = None
+            provider = BasicAuthenticationProvider.BEARER_TOKEN_PROVIDER_NAME
+            token_type = BasicAuthenticationProvider.TOKEN_TYPE
+            inner_token, _ = Credential.temporary_token_create(
```
leonardr, line 2662: I like the 'inner' terminology; it really clarifies what's going on.

api/controller.py
```diff
@@ -452,6 +453,7 @@ def setup_configuration_dependent_controllers(self):
         This method will be called fresh every time the site
         configuration changes.
         """
+        self.basic_auth_controller = BasicAuthTempTokenController(self.auth)
```
leonardr, line 456: I would like to call this basic_auth_token_controller because this makes it sound like the only controller that responds to Basic Auth.

leonardr (approved):
This looks good; my suggested changes are very minor.

leonardr (approved):

PR #1660, opened 2021-12-15
OE-301 & OE-303: Adding new First Book login button image

samanthaandrews:
## Description
Updating First Book login button that is used in https://beta.openebooks.us/ and other frontend contexts.

## Motivation and Context
We were using an outdated logo in this button and the First Book team reached out to us to update to the latest. Also, we needed to change the text from FirstBook to First Book.

https://jira.nypl.org/browse/OE-301
https://jira.nypl.org/browse/OE-303

## Checklist:


leonardr:
It looks like you deleted the old image without adding the new one--probably just an oversight. But it would be best to replace the old image with the new one, rather than giving it a different name, so you don't need to change LOGIN_BUTTON_IMAGE in api/firstbook2.py. (I also can't guarantee the mobile apps don't have the old URL hard-coded, though hopefully they don't.)

samanthaandrews:
> It looks like you deleted the old image without adding the new one--probably just an oversight. But it would be best to replace the old image with the new one, rather than giving it a different name, so you don't need to change LOGIN_BUTTON_IMAGE in api/firstbook2.py. (I also can't guarantee the mobile apps don't have the old URL hard-coded, though hopefully they don't.)

I think I did add the new image with the same name, so the binary shows a decrease in size, but not the addition of a new file. I am unsure why tests are failing though. It looks like similar failures are happening on Keith's recent PR though?

leonardr:
I see now -- I was misreading the diff.

Here's the test failure that's probably causing the problem:

```
Env var SIMPLIFIED_STATIC_DIR unset or empty, defaulted to '/simplified_static'. Static resources directory path '/simplified_static' does not exist.
```

My guess is this is related to @nballenger's recent one-Docker-file change, which I think moved or removed that directory. I'll take a look tomorrow if Nick hasn't resolved the issue by then.

nballenger:
I think the test issue should be resolved now. I'll check again in the AM.

nballenger (approved):
Looks good to me. I fixed the Github Actions test file in this branch, and looks like all tests pass fine. It does have a merge conflict with develop now, but it's very minor. If you need any help getting it merged, ping me and we'll work it out.

keithbauer (approved):

leonardr:
@nballenger The tests are passing, but I'm still getting error messages from DockerHub once it's time to actually build the Docker image. Can you investigate?

The error is: `Dockerfile not found at ./docker/Dockerfile.exec`

nballenger:
Yes! Sorry about that, it's on my list to fix.

PR #1662, opened 2022-01-05
SIMPLY-3929: Fix comma + space issue in access-control-allow-origin headers

samanthaandrews:
## Description

- Just a small fix to the access-control-allow-origin header that was erroneously adding a space before the comma in the list instead of after.

## Motivation and Context

https://jira.nypl.org/browse/SIMPLY-3929

## Checklist:

### DIFF
--- a/api/routes.py
+++ b/api/routes.py
@@ -128,7 +128,7 @@ def allows_cors(allowed_domain_type):
 
             if web_domains:
                 options = get_cors_options(
-                    app, dict(origins=" ,".join(web_domains),
+                    app, dict(origins=", ".join(web_domains),
                             supports_credentials=True)
                 )
                 set_cors_headers(resp, options)


leonardr (approved):
Let's merge this because it's better than what we have now, but I believe this feature is broken.

First, the unit test assertion functions like `assert_cors_headers_patron_and_admin` use a space to generate the expected value, not a comma and a space. That indicates we don't actually test this functionality; if we had we'd have seen the discrepancy earlier.

More importantly, I think this feature was implemented wrong originally and has never worked. The spec doesn't allow more than one hostname in Access-Control-Allow-Origin. It's supposed to be `*`, or empty, or the value of the incoming `Origin` header.

https://fetch.spec.whatwg.org/#http-responses
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin

So we can keep a list of multiple allowed hostnames internally, but the way it ought to work is we check the incoming hostname and set `Access-Control-Allow-Origin` to _the same value_ but only if it's in the list of approved hostnames.

We are fine for now, because I don't think anyone actually uses this feature, but please file a ticket about this so we can clean it up later.

samanthaandrews:
Thanks @leonardr . I created https://jira.nypl.org/browse/SIMPLY-3983.

PR #1664, opened 2022-01-12
OE-317 Advertise OAuth endpoint for HTTP Basic Auth

keithbauer:
## Description

Advertises the new basic auth bearer token URL in the authentication document.

## Motivation and Context

Addresses [OE-317](https://jira.nypl.org/browse/OE-317). This was something that I missed while working on [OE-162](https://jira.nypl.org/browse/OE-162).

## How Has This Been Tested?

Add a Simple Authentication Provider to the Patron Authentication. Ensure that the new endpoint is being advertised in the authentication document. Additional asserts have been added `test_authentication_flow_document` to check that both documents are present and contains the correct attributes.

## Checklist:

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -2177,10 +2177,12 @@ class BasicAuthenticationProvider(AuthenticationProvider, HasSelfTests):
         basic_doc = self._generate_authentication_flow_document(_db, type="http://opds-spec.org/auth/basic")
         oauth_doc = self._generate_authentication_flow_document(_db, type="http://librarysimplified.org/authtype/OAuth-Client-Credentials")
         oauth_doc.update(dict(
-            links=dict(
-                rel="authenticate",
-                href=url_for("http_basic_auth_token", _external=True)
-            )
+            links=[
+                dict(
+                    rel="authenticate",
+                    href=url_for("http_basic_auth_token", _external=True)
+                )
+            ]
         ))
 
         return [basic_doc, oauth_doc]
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -2666,4 +2666,10 @@ class BasicAuthTempTokenController(object):
             # Wrap the inner token with the provider name
             outer_token = self.authenticator.create_bearer_token(provider, inner_token.credential)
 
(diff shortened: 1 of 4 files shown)


tests/test_authenticator.py
```diff
@@ -2166,6 +2167,8 @@ def test_authentication_flow_document(self):
         provider.identifier_maximum_length=22
         provider.password_maximum_length=7
         provider.identifier_barcode_format = provider.BARCODE_FORMAT_CODABAR
+        FLOW_TYPE_BASIC = 'http://opds-spec.org/auth/basic'
```
keithbauer, line 2170: I also define this in First Book tests. Is this fine as it is or should I move these constants to `BasicAuthenticationProvider`?

tests/test_authenticator.py
```diff
@@ -1241,7 +1241,8 @@ def annotate_authentication_document(
             # authentication sub-documents are assembled properly and
             # placed in the right position.
             flows = doc['authentication']
-            oauth_doc, basic_doc = sorted(flows, key=lambda x: x['type'])
+            [oauth_doc] = list(filter(lambda x: isinstance(x, dict), flows))
```
keithbauer, line 1244: I'm not sure how to sort a mixed list that now contains lists and a dictionary, so I filtered by the type instead. I don't know if the order is now guaranteed as it was before.

api/authenticator.py
```diff
@@ -2174,6 +2174,21 @@ def _authentication_flow_document(self, _db):
         OPDS document.
         """
 
+        basic_doc = self._generate_authentication_flow_document(_db, type="http://opds-spec.org/auth/basic")
+        oauth_doc = self._generate_authentication_flow_document(_db, type="http://librarysimplified.org/authtype/OAuth-Client-Credentials")
```
leonardr, line 2178: Per your other comment, yes, both of these should be constants in this class. Just not FLOW_TYPE, which you're right to have stopped using.

tests/test_authenticator.py
leonardr: This is OK since there's only one item in each list.

leonardr (approved):

PR #1672, opened 2022-01-19
OE 318 Require OAuth flow to be explicitly enabled

keithbauer:
## Description

Adds a `LIBRARY_SETTINGS` configuration to `BasicAuthenticationProvider` that allows libraries to opt in to the new OAuth flow.

## Motivation and Context

[OE-318](https://jira.nypl.org/browse/OE-318). Allows libraries to opt in to the new flow, with the default being opted out.

## Checklist:

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -1822,6 +1822,13 @@ class BasicAuthenticationProvider(AuthenticationProvider, HasSelfTests):
             or self.DEFAULT_PASSWORD_LABEL
         )
 
+        self.oauth_enabled = self.HTTP_BASIC_OAUTH_ENABLED_DEFAULT
+        oauth_enabled = list(
+            filter(lambda setting: setting.key == self.HTTP_BASIC_OAUTH_ENABLED, library.settings)) \
+            or self.HTTP_BASIC_OAUTH_ENABLED_DEFAULT
+        if oauth_enabled:
+            self.oauth_enabled = oauth_enabled[0].bool_value or self.HTTP_BASIC_OAUTH_ENABLED_DEFAULT
+
     def remote_patron_lookup(self, patron_or_patrondata):
         """Ask the remote for information about this patron, and then make sure
         the patron belongs to the library associated with thie BasicAuthenticationProvider."""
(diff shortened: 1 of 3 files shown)


api/authenticator.py
```diff
@@ -1804,6 +1822,13 @@ def __init__(self, library, integration, analytics=None):
             or self.DEFAULT_PASSWORD_LABEL
         )
 
+        self.oauth_enabled = self.HTTP_BASIC_OAUTH_ENABLED_DEFAULT
+        oauth_enabled = list(
+            filter(lambda setting: setting.key == self.HTTP_BASIC_OAUTH_ENABLED, library.settings)) \
+            or self.HTTP_BASIC_OAUTH_ENABLED_DEFAULT
+        if oauth_enabled:
+            self.oauth_enabled = oauth_enabled[0].bool_value or self.HTTP_BASIC_OAUTH_ENABLED_DEFAULT
```
keithbauer, line 1830: Is there a better way to get this setting? I tried calling it like the other settings here, `integration.setting(self.HTTP_BASIC_OAUTH_ENABLED).value` and `.bool_value` but nothing is returned.

api/authenticator.py
```diff
@@ -1674,6 +1679,19 @@ class BasicAuthenticationProvider(AuthenticationProvider, HasSelfTests):
...
+          "label": _("Enable OAuth for HTTP Basic Auth"),
+          "description": _("Enable authentication with bearer tokens generated via basic auth credentials"),
+          "type": "select",
+          "options": [
+              { "key": "false", "label": _("Disabled") },
+              { "key": "true", "label": _("Enabled") },
+          ],
+          "default": "false",
+        },
+    ] + AuthenticationProvider.LIBRARY_SETTINGS
```
keithbauer, line 1693: I added this setting to get tests to pass (`tests/admin/controller/test_patron_auth.py`), but noticed that it is causing them to also show up in the Library Settings on the Patron Authentication page, is this fine or is there something else I should be doing?

keithbauer: This looks like it's the correct behavior. I checked out the `develop` branch and added a basic authenticator and looked at the settings and these were there sans the new OAuth setting. Marking as resolved.

api/authenticator.py
nballenger: There are a bunch of class methods on `ConfigurationSetting` like `ConfigurationSetting.for_library(key, library)`, but they all flow through `ConfigurationSetting.for_library_and_externalintegration()`, which will create the setting if it doesn't already exist. I'm assuming that you just want to access the setting if it's there, not create it, so probably you do need to iterate / filter the settings on the library object you've got. That said, it may be more efficient to run the filter on the db side, so you may want to switch to using a chained `.filter()` SQLAlchemy call instead of a native Python iterable filter.

nballenger (approved):
👍

PR #1686, opened 2022-02-11
Switched saml validation tests to pytest, shrunk xml file

nballenger:
## Description

IT reported an automated scanner found email addresses in a source file. That turned out to be a locally stored copy of a timestamped list of identity providers from incommon.org, that we use to test some of our SAML code.

I noticed it was a 37M XML file, which seems on the large side for briefly referenced test data. I cut it down to 64k by removing a bunch of the entities, and all the tests pass except the one that checks for a valid signature on that file. I assume by changing the file I borked some checksum that informed the original signature value.

I've skipped the test for now, because I don't want to take the time to figure out how to generate a new signature. I'll take another look before this merges, but given that most of what it's testing is just a direct passthrough to the validation routine of the third party SAML library we're using, I'm not sure it strictly needs a test.

### DIFF
--- a/tests/saml/metadata/federations/test_validator.py
+++ b/tests/saml/metadata/federations/test_validator.py
@@ -13,89 +13,141 @@ from api.saml.metadata.federations.validator import (
     SAMLFederatedMetadataValidationError,
     SAMLMetadataSignatureValidator,
 )
-from core.util.datetime_helpers import (
-    datetime_utc,
-    from_timestamp,
-    utc_now
-)
+from core.util.datetime_helpers import datetime_utc, utc_now
 
 
-class TestSAMLFederatedMetadataExpirationValidator(object):
-    @parameterized.expand(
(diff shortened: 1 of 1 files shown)


leonardr:
@mwbenowitz Thanks for reminding me of this branch. Looking at the test that Nick skipped, all it does is verify that 

`SAMLMetadataSignatureValidator.validate` calls `OneLogin_Saml2_Utils.validate_metadata_sign` with the appropriate options. You can test it by mocking validate_metadata_sign, there's no need to get an actual signature. (and I don't think we _can_ get an actual signature because we would presumably need someone's private key).

mwbenowitz:
Great! I’ll try to get a mock set up for that shortly if that’s all that’s blocking this from being merged!

mwbenowitz:
The changes included in this PR were folded into #1724 in order to incorporate the large change of removing the `core` submodule. As this code has made it into `develop` this can be safely closed!

PR #1690, opened 2022-03-18
Merge server_core code instead of using a git submodule

mattkoch614:
## Description
**This PR:** 

- Removes the git submodule [`server_core`](https://github.com/NYPL-Simplified/server_core) and merges its code into this repository instead, inside of the `core` folder (where the submodule code was being fetched.)
- Removes the `.gitmodules` file, since there was only a single submodule inside of it.
- Retains the entire commit history of `server_core` 

This [gist](https://gist.github.com/ryaninvents/9fb333760e217c3425c728a1c2a99db4) proved helpful in performing the actions necessary.

**This PR does not:**

- Add the suite of `server_core` tests (inside of `core/tests`) to the parent Makefile so that they are also run when the `make test` command is executed. 
- Change the `Makefile` to remove the need for the `make setup` command, since we no longer have submodules. This doesn't hurt anything, and could still be used in the future, if another submodule is added.
- Modify any Github workflows or CI/CD tasks.
- Update the nested `core/README.md` file

## Motivation and Context

This change addresses spike investigations researched in [SIMPLY-3979](https://jira.nypl.org/browse/SIMPLY-3979), where the basic consensus is that to simplify development tasks, combining `server_core` into `circulation` helps with testability and reducing cognitive load, specifically with the current need to jump between two repos. 

## How Has This Been Tested?

This was tested using the `Makefile` in the root of this repository, and running the `make test` command. All tests continue to pass locally, after the submodule merge.

## Checklist:


keithbauer (approved):
👍  This looks good to me.

leonardr:
This is a great solution, since it lets us keep the history. I haven't been through the entire diff, but I went through the same process on a fresh circulation checkout and got a similar result. Here are the commands I used, adapted from the Gist you linked:

```
git remote add submodule_origin https://github.com/NYPL-Simplified/server_core
git fetch submodule_origin
git merge --allow-unrelated-histories -s ours --no-commit submodule_origin/develop
git rm --cached core
git rm .gitmodules
rm -rf core/.git
git add core
git commit -am "Removed submodule"
```

What I'm about to recommend should go into a separate branch based on this one, for (I'm assuming) the same reason you didn't do it as part of this branch in the first place. Review is easier if we can distinguish new work from moving old work around, and this branch moves a whole lot of old work around.

I'd like to see three things fixed/verified and merged into this branch before we merge this into develop:

1. `make test` should run the core tests as well. We can't merge a branch where we haven't done an automated test of the entire affected code base, and this branch makes the affected code base much bigger. If this causes weird Python scope issues, I can take a look -- I've dealt with stuff like that in the past.
2. Github workflows should be modified. I think this will be a no-op if `make test` is fixed, because there are no explicit references to core in the workflows. 
3. Verify that the Docker builds still work. I'm pretty sure this is also a no-op since they depend on the 'core' directory already being present and don't care how it got there.

mattkoch614:
@leonardr - Thanks for reproducing the gist - I did indeed need the ` --allow-unrelated-histories` flag to get things to work.
To your points:
1. OK. Will create a new branch for this.
2. Can we revisit after 1. is resolved?
3. I can verify, locally at least, that the rest of the make commands produced a working web application for me (`make build, make up`)

nballenger:
Just adding a reference here to an issue I raised in Slack--we may want to wait on actually pulling the trigger on this till @keithbauer  is able to wrap and merge his PR on converting tests in core to pytest fixtures:

https://github.com/NYPL-Simplified/server_core/pull/1278

An alternative to waiting would be to have Keith merge the current head of core's `develop` into his branch, then base this PR on merging from Keith's branch rather than `develop`.

mwbenowitz (commented):
This is piggybacking on top of what @leonardr has commented. There are a few things that I think would make this merge cleaner and perhaps address his thoughts!

1. Merge the `core/.gitignore` into the root `.gitignore` as having both files could result in some odd behavior in the future (I think?)
2. Remove the `core/.github` directory as we should fold the tests into the root's actions. As part of this I would make sure that the matrix/versions being tested align between the actions
3. Migrate the `core/tests` dir into `tests/core/*` which might not be as simple as all of that, but should address the concerns about making sure these tests run without the need to remember where they all live.

Happy to hear if these are bad ideas! Just trying to look towards what will be the most maintainable thing here.

nballenger:
> 1. Merge the `core/.gitignore` into the root `.gitignore` as having both files could result in some odd behavior in the future (I think?)

I think you can just delete the one from core--the one in the root of `circulation` is a superset of what's in `core/.gitignore`. Also, I think they'd place nicely together anyway, since there's this in the gitignore docs:

> _Patterns read from a .gitignore file in the same directory as the path, or in any parent directory (up to the top-level of the working tree), with patterns in the higher level files being overridden by those in lower level files down to the directory containing the file. These patterns match relative to the location of the .gitignore file. A project normally includes such .gitignore files in its repository, containing patterns for files generated as part of the project build._

But we should only have one if we can to avoid confusion, so lets ditch the one in core.

> 2. Remove the `core/.github` directory as we should fold the tests into the root's actions. As part of this I would make sure that the matrix/versions being tested align between the actions

Agreed on this. A while back I dropped the test matrix for `circulation` back to just Python 3.6, where before we were testing 3.6 through 3.9. We can't run the service using anything past 3.6 until we resolve the dependency situation around whichever module it is that's stuck in the past (maybe `dm.xmlsec.binding`? I can't remember off the top of my head), so there's no reason to test in any further versions.

> 3. Migrate the `core/tests` dir into `tests/core/*` which might not be as simple as all of that, but should address the concerns about making sure these tests run without the need to remember where they all live.

It's a good idea, but I think is worth deferring to a separate process at a later date. We're not only merging the codebase, but two previously separate pytest environments. We could end up forced to run the test sets separately for a while if that involves weirdness around objects/fixtures scoped to the entire test session. Keeping the tests in different directories will make that easier until we're sure everything is working.

nballenger:
Adjusted the Makefile. Building, running, and testing work on my machine, though with the following failures:

```
============================== short test summary info ===============================
FAILED core/tests/test_app_server.py::TestHeartbeatController::test_heartbeat - OSE...
FAILED core/tests/test_app_server.py::TestErrorHandler::test_unhandled_error - asse...
FAILED core/tests/test_app_server.py::TestErrorHandler::test_handle_error_as_problem_detail_document
FAILED core/tests/test_config.py::TestConfiguration::test_app_version - OSError: [E...
FAILED core/tests/test_external_search.py::TestExternalSearch::test_works_index_name
FAILED core/tests/test_external_search.py::TestExternalSearch::test_setup_index_creates_new_index
FAILED core/tests/test_external_search.py::TestExternalSearch::test_set_works_index_and_alias
FAILED core/tests/test_external_search.py::TestExternalSearch::test_setup_current_alias
FAILED core/tests/test_external_search.py::TestExternalSearch::test_transfer_current_alias
FAILED core/tests/test_external_search.py::TestExternalSearchWithWorks::test_query_works
```

mwbenowitz:
thanks for the feedback @nballenger that all sounds good to me!

mattkoch614:
Hey @nballenger  / @keithbauer - what do we need to do to merge this, any additional tweaks or pre-reqs?

leonardr:
First, we've got the test failures @nballenger mentioned -- that's tracked in https://jira.nypl.org/browse/SIMPLY-4046. My guess is some code needs to be moved from the server_core tox.ini to the circulation tox.ini, to set up the Minio image. The Elasticsearch image should be set up already, so I'm not sure why those tests are failing.

Second, when Github Actions runs the tests, it's only running the circulation tests. (That's why the tests pass -- otherwise I'd expect the same problems Nick is seeing running the tests with `make`.) In other words, my item #2 is wrong -- `make test` isn't what Github Actions is using to run the tests. At any reate, Github Actions should run the circulation tests + the core tests.

keithbauer:
I opened #1691 which adds MinIO support and adds a missing Elasticsearch environment variable for local docker setup, but perhaps I should've added that to this PR. That fixes the failing S3 and Elasticsearch tests, but still leaves the following tests failing:

```
FAILED core/tests/test_app_server.py::TestHeartbeatController::test_heartbeat - OSE...
FAILED core/tests/test_app_server.py::TestErrorHandler::test_unhandled_error - asse...
FAILED core/tests/test_app_server.py::TestErrorHandler::test_handle_error_as_problem_detail_document
FAILED core/tests/test_config.py::TestConfiguration::test_app_version - OSError: [E...
```

mattkoch614:
Looking good! Only 4 tests left failing? @keithbauer ?

mattkoch614:
@leonardr / @nballenger / @keithbauer - FYI I'll close this as Keith is going to merge a different branch in to achieve the same end result.

PR #1691, opened 2022-03-22
Add MinIO to docker-compose

keithbauer:
## Description

Adds a MinIO container and fixes test Elasticsearch endpoint.

## Motivation and Context

Adds a MinIO container for S3 tests in `core` and fixes the Elasticsearch test endpoint.

## How Has This Been Tested?

Locally with `make up` and running the tests separately via `docker exec -it --env TESTING=1 cm_local_webapp /usr/local/bin/runinvenv /simplified_venv pytest tests --disable-pytest-warnings` and `docker exec -it --env TESTING=1 cm_local_webapp /usr/local/bin/runinvenv /simplified_venv pytest core/tests/ --disable-pytest-warnings`.

There are four tests that fail for me locally from `core`. Two deal with read-only settings and two that have some sort of mismatched exceptions from `core/tests/test_app_server.py`.

## Checklist:

### DIFF
--- a/docker-compose.arm64.yml
+++ b/docker-compose.arm64.yml
@@ -37,8 +37,6 @@ services:
     ports:
       - "9000:9000"
       - "9001:9001"
-    volumes:
-      - cm_local_minio:/data
 
   scripts:
     container_name: cm_local_scripts
@@ -90,4 +88,3 @@ volumes:
   cm_local_gunicorn_log:
   cm_local_script_logs:
   es_data:
-  cm_local_minio:
--- a/docker-compose.arm64.yml
+++ b/docker-compose.arm64.yml
@@ -27,7 +27,19 @@ services:
     ports:
       - "9200:9200"
       - "9300:9300"
-    
+
+  minio:
+    container_name: minio
+    image: bitnami/minio:latest
+    environment:
+      MINIO_ROOT_USER: simplified
+      MINIO_ROOT_PASSWORD: 12345678901234567890
+    ports:
+      - "9000:9000"
+      - "9001:9001"
+    volumes:
+      - cm_local_minio:/data
+
   scripts:
     container_name: cm_local_scripts
     depends_on:
@@ -52,6 +64,7 @@ services:
     depends_on:
       - database
       - elasticsearch
+      - minio
     build:
       context: .
       target: cm_webapp_local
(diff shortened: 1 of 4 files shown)


leonardr:
This sounds like a good start on making https://github.com/NYPL-Simplified/circulation/pull/1690 mergeable.

I think if you make this change to tox.ini, Github Actions will start running the core tests as well as the circulation tests. That will not only take care of the other thing we need to make the other branch mergeable, it will hopefully make those errors start showing up so I can take a look at exactly what's going on.

```
diff --git a/tox.ini b/tox.ini
index 083814d31..0f8fbb8f1 100644
--- a/tox.ini
+++ b/tox.ini
@@ -9,7 +9,7 @@ commands_pre =
     docker: docker restart es
     python -m textblob.download_corpora
 commands =
-    pytest --disable-warnings {posargs:"tests"}
+    pytest --disable-warnings {posargs:"tests core/tests"}
 passenv = SIMPLIFIED_*
 setenv =
     docker: SIMPLIFIED_TEST_DATABASE=postgres://simplified_test:test@localhost:9005/simplified_circulation_test
```

leonardr:
Good news: I was able to get tox to run the core tests as well as the circulation tests.
Bad news: in the tox context it looks like the Elasticsearch and MinIO images are not available, because those tests are failing when run through Github Actions.

keithbauer:
Thanks for getting the pytest command updated. I've added MinIO to tox so the S3 tests are now passing. The failing tests I mentioned in the PR body above are from the `SIMPLY-4025-pytest-core-models` branch in `server_core`. 

There's a different set of tests that are failing here that is based off the current `develop` branch of `server_core`:
```
FAILED core/tests/test_selftest.py::TestHasSelfTests::test_run_self_tests
FAILED core/tests/test_selftest.py::TestHasSelfTests::test_exception_in_has_self_tests
FAILED core/tests/models/test_admin.py::TestAdmin::test_with_password
FAILED core/tests/models/test_configuration.py::TestConfigurationSetting::test_explain
```

leonardr:
I suspect the four failing tests are caused by circulation setup code that modifies the dataset a test sees in the database when it starts up, making it different than the dataset the "core" module sees when _it_ starts up. Either this setup code is being called when it shouldn't be, or (more likely) it's being called outside the normal subtransaction-rollback process, probably in a setup method, such that the data persists across tests.

This is most obvious with the TestConfigurationSetting.test_explain failure, where I think the configuration settings are being "polluted" by ControllerTest.setup_method. But it also explains why all the tests in circulation/tests/admin/controller/test_individual_admins.py start off by deleting all existing Admin objects, including the one that screws up TestAdmin.test_with_password. It was probably created during a setup step.

Basically I'd take a look at ControllerTest.setup_method -- it's responsible for at least one of these, maybe two, and possibly all of them.

The two failures in `TestHasSelfTests` are consistent with the first (or second) test in the self-test suite raising an exception, such that subsequent tests don't even get a chance to run. We're trying to pull out two (or three) test results but only getting one (or two). If you also get these test failures locally, putting a breakpoint in HasSelfTests.run_self_test should let you isolate the problem.

PR #1693, opened 2022-03-30
Feature/merge server core

keithbauer:
## Description
This is an update to PR #1690 that pulls in the latest changes from core (pytest fixtures).

**This PR:** 

- Removes the git submodule [`server_core`](https://github.com/NYPL-Simplified/server_core) and merges its code into this repository instead, inside of the `core` folder (where the submodule code was being fetched.)
- Removes the `.gitmodules` file, since there was only a single submodule inside of it.
- Retains the entire commit history of `server_core` 

This [gist](https://gist.github.com/ryaninvents/9fb333760e217c3425c728a1c2a99db4) proved helpful in performing the actions necessary.

**This PR does not:**

- Update the nested `core/README.md` file

## Motivation and Context

This change addresses spike investigations researched in [SIMPLY-3979](https://jira.nypl.org/browse/SIMPLY-3979), where the basic consensus is that to simplify development tasks, combining `server_core` into `circulation` helps with testability and reducing cognitive load, specifically with the current need to jump between two repos. 

## How Has This Been Tested?

This was tested using the `Makefile` in the root of this repository, and running the `make test` command. All tests continue to pass locally, after the submodule merge.

## Checklist:


core/tests/test_selftest.py
```diff
@@ -127,9 +127,8 @@ def _run_self_tests(self, _db):
 
         # By default, the default constructor is instantiated and its
         # _run_self_tests method is called.
-        data, [setup, test] = Tester.run_self_tests(
-            mock_db, extra_arg="a value"
-        )
+        data, [setup, test] = Tester.run_self_tests(mock_db, constructor_method=Tester, extra_arg="a value")
```
keithbauer, line 130: I'm not entirely sure why `constructor_method` needs to be passed when running both test suites through pytest, but not when running the core tests separately. Without passing `constructor_method` this exception is raised: `TypeError: object() takes no parameters`.

leonardr: I can duplicate this problem, or at least a very similar problem. the issue is that at some point inside `Tester.run_self_tests`, the class variable `cls` starts to refer to `HasSelfTests` instead of `Tester`. Instantiating `HasSelfTests` results in an error when actually running the tests, since `HasSelfTests` doesn't have any tests to run.

leonardr: Still not sure what's going on here, but I've gotten a little more detail on the problem.

Normal Python behavior is that an inherited class method is bound to the subclass, not the superclass:

```
class Sup(object):
    @classmethod
    def foo(cls, arg):
        print(cls, arg)

class Subclass(Sup):
    pass

print(Subclass.foo)
# <bound method Sup.foo of <class '__main__.Subclass'>>

Subclass.foo("arg1")
# <class '__main__.Subclass'> arg1
```
But when this test suite is being run, a subclass of `HasSelfTests` has its `run_self_tests` method bound to `HasSelfTests.run_self_tests`.

```
(Pdb) Tester.run_self_tests
<bound method HasSelfTests.run_self_tests of <class 'core.selftest.HasSelfTests'>>
```

Calling `HasSelfTests.run_self_tests()` is always the wrong call, but that's what's happening here.

If I make a _new_ `HasSelfTests` subclass in my pdb session, it has the same problem:

```
(Pdb) class Foo(HasSelfTests): pass
(Pdb) Foo
<class 'core.tests.test_selftest.Foo'>
(Pdb) Foo.run_self_tests
<bound method HasSelfTests.run_self_tests of <class 'core.selftest.HasSelfTests'>>
```
The same is true of the other class methods in `Tester`.

Kind of grasping at straws here, but there's another `HasSelfTests` class in api/selftest.py. The class here is clearly `core.selftest.HasSelfTests`, but the two classes might be interfering with each other, and merging them might solve the test failures.

keithbauer: I've tried to merge the two `HasSelfTests` classes into one that remains in `core.selftest.HasSelfTests` and removed `api.selftest.HasSelfTests`, but the `TypeError` is still raising when `constructor_method` isn’t specified.

leonardr: OK, let's let it go for now. If this shows up in a different context we'll do a more thorough investigation.

leonardr (approved):

PR #1698, opened 2022-04-07
Return an existing HTTP Basic Auth token if requested within a minute since creation

keithbauer:
## Description

Return a patron's existing HTTP Basic temp token if it was requested within a minute since the token's creation. Otherwise a new token is generated and returned.

## Motivation and Context

[[SIMPLY-4057](https://jira.nypl.org/browse/SIMPLY-4057)] Returning an existing token if it was requested within a minute of the token's creation rather than generating a new one will help avoid race conditions for the consuming clients.

## How Has This Been Tested?

Added a test to confirm that the same token is returned when authenticating with HTTP Basic within a minute of the initial token's creation.

## Checklist:

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -30,7 +30,6 @@ from .config import (
     IntegrationException,
 )
 from core.model import (
-    credential,
     get_one,
     get_one_or_create,
     CirculationEvent,
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -30,6 +30,7 @@ from .config import (
     IntegrationException,
 )
 from core.model import (
+    credential,
     get_one,
     get_one_or_create,
     CirculationEvent,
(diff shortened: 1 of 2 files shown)


api/authenticator.py
```diff
@@ -2685,37 +2685,61 @@ class BasicAuthTempTokenController(object):
     """A controller that handles requests for issuing temporary tokens
     to HTTP Basic Auth credentials.
     """
+    DO_NOT_GENERATE_NEW_TOKEN_PERIOD = 60 * 59
+    TOKEN_DURATION = datetime.timedelta(seconds=3600)
```
leonardr, line 2688: Since the rule is "don't generate a token in the first minute of its lifetime", DO_NOT_GENERATE_NEW_TOKEN_PERIOD should be derived from TOKEN_DURATION, i.e. `TOKEN_DURATION.total_seconds() - 60`. As it is, if TOKEN_DURATION changes, DO_NOT_GENERATE_NEW_TOKEN_PERIOD will stay the same and possibly end up making no sense.

tests/test_authenticator.py
```diff
@@ -2947,3 +2946,38 @@ def test_basic_auth_temp_token(self):
...
+
+            token = response.json.get('access_token')
+            assert token
+
+            another_response = self.controller.basic_auth_temp_token({}, self._db)
+            assert another_response.status_code == 200
+
+            another_token = another_response.json.get('access_token')
+            assert another_token == token
+
```
leonardr, line 3002: In the code as it stood before, there was a case that wasn't explicitly tested: the idea that making a second request to the token endpoint results in a new token. It probably should have been tested before, but it definitely needs to be tested now.

We need to verify that making this request 1 second before the 1-minute cutoff results in the old token, and that making the request 1 second after the cutoff results in a new token. You can use [the `freezegun` module](https://pypi.org/project/pytest-freezegun/) to set appropriate times for these tests.

There are some freezegun examples in test_odl.py and a few other tests, but they use the `freeze_time` decorator, which is designed for running an entire test as though at a certain time. You need to simulate the passage of time, so I think the guidance on the PyPI page will be more useful.

leonardr (changes requested):
This looks good and should be ready to merge with some minor changes.

tests/test_authenticator.py
```diff
@@ -2966,11 +2966,12 @@ def test_basic_auth_temp_token_returns_existing(self):
             token = response.json.get('access_token')
             assert token
 
-            another_response = self.controller.basic_auth_temp_token({}, self._db)
-            assert another_response.status_code == 200
+            with freeze_time(lambda: utc_now() + datetime.timedelta(seconds=59)):
+                another_response = self.controller.basic_auth_temp_token({}, self._db)
+                assert another_response.status_code == 200
```
leonardr, line 2973: This will work most of the time, but I think it's likely to fail unpredictably if there's a second boundary between specific lines of code.

Rather than freezing time to a time relative to the present, can you freeze time to a specific second (2022-01-01 00:00:00) before issuing the initial token, and then move it around to specific other seconds?

leonardr (approved):

PR #1700, opened 2022-04-11
Improvements to crontab and run script

nballenger:
## Description

* Updates the crontab and `/core/bin/run` scripts to emit more logging
* Log messages emitted by the `run` script are now JSON formatted, and look like the following:

    ```
    {"timestamp": "2022-04-11T19:09:02+0000", "level": "INFO", "pid": "18", "script": "cache_opds_blocks", "status": "PID file lock acquired"}
    {"timestamp": "2022-04-11T19:09:02+0000", "level": "INFO", "pid": "18", "script": "cache_opds_blocks", "status": "started"}
    {"timestamp": "2022-04-11T19:09:03+0000", "level": "INFO", "pid": "18", "script": "cache_opds_blocks", "status": "completed"}
    {"timestamp": "2022-04-11T19:09:03+0000", "level": "INFO", "pid": "18", "script": "cache_opds_blocks", "status": "PID file lock released"}
    ```

* Those log messages are simultaneously sent to two locations:
    * The file `/var/log/cron.log`
    * The `STDOUT` of the container's Process 1, which allows them to propagate to the Docker host (and therefore to AWS CloudWatch)
* The following scheduled scripts are now deprecated, though they still appear in the crontab as comments:
    * `metadata_wrangler_collection_registrar`
    * `metadata_wrangler_collection_reaper`
    * `metadata_wrangler_collection_updates`
    * `metadata_upload_coverage`
    * `metadata_wrangler_auxiliary_metadata`
    * `rbdigital_availability_monitor`
    * `rbdigital_collection_delta`
    * `rbdigital_initial_import`

## Motivation and Context

The periodic scripts for the Circulation Manager are executed by `cron`, using these files:

* `docker/simplified_crontab` <-- Installed as the root user's crontab in the `scripts` container
* `core/bin/run` <-- Wrapper that does mutex locking via PID files, to ensure only one copy of a script runs at a time

While we have log files in `/var/log/simplified` for the individual scripts that the `run` wrapper executes, we have not to date had reliable error logging when either `cron` or the `run` script experiences a failure.

That made [SIMPLY-4054](https://jira.nypl.org/browse/SIMPLY-4054) difficult to diagnose and fix. I'm hoping that this overhaul either makes that problem more apparent, or fixes it outright.

## How Has This Been Tested?

Since the changes are to a crontab and shell script, the testing has been manual, and I'll have to deploy it to determine whether it improves the situation on the remote environments.


core/bin/run
```diff
@@ -143,10 +168,22 @@ fi
...
+log_info_to_stdout "$SCRIPT_NAME" "started"
+
 source $SIMPLIFIED_VENV/bin/activate && $FULL_SCRIPT_PATH "$@" >> $LOG_FILE 2>&1
 
+if [ $? -eq 0 ]; then
+  log_info_to_stdout "$SCRIPT_NAME" "completed"
+else
+  log_critical_to_stderr "$SCRIPT_NAME" "could not complete successfully, see $LOG_FILE"
+fi
+
```
nballenger, line 185: This adds a CRITICAL log entry (to CloudWatch and `/var/log/cron.log`) if the script exits non-zero, so we can support alarms on script failures if we want to. It doesn't tell you what actually went wrong, though you can trace it back by looking at the log file and comparing timestamps.

Future improvement would be to make sure each script emits either a simple summary message on success, or a detailed error for any non-zero exit, so we can capture the failure reason directly in the container output.

nballenger:
I've manually deployed the updated crontab and `core/bin/run` files to the NYPL production script runner, and confirmed that it's emitting usable logs to CloudWatch:

```JSON
{
  "timestamp": "2022-04-14T16:54:01+0000",
  "level": "INFO",
  "pid": "27781",
  "script": "search_index_refresh",
  "status": "PID file lock acquired"
}
{
  "timestamp": "2022-04-14T16:54:01+0000",
  "level": "INFO",
  "pid": "27781",
  "script": "search_index_refresh",
  "status": "started"
}
{
  "timestamp": "2022-04-14T16:54:23+0000",
  "level": "INFO",
  "pid": "27781",
  "script": "search_index_refresh",
  "status": "completed"
}
{
  "timestamp": "2022-04-14T16:54:23+0000",
  "level": "INFO",
  "pid": "27781",
  "script": "search_index_refresh",
  "status": "PID file lock released"
}
```

So far I haven't seen any messages at the ERROR or CRITICAL levels, though some WARN messages have come through (WARN is only used for "it's already running, aborting"):

```JSON
{
  "timestamp": "2022-04-14T17:12:01+0000",
  "level": "WARN",
  "pid": "28077",
  "script": "search_index_refresh",
  "status": "search_index_refresh is already running, exiting"
}
{
  "timestamp": "2022-04-14T17:15:01+0000",
  "level": "WARN",
  "pid": "28108",
  "script": "axis_monitor",
  "status": "axis_monitor is already running, exiting"
}
{
  "timestamp": "2022-04-14T17:15:01+0000",
  "level": "WARN",
  "pid": "28102",
  "script": "overdrive_reaper",
  "status": "overdrive_reaper is already running, exiting"
}
```

I'll check in on it daily until we've got this merged, since it may take some time for real errors to show up. When I rewrote the crontab locally to run each of the scripts at a different minute of the hour, all of them completed successfully except for:

* `update_nyt_best_seller_lists` - because I don't have a local integration, I think?
* `work_classify_unchecked_subjects` - This one is more puzzling, and feels like maybe an ORM related bug:

    ```
      File "/home/simplified/circulation/core/scripts.py", line 163, in run
        self.update_timestamp(timestamp_data, start_time, None)
      File "/home/simplified/circulation/core/scripts.py", line 217, in update_timestamp
        self.script_name, Timestamp.SCRIPT_TYPE, self.timestamp_collection,
    AttributeError: 'ReclassifyWorksForUncheckedSubjectsScript' object has no attribute 'timestamp_collection'

    During handling of the above exception, another exception occurred:

      File "/home/simplified/circulation/bin/work_classify_unchecked_subjects", line 13, in <module>
        ReclassifyWorksForUncheckedSubjectsScript().run()
      File "/home/simplified/circulation/core/scripts.py", line 170, in run
        self.update_timestamp(None, start_time, stack_trace)
      File "/home/simplified/circulation/core/scripts.py", line 217, in update_timestamp
        self.script_name, Timestamp.SCRIPT_TYPE, self.timestamp_collection,
    AttributeError: 'ReclassifyWorksForUncheckedSubjectsScript' object has no attribute 'timestamp_collection'
    ```

leonardr (approved):
This adds a lot of useful context and functionality to the crontab. In private conversation we verified that when deployed to a running system, the scripts do run (which they weren't reliably doing before) and that the logs you've added go to the right place. There are still some problems, but they're problems that existed before this much larger problem took over.

leonardr:
The work_classify_unchecked_subjects problem is definitely unrelated -- not an ORM issue but possibly the script is subclassing the wrong `Script` subclass.

We also discussed an issue with the Axis 360 Monitor which means it can't make incremental progress -- it has to process the entire timeline without an HTTP error, or it crashes and has to start over again.

I think a good remediation step would be to verify that every one of our non-deprecated scripts a) doesn't have that kind of crash, b) makes incremental progress even if it's interrupted in the middle of its work, and can pick up from where it left off.

Sometimes b) is just not possible, and in the Axis 360 case it may require some new infrastructure on our end.

PR #1701, opened 2022-04-15
Run coverage and generate reports

keithbauer:
## Description

Update tox commands to run pytest through coverage and generate a command line report and an HTML report. The HTML report is uploaded as an artifact which can be downloaded and viewed. 

## Motivation and Context

[SIMPLY-4051](https://jira.nypl.org/browse/SIMPLY-4051)

## How Has This Been Tested?

Through GitHub Actions.

## Checklist:

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -5,7 +5,7 @@ jobs:
   test-circulation:
     name: Run Circulation Tests
     runs-on: ubuntu-latest
-    timeout-minutes: 40
+    timeout-minutes: 60
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -46,3 +46,4 @@ jobs:
         with:
           name: coverage-report
           path: htmlcov
+          retention-days: 30
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -40,3 +40,9 @@ jobs:
         run: |
           export SIMPLIFIED_STATIC_DIR="$(pwd)/resources"
           tox
+
+      - name: Coverage HTML Report
+        uses: actions/upload-artifact@v3
+        with:
+          name: coverage-report
+          path: htmlcov
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -41,9 +41,8 @@ jobs:
           export SIMPLIFIED_STATIC_DIR="$(pwd)/resources"
           tox
 
-      - name: Coverage HTML Report
-        uses: actions/upload-artifact@v3
+      - name: Upload coverage to Codecov
+        uses: codecov/codecov-action@v2
         with:
-          name: coverage-report
-          path: htmlcov
-          retention-days: 30
+          files: ./coverage.xml
+          verbose: true
--- a/core/requirements-base.txt
+++ b/core/requirements-base.txt
@@ -66,6 +66,7 @@ webpub-manifest-parser==0.0.4
 # Theses dependencies are only used in our tests. Ideally they would get moved out to
 # the dev requirements file, but the mock classes for our tests are included alongside
 # the production classes, so these have to be installed all the time for now.
+coverage==6.2
 mock==3.0.5
 parameterized==0.7.4
 pytest<7.0
--- a/tox.ini
+++ b/tox.ini
@@ -12,8 +12,7 @@ commands =
     coverage erase
     coverage run -m pytest tests core/tests --disable-warnings
 commands_post =
-    coverage report --omit="*test*"
-    coverage html --omit="*test*"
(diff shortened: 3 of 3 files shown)


.github/workflows/test.yml
```diff
@@ -40,3 +40,9 @@ jobs:
         run: |
           export SIMPLIFIED_STATIC_DIR="$(pwd)/resources"
           tox
+
+      - name: Coverage HTML Report
+        uses: actions/upload-artifact@v3
```
keithbauer: The default retention period is 90 days which I think is longer than it needs to be for these artifacts. I'm not sure what a good retention period would be, maybe 7 or 14 days?

leonardr: 14 days is good since it guarantees the artifact is available for the rest of the current sprint. 30 days would also be fine. I agree that 90 is excessive.

leonardr (approved):
This is definitely working and can be merged, but I don't think it's the end of the story for SIMPLY-4051. I say this mainly because the original ticket doesn't make it clear _why_ we want the coverage package integrated. I found the artifact created by the test run [here](https://github.com/NYPL-Simplified/circulation/actions/runs/2173136099), but it's a zipped-up website that's not terribly useful on its own. The information is only useful if we can use it to make decisions. Right now it's similar to if we ran the unit tests with every PR, but didn't check whether the tests pass before merge.

Some requirements like these might flesh out SIMPLY-4051 better:

* Sound an alarm, somewhere in the web page for a given PR, if the PR makes the code coverage percentage drop, because that's usually a reason to send the PR back or keep working on it.
* Publish the coverage report where it can be consumed as a website (on Github Pages?) so that it's easy to review and determine exactly how coverage has dropped.

I'm not sure how those changes would actually work (in particular, how would we know if coverage has _decreased_ vis-a-vis `develop`? We'd need to keep coverage statistics as state in Git, and compare them). And this is a clear step forward, so again, not a reason to hold up this merge. But I don't see how to use this information in a useful way right now, and this is such a common setup that there must be best practices out there.

leonardr:
Are you getting value out of the command-line report that happens when you run pytest during development?

keithbauer:
I don't think the command-line report is very useful, but would provide some insight if you wanted to download the artifact to view it in greater depth.

I couldn't elegantly reconcile hosting both the Sphinx documentation and the HTML report on GitHub Pages so I chose to use the artifact upload for this. If the Sphinx documentation isn't important then I can update this to host the coverage report on GitHub Pages.

I think the primary reason behind 4051 is to have some baseline metrics as we move forward with refactors. I agree that it's not very useful on its own at the moment and I could look into failing a pull request if coverage drops below a specified percentage.

leonardr:
The Github Pages site is an appropriate place for artifacts generated from the _most recent_ version of the software. So that probably means `develop`. Sphinx documentation, absolutely -- people will be looking for it because they're doing work on a client. Coverage reports are less important, but would become more important if we needed to show off that topline number.

For any kind of software check associated with a pull request, we don't really need permanent storage of the artifacts produced by that check, because the PR itself is transitory. We run these checks to automate as much of the job of code review as possible. This includes reporting when tests fail of course, but it also includes reporting new uncovered code or declines in coverage (which saves the reviewer some of the mental work of thinking "is there a test for this?") and (for Sphinx) reporting new bits of code that aren't properly documented.

When the tests fail, the corresponding Github Actions task fails and the branch can't be merged. That's the topline result. To get into which tests failed and why, you have to drill into the action log. I can see "did code coverage decline?" and "is there new undocumented code?" as separate checks that turn red if they fail (inspiring people to drill down into why) but don't necessarily block merge. But how do you measure that stuff and make the action log useful for seeing where the problem is? I've got no clue. It's much more obvious with unit tests.

Basically, don't worry about publishing that big webpage, it's not terribly useful, but the thing that would be useful is outside my experience and I'd need to do some research to see how to get there.

leonardr:
I looked into this a bit and any solution requires storing the "known good" results for `develop` somewhere and comparing the new code against the old. Github Actions isn't really set up for that. There are tools like Codecov, which has a [free Github integration for open source projects](https://about.codecov.io/for/open-source/), which seem like they might do what we want. You might take a look at Codeconv and maybe one or two competing tools, and see which one comes the closest to the goal of "just tell me which untested lines of code I added/edited in this PR."

keithbauer:
I'll look at Codecov and other viable alternatives. From a quick glance it looks more useful than what I've set up here.

leonardr:
I think Codeconv just takes the `coverage` report as input and turns it into comments on the PR, so it would build on your work here -- you'd just be sending the report to Codeconv rather than publishing it.

keithbauer:
Reading through more of the documentation, I think that merging this in will establish the baseline for coverage comparison moving forward.

PR #1702, opened 2022-04-26
OE-259 Store transaction_id in loans.external_identifier

keithbauer:
## Description

When parsing the availability response, store the `transaction_id` as `external_identifier` in `LoanInfo`. Saves this information to the patron's `Loan` so it can be retrieved in future fulfillments, skipping a call to the availability API.

## Motivation and Context

[OE-259](https://jira.nypl.org/browse/OE-259)

## How Has This Been Tested?

Added a new test to verify that what is stored in `loan.external_identifier` matches the `transaction_id`. The test also checks that only 1 request was made (to the availability endpoint).

## Checklist:

### DIFF
--- a/api/axis.py
+++ b/api/axis.py
@@ -435,6 +435,19 @@ class Axis360API(Authenticator, BaseCirculationAPI, HasCollectionSelfTests):
             )
 
         identifier = licensepool.identifier
+        if isinstance(patron, Patron):
+            # Check if the Patron's Loan has
+            # a transaction ID as an external_identifier
+            loan, _ = licensepool.loan_to(patron)
+            if loan and loan.external_identifier:
+                # Skip the availability API call and return fulfillment
+                return Axis360FulfillmentInfo(
+                    api=self, key=loan.external_identifier,
+                    data_source_name=DataSource.AXIS_360,
+                    identifier_type=Identifier.AXIS_360_ID,
+                    identifier=identifier.identifier
+                )
+
         # This should include only one 'activity'.
         activities = self.patron_activity(patron, pin, licensepool.identifier, internal_format)
         for loan in activities:
--- a/api/axis.py
+++ b/api/axis.py
@@ -53,6 +53,7 @@ from core.model import (
     Library,
     LicensePool,
     LinkRelations,
+    Loan,
     MediaTypes,
     Patron,
     Representation,
@@ -438,7 +439,7 @@ class Axis360API(Authenticator, BaseCirculationAPI, HasCollectionSelfTests):
         if isinstance(patron, Patron):
             # Check if the Patron's Loan has
             # a transaction ID as an external_identifier
-            loan, _ = licensepool.loan_to(patron)
+            loan = get_one(self._db, Loan, patron=patron, license_pool=licensepool)
             if loan and loan.external_identifier:
                 # Skip the availability API call and return fulfillment
                 return Axis360FulfillmentInfo(
@@ -464,7 +465,9 @@ class Axis360API(Authenticator, BaseCirculationAPI, HasCollectionSelfTests):
 
             if isinstance(patron, Patron) and loan.external_identifier:
                 # Save the external_identifier to the Patron's Loan for future retrieval
-                licensepool.loan_to(patron, external_identifier=loan.external_identifier)
+                patron_loan = get_one(self._db, Loan, patron=patron, license_pool=licensepool)
+                if patron_loan:
+                    patron_loan.external_identifier = loan.external_identifier
 
             return fulfillment
         # If we made it to this point, the patron does not have this
--- a/api/axis.py
+++ b/api/axis.py
@@ -1553,7 +1553,8 @@ class AvailabilityResponseParser(ResponseParser):
                 identifier_type=self.id_type,
                 identifier=axis_identifier,
                 start_date=start_date, end_date=end_date,
-                fulfillment_info=fulfillment
+                fulfillment_info=fulfillment,
+                external_identifier=transaction_id,
             )
 
         elif reserved:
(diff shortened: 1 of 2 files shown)


api/axis.py
```diff
@@ -435,6 +435,19 @@ def fulfill(self, patron, pin, licensepool, internal_format, **kwargs):
             )
 
         identifier = licensepool.identifier
+        if isinstance(patron, Patron):
+            # Check if the Patron's Loan has
+            # a transaction ID as an external_identifier
+            loan, _ = licensepool.loan_to(patron)
```
leonardr: This will create a local Loan object starting at the current moment, even if no loan actually exists. Instead, use `get_one(_db, Loan, patron=patron, licensepool=licensepool)` to fetch a loan only if it already exists on our side. If it doesn't, then rely on the availability call to either find it (if it was somehow created on the Axis side and we didn't hear about it) or raise an exception.

api/axis.py
```diff
@@ -448,6 +461,11 @@ def fulfill(self, patron, pin, licensepool, internal_format, **kwargs):
             fulfillment = loan.fulfillment_info
             if not fulfillment or not isinstance(fulfillment, FulfillmentInfo):
                 raise CannotFulfill()
+
+            if isinstance(patron, Patron) and loan.external_identifier:
+                # Save the external_identifier to the Patron's Loan for future retrieval
+                licensepool.loan_to(patron, external_identifier=loan.external_identifier)
```
leonardr: This usage of `loan_to` isn't as bad because if you were able to fulfill a loan, you really should have a local Loan object by now. But I do think it would be better to fetch the loan (using get_one like in my earlier suggestion) and skip this step if for whatever reason (maybe a race condition) it's not found in the database. If nothing else, calling `loan_to` overwrites `Loan.start` and makes it look like the loan started right now.

leonardr:
It looks like you get that test failure because the new code is working! Since fulfillment succeeded the first time (earlier in the test), it doesn't matter that Axis didn't provide any fulfillment information, because the fulfillment ID was cached and we never even requested it the second time. I think by changing the comments and adding some more assertions this can become a test of the new functionality:

1. Verify that we never made the authentication call.
2. Verify that we get fulfillment information, rather than just that no exception is raised.
3. Verify that the fulfillment ID from the previous availability call is cached in Loan.external_identifier.

Of course, then the "Axis doesn't provide any fulfillment information" case needs to be moved to a separate test, where Axis forgets to provide this information the _first_ time we try to fulfill a book -- in that case we can't get started because there's no data to cache.

api/axis.py
keithbauer: I've updated this to use `get_one` instead of `licensepool.loan_to()`, definitely don't want to be creating a new loan here.

api/axis.py
```diff
@@ -448,6 +462,13 @@ def fulfill(self, patron, pin, licensepool, internal_format, **kwargs):
             fulfillment = loan.fulfillment_info
             if not fulfillment or not isinstance(fulfillment, FulfillmentInfo):
                 raise CannotFulfill()
+
+            if isinstance(patron, Patron) and loan.external_identifier:
+                # Save the external_identifier to the Patron's Loan for future retrieval
+                patron_loan = get_one(self._db, Loan, patron=patron, license_pool=licensepool)
+                if patron_loan:
+                    patron_loan.external_identifier = loan.external_identifier
```
leonardr, line 470: I was wondering why this code was necessary, because some part of me had assumed this would be taken care of in the API-agnostic code in circulation.py. After looking into it I do think this code is still necessary, but there's a big optimization opportunity in circulation.py, specifically [right here](https://github.com/NYPL-Simplified/circulation/blob/develop/api/circulation.py#L1285).

We can go over this live if you want, since I realize this isn't code we've talked about before, but: `sync_bookshelf` is called periodically by the client refreshing its view of the end-user's loans. It works by calling out to every API that's integrated into the system. Those APIs give it a bunch of `LoanInfo` objects, and it makes sure that the patrons' entries in `loans` are created, deleted, or updated appropriately.

For Axis 360 API, "calling out to the API" means making a call to `availability/v2`. You're changing the `AvailabilityResponseParser` to attach `external_identifier` to every `LoanInfo` that gets parsed.

So, let's say you change `sync_bookshelf` to set `external_identifier` on a `Loan` if there's one in the corresponding `LoanInfo`, similar to how we already set `locked_to` on the `Loan` if the corresponding `LoanInfo` is locked. Now a `sync_bookshelf` call sets `external_identifier` on _every_ active loan that doesn't already have one set. This potentially saves an extra individual `availability/v2` call for each of those loans.

leonardr: Since the known client behavior is to borrow a book and then immediately try to fulfill it (which will set external_identifier for that book), I think this will only improve things in marginal cases. Unless you want to learn more about that part of the code right now, I'm OK with not doing this work in this branch.

tests/test_axis.py
```diff
@@ -543,6 +538,70 @@ def fulfill(internal_format="not AxisNow"):
...
+        )
+
+        patron = self._patron()
+        patron.authorization_identifier = "a barcode"
+        # Ensure a Loan exists for this patron for future calls
+        create(self._db, Loan, patron=patron, license_pool=pool) 
+        data = self.sample_data("availability_with_axisnow_fulfillment.xml")
+        data = data.replace(b"0016820953", pool.identifier.identifier.encode("utf8"))
+        self.api.queue_response(200, content=data)
+        first_fulfillment = self.api.fulfill(patron, "pin", pool, "AxisNow")
```
leonardr, line 564: Verify that the external_identifier has been set after the first fulfillment, rather than waiting for the end of the test.

leonardr (approved):

api/axis.py
keithbauer: This would be good to add. I did a quick attempt to see if any of the tests from `test_circulationapi.py` had any loans with an `external_identifier` set, but no loans had one set. I think the work for this could be done in a future branch / ticket.

leonardr: The only other API that _might_ do this is ODL. The `external_identifier` was added to the `loans` table specifically for ODL.

PR #1704, opened 2022-04-29
OE-260 Cache result of "getFulfillmentInfo" call

keithbauer:
## Description

This adds two new fields to the Loans table, `cached_manifest` and `cached_content_type`. A `can_cache_manifest` flag has been added to `FulfillmentInfo` and is set to `False` by default. Axis360 fulfillments have this flag enabled, allowing for the manifest and content_type to be stored in the patron's loan.

When fulfilling a loan in `CirculationAPI.fulfill()`, a `FulfillmentInfo` will be created with the cached manifest and content_type if available, skipping additional API calls.  If there is no cached manifest then fulfillment will as proceed to make those API calls as it did before.

## Motivation and Context

[OE-260](https://jira.nypl.org/browse/OE-260)

## How Has This Been Tested?

Two new tests have been added. A test to fulfill a loan when the loan has a cached manifest. A test to fulfill a loan that does not currently have a cached manifest, but the manifest is allowed to be cached. Additionally, new assertions have been added to ensure that only AxisNow manifests can be cached.

I have also tested this manually by setting up an Axis 360 collection locally, borrowing a book, then fulfilling the (AxisNow) book. Checking the database after fulfillment shows that the transaction ID was cached to the loan's external_identifier, the manifest was cached to cached_manifest, and the content_type was cached to cached_content_type.

## Checklist:

### DIFF
--- a/api/axis.py
+++ b/api/axis.py
@@ -442,12 +442,14 @@ class Axis360API(Authenticator, BaseCirculationAPI, HasCollectionSelfTests):
             loan = get_one(self._db, Loan, patron=patron, license_pool=licensepool)
             if loan and loan.external_identifier:
                 # Skip the availability API call and return fulfillment
-                return Axis360FulfillmentInfo(
+                fulfillment = Axis360FulfillmentInfo(
                     api=self, key=loan.external_identifier,
                     data_source_name=DataSource.AXIS_360,
                     identifier_type=Identifier.AXIS_360_ID,
                     identifier=identifier.identifier
                 )
+                fulfillment.can_cache_manifest = True
+                return fulfillment
 
         # This should include only one 'activity'.
         activities = self.patron_activity(patron, pin, licensepool.identifier, internal_format)
@@ -469,6 +471,7 @@ class Axis360API(Authenticator, BaseCirculationAPI, HasCollectionSelfTests):
                 if patron_loan:
                     patron_loan.external_identifier = loan.external_identifier
 
+            fulfillment.can_cache_manifest = True
             return fulfillment
         # If we made it to this point, the patron does not have this
         # book checked out.
--- a/api/axis.py
+++ b/api/axis.py
@@ -448,7 +448,6 @@ class Axis360API(Authenticator, BaseCirculationAPI, HasCollectionSelfTests):
                     identifier_type=Identifier.AXIS_360_ID,
                     identifier=identifier.identifier
                 )
-                fulfillment.can_cache_manifest = True
                 return fulfillment
 
         # This should include only one 'activity'.
@@ -471,7 +470,6 @@ class Axis360API(Authenticator, BaseCirculationAPI, HasCollectionSelfTests):
                 if patron_loan:
                     patron_loan.external_identifier = loan.external_identifier
 
-            fulfillment.can_cache_manifest = True
             return fulfillment
         # If we made it to this point, the patron does not have this
         # book checked out.
@@ -1813,3 +1811,6 @@ class Axis360FulfillmentInfo(APIAwareFulfillmentInfo):
         self._content = str(manifest)
         self._content_type = manifest.MEDIA_TYPE
         self._content_expires = expires
+
+        if manifest.MEDIA_TYPE == AxisNowManifest.MEDIA_TYPE:
+            self.can_cache_manifest = True
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -993,7 +993,7 @@ class CirculationAPI(object):
                     part=part, fulfill_part_url=fulfill_part_url
                 )
 
-                if getattr(fulfillment, "can_cache_manifest", False):
+                if fulfillment.can_cache_manifest:
                     loan.cached_manifest = fulfillment.content.encode("raw_unicode_escape")
                     loan.cached_content_type = fulfillment.content_type
 
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -993,14 +993,16 @@ class CirculationAPI(object):
                     part=part, fulfill_part_url=fulfill_part_url
                 )
 
-                if loan and fulfillment.can_cache_manifest:
-                    loan.cached_manifest = fulfillment.content.encode("raw_unicode_escape")
-                    loan.cached_content_type = fulfillment.content_type
-
             if not fulfillment or not (
                 fulfillment.content_link or fulfillment.content
             ):
                 raise NoAcceptableFormat()
+
+            # Now that the additional API calls have been made,
+            # fulfillment data can be saved to the loan for future retrieval
+            if loan and fulfillment.can_cache_manifest:
+                loan.cached_manifest = fulfillment.content.encode("raw_unicode_escape")
+                loan.cached_content_type = fulfillment.content_type
 
         # Send out an analytics event to record the fact that
         # a fulfillment was initiated through the circulation
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -993,7 +993,7 @@ class CirculationAPI(object):
                     part=part, fulfill_part_url=fulfill_part_url
                 )
 
-                if fulfillment.can_cache_manifest:
+                if loan and fulfillment.can_cache_manifest:
                     loan.cached_manifest = fulfillment.content.encode("raw_unicode_escape")
                     loan.cached_content_type = fulfillment.content_type
 
--- a/api/circulation.py
+++ b/api/circulation.py
@@ -163,6 +163,7 @@ class FulfillmentInfo(CirculationInfo):
     """A record of a technique that can be used *right now* to fulfill
     a loan.
     """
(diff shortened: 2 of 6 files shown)


core/model/patron.py
```diff
@@ -511,6 +511,8 @@ class Loan(Base, LoanAndHoldMixin):
     # Some distributors (e.g. Feedbooks) may have an identifier that can
     # be used to check the status of a specific Loan.
     external_identifier = Column(Unicode, unique=True, nullable=True)
+    cached_manifest = Column(Unicode, unique=True, nullable=True)
```
mwbenowitz: Does it make sense to use a different data type for this column? I'm assuming that we'll only ever be retrieving this and never doing any query operations with it, so maybe a `blob` or something else might be more efficient?

Also should `unique` be true?

core/model/patron.py
```diff
@@ -511,6 +511,8 @@ class Loan(Base, LoanAndHoldMixin):
...
+    cached_content_type = Column(Unicode, unique=True, nullable=True)
```
mwbenowitz: I feel like `unique` shouldn't be true here? A user should be able to have multiple loans with the same format.

keithbauer: I think you're right. I'll remove the unique constraint.

core/model/patron.py
leonardr: Yes, this should be a `Binary` field. The model I was thinking of here was `Representation.content`. 

I considered suggesting that this be done by connecting `Loan` to `Representation`, but that opens up a can of worms -- so far `Representation` has only been used for documents that can be made public, like cover images.

core/migration/20220427-add-loan-cached-manifest-content-type.sql
```diff
@@ -0,0 +1,17 @@
+ BEGIN
+  -- Add the 'cached_manifest' column
+  BEGIN
+   ALTER TABLE loans ADD COLUMN cached_manifest varchar;
```
leonardr: When you change the data type of this field, be sure to change the migration script as well.

api/axis.py
```diff
@@ -469,6 +471,7 @@ def fulfill(self, patron, pin, licensepool, internal_format, **kwargs):
                 if patron_loan:
                     patron_loan.external_identifier = loan.external_identifier
 
+            fulfillment.can_cache_manifest = True
```
leonardr: I think it's better to set `can_cache_manifest` right at the point where you set `content` to a cacheable value. That would be in `Axis360FulfillmentInfo.do_fetch`. Right you you should only do this for AxisNow manifests,which means `if manifest.MEDIA_TYPE==AxisNowManifest.MEDIA_TYPE`.

api/axis.py
```diff
@@ -442,12 +442,14 @@ def fulfill(self, patron, pin, licensepool, internal_format, **kwargs):
...
             if loan and loan.external_identifier:
                 # Skip the availability API call and return fulfillment
-                return Axis360FulfillmentInfo(
+                fulfillment = Axis360FulfillmentInfo(
                     api=self, key=loan.external_identifier,
                     data_source_name=DataSource.AXIS_360,
                     identifier_type=Identifier.AXIS_360_ID,
                     identifier=identifier.identifier
                 )
+                fulfillment.can_cache_manifest = True
```
leonardr: At this point you don't have a manifest yet, so I think it's premature to set `can_cache_manifest`.

api/circulation.py
```diff
@@ -966,17 +967,36 @@ def fulfill(self, patron, pin, licensepool, delivery_mechanism, part=None, fulfi
...
+                # impact on implementation signatures. Most vendor APIs
+                # will ignore one or more of these arguments.
+                fulfillment = api.fulfill(
+                    patron, pin, licensepool, internal_format=internal_format,
+                    part=part, fulfill_part_url=fulfill_part_url
+                )
+
+                if getattr(fulfillment, "can_cache_manifest", False):
+                    loan.cached_manifest = fulfillment.content
+                    loan.cached_content_type = fulfillment.content_type
```
leonardr: `getattr` is often a code smell to me. Since you're using `can_cache_manifest` to transmit information between modules, it should be made an official part of the `FulfillmentInfo` API, with a default value of False.

keithbauer: Some of the tests were failing since they had a `LicensePoolDeliveryMechanisms` that was acting as `fulfillment` here instead of a `FulfillmentInfo`. Due to that I wasn't sure if a `FulfillmentInfo` is always present.

keithbauer: I'll look into this further and double check. I had some that were failing yesterday, but just changed this locally and it seems ok.

tests/test_circulationapi.py
```diff
@@ -929,6 +931,45 @@ def yes_we_can(*args, **kwargs):
...
+
+        def mock_axis_api(licensepool):
+            api = MockAxis360API(self._db, self.collection)
+            api.queue_response(
+                200, {}, sample_data("availability_with_axisnow_fulfillment.xml", "axis"))
+            api.queue_response(
+                200, {}, sample_data("ebook_fulfillment_info.json", "axis"))
+            return api
+
+        self.circulation.api_for_license_pool = mock_axis_api
```
leonardr: This needs to be split up into a few different tests, some in api/axis.py and some here in circulation.py.

For axis.py, you need to verify that an outgoing AxisNow manifest has `can_cache_manifest` set to True, but that Adobe and Findaway manifests have the field set to False. You should be able to do this by adding assertions to the existing tests, instead of writing new ones.

For the test of `CirculationAPI`, you don't need any AxisNow-specific details, because the only thing that matters is whether `can_cache_manifest` is true. You need to test three cases:

* If there's already a Loan object with `.cached_manifest` set, then `fulfill` returns a `FulfillmentInfo` with the appropriate media type and content, without making any API requests at all.
* If there's no Loan object, then `fulfill` makes the API call and sets `Loan.cached_manifest` and `Loan.cached_content_type`, but only if `Fulfillment.can_cache_manifest` is true. You can queue the FulfillmentInfo object directly with `self.remote.queue_fulfillment` (lots of examples in test_circulationapi.py) and you don't have to set up a mock API.
* If there's no Loan object and `Fulfillment.can_cache_manifest` is false, then `fulfill` leaves `Loan.cached_manifest` and `Loan.cached_content_type` alone. You should be able to add assertions to an existing test, since the default behavior is that `Fulfillment.can_cache_manifest` is False.

api/circulation.py
leonardr: I see... that sounds a little weird to me, so it could be a problem with the test or the code. Let me know what you find.

keithbauer: Instead of `if getattr(...):` I tried using `if fulfillment.can_cache_manifest:` and these tests failed:
* tests/test_circulationapi.py::TestCirculationAPI::test_fulfill - AttributeError: 'LicensePoolDeliveryMechanism' object has no attribute 'can_cache_manifest'
* tests/test_circulationapi.py::TestCirculationAPI::test_fulfill_without_loan - AttributeError: 'LicensePoolDeliveryMechanism' object has no attribute 'can_cache_manifest'

I think these are failing due to fulfillment being set as `fulfillment = self.pool.delivery_mechanisms[0]` which sets it as a`LicensePoolDeliveryMechanism` instead of a `FulfillmentInfo`. I can update these tests to use `FulfillmentInfo` instead.

leonardr: Yeah, it looks like you exposed an error or shortcut in those tests that we can do without.

keithbauer: I've updated these tests to use a `FulfillmentInfo` and replaced the `getattr()`.

api/circulation.py
```diff
@@ -993,15 +993,17 @@ def fulfill(self, patron, pin, licensepool, delivery_mechanism, part=None, fulfi
...
                 fulfillment.content_link or fulfillment.content
             ):
                 raise NoAcceptableFormat()
 
+            # Now that the additional API calls have been made,
+            # fulfillment data can be saved to the loan for future retrieval
+            if loan and fulfillment.can_cache_manifest:
+                loan.cached_manifest = fulfillment.content.encode("raw_unicode_escape")
+                loan.cached_content_type = fulfillment.content_type
+
```
leonardr, line 1006: Did this fix the problem you were seeing? I don't understand why it would have.

keithbauer: Yes, I think it's because `fulfillment.can_cache_manifest` isn't set to `True` until `fulfillment.content` is accessed, which kicks off `Axis360FulfillmentInfo.do_fetch()` that sets `fulfillment.can_cache_manifest=True`.

If this block of code fits better where it was then I think something like `if loan and fulfillment.content and fulfillment.can_cache_manifest:` would work too.

leonardr: No, your explanation makes sense.

leonardr (approved):

PR #1705, opened 2022-05-02
OpenAPI documentation for /admin/* endpoints

mwbenowitz:
## Description

This is an early working draft of adding OpenAPI support to the circulation-manager endpoints. It includes documentation for some endpoints as an example and can be extended to cover all API endpoints with the addition of further docstrings (and relevant schemas).

There is room for a lot of improvement in the code:

- the OpenAPI spec is generated for each request. This should be generated when the container comes up and stored somewhere. (there are circular dependencies that make this difficult at the moment)
- The schemas generated currently happen in one large method. This could/should be refactored to be more manageable
- Authentication is not currently correctly configured

This is a test PR to solitit feedback and evaluate if this is a viable way of documenting the API structure

## Motivation and Context

This is intended to help unblock/speed other work that is dependent on understanding the structure of the circulation-manager, and especially the admin endpoint structure.

## How Has This Been Tested?

I've tested the code by running the app locally and verifying that the SwaggerUI documentation comes up and includes the intended descriptions. I have not validated actually running any requests via OpenAPI/Swagger

## Checklist:


keithbauer:
I think this is a great start! I was able to add localhost to the servers, add a BasicAuth security schema, and test the GET endpoint. I had some CSRF token issues with the POST and DELETE endpoints. I can see the `routes.py` getting large with these docstrings for every route, but I don't think that should be a deterrent.

mwbenowitz:
@keithbauer Check out the latest commit for how I'm dealing with the CSRF issues. As with a lot of stuff here this remains experimental and could be improved.

api/admin/routes.py
```diff
@@ -729,6 +950,86 @@ def discovery_service_library_registrations():
...
+        201:
+          description: A text response containing only the id of a newly created list
+          content:
+            text/html:
+              example: 1234
+        4XX:
+          description: |
+            One of the following endpoint-specific errors
+            * `AdminNotAuthorized`: Current admin cannot perform this operation for this library
+            * `MISSING_CUSTOM_LIST`: No list could be found for provided identifier
```
leonardr, line 1088: I don't know how to address this problem, but these are different types of things. AdminNotAuthorized is an exception class that can be converted into a ProblemDetail object, and MISSING_CUSTOM_LIST is a ProblemDetail constant.

As you note, the thing that actually gets returned is a ProblemResponse containing the information conveyed by a ProblemDetail object. I don't know what to do here since that ProblemDetail could have been one of the constants, derived from one of the constants, instantiated via an exception,  or instantiated some other way.

1. The distinguishing feature of a ProblemDetail is the URI which uniquely identifies the type of problem, so maybe that should be the identifying factor.
2. The ProblemError class (in core/util/problem_detail.py) might be a useful way to think about this. Maybe making it so it can be treated _either_ as a raised exception _or_ a `ProblemResponse` would simplify things.

leonardr (commented):
Re: the tests, I feel like unit test coverage for that code isn't necessary because we have a the doc build process also runs on every branch and exercises that code, effectively forming a second test suite. Does that sound right to you?

If you can write tests for that, then great, but it seems like it requires combining the code with the actual docstrings in use, in which case you're recreating the doc build process.

api/admin/routes.py
mwbenowitz: I think using the URI as the identifier for the type of response would be the best bet, since that is more descriptive and unique. I tried, for all of these endpoints, to follow the logic as far as I could in looking what could raise (anticipated) exceptions. But I missed the `ErrorHandler` which catches unanticipated errors. So I think for these codes I would update this section to be:
- 403: An authentication/permission error that returns an HTML page
- 4XX: Anticipated exceptions identified by URI
- 5XX: Unanticipated error that can return a `ProblemResponse`
- 5XX: Unanticipated error due to a bug that returns a text resopnse

mwbenowitz:
@leonardr That makes sense, I might add some basic tests around the new controller I created since I added some helper methods that build OpenAPI schemas and components. I think that unit testing that will help ensure that the controller class is extensible and supports the use case that we need.

leonardr (approved):
This has shaped up really well and I think it's ready to merge.

I'd like you to explore the possibility of centralizing the documentation about error conditions even more than you have. What you have seems repetitive and a prime candidate for refactoring, assuming further refactoring is possible.

mwbenowitz:
@leonardr yeah, I think the `responses` blocks are prime targets and should be easy to refactor out to the controller.

PR #1708, opened 2022-05-05
Add logging for `initialize_database` and `migrate_database` scripts

keithbauer:
## Description

Fixes the path for `db_init_script` to point to the core bin dir where the `initialize_database` script currently resides. Runs `initialize_database` and `migrate_database` with `core/bin/run` which does a bunch of stuff but mainly interested in the logging format that it establishes.

## Motivation and Context

[SIMPLY-4071](https://jira.nypl.org/browse/SIMPLY-4071) During a recent deploy the migrations failed but there was no logged output for the failure. This adds logging for the `docker-entrypoint.sh` script.

## How Has This Been Tested?

Tested this locally by doing `make full-clean` then `make up` and monitoring the logging output in docker desktop. I can see log entries for `initialize_database` and `migrate_database`.

## Checklist:

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -256,7 +256,7 @@ RUN chmod 664 /etc/cron.d/circulation \
  && crontab /etc/cron.d/circulation \
  && touch /var/log/cron.log
 
-CMD ["scripts"]
+CMD ["scripts", "|& tee -a /var/log/cron.log 2>$1"]
 
 ###############################################################################
 ## cm_scripts_local - local dev version of scripts, relies on host mounted code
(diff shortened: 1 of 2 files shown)


nyplwesclark (approved):

mwbenowitz (approved):
These changes look good to me. For the update to the Dockerfile, is that necessary for logging the database scripts? Or is it a related change to improve script logging more generally?

keithbauer:
I moved the log redirection from `docker-entrypoint.sh` to `Dockerfile` since it looks cleaner and accomplishes the same thing (at least locally). Next step would be to deploy this to a QA environment and make sure the logs show up in CloudWatch.

PR #1709, opened 2022-05-06
Fix docker elasticsearch for non-arm64 

keithbauer:
## Description

Manually run `/usr/local/bin/docker-entrypoint.sh` in the `command:` section in `docker-comopse.yml`. Also includes some `README.md` cleanup.

## Motivation and Context

Due to the differences between arm64 and amd64, the elasticsearch container needed to be handled differently. The arm64 builds uses `docker/Dockerfile-es654.arm64` to build the image while amd64 pulls the elasticsearch image from dockerhub. Due to this the arm64 Dockerfile can install elasticsearch plugins, but the amd64 version uses a command in `docker-compose.yml` to install the plugin. It seems like the entrypoint script wasn't getting ran so I've added it to the command. I'm not sure if it also needs to be added to `docker-compose.cicd.yml`.

## How Has This Been Tested?

Tested on my windows machine and @oliviawongnyc tested on her non-m1 mac.

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -62,7 +62,6 @@ The following shell commands will clone this repo and build Docker images for th
 ```shell
 git clone https://github.com/NYPL-Simplified/circulation.git
 cd ./circulation
-make setup
 make build
 ```
 
--- a/README.md
+++ b/README.md
@@ -31,7 +31,7 @@ Branches whose name follows the pattern `<organization>-deploy-<env>` are protec
 * `nypl-deploy-qa`
 * `nypl-deploy-production`
 * `openebooks-deploy-qa`
-* `openebooks-deploy-qa`
+* `openebooks-deploy-production`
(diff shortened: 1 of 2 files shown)


mwbenowitz:
@keithbauer So this might be a bit of a left-field solution but I did some poking around and it turns out that bitnami publishes ES images that closely track the official ones but with some extra tooling around them: https://registry.hub.docker.com/r/bitnami/elasticsearch

In that is the ability to specify plugins you want install as env variables, which then get installed when the container starts up.

Since we're already pulling a bitnami image for minio, maybe this wouldn't be so bad? Just did a quick test and it seemed to solve the problem for me.

leonardr (approved):
Very happy with the simplification of moving to the bitnami image. I tested this on my AMD machine and am happy with the results.

PR #1710, opened 2022-05-09
Don't return expired HTTP Basic tokens

keithbauer:
## Description

When returning an existing HTTP Basic token, ensure that the remaining token time is between the upper boundary of 1 hour and lower boundary of 59 minutes. Use `.total_seconds()` instead of `.seconds` to get a proper duration that takes days into account.

## Motivation and Context

Expired HTTP Basic tokens were being returned in QA, this fix will now properly generate new tokens. When using `.seconds` days were not taken into account when checking whether to return an existing token or generate a new one.

## How Has This Been Tested?

Parametrized an existing test that returns new tokens to include a request at a time beyond the lifespan of a token. Test was failing locally before changes were implemented and passed afterwards.

## Checklist:

### DIFF
--- a/tests/test_authenticator.py
+++ b/tests/test_authenticator.py
@@ -2950,7 +2950,15 @@ class TestBasicAuthTempTokenController(AuthenticatorTest):
             patron = self.controller.authenticator.authenticated_patron(self._db, headers_bearer)
             assert 'unittestuser' == patron.username
 
-    def test_basic_auth_temp_token_returns_existing(self):
+    @pytest.mark.parametrize(
+        'delta',
+        [
+            pytest.param(1, id="after_1_second"),
+            pytest.param(59, id="after_59_seconds"),
+            pytest.param(60, id="after_60_seconds"),
+        ]
+    )
+    def test_basic_auth_temp_token_returns_existing(self, delta):
         """"
         GIVEN: A request to authenticate a patron with a base64 encoded username:password to recieve a token
         WHEN:  Re-requesting authentication with the same credentials within a minute of the token's creation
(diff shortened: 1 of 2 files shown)


tests/test_authenticator.py
```diff
@@ -2975,7 +2975,15 @@ def test_basic_auth_temp_token_returns_existing(self):
                 another_token = another_response.json.get('access_token')
                 assert another_token == token
 
-    def test_basic_auth_temp_token_returns_new_token(self):
+
+    @pytest.mark.parametrize(
+        'delta',
+        [
+            pytest.param(61, id='after_one_minute'),
+            pytest.param(3601, id='after_one_hour')
```
mwbenowitz: Shouldn't the authenticator return an error if trying to retrieve after more than one hour? Or does changing the `frozen_time` reset things in a way that I'm not following?

keithbauer: You're right that an error is raised, I think it does it [here](https://github.com/NYPL-Simplified/circulation/blob/develop/api/authenticator.py#L825). I don't think this is a good/accurate test now, I'll revert it.

mwbenowitz: okay, I see, so the check for an expired token happens there, _before_ we get to this method. So we can assume that this method is only dealing with the renewal of a valid token or creation of a new one. Makes sense to eliminate this test here then, unless we want to make sure that this also guards against that.

leonardr: I think this test was fine. A controller gated by an authentication token should raise an error when you send it an expired token (e.g. one more than an hour old), but this controller isn't of that type. This controller doesn't take a token for authentication, it only takes username/password. It uses that information to _issue_ a token, and the test is verifying that if the token already in the database is a _minute_ old, this controller issues a new token instead of recycling the old one.

leonardr:
The code looks right, but you do need a test to verify that the problem Sam reported has been fixed.

Fortunately you already have tests that make the right assertions, they just need to be parameterized, as you did initially (although you need to go further than you did initially).

First, `test_basic_auth_temp_token_returns_existing`
`WHEN Re-requesting authentication with the same credentials within a minute of the token's creation`

Parameterize this to make the second request:

1. after 1 second
2. after 59 seconds
3. after 60 seconds

Second, `test_basic_auth_temp_token_returns_new`
`WHEN:  Re-requesting authentication with the same credentials after a minute of the token's creation`

Parameterize this to make the second request:
4. After 1 minute and 1 second
5. After 60 minutes and 1 second
6. After 1 day and 1 second

My understanding is that the problem reported here is case #6, because of the following timedelta quirk (which has bitten me many times):

```
one_day = datetime.timedelta(days=1, seconds=1)
one_day.seconds
# 1
one_day.total_seconds()
# 86401.0
```

leonardr (approved):

PR #1711, opened 2022-05-11
Fixed misplaced parentheses in a bin/informational script.

leonardr:
## Description

A minor change to fix a script I plan to use in a demo.

## Motivation and Context

Nick changed this script in May to fix it, but it looks like he missed some parentheses. The parentheses are used to extract the part of the Overdrive API call we actually need -- the JSON document sent by the server.

## How Has This Been Tested?

Manually tested. Since this is a utility script we don't have good test coverage for it. You can test this by hooking up the branch to NYPL's QA circulation manager database and issuing a command like this:

```
bin/informational/overdrive-raw-circulation --identifier-type="Overdrive ID" 1dc5b1c1-79f8-42ad-8e1e-912a37a8f835
```

For the record, the ignored variables are an ID mapping, the HTTP response code, and the HTTP response headers.

## Checklist:

### DIFF
--- a/bin/informational/overdrive-raw-circulation
+++ b/bin/informational/overdrive-raw-circulation
@@ -19,7 +19,7 @@ class OverdriveRawCirculationScript(IdentifierInputScript):
         for collection in Collection.by_protocol(self._db, ExternalIntegration.OVERDRIVE):
             overdrive = OverdriveAPI(self._db, collection)
             for identifier in args.identifiers:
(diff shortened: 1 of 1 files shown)


nyplwesclark:
Hey Leonard. The fix for this line has been merged. I saw you tested against the QA db, but I wanted to check with you since the local tests and building api documentation are failing on github. Were you either expecting that or already noticing it? Let me know if you want me to look more into it.

PR #1717, opened 2022-05-24
SIMPLY-4089 Add Unsubscribe link option

mwbenowitz:
## Description

This adds an option for a library to have an (optional) link to unsubscribe to email notifications/communications. This is part of the `Patron Support` page and returns the URI as part of the `links` array as part of the `authentication_document`.

In addition this adds a new convenience method `_help_uri_with_fallback` that will return a help URI, falling back to the general help link configured for the library. This compliments the `_email_uri_with_fallback` method.

In addition one test is updated to reflect the new state of the `authentication_document`.
## Motivation and Context

This change is necessary to meet new iOS requirements for email subscription management as described in [SIMPLY-4089](https://jira.nypl.org/browse/SIMPLY-4089)

## How Has This Been Tested?

I updated the test suite to validate that the output appears as desired and updated this variable for my local testing configuration, validating it can be created in the circ admin and returned in an authentication document.

## Checklist:

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -989,6 +989,13 @@ class LibraryAuthenticator(object):
                 )
             )
 
+        # If there is an unsubscribe link, add it here
+        unsubscribe_uri = Configuration.unsubscribe_email_uri(library)
+        if unsubscribe_uri:
+            links.append(
+                dict(rel=Configuration.HELP_UNSUBSCRIBE_URI, href=unsubscribe_uri)
+            )
+
         # Add a rel="help" link for every type of URL scheme that
         # leads to library-specific help.
         for type, uri in Configuration.help_uris(library):
--- a/api/config.py
+++ b/api/config.py
@@ -165,6 +165,10 @@ class Configuration(CoreConfiguration):
     # OPDS documents but are not registered with IANA.
     AUTHENTICATION_FOR_OPDS_LINKS = ['register']
 
+    # The link relation used to indicate the related link provides marketing email
+    # unsubscribe options. This is distinguished from a "1-click" unsubscribe option
+    HELP_UNSUBSCRIBE_URI = "http://librarysimplified.org/rel/email/unsubscribe/options"
+
     # We support three different ways of integrating help processes.
     # All three of these will be sent out as links with rel='help'
     HELP_EMAIL = 'help-email'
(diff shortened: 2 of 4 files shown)


leonardr (changes requested):
This looks good. I have two comments, both around the idea that the generic help URI can be a fallback for the unsubscribe URI.

First, has product management signed off on this idea? If libraries aren't aware that we're offering their help URIs for "unsubscribe" purposes, they might be confused by some of the emails they get, especially libraries that don't send their patrons email at all. I want to make sure we're ready to communicate the new Apple requirements to our partners. If an "unsubscribe" URL is a hard requirement, I think this is the right decision, I just want to make sure.

Second, the _email_url_with_fallback method needs a test of the fallback(s). If you do that I don't think you need to test that unsubscribe_email_url also falls back.

killerpuppytails:
> This looks good. I have two comments, both around the idea that the generic help URI can be a fallback for the unsubscribe URI.
> 
> First, has product management signed off on this idea? 

I hadn't, and I don't think it's needed, but I'm willing to be convinced.

> If libraries aren't aware that we're offering their help URIs for "unsubscribe" purposes, they might be confused by some of the emails they get, especially libraries that don't send their patrons email at all. I want to make sure we're ready to communicate the new Apple requirements to our partners. If an "unsubscribe" URL is a hard requirement, I think this is the right decision, I just want to make sure.

To be clear, other libraries don't need to use this. The same way they don't need to add Copyright URLs. If they add a Patron registration URL, I'd strongly recommend it, but it's not required.

mwbenowitz:
After reviewing this I think there are a few issues here (one of which might be an issue with the circ-admin interface):

1. 100% agree that tests need to be added for the fallbacks. Will do that!
2. Assuming that all/most libraries have a help URI, this will frequently fall back on to that value. I think Leonard's question is well taken there since I don't think I fully considered that
3. Right now it seems that there is a bug in the admin interface where I cannot _delete_ any value once entered. I can update a value, but I cannot completely remove the unsubscribe URI once I've provided it. This seems to be coming from the front-end app, by the time the value gets to the API layer it's been reset to the old value. I don't think this blocks this work but wanted to bring it up.

killerpuppytails:
> 2. Assuming that all/most libraries have a help URI, this will frequently fall back on to that value. I think Leonard's question is well taken there since I don't think I fully considered that

A few things:
- A help email is a required field in the app, so all libraries will have one.
- The only libraries for whom this unsubscribe feature will display in the front end are those who offer library card signup within the app. No other library will display this feature, so it won't be an issue for them. 

> 3. Right now it seems that there is a bug in the admin interface where I cannot _delete_ any value once entered. I can update a value, but I cannot completely remove the unsubscribe URI once I've provided it. This seems to be coming from the front-end app, by the time the value gets to the API layer it's been reset to the old value. I don't think this blocks this work but wanted to bring it up.

Definitely not a blocker from my perspective.

leonardr (approved):

PR #1719, opened 2022-05-24
Ubuntu 22.04 & Python 3.10

keithbauer:
## Description

Brings the stack up to date with current packages. Update Ubuntu from 18.04 to 22.04 and Python from 3.6 to 3.10. Combine and upgrade the requirements from core into the respective requirements in the root.

## Motivation and Context

Update to Python 3.10 since 3.6 is now EOL. Update requirements so they don't fall too far behind and address depreciations. Most of the requirements were easily upgradable.

## How Has This Been Tested?

Some failing tests need to be fixed.

## Checklist:


Dockerfile
```diff
@@ -75,15 +75,16 @@ RUN apt-get update \
...
+ && echo "deb-src https://deb.nodesource.com/node_14.x jammy main" >> /etc/apt/sources.list.d/nodesource.list \
  && apt-get update \
  && apt-get install --yes --no-install-recommends \
     build-essential \
+    pkg-config \
     software-properties-common \
     language-pack-en \
     git \
-    python3.6 \
+    python3.10 \
```
keithbauer, line 87: If we don't feel comfortable moving to 3.10 yet then this could be downgraded.

api/admin/geographic_validator.py
```diff
@@ -74,27 +74,31 @@ def validate_geographic_areas(self, values, db):
...
-                        # Flag this as needing to be checked with the registry
+                    try:
+                        if us_search.by_city_and_state(city_or_county, state):
+                            locations["US"].append(value)
+                        elif len([x for x in us_search.query(state=state, returns=None) if x.county == city_or_county]):
+                            locations["US"].append(value)
+                        else:
+                            # Flag this as needing to be checked with the registry
+                            flagged = True
+                    except ValueError:
```
keithbauer, line 95: `uszipcode` added fuzzy searching that can't be toggled off. An invalid city/state search will attempt to find something while raising a `ValueError`.

requirements-base.txt
```diff
@@ -1,61 +1,118 @@
...
 unicodecsv==0.14.1
 uritemplate==3.0.1
-uszipcode==0.2.4
+# urllib3 is pinned to a pre-1.25 version to work around https://jira.nypl.org/browse/SIMPLY-3477
+urllib3==1.26.9
+uszipcode==1.0.1
+# watchtower is for Cloudwatch logging integration
+watchtower==0.8.0
 wcag-contrast-ratio==0.9
+webpub-manifest-parser==0.0.4
```
keithbauer: This package needs an [update for the path](https://github.com/NYPL-Simplified/python-webpub-manifest-parser/blob/master/src/webpub_manifest_parser/core/registry.py#L25) to `MutableMapping`. Instead of `collections.MutableMapping` it is now located at `collections.abc.MutableMapping`.

requirements.txt
```diff
@@ -1,2 +1,4 @@
--r core/requirements.txt
+# All the requirements should be listed in requirements-base.txt except
+# for psycopg2, which is listed here.
 -r requirements-base.txt
+psycopg2==2.8.6
```
keithbauer, line 4: I think I've put this in the right place after combining requirements. Same for `psycopg2-binary`.

tests/saml/configuration/test_model.py
```diff
@@ -499,22 +499,24 @@ def test_get_settings_returns_correct_result(self):
...
                 "signMetadata": False,
-                "digestAlgorithm": "http://www.w3.org/2000/09/xmldsig#sha1",
+                "digestAlgorithm": "http://www.w3.org/2001/04/xmlenc#sha256",
                 "metadataValidUntil": None,
                 "wantAssertionsSigned": False,
                 "wantNameId": True,
                 "wantAssertionsEncrypted": False,
                 "nameIdEncrypted": False,
-                "signatureAlgorithm": "http://www.w3.org/2000/09/xmldsig#rsa-sha1",
+                "signatureAlgorithm": "http://www.w3.org/2001/04/xmldsig-more#rsa-sha256",
```
keithbauer, line 519: I think upgrading the `certifi` package updated these fields. I'm not very familiar with the SAML configuration, but hopefully this is ok.

tests/test_opds.py
```diff
@@ -1015,7 +1015,7 @@ def test_active_loan_feed(self):
         # Putting the last loan activity sync into an Flask Response
         # strips timezone information from it,
         # so to verify we have the right value we must do the same.
-        last_sync_naive = patron.last_loan_activity_sync.replace(tzinfo=None)
+        last_sync_naive = patron.last_loan_activity_sync
```
keithbauer, line 1018: As of Flask 2.0 the datetime object is timezone-aware

keithbauer:
At the moment the tests fail due to the import error for `MutableMapping` from `webpub_manifest_parser`, which won't expose the additional errors from upgrades. I edited this locally so I could continue with the upgrades and fixes.

requirements-base.txt
```diff
@@ -1,61 +1,118 @@
...
+botocore==1.26.2
+certifi==2021.10.8
+cffi==1.15.0
+chardet==4.0.0
+click==8.1.3
+contextlib2==21.6.0
+cryptography==37.0.2
 # defusedxml is required for SAML authentication
-defusedxml==0.6.0
-dm.xmlsec.binding==2.1
```
keithbauer, line 13: I don't think `dm.xmlsec.binding` is required now with `python3-saml`. I think it was required `python-saml`. If that's not the case though then this will need to wait until support for versions of Python > 3.6 is pushed to PyPI.

requirements-base.txt
leonardr: You can fix this yourself pretty easily with a PR. However, the only developer with PyPI upload permissions is a former Lyrasis contractor who lives in Russia, so it might be tricky to contact him. 

Since we're not planning on using this for anything else, I think it'd be OK to copy this into the circulation repo and make the change there, at least as a temporary measure.

requirements-base.txt
leonardr: That's really good news, because these two packages were the reason we couldn't upgrade Python earlier.

api/adobe_vendor_id.py
```diff
@@ -801,7 +801,7 @@ def _encode(self, iss=None, sub=None, iat=None, exp=None):
         if exp:
             payload['exp'] = self.numericdate(exp) # Expiration Time
         return base64.encodebytes(
-            jwt.encode(payload, self.secret, algorithm=self.ALGORITHM)
+            jwt.encode(payload, self.secret, algorithm=self.ALGORITHM).encode()
```
leonardr: Did jwt change such that encode now returns a Unicode string?

core/metadata_layer.py
```diff
@@ -805,6 +805,10 @@ def mirror_link(self, model_object, data_source, link, link_obj, policy):
                     self.log.error(pool.license_exception)
             return
 
+        # If the collection is accessed later than here
+        # a TypeError: can't escape str to binary is raised
+        collection = pools[0].collection if pools else None
+
```
leonardr, line 811: This is weird.

core/model/resource.py
```diff
@@ -565,7 +565,7 @@ class Representation(Base, MediaTypes):
     image_width = Column(Integer, index=True)
 
     # The content of the representation itself.
-    content = Column(Binary)
+    content = Column(LargeBinary)
```
leonardr, line 568: Do these turn into the same underlying postgresql data type as before?

docker/docker-compose.yml
```diff
@@ -36,8 +36,8 @@ services:
     environment:
       UWSGI_PROCESSES: 6
       UWSGI_THREADS: 2
-      SIMPLIFIED_PRODUCTION_DATABASE: postgres://simplified:password@db:5432/simplified_circulation_dev
-      SIMPLIFIED_TEST_DATABASE: postgres://simplified_test:password@db:5432/simplified_circulation_test
+      SIMPLIFIED_PRODUCTION_DATABASE: postgresql://simplified:password@db:5432/simplified_circulation_dev
+      SIMPLIFIED_TEST_DATABASE: postgresql://simplified_test:password@db:5432/simplified_circulation_test
```
leonardr, line 40: It occurs to me we're going to have to mirror this change in our build processes, since that's where the environment variables come from.

api/adobe_vendor_id.py
keithbauer: Yes, with [v2.0.0](https://pyjwt.readthedocs.io/en/stable/changelog.html#jwt-encode-return-type) a string is returned instead of a byte string.

core/metadata_layer.py
keithbauer: This one was very strange and I don't know why the TypeError was raising or why this fixes it...

core/model/resource.py
keithbauer: Yes this is the same data type as before. https://docs.sqlalchemy.org/en/14/changelog/changelog_14.html#change-214869a306f72f33237772d33fc332ec

tests/saml/configuration/test_model.py
leonardr: Yeah, it should be, SHA1 has been deprecated in favor of SHA256 in general.

keithbauer:
This error occurs a few times and is one I don't really understand.

- `AssertionError: Collection was loaded during event handling.`: These errors seem to be raised when assigning works to licensepools. Putting the assignments in a try/except block and re-trying the assignment in the except block seems to get them to pass, but this is not a good solution.

I also enabled warnings for the most recent test outputs and noticed some additional SQLAlchemy warnings, I'm not sure if they're really issues:

- `SAWarning: Coercing Subquery object into a select() for use in IN(); please pass a select() construct explicitly`. This occurs in the Work when generating search documents [here](https://github.com/NYPL-Simplified/circulation/blob/develop/core/model/work.py#L1659) and [here](https://github.com/NYPL-Simplified/circulation/blob/develop/core/model/work.py#L1689).
- `SAWarning: SELECT statement has a cartesian product between FROM element(s) ... Apply join condition(s) between each element to resolve.` This also occurs when [generating the search document](https://github.com/NYPL-Simplified/circulation/blob/develop/core/model/work.py#L1766) and [querying for the ElasticSearch integration](https://github.com/NYPL-Simplified/circulation/blob/develop/api/app.py#L54).
- `SA Warning: relationship X will copy column Q to column P, which conflicts with relationship(s): ‘Y’` [This test](https://github.com/NYPL-Simplified/circulation/blob/develop/core/tests/models/test_resource.py#L566) displays several relationship warnings and the warning message has a link to [this resource](https://github.com/NYPL-Simplified/circulation/blob/ERD-36/core/tests/models/test_resource.py#L566) and mentions configuring an overlaps parameter to some relationships.

I also noticed that there were some `PytestCollectionWarning cannot collect test class '<TestClass>' because it has a __init__ constructor (from: <test_file.py>)`. There are 10 of these warnings and it does look like these tests were not collected or ran. I'll dig into this further since every test should run.

keithbauer:
PyJWT made some changes that affect the tests for `ExpiredSignatureError` and `InvalidIssuedAtError`. This [pull request](https://github.com/jpadilla/pyjwt/pull/252/commits/8f3a2a8a4098693357b69d63a1dbec514ed7c701) no longer rejects tokens with a future `iat` value. [This test](https://github.com/NYPL-Simplified/circulation/blob/develop/tests/test_adobe_vendor_id.py#L803) and [this test](https://github.com/NYPL-Simplified/circulation/blob/develop/tests/test_adobe_vendor_id.py#L812) now raise a `DecodeError: Unknown library: Patron identifier` instead of their respective errors, should these tests raise a `DecodeError` instead?

Looking at the [documentation](https://pyjwt.readthedocs.io/en/latest/api.html#jwt.decode), `iat` and `exp` are only verified if present in a `require` key passed in the `options` dict. It doesn't look like they'll be verified if `verify_signature=False`, which is set [here](https://github.com/NYPL-Simplified/circulation/blob/develop/api/adobe_vendor_id.py#L866), unless `verify_iat` or `verify_exp` is explicitly set to True and passed in as part of the `options` otherwise they are [set as False](https://github.com/jpadilla/pyjwt/blob/master/jwt/api_jwt.py#L88).

keithbauer:
After looking into the `PytestCollectionWarning` warnings, I think they're ok. All the tests are being ran. I think the collection warning is due to some test classes being used as a sort of mock class that don't have test methods, e.g. `TestLicense` or `TestLicenseInfo` used in `TestODLExpiredItemsReaperSingleLicense`.

leonardr:
I recently realized that the entire adobe_vendor_id.py file is obsolete ever since we moved Open eBooks off of Adobe DRM in 2021. Rather than spend a lot of time trying to figure out the jwt test failures, we could remove that whole file and its tests, and potentially the jwt dependency. However, similar code is present in the library registry so we might run into the problem again and have to solve it for real.

I guess I could live with an implementation that allowed for a future `iat`, since I don't think that case would realistically happen, even if this code was still in use. But a token that's more than an hour old shouldn't be accepted.

Talk to Risa about removing the adobe_vendor_id.py file altogether, rather than fixing those bugs. I don't think we'll ever need that code again, and if we do we can model it after the code in the library registry.

leonardr:
If you end up removing that file, do it in a separate PR that lands before this one does.

leonardr:
I don't have a clue about the AssertionError while adding something to a collection. The method where the assertion is failing hasn't changed recently.

It looks like it's got something to do with this relationship on Work:

```
license_pools = relationship("LicensePool", backref="work", lazy='joined')
```

Setting LicensePool.work adds the LicensePool to this list, and that ends up in the bit of code that's failing.

As a way to get more information I'd try first changing the `lazy` option, which is a performance optimization, and see if that solves the problem. Then I'd put a breakpoint inside that append method and look at all the state, especially `self.key`, the thing that's not supposed to be in `dict_` but apparently is.

keithbauer:
Would  `adobe_vendor_id.py` still be needed for SimplyE if it is serving Adobe DRM content?

leonardr:
I don't think so -- adobe_vendor_id.py is an implementation of Adobe's Authentication Web Service, and we've told Adobe to use the library registry implementation at libraryregistry.librarysimplified.org instead.

To test this out, you could check the logs to if there are any hits coming in to /AdobeAuth/* on circulation.librarysimplified.org. If there are any, we probably need to have a talk with Adobe. If not, we should be good to remove those endpoints and all the code behind them.

keithbauer:
I looked through the logs for circulation manager and didn't see any instances of /AdobeAuth/*, but did see them in the logs for registry so I also think this is safe to remove now. I'll work on that in a separate PR.

requirements-base.txt
```diff
@@ -1,64 +1,125 @@
...
+Sphinx==5.0.0b1
+sphinxcontrib-websupport==1.2.4
+SQLAlchemy==1.4.36
+strict-rfc3339==0.7
+textblob==0.15.3
 typing==3.7.4.3
 unicodecsv==0.14.1
 uritemplate==3.0.1
-uszipcode==0.2.4
+# urllib3 is pinned to a pre-1.25 version to work around https://jira.nypl.org/browse/SIMPLY-3477
```
leonardr: This comment can now be removed.

leonardr (approved):
Looks good to me.

PR #1720, opened 2022-05-25
Remove check for http/https scheme in cors domain header

samanthaandrews:
## Description

Allows any url scheme to be returned an allowable Access-Control-Allow-Origin header, not just `http` or `https`.

## Motivation and Context

@kristojorg is working on a proof of concept hybrid web app that makes requests from `capacitor://<something>` which were previously being blocked by CORS access-control-allow-origin headers. This makes it so that any scheme is allowed, not just `http` or `https`.

## How Has This Been Tested?

I added a couple new test cases to `test_url.py::URLUtiliity::url_match_in_domain_list`. 

## Checklist:

### DIFF
--- a/README.md
+++ b/README.md
@@ -137,16 +137,16 @@ That gives you access to all of the standard options to `pytest`, allowing varia
 
 ```shell
 # Fail fast with -x
-docker exec -it --env TESTING=1 cm_local_webapp pipenv run pytest -x tests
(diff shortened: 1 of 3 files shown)


keithbauer (approved):
LGTM!

leonardr (approved):
I'm fine with this implementation. You could also add `capacitor:` to the list of allowed URI schemes, but if you think you might be adding more, this is a lot less maintenance.

PR #1724, opened 2022-06-01
NOREF Reduce SAML test file and update tests

mwbenowitz:
## Description

This is a clone of Nick Ballenger's PR #1686 which reduces the SAML test fixture size and updates tests to match. It is being re-opened here based off the latest develop code, which importantly includes the merged core module.

To ensure that this appropriately tests the validator class we mock out the signature calculation method, as this is an external dependency.

## Motivation and Context

This helps simplify and streamline our unit test suite, with a side-effect benefit of slimming down a 37M file by 99%.

## How Has This Been Tested?
The test suite has been run to ensure that the tests pass as expected and should exercise all relevant parts of the code.

## Checklist:

### DIFF
--- a/tests/saml/metadata/federations/test_validator.py
+++ b/tests/saml/metadata/federations/test_validator.py
@@ -1,8 +1,10 @@
 import datetime
 import os
 
+from mock import patch
 import pytest
 from freezegun import freeze_time
+from onelogin.saml2.utils import OneLogin_Saml2_Utils
 from parameterized import parameterized
 
 import tests.saml.fixtures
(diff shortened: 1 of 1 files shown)


tests/saml/metadata/federations/test_validator.py
```diff
@@ -13,89 +15,141 @@
...
                 with pytest.raises(expected_exception):
-                    validator.validate(federation, metadata)
+                    metadata_expiration_validator.validate(incommon_federation, metadata)
             else:
-                validator.validate(federation, metadata)
+                metadata_expiration_validator.validate(incommon_federation, metadata)
 
 
 class TestSAMLMetadataSignatureValidator(object):
+    @pytest.mark.skip(reason="TODO: Needs new signature for shortened version of metadata file.")
```
leonardr: You're still skipping this test. I'd expect either this test to run with the mocked code and then check that validate_metadata_sign is called with the right options -- OR, delete this test altogether and put that check inside test_validate.

mwbenowitz: thanks for catching that! I was focused on the test condition that had been deleted and forgot to remove that. I think this is a useful test so I kept it and added the mock assertion. Passed locally, so hopefully that wraps this up!

leonardr (approved):

PR #1729, opened 2022-06-07
SIMPLY-4096 Remove Adobe's Authentication Web Service

keithbauer:
## Description

This PR removes Adobe's Authentication Web Service from the circulation manager. This functionality exists and is used in the Library Registry now. I had a call with Leonard on 2022-06-02 and the following was discussed:

---
**Can be removed now**

These are no longer necessary now that the library registry issues Adobe IDs:
class AdobeVendorIDController(object):
class AdobeVendorIDRequestHandler(object):
class AdobeRequestParser(XMLParser):
class AdobeSignInRequestParser(AdobeRequestParser):
class AdobeAccountInfoRequestParser(AdobeRequestParser):
class AdobeVendorIDModel(object):

**These are no longer necessary now that libraries can register directly with the library registry:**
class VendorIDLibraryConfigurationScript(Script):
class ShortClientTokenLibraryConfigurationScript(Script):

---
**Can be removed later**

First, need to verify with Risa that this feature is dead
class DeviceManagementProtocolController(BaseCirculationManagerController):
class DeviceManagementRequestHandler(object):

---
**The tough one**

class AuthdataUtility(object):

Once the changes below have been made, the AuthdataUtility could be renamed to something more accurate like ShortClientTokenUtility.

---
**These methods can be removed:**

def adobe_relevant_credentials(self, patron):
def encode(self, patron_identifier):
def _encode(self, iss=None, sub=None, iat=None, exp=None):
def decode(self, authdata):
def _decode(self, authdata):
def migrate_adobe_id(self, patron):

---
**References to "other_libraries" can be removed from these methods, but the methods themselves should be kept:**

def __init__(self, vendor_id, library_uri, library_short_name, secret, other_libraries={}):
def from_config(cls, library, _db=None):

---
**These methods should be kept.**

def adobe_base64_encode(cls, str_to_encode):
def adobe_base64_decode(cls, str):
def _adobe_patron_identifier(cls, patron):
def short_client_token_for_patron(self, patron_information):
def encode_short_client_token(self, patron_identifier):
def _encode_short_client_token(self, library_short_name, patron_identifier, expires):
def decode_short_client_token(self, token):
def decode_two_part_short_client_token(self, username, password):
def _decode_short_client_token(self, token, supposed_signature):
def numericdate(cls, d):

## Motivation and Context

[SIMPLY-4096](https://jira.nypl.org/browse/SIMPLY-4096). While working on the stack upgrade #1719, it was mentioned that this functionality is deprecated and can be removed before the stack upgrade is merged in.

## How Has This Been Tested?

Ran through test suite.

## Checklist:


nyplwesclark:
Adding for future reference in case it is needed, the Adobe DRM in the library registry is located in the drm blueprint: [https://github.com/NYPL-Simplified/library_registry/tree/develop/library_registry/drm](https://github.com/NYPL-Simplified/library_registry/tree/develop/library_registry/drm)

api/adobe_vendor_id.py
```diff
@@ -39,47 +39,6 @@
...
-        """Process an incoming userInfoRequest document."""
-        output = self.request_handler.handle_accountinfo_request(
-            flask.request.data, self.model.urn_to_label)
-        return Response(output, 200, {"Content-Type": "application/xml"})
-
-    def status_handler(self):
-        return Response("UP", 200, {"Content-Type": "text/plain"})
-
 
 class DeviceManagementProtocolController(BaseCirculationManagerController):
```
keithbauer: @killerpuppytails Do you know if this feature is no longer in use here (`DeviceManagementProtocolController`)?

api/adobe_vendor_id.py
```diff
@@ -162,91 +133,6 @@ def device_id_handler(self, device_id):
...
-                self.ACCOUNT_INFO_ERROR_TYPE,
-                self.URN_LOOKUP_FAILURE % data['user']
-            )
-
-    def error_document(self, type, message):
-        return self.ERROR_RESPONSE_TEMPLATE % dict(
-            vendor_id=self.vendor_id, type=type, message=message)
-
-
 class DeviceManagementRequestHandler(object):
```
keithbauer: @killerpuppytails And this one too, is this no longer in use (`DeviceManagementRequestHandler`)?

api/adobe_vendor_id.py
killerpuppytails: Per our Slack convo, since Adobe is a black box, please do an initial PR with this intact, and we'll remove this one in the next PR.

api/adobe_vendor_id.py
killerpuppytails: Per our Slack convo, since Adobe is a black box, please do an initial PR with this intact, and we'll remove this one in the PR after the next one.

api/controller.py
```diff
@@ -484,17 +483,8 @@ def setup_adobe_vendor_id(self, _db, library):
                     self.log.warn(
```
mwbenowitz: It seems like `new_adobe_vendor_id` will always be `None` so Line 481 could just be `if not vendor_id or not node_value:` and eliminate the `else:` below?

Just trying to make the logic here as clear as possible

mwbenowitz (approved):
This PR looks good in terms of the code. I'm a little hesitant to approve it because I don't have a great sense of the effect, but it does line up with the changes as described in the ticket. I'm not sure there's a way around that though. Maybe I could get a walk through of this system to help my understanding?

One additional comment I have is perhaps it would make sense to rename the api/adobe_vendor_id.py module to something more generic, because I don't think it contains any logic specific to Adobe now?

api/controller.py
keithbauer: Agreed, I'll clean this up.

keithbauer:
I'm open to re-naming this file, but I'm not sure what. Move it to `api/util` and call it `short_client_token.py`?

mwbenowitz:
@keithbauer I think that would make sense!

PR #1730, opened 2022-06-07
Adjust cron times

keithbauer:
## Description

Increase cron frequency for a few scripts. Ensures data integrity by running more often in case of transient failures. This is a summary of the call:

- axis reaper every 30 minutes to start with
- bibliotecha
    - purchase monitor once an hour
    - circulation sweep every 30 minutes
- Overdrive
    - new_titles minute 30 every fifth hour
    - overdrive_reaper
    - format_sweep every 30 minutes, reaper goes away because this should do the same job
- Enki
    - reaper often gives 500 errors, run this at least as often of the import, so maybe minute 30 of every 6th hour
- Spread out the OPDS for distributors a little bit
    - both should run once a day at different times of the day
- OPDS import from feedbooks should run once a day
    - think they moved from OPDS 1 to OPDS 2
- Next time we do work on the DPLA exchange we should look at the ODL import scripts
    - We’ll want these frequencies to be more like the overdrive and axis360 ones to run every 15 minutes, especially the `odl_import_monitor`
- Odilo doesn’t work, leave it alone
- ProQuest likely puts a ton of information into these OPDS feeds
    - `proquest_import_monitor` — maybe do once a day, but investigate
- SAML monitor is fine

## Motivation and Context

[SIMPLY-4085](https://jira.nypl.org/browse/SIMPLY-4085). The times were chosen on a call with Leonard on 2022-06-02.

## How Has This Been Tested?

I don't know how to test this. I double checked my times via https://crontab.guru/.

## Checklist:

### DIFF
--- a/docker/simplified_crontab
+++ b/docker/simplified_crontab
@@ -261,12 +261,6 @@ PID1_STDERR=/proc/1/fd/2
 #   Frequency: Minute 5 of every 6th hour
 5 */6 * * * core/bin/run shared_odl_import_monitor |& tee -a /var/log/cron.log > $PID1_STDOUT 2>$PID1_STDERR
 
-#--- Odilo -------------------------------------------------------------------
-
-# odilo_monitor_recent - Updates an Odilo collection.
-#   Frequency: Every 15th minute
-*/15 * * * * core/bin/run odilo_monitor_recent |& tee -a /var/log/cron.log > $PID1_STDOUT 2>$PID1_STDERR
-
 #--- ProQuest ----------------------------------------------------------------
 
 # proquest_import_monitor - Import ProQuest OPDS 2.0 feeds into Circulation Manager.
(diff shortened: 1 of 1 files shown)


mwbenowitz (approved):
These changes look great as far as the actual cron statements. A few questions about whats described in the PR:
- If feedbooks moved to OPDS2 do we need to update that script?
- I don't see an update for `odl_import_monitor` and `proquest_import_monitor`, should these be updated? Or do we need tickets to investigate what should be done
- What should we do for Odilo? Fix it or remove it would be the two options I think

I don't think these questions block merging this but I want to be sure we capture any questions and feature work somewhere!

keithbauer:
Great questions!
* After some light investigation, I think feedbooks is still on OPDS1, although I do wonder about [this URL](https://github.com/NYPL-Simplified/circulation/blob/main/api/feedbooks.py#L69). I can't get this URL to resolve, but removing `/books/` from it resolves. I think this warrants further investigation.
*  I've adjusted ODL & Proquest to run more frequently.
* I've moved the Odilo script to the `DEPRECATED` section next to the RBdigital timings. I'm a bit hesitant to outright remove it, but if we think that's best I can do that.

mwbenowitz (approved):
LGTM!

PR #1736, opened 2022-06-30
Simply-4099: Remove DeviceManagementProtocolController

keithbauer:
## Description

Removes DeviceManagementProtocolController.

## Motivation and Context

[SIMPLY-4099](https://jira.nypl.org/browse/SIMPLY-4099)

## How Has This Been Tested?

Existing test suite. Also searched CloudWatch logs for `AdobeAuth` which had 0 results for the circulation manager but was present in the library registry.

## Checklist:


nyplwesclark (approved):

nyplwesclark:
Unrelated to the goal of the PR, do you think [this problem with no threshold set](https://github.com/spinalcordtoolbox/spinalcordtoolbox/issues/3654) is why we are seeing the codecov/project failure here, which actually doesn't seem like a failure when I look closer?

keithbauer:
That seems likely. I haven't dug too deep into the codecov configuration, but we could definitely customize the threshold or set it to just be informational.

PR #1740, opened 2022-07-11
SIMPLY-4100 Remove DeviceManagementRequestHandler

keithbauer:
## Description

Second part of removing the Adobe Device DRM. This is all currently handled in the Library Registry.

## Motivation and Context

[SIMPLY-4100](https://jira.nypl.org/browse/SIMPLY-4100)

## How Has This Been Tested?

Current test suite

## Checklist:

### DIFF
--- a/api/util/short_client_token.py
+++ b/api/util/short_client_token.py
@@ -25,36 +25,6 @@ from core.model import (
 )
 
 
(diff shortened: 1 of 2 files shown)


nyplwesclark (approved):
Thanks, Keith. All in the library registry as you said!

PR #1757, opened 2022-07-19
Documentation simply 4082

thinkwebengineer:
## Description

Add openAPI documentation for admin routes listed in Simply-4082
Moved into PR #1760 

## Motivation and Context

To improve the openAPI documentation for all developers
As requested in Simply-4082

## How Has This Been Tested?

No tests required, documentation update only

## Checklist:

### DIFF
--- a/api/admin/routes.py
+++ b/api/admin/routes.py
@@ -575,6 +575,30 @@ def languages():
 @allows_cors(allowed_domain_type=set({"admin"}))
 @returns_json_or_response_or_problem_detail
 def media():
+    """Return the supported media types for a work and link to their schema.org values.
+    ---
+    get:
+      tags:
+        - administration
+      description: Return links to schema.org for associated media type.
+      security:
+        - BasicAuth: []
+      responses:
+        200:
+          description: dict of schema.org links and associated media types
+          content:
+            application/json:
+              schema: MediaSchemaDict
+              example: |
+                {"http://bib.schema.org/Audiobook":"Audio",
+                "http://schema.org/Book":"Book",
+                "http://schema.org/Course":"Courseware",
+                "http://schema.org/EBook":"Book",
+                "http://schema.org/ImageObject":"Image",
+                "http://schema.org/MusicRecording":"Music",
+                "http://schema.org/PublicationIssue":"Periodical",
+                "http://schema.org/VideoObject":"Video"}
+    """
     return app.manager.admin_work_controller.media()
 
 
--- a/api/admin/routes.py
+++ b/api/admin/routes.py
@@ -582,7 +582,9 @@ def roles():
     get:
       tags: 
         - administration
-      description: Returns a JSON of properly mapped MARC codes that are currently defined in this system.
+      summary: Return a mapping from MARC codes to contributor roles.
+      description: |
+        This end point returns a map of MARC Codes to contributor roles that are currently available with this system.
       security: 
         - BasicAuth: []
       responses:
@@ -630,7 +632,7 @@ def languages():
     get:
       tags: 
         - administration
-      description: Returns a JSON of language_codes and associated list of language names
+      summary: Returns a JSON of language_codes and associated list of language names
       security: 
         - BasicAuth: []
       responses:
@@ -663,7 +665,7 @@ def media():
     get:
       tags:
         - administration
-      description: Return links to schema.org for associated media type.
+      summary: Return links to schema.org for associated media type.
       security:
         - BasicAuth: []
       responses:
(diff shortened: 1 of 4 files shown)


api/admin/routes.py
```diff
@@ -401,15 +420,85 @@ def work_change_book_cover(identifier_type, identifier):
...
 @allows_cors(allowed_domain_type=set({"admin"}))
 @returns_json_or_response_or_problem_detail
 @requires_admin
 @requires_csrf_token
 def work_custom_lists(identifier_type, identifier):
+    """Returns or updates custom lists
+    ---
+    args:
+        identifier_type (str): A type of identifier, e.g. "ISBN"
+        identifier (str): An identifier string, used with `identifier_type` to look up an Identifier.
```
mwbenowitz: These should be set as parameters in the `get` and `post` block so they show up in the the API docs. If it's useful they can be put into the controller and re-used for both. There should be a few examples available.

api/admin/routes.py
```diff
@@ -401,15 +420,85 @@ def work_change_book_cover(identifier_type, identifier):
...
+
+    get:
+      tags: 
+        - works
+      summary: A custom list of works
```
mwbenowitz, line 436: I think this returns an array of all the lists a work belongs to!

api/admin/routes.py
```diff
@@ -401,15 +420,85 @@ def work_change_book_cover(identifier_type, identifier):
...
+
+        This returns a Dict of custom lists of works.
+      security:
+        - BasicAuth: []
+      responses:
+        200:
+          description: A dict of a custom list of works
+          content:
+            application/json:
+              schema: CustomListCollection
```
mwbenowitz: Looks like this is simply returning an object right now, should return an array of objects! Easiest way to do this is to create one schema that is simply an array and that references this `CustomListCollection`.

api/admin/routes.py
```diff
@@ -401,15 +420,85 @@ def work_change_book_cover(identifier_type, identifier):
...
+
+    post:
+      tags:
+        - works
+      summary: |
+        To update a list of custom works and affected lanes.
+      description: |
+        An end point to update and create custom lists of works.
+      security: 
+        - BasicAuth: []
```
mwbenowitz, line 482: With our current setup this should accept the two parameters set above and a `X-CSRF-Token` parameter that is required for all `post` requests.

api/admin/routes.py
```diff
@@ -463,30 +557,126 @@ def resolve_complaints(identifier_type, identifier):
...
 @app.route('/admin/roles')
 @allows_cors(allowed_domain_type=set({"admin"}))
 @returns_json_or_response_or_problem_detail
 def roles():
+    """Return a mapping from MARC codes to contributor roles.
+    ---
+    get:
+      tags: 
+        - administration
+      description: Returns a JSON of properly mapped MARC codes
```
mwbenowitz: a nit, but add "that are currently defined in this system" or something like that?

api/admin/routes.py
```diff
@@ -512,14 +704,25 @@ def genres():
...
+    ---
+    get:
+      tags:
+        - administration
+      description: Return CSV of circulation events
+      responses:
+        200:
+          description: CSV of circulation events
+          schema: BulkCirculation
+          example: [time,	event, identifier, identifier_type, title, author, fiction, audience, publisher, imprint, language, target_age, genres, location]
```
mwbenowitz: These two attributes are not currently showing up in the API docs, but I'm also not sure how best to represent a CSV return type here. Maybe as a markdown table?

api/documentation/controller.py
```diff
@@ -405,6 +414,50 @@ def addSchemes(self):
...
+                      'properties': {
+                          'href': {'type': 'string'},
+                          'media_type': {'type': 'string'},
+                          'href': 'media_type'
+                      }
+                }
+            }
+        )
+        self.addComponent(
+            'schema', 'BulkCirculation', 'list',
```
mwbenowitz: Should be `array` here, instead of `list` but see other comment. I'm not sure what the best way to represent a CSV download response is, I've never documented that kind of endpoint before.

api/documentation/controller.py
```diff
@@ -405,6 +414,50 @@ def addSchemes(self):
...
+        )
+        self.addComponent(
+            'schema', 'LanguageCodes', 'dict',
+            {
+                'items': {
+                      'type': 'object',
+                      'properties': {
+                          'language_code': {'type': 'string'},
+                          'languages': {'type': "List['string']"},
+                          'language_code': 'languages'
```
mwbenowitz, line 438: Same comment as above. And if this should be an enumerated list (for the purposes of this documentation I don't think it needs to be) that should be it's own component!

mwbenowitz (changes requested):
This is a really strong first PR to create documentation against a huge codebase using a pretty finicky tool! A few requests to expand on some documentation and move a few things around so they show up in the UI but overall this is great and will be a huge help.

api/admin/routes.py
```diff
@@ -443,13 +439,23 @@ def work_custom_lists(identifier_type, identifier):
 
         This returns a Dict of custom lists of works.
       security:
-        - BasicAuth: []
+        - BasicAuth: [X-CSRF-Token]
```
mwbenowitz, line 442: I'm actually not familiar with this format, what does setting the CSRF token component here do? And I think this should be on the POST method here, not GET!

api/admin/routes.py
```diff
@@ -443,13 +439,23 @@ def work_custom_lists(identifier_type, identifier):
...
+      parameters:
```
mwbenowitz, line 443: Should have specified also need the `library_short_name` parameter! Can just copy it from the OPDS work return endpoint.

api/admin/routes.py
```diff
@@ -475,6 +480,16 @@ def work_custom_lists(identifier_type, identifier):
         An end point to update and create custom lists of works.
       security: 
         - BasicAuth: []
+      parameters:
```
mwbenowitz, line 483: Same comment regarding library name here!

api/admin/routes.py
```diff
@@ -711,18 +804,49 @@ def genres():
...
-    """
+    """Return CSV array of the library bulk circulation events.
+        ---
+        get:
+          tags:
+            - files
+          summary: Return CSV array of library bulk circulation events.
+          description: |
+            Returns a CSV array of bulk circulation events between 2 dates.
+          security: BasicAuth[]
```
mwbenowitz, line 815: Should be (throws an error in the apidocs right now):
```
security:
  - BasicAuth: []
```

api/documentation/controller.py
```diff
@@ -454,9 +454,35 @@ def addSchemes(self):
...
+            'schema', 'BulkCirculationEvents', 'array',
             {'items': 'string'}
         )
+        self.addComponent(
+            'schema', 'CustomListCollectionArray', 'array',
+            {
+                'array': {
+                    'type': 'array',
+                    'items': {'$ref': '#/components/schemas/CustomListCollection'}
+                }
```
mwbenowitz, line 466: This can be simplified down to just `items: {'$ref': ...}`, right now it's causing some odd display issues.

api/documentation/controller.py
```diff
@@ -454,9 +454,35 @@ def addSchemes(self):
...
+                'items': {
+                      'type': 'object',
+                      'key': {
+                          'type': 'string',
+                          'description': 'URI string'
+                      },
+                    'properties': {
+                          'allows_derivitives': {'type': 'boolean'},
+                          'name': {'type': 'string'},
+                          'open_access': 'boolean'
```
mwbenowitz, line 481: should be `type: boolean` here

api/documentation/controller.py
```diff
@@ -454,9 +454,35 @@ def addSchemes(self):
             }
         )
         self.addComponent(
-            'schema', 'BulkCirculation', 'list',
+            'schema', 'BulkCirculationEvents', 'array',
             {'items': 'string'}
```
mwbenowitz, line 458: Should be `type: string` here. The documentation actually works, just throws an error in the UI!

mwbenowitz (approved):
This looks good! A few minor nits/fixes to the openAPI code but this is a really useful extension to our documented endpoints.

api/admin/routes.py
thinkwebengineer: Fixed this.

api/admin/routes.py
thinkwebengineer: fixed

api/admin/routes.py
thinkwebengineer: fixed

api/admin/routes.py
thinkwebengineer: fixed

api/documentation/controller.py
thinkwebengineer: fixed

api/documentation/controller.py
thinkwebengineer: fixed

mwbenowitz:
@thinkwebengineer Are these changes rolled up in the newer PR(s)? If they are we should reference them here and close this one!

PR #1762, opened 2022-08-04
Documentation simply 4150

thinkwebengineer:
## Description

Documentation of admin API end points for ticket 4150

## Motivation and Context

To improve utilization with proper API documentation.

## Checklist:

### DIFF
--- a/api/admin/routes.py
+++ b/api/admin/routes.py
@@ -997,6 +997,60 @@ def lookup_patron():
 @requires_admin
 @requires_csrf_token
 def reset_adobe_id():
+    """Delete all Credentials for a patron that are relevant to the patron's Adobe Account ID.
+    ---
+    post:
+      tags:
+        - administration
+      summary: Delete all Credentials for a patron that are relevant to the patron's Adobe Account ID.
+      security:
+        - BasicAuth: []
+      parameters:
+        - X-CSRF-Token
+        - in: path
+          name: library_short_name
+          description: Short identifying code for a library
+          schema:
+            type: string
+      requestBody:
+        required: true
+        content:
+          multipart/form-data:
+            schema: String
+            example: Identifier for patron
+      responses:
+        200:
+          description: A text confirmation that the password was updated
+          content:
+            text/html:
+              schema: string
+              example: |
+                "Adobe ID for patron %(name_or_auth_id)s has been reset."
+        4XX:
+          description: |
+            These are anticipated errors due to a malformed request, invalid option, 
+            or other issue with the current request. These are returned as JSON objects with 
+            a uniquely identifying URI. Possible URIs for this endpoint are:
+            * `http://librarysimplified.org/terms/problem/incomplete-configuration`
+            * `http://librarysimplified.org/terms/problem/invalid-email`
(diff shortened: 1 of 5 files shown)


api/documentation/controller.py
```diff
@@ -405,6 +405,95 @@ def addSchemes(self):
...
+            {
+                'admin_auth_services': {'$ref': '#/components/schemas/AuthServicesDictSchema'},
+                'protocols': {'$ref': '#/components/schemas/ProtocolDictSchema'}
+            }
+        )
+
+        self.addComponent(
+            'schema', 'AdminAuthPost', 'object',
+            {
+                'protocol': {'type': 'string'},
```
mwbenowitz: I think this should be some kind of JSON object. We should capture that even if we don't fully capture everything that should be submitted here. The general shape of what the protocol should look like would be really useful!

api/admin/routes.py
```diff
@@ -803,6 +936,58 @@ def patron_auth_self_tests(identifier):
 @requires_admin
 @requires_csrf_token
 def lookup_patron():
+    """Look up personal information about a patron via the ILS
```
mwbenowitz, line 939: The `200` response code says "A confirmation that the password was updated" which doesn't match this description. Could you clarify specifically what this does? The fact that it's a POST makes me feel like it is actually probably doing something like resetting a password.

api/admin/routes.py
```diff
@@ -812,6 +997,60 @@ def lookup_patron():
...
+            type: string
+      requestBody:
+        required: true
+        content:
+          multipart/form-data:
+            schema: String
+            example: Identifier for patron
+      responses:
+        200:
+          description: A text confirmation that the password was updated
```
mwbenowitz: Should be a text confirmation that the Adobe ID was successfully reset!

api/admin/routes.py
```diff
@@ -965,6 +1204,94 @@ def sitewide_setting(key):
...
+        Create or update logging services and associated protocols.
+      security:
+        - BasicAuth: []
+      parameters:
+        - X-CSRF-Token
+      requestBody:
+        required: true
+        content:
+          multipart/form-data:
+            schema: AdminAuthPost
```
mwbenowitz, line 1257: Is this the same as the `Protocol` object used for the `admin_auth_services` requests? I'm not really familiar with either of these endpoints!

mwbenowitz (approved):
A few questions but otherwise looks great!

api/documentation/controller.py
thinkwebengineer: This is for the POST request on admin_auth_services and is grabbing a protocol name from a form to then search for the protocols objects.

api/admin/routes.py
thinkwebengineer: These are asking for the same information from the requests in 2 different controller areas. the protocol that is asked for in the form is a string that is used to search protocols by name.

api/documentation/controller.py
thinkwebengineer: Added description in the schema for clarity.

PR #1763, opened 2022-08-08
Documentation simply 4158

thinkwebengineer:
## Description

Update some missing/invalid schemas.
Add documentation listed in simply 4158.

## Motivation and Context

Add documentation for increased API usage and understanding.

## Checklist:

### DIFF
--- a/api/documentation/controller.py
+++ b/api/documentation/controller.py
@@ -269,7 +269,7 @@ class OpenAPIController:
         )
 
         self.addComponent(
-            'schema', 'CustomListResponseSchema', 'object',
+            'schema', 'CustomListResponse', 'object',
             {
                 'custom_lists': {
                     'type': 'array',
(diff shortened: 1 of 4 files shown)


api/documentation/controller.py
```diff
@@ -696,6 +696,48 @@ def addSchemes(self):
...
+        self.addComponent(
+            'schema', 'ChangeBookCoverForm', 'object',
+            {
+                'rights_status': {'type': 'string'},
+                'rights_explanation': {'type': 'string'},
+                'cover_file': {
+                    'type': 'string',
+                    'format': 'binary'
+                },
+                'cover_url': {'type': 'string'},
```
mwbenowitz, line 736: Does the endpoint expect one or the other of these? I don't think we can capture that as a requirement, but a comment along those lines could help clarify things!

(Or if that assumption is incorrect, what the functionality of both fields are)

mwbenowitz (approved):
Looks really good! Just one question about a POST body element

api/documentation/controller.py
thinkwebengineer: Comment added.

PR #1765, opened 2022-08-11
Documentation simply 4165

thinkwebengineer:
## Description

Added documentation for admin end points listed in Simply 4165 and 4167

## Motivation and Context

Improve api understanding and utilization

## Checklist:

### DIFF
--- a/api/admin/routes.py
+++ b/api/admin/routes.py
@@ -605,7 +605,7 @@ def work_custom_lists(identifier_type, identifier):
           description: An array of lists a work belongs to.
           content:
             application/json:
-              schema: CustomListResponseSchema
+              schema: CustomListResponse
         404:
           description: |
             NO_LICENSE error
@@ -2315,7 +2315,7 @@ def custom_lists():
           description: Returns a simple object that contains an array of list objects
           content:
             application/json:
-              schema: CustomListResponseSchema
+              schema: CustomListResponse
         403:
           description: The current admin is not an authorized librarian for the current library
           content:
--- a/api/admin/routes.py
+++ b/api/admin/routes.py
@@ -2261,6 +2261,52 @@ def analytics_services():
 @requires_admin
 @requires_csrf_token
 def analytics_service(service_id):
+    """Delete an analytics service.
+    ---
+    delete:
+      tags:
+        - analytics
+      summary: Delete an analytics service.
+      security:
+        - BasicAuth: []
+      parameters:
+        - X-CSRF-Token
+        - in: path
+          name: service_id
+          description: Id of analytics service to be deleted.
+          schema:
+            type: string
+      responses:
+        200:
+          description: Analytics service has been deleted.
+          content:
+            text/html:
+              schema: |
+                type: string
+        4XX:
+          description: |
+            These are anticipated errors due to a malformed request, invalid option, 
+            or other issue with the current request. These are returned as JSON objects with 
+            a uniquely identifying URI. Possible URIs for this endpoint are:
+            * `http://librarysimplified.org/terms/problem/incomplete-configuration`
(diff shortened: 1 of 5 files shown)


api/admin/routes.py
```diff
@@ -1127,7 +1127,37 @@ def suppressed():
 @returns_json_or_response_or_problem_detail
 @requires_admin
 def genres():
-    """Returns a JSON representation of complete genre tree."""
+    """Returns a JSON representation of complete genre tree.
+    ---
+    get:
+      tags: 
+        - works
```
mwbenowitz: Every other endpoint in this tag concerns fetching/updating an individual work, so I'm not sure this fits here, but not sure where else to put it... `administration` is getting pretty big. Maybe we could make a new one `standard lists` or something like that? It would include:
- this
- `admin/languages`
- `admin/roles`
- `admin/rights_status`
- `admin/media`
- anything else?

api/admin/routes.py
```diff
@@ -1313,6 +1343,70 @@ def library(library_uuid):
 @requires_admin
 @requires_csrf_token
 def collections():
+    """Return collections and protocols object.
```
mwbenowitz: Does this return _all_ the collections and protocols?

api/admin/routes.py
```diff
@@ -1322,6 +1416,44 @@ def collections():
...
+          parameters:
+            - X-CSRF-Token
+            - in: path
+              name: collection_id
+              schema:
+                type: string
+                description: The id of the collection to be deleted
+          responses:
+            200:
+              description: Collections and protocols object
```
mwbenowitz: nit: update description

api/admin/routes.py
```diff
@@ -1340,6 +1472,70 @@ def collection_self_tests(identifier):
...
+        ---
+        get:
+          tags:
+            - administration
+          summary: Return a collection of library registrations.
+          security:
+            - BasicAuth: []
+          responses:
+            200:
+              description: Collections and protocols object
```
mwbenowitz: same comment, update description here!

api/admin/routes.py
```diff
@@ -1340,6 +1472,70 @@ def collection_self_tests(identifier):
...
+          parameters:
+            - X-CSRF-Token
+          requestBody:
+            required: true
+            content:
+              multipart/form-data:
+                schema: CollectionsLibraryRegistrationPost
+          responses:
+            200:
+              description: Collections and protocols object
```
mwbenowitz: another copied description

api/documentation/controller.py
```diff
@@ -738,6 +760,110 @@ def addSchemes(self):
...
+                        }
+                    }
+                }
+            }
+        )
+
+        self.addComponent(
+            'schema', 'CollectionsLibraryRegistrationPost', 'object',
+            {
+                'collection_id': {'type': 'string'},
```
mwbenowitz, line 833: Is this a single string? Or does it accept an array? If it's just a single `collection_id` we should update the description to state that we can only add collections to a library one at a time.

mwbenowitz (commented):
Overall looks great as usual. Just have a question about perhaps reorganizing a few endpoints as we start to have this built out more and some of the sections have grown larger.

api/admin/routes.py
thinkwebengineer: Nothing else I have come across so far.

mwbenowitz: cool, let's shift them into a category called `taxonomies`. That might be a touch inaccurate, but I think it captures what we want to convey to people!

api/admin/routes.py
thinkwebengineer: Updated for clarity, but yes.

mwbenowitz (approved):
👍

PR #1766, opened 2022-08-17
Documentation simply 4173

thinkwebengineer:
## Description

Add documentation for endpoints listed in Simply 4173.

## Motivation and Context

Improve api documentation and utilization.

## Checklist:

### DIFF
--- a/api/admin/routes.py
+++ b/api/admin/routes.py
@@ -1287,35 +1287,35 @@ def circulation_events():
 @requires_admin
 def stats():
     """Return object of library statisticss by library name and total statistics.
-        ---
-        get:
-          tags:
-            - analytics
-          summary: Return object of library statistics by library name and total stattistics.
-          security:
-            - BasicAuth: []
-          responses:
-            200:
-              description: Object of library statistics.
-              content:
-                application/json:
-                  schema: LibraryStatsSchema
-            4XX:
-              description: |
-                An error including:
-                * `LIBRARY_NOT_FOUND`: Library was not found.
-                * `INVALID_ADMIN_CREDENTIALS`: Auth was unable to validate the authenticated email address
-              content:
-                application/json:
-                  schema: ProblemResponse
-            5XX:
-              description: |
-                An error including:
-                * `ADMIN_AUTH_NOT_CONFIGURED`: No admin auth systems set up
-              content:
-                application/json:
-                  schema: ProblemResponse
-        """
+    ---
+    get:
+      tags:
+        - analytics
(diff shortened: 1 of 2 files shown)


api/documentation/controller.py
```diff
@@ -738,6 +760,127 @@ def addSchemes(self):
...
+                'book': {
+                    'type': 'object',
+                    'properties': {
+                        'identifier_type': {'type': 'string'},
+                        'identifier': {'type': 'string'}
+                    }
+                },
+                'complaints': {
+                    'type': 'object',
+                    'description': 'A counter object of complaint types for a given work'
```
mwbenowitz: Is it possible what the counter object looks like? Or is that not obvious from the codebase? Knowing what kind of types of complaints can be here would be helpful.

api/admin/routes.py
```diff
@@ -906,6 +950,57 @@ def refresh(identifier_type, identifier):
...
+          schema:
+            type: string
+          description: An identifier for a work record
+      requestBody:
+        required: true
+        content:
+          multipart/form-data:
+            schema: 
+              type: string
+              name: type
```
mwbenowitz, line 984: What's required to go here? If not comprehensive an example would be really helpful.

api/admin/routes.py
```diff
@@ -1127,7 +1287,37 @@ def suppressed():
 @returns_json_or_response_or_problem_detail
 @requires_admin
 def genres():
-    """Returns a JSON representation of complete genre tree."""
+    """Returns a JSON representation of complete genre tree.
+    ---
+    get:
+      tags: 
+        - works
```
mwbenowitz: Should probably be moved to the `standard lists` section we created in for SIMPLY-4165. Moving it there now will cause a PR conflict I think so we'll just have to resolve that when it comes time to merge.

api/admin/routes.py
```diff
@@ -1257,23 +1447,79 @@ def circulation_events():
 @requires_admin
 def stats():
     """Return object of library statisticss by library name and total statistics.
+    ---
+    get:
+      tags:
+        - analytics
+      summary: Return object of library statistics by library name and total stattistics.
```
mwbenowitz: nit: typo in 2nd statistics!

mwbenowitz (commented):
A few minor things but 99% of the way there!

api/admin/routes.py
thinkwebengineer: Merged develop into this branch to bring all commits up to date.

mwbenowitz (approved):
👍

PR #1774, opened 2022-08-26
HOTFIX Manage Hanging Sessions

mwbenowitz:
## Description

This updates the codebase to more stringently remove database connections and in some places prevent them from being created. Some strategies implemented:

- Removing explicit calls to `engine.connect()` and instead rely on the `Session` object to manage connections
- Explicitly call `close()` when possible

## Motivation and Context

Database connections slowly climb and occasionally cause `QueuePool` errors.

## How Has This Been Tested?

This has been tested locally to ensure that both scripts and the webapp contain now regressions

## Checklist:

### DIFF
--- a/api/app.py
+++ b/api/app.py
@@ -9,7 +9,6 @@ from flask import (
 )
 from flask_swagger_ui import get_swaggerui_blueprint
 from flask_sqlalchemy_session import flask_scoped_session
-from sqlalchemy.orm import sessionmaker
 from .config import Configuration
 from core.model import (
     ConfigurationSetting,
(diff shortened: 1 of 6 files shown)


nyplwesclark (commented):
I'm looking at calls like `self._db.close()` and thinking that we always want code like this that cleans up resources to run. I don't really see this throughout the codebase, and I'm not sure why, but especially for database code, wouldn't try/finally blocks be useful here to guarantee that code like `self._db.close()` always runs? We could create a refactor ticket for this.

mwbenowitz:
@nyplwesclark I think a `try/except/finally` block is a great idea, but the issue is that generally there is no one entry/exit point for scripts, which makes it hard to encapsulate things in that block. If we can figure out a good way to refactor and capture the database connection creation/destruction there I'm all for it

keithbauer (approved):

PR #1775, opened 2022-08-26
Firstbook auth

thinkwebengineer:
## Description

Update firstbook authorization procedure for API calls.

## Motivation and Context

The provider Firstbook is changing the way API calls need to be made to retrieve books and data.  This is related to ticket Simply 4176.

## How Has This Been Tested?

The test_firstbook.py file was updated with new parameters to verify proper url building and sending of request.

## Checklist:

### DIFF
--- a/api/firstbook.py
+++ b/api/firstbook.py
@@ -151,13 +151,15 @@ class MockFirstBookAuthenticationAPI(FirstBookAuthenticationAPI):
                 self.failure_status_code, "Error %s" % self.failure_status_code
             )
         qa = urllib.parse.parse_qs(url)
-        if 'accesscode' in qa and 'pin' in qa:
-            [code] = qa['accesscode']
-            [pin] = qa['pin']
-            if code in self.valid and self.valid[code] == pin:
-                return MockFirstBookResponse(200, self.SUCCESS)
+        for code in qa:
+            if code == 'pin':
+                (pin,) = qa['pin']
             else:
-                return MockFirstBookResponse(200, self.FAILURE)
+                (code,) = qa[code]
+        if code in self.valid and self.valid[code] == pin:
+            return MockFirstBookResponse(200, self.SUCCESS)
+        else:
+            return MockFirstBookResponse(200, self.FAILURE)
 
 
 # Specify which of the classes defined in this module is the
--- a/api/firstbook.py
+++ b/api/firstbook.py
@@ -151,11 +151,11 @@ class MockFirstBookAuthenticationAPI(FirstBookAuthenticationAPI):
                 self.failure_status_code, "Error %s" % self.failure_status_code
             )
         qa = urllib.parse.parse_qs(url)
-        for code in qa:
-            if code == 'pin':
+        for key in qa:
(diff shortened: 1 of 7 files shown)


api/firstbook.py
```diff
@@ -40,26 +41,26 @@ class FirstBookAuthenticationAPI(BasicAuthenticationProvider):
...
                 raise CannotLoadConfiguration(
                     "First Book server not configured."
                 )
-            if '?' in url:
-                url += '&'
-            else:
-                url += '?'
-            root = url + 'key=' + key
+        if root[-1] == '/':
+            root = root[:-1]
```
mwbenowitz: I see the utility in this check, what prompted its inclusion though? We should be able to trust the URL stored in the database.

api/firstbook.py
```diff
@@ -84,7 +85,7 @@ def remote_authenticate(self, username, password):
     # End implementation of BasicAuthenticationProvider abstract methods.
 
     def remote_pin_test(self, barcode, pin):
-        url = self.root + "&accesscode=%s&pin=%s" % tuple(map(
+        url = self.root + "/rest/V1/serialcode?code=%s&pin=%s" % tuple(map(
```
mwbenowitz: Two comments here
1. How do you feel about using a newer string formatting approach here -- either `.format` or an f-string? Trade-offs of consistency v clarity I think
2. Can we store the path as a class variable? I think that's a pattern elsewhere and I like it for clarity's sake.

api/firstbook.py
```diff
@@ -109,7 +110,8 @@ def request(self, url):
 
         Defined solely so it can be overridden in the mock.
         """
-        return requests.get(url)
```
mwbenowitz: If we're constructing the auth header here this is no longer defined solely for that purpose!

mwbenowitz (commented):
FirstBook changes look good! A few questions just about formatting but those are about general standards.

The documentation changes made here don't seem to pertain to the FirstBook updates. Generally would like to keep PRs separate if they address separate issues/tickets. Makes it easier to review and make sure we don't get blocked on merging something we want/need by an unrelated issue! Of course if they are related I stand corrected.

api/firstbook.py
thinkwebengineer: The original firstbook auth seemed to require significant url manipulation.  I added that for a test that was failing in the original test suite.  The tests have been updated as this version does not require this. And this was removed.

api/firstbook.py
```diff
@@ -104,12 +104,12 @@ def remote_pin_test(self, barcode, pin):
             return True
         return False
 
-    def request(self, url):
+    def request(self, url, header):
```
mwbenowitz: Probably not necessary, but could we make this `header={}` here? That way if we need to use this method elsewhere the headers default to an empty object and the calling method doesn't need to worry about it.

Also slightly worried this gets called elsewhere in the codebase and right now this could break those invocations.

mwbenowitz (approved):
One comment to slightly future-proof an update to a method, otherwise this should be good to go now!

PR #1785, opened 2022-09-07
Halt Cron jobs between 01:00 and 03:00 on wednesdays

thinkwebengineer:
## Description

Added an if elif bash statement to check for date and time of the server and exit the cron job if it is between 01:00 and 03:00 server time on Wednesdays.

## Motivation and Context

A manual maintenance window is needed for the database between 01:00 and 03:00 on Wednesdays.  This requires cron jobs to be halted for clean handling. 

## Checklist:

### DIFF
--- a/core/bin/run
+++ b/core/bin/run
@@ -1,13 +1,14 @@
 #!/bin/bash
 
 #### Housekeeping functions and variables ####################################
+# This is to halt the script runner create a manual database maintenance window between 01:00 and 03:00 on Wendesdays.
 
 if [[ $(date +%u) -eq 3 && $(date +%H) -eq 1 ]]; then
   echo "I don't run at this time"
-  exit 1
+  exit 0
 elif [[ $(date +%u) -eq 3 && $(date +%H) -eq 2 ]]; then
   echo "I don't run at this time"
-  exit 1
+  exit 0
 fi
 
 THIS_SCRIPTS_NAME="$(basename $0)"
--- a/core/bin/run
+++ b/core/bin/run
@@ -2,6 +2,14 @@
 
 #### Housekeeping functions and variables ####################################
 
+if [[ $(date +%u) -eq 3 && $(date +%H) -eq 1 ]]; then
+  echo "I don't run at this time"
+  exit 1
+elif [[ $(date +%u) -eq 3 && $(date +%H) -eq 2 ]]; then
+  echo "I don't run at this time"
+  exit 1
+fi
+
 THIS_SCRIPTS_NAME="$(basename $0)"
 
 function display_help () {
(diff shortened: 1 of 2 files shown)


docker/simplified_crontab
```diff
@@ -78,7 +78,14 @@
 #   /var/log/simplified that's named after the script it was asked to run. Those
 #   messages will typically NOT appear in the container output or the cron.log file.
 #
-##############################################################################
+############################################################################### # A database maintenance window is needed between 01:00 and 03:00 on Wednesdays
+if [[ $(date +%u) -eq 3 && $(date +%H) -eq 1 ]]; then
+  echo "I don't run at this time"
+  exit 1
+elif [[ $(date +%u) -eq 3 && $(date +%H) -eq 1 ]]; then
```
keithbauer: I think this line's hour (L#85) should be a 2.

keithbauer:
I like this approach since it's tidy and I think we can confirm if this works locally by changing the hours to the current one then checking the docker logs to see if anything ran with `docker logs cm_local_scripts`.

docker/simplified_crontab
thinkwebengineer: devil is in the details lol. Thanks for the catch.

thinkwebengineer:
This beautiful easy simple fix... breaks the docker cron build... closing PR to fix.

thinkwebengineer:
Updated correct file with time check.  Files now build and the cron manager exits every time it is called for those 2 hours

keithbauer (approved):
This looks good to me now!

core/bin/run
```diff
@@ -2,6 +2,14 @@
 
 #### Housekeeping functions and variables ####################################
 
+if [[ $(date +%u) -eq 3 && $(date +%H) -eq 1 ]]; then
+  echo "I don't run at this time"
+  exit 1
```
mwbenowitz: should our exit code be `0` here? I don't think we consider this an exception, this is our intended functionality during this window.

core/bin/run
```diff
@@ -2,6 +2,14 @@
 
 #### Housekeeping functions and variables ####################################
 
+if [[ $(date +%u) -eq 3 && $(date +%H) -eq 1 ]]; then
```
mwbenowitz, line 6: Let's add a comment up top here explaining what this does and why for future reference!

mwbenowitz (commented):
Functionality looks great! Two QoL questions/requests

core/bin/run
thinkwebengineer: Gotta love how fast the documentation guy forgets documentation :-/. Thanks!

mwbenowitz (approved):

PR #1787, opened 2022-09-09
SIMPLY-4163 GitHub Action to push QA images to ECR

keithbauer:
## Description

Add a new Action to build and publish QA images to ECR.

## Motivation and Context

[SIMPLY-4163](https://jira.nypl.org/browse/SIMPLY-4163) Moving forward with a single QA branch helps the overhead of maintaining three branches and removes the dependency on Bamboo to manage deployments.

## How Has This Been Tested?

This has not been tested yet, but will be on merged pull requests into the new `qa` branch.

## Checklist:

### DIFF
--- a/.github/workflows/publishQA.yml
+++ b/.github/workflows/publishQA.yml
@@ -29,7 +29,7 @@ jobs:
           AWS_SECRET_ACCESS_KEY: ${{ secrets.SE_AWS_SECRET_ACCESS_KEY }}
         - stage: cm_scripts_active
           container_tag: qa-cron-latest
-        - stage: cm_webapp_active:
+        - stage: cm_webapp_active
           container_tag: qa-latest
     steps:
       - name: checkout
--- a/.github/workflows/publishQA.yml
+++ b/.github/workflows/publishQA.yml
@@ -16,17 +16,17 @@ jobs:
         stage: [cm_scripts_active, cm_webapp_active, cm_exec_active]
         include:
         - project: BPL
-            ECR_REPOSITORY: ${{ secrets.BPL_ECR_REPOSITORY }}
-            AWS_ACCESS_KEY_ID: ${{ secrets.SE_AWS_ACCESS_KEY_ID }}
-            AWS_SECRET_ACCESS_KEY: ${{ secrets.SE_AWS_SECRET_ACCESS_KEY }}
+          ECR_REPOSITORY: ${{ secrets.BPL_ECR_REPOSITORY }}
+          AWS_ACCESS_KEY_ID: ${{ secrets.SE_AWS_ACCESS_KEY_ID }}
+          AWS_SECRET_ACCESS_KEY: ${{ secrets.SE_AWS_SECRET_ACCESS_KEY }}
         - project: OE
-            ECR_REPOSITORY: ${{ secrets.OE_ECR_REPOSITORY }}
-            AWS_ACCESS_KEY_ID: ${{ secrets.OE_AWS_ACCESS_KEY_ID }}
-            AWS_SECRET_ACCESS_KEY: ${{ secrets.OE_AWS_SECRET_ACCESS_KEY }}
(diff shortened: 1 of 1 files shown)


.github/workflows/publishQA.yml
```diff
@@ -0,0 +1,61 @@
...
+
+jobs:
+  publish_and_deploy_qa:
+    name: Publish image to ECR
+    if: github.event.pull_request.merged
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        project: [BPL, OE, SE]
+        stage: [cm_scripts_active, cm_webapp_active, cm_exec_active]
```
mwbenowitz: Do we push `cm_exec_active` to ECR? I thought it was just scripts and webapp

.github/workflows/publishQA.yml
```diff
@@ -0,0 +1,61 @@
...
+      - name: Build, tag, and push image to Amazon ECR
+        env:
+          ECR_REGISTRY: ${{ steps.login-ecr.outputs.registry }}
+          ECR_REPOSITORY: ${{ matrix.ECR_REPOSITORY }}
+          IMAGE_TAG: ${{ github.sha }}
+        run: |
+          docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG . --target ${{ matrix.stage }}
+          docker push $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG
+          docker tag $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG $ECR_REGISTRY/$ECR_REPOSITORY:qa-latest
+          docker push $ECR_REGISTRY/$ECR_REPOSITORY:qa-latest
```
mwbenowitz: I think we need the tag to be `qa-cron-latest` for the cm_scripts_active stages otherwise we'll have built the wrong target for this. Not sure how to set that up through the matrix approach though...

mwbenowitz (commented):
Just two things, main question is how do we manage the tags between the webapp and scripts because in bamboo those are just entirely separate built processes.

.github/workflows/publishQA.yml
keithbauer: I checked ECR and you're right, we don't push exec. I'll remove that.

.github/workflows/publishQA.yml
keithbauer: Hmm that does seem tricky. I'll have to think about this some more.

keithbauer: I think this is handled with the updated matrix includes. Another approach, though less ideal, could be to split the scripts and webapp into separate actions.

.github/workflows/publishQA.yml
```diff
@@ -0,0 +1,66 @@
...
+          CONTAINER_TAG: ${{ matrix.container_tag }}
+        run: |
+          docker build -t $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG . --target ${{ matrix.stage }}
+          docker push $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG
+          docker tag $ECR_REGISTRY/$ECR_REPOSITORY:$IMAGE_TAG $ECR_REGISTRY/$ECR_REPOSITORY:$CONTAINER_TAG
+          docker push $ECR_REGISTRY/$ECR_REPOSITORY:$CONTAINER_TAG
+
+      - name: Force ECS Update
+        run: |
+          aws ecs update-service --cluster ${{ matrix.ECR_REPOSITORY }} --service <service_name> --force-new-deployment
```
mwbenowitz: `<service_name>` looks like a placeholder here, but I think for these clusters it's always the same as the cluster name, which is the same as the ECR repo? Would want to confirm that but pretty sure that's true

mwbenowitz (approved):
Looks good, pending resolution of the last comment here!

.github/workflows/publishQA.yml
keithbauer: Whoops, good catch!

PR #1789, opened 2022-09-12
SIMPLY-4112 Update NR INI file for log forwarding

mwbenowitz:
## Description

The original .ini file configuring our Python NewRelic agent was missing logging configuration. Enabling logging and log forwarding should surface existing application logs in the NewRelic console.

## Motivation and Context

Improves our experience using the NewRelic instrumentation

## How Has This Been Tested?

This does not affect any other parts of the codebase, but cannot be tested until deployed to QA (where we have an instrumented server environment).

## Checklist:


thinkwebengineer:
This looks good to me.

keithbauer (approved):

PR #1803, opened 2022-09-28
Documentation_4175

thinkwebengineer:
## Description

Add documentation for admin end point listed in Simply 4175 and fix Sphinx documentation build

## Motivation and Context

Improve use and readability of circ manager admin APIs

## Checklist:

### DIFF
--- a/api/admin/routes.py
+++ b/api/admin/routes.py
@@ -4003,16 +4003,16 @@ def admin_view(collection=None, book=None, etc=None, **kwargs):
           description: Catchall for single-page URLs
       responses:
         200:
-          content:
-            text/html:
-              example: "Your admin account doesn't have access to any libraries. Contact your library manager for assistance."
-        200:
           description: Rendered html template
           content:
             text/html:
               schema: AdminViewPageSchema
         302:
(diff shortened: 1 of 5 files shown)


api/admin/routes.py
```diff
@@ -3971,6 +3971,56 @@ def admin_sign_in_again():
...
+        - in: path
+          name: etc
+          schema:
+            type: string
+          description: Catchall for single-page URLs
+      responses:
+        200:
+          content:
+            text/html:
+              example: "Your admin account doesn't have access to any libraries. Contact your library manager for assistance."
```
mwbenowitz: Unfortunately this gets clobbered by the `200` response below and doesn't show up, not sure if/how to get both to show up but a relatively minor bug if there isn't an obvious solution.

mwbenowitz (approved):
Appreciate the comment explaining the situation. I think this is good to go, but might want to look to see if there's an easy fix for the thing I highlighted.

api/admin/routes.py
thinkwebengineer: Obvious solution would be to change the code from a (incorrect) 200 response to a 401 not authorized.  Though I am not sure what that would break on the other parts of the code/front end.

PR #1807, opened 2022-10-12
Move docker hooks folder to be alongside Dockerfile

keithbauer:
## Description

Moves the `docker/hooks` folder to the root so it is alongside the `Dockerfile`. Also simplifies the build command to use a variable, `IMAGE_BRANCH`, from Docker Hub rather than `sed`ing it from the Docker Hub repository name.

## Motivation and Context

[SIMPLY-4215](https://jira.nypl.org/browse/SIMPLY-4215). Currently targeting isn't working for Docker Hub builds and the three Docker Hub repositories (`circ-exec`, `circ-scripts`, and `circ-webapp`) all build the `cm_exec_active` stage. My best guess for this is due to `cm_exec_active` being the final stage listed in the `Dockerfile` and since targeting isn't functional this is the stage that gets built for all repositories. With this in mind, I don't think that the custom build command that was located at `docker/hooks/build` was actually the build command being executed and instead the default build command from Docker Hub was used. It's difficult to tell if this is actually the case though since the [logs](https://hub.docker.com/repository/registry-1.docker.io/nypl/circ-webapp/builds/87906492-bc05-484c-ad1f-1e714754cf34) from Docker Hub don't show the build command.

The Docker Hub Autobuild documentation states that the `hooks` folder should reside at the same level as the `Dockerfile` [[source](https://docs.docker.com/docker-hub/builds/advanced/#custom-build-phase-hooks)]. By moving the `hooks` folder to be alongside the `Dockerfile` hopefully Docker Hub picks up and uses the custom build command that specifies targeting. With this change I'm not sure if `hooks` should be added to `.dockerignore` so it's not included in the resulting images or if it matters.

## How Has This Been Tested?

This will be tested by the autobuild feature on Docker Hub when it is merged into `develop`.

## Checklist:


mwbenowitz (approved):
This looks good to me. Appreciate the simplification.

I don't think it would hurt anything by adding this to `.dockerignore` since it's only relevant in the build context, but nothing should invoke it so I don't think it will do any harm.

PR #1808, opened 2022-10-13
Simply-4215 Fix Docker Hub builds [hooks/test]

keithbauer:
## Description

Update `hooks/test` to match the version of postgres that we currently use. Also updates the service status check commands to reflect the current architecture.

## Motivation and Context

[SIMPLY-4215](https://jira.nypl.org/browse/SIMPLY-4215). Docker Hub is now picking up the hooks and the test hook was not updated which caused the builds to fail. The test hook has been updated to reflect the current state of services.

`nginx` and `cron` can be checked through `service <nginx|cron> status`, but `gunicorn` has to be checked through `supervisorctl`.

## How Has This Been Tested?

I couldn't find a way to run the test hook, so I tested it piece by piece. The `check_service_status` function with `cron`, `nginx`, and `gunicorn` all reported the expected status. I also curled the healthcheck endpoints and the output looks unchanged.

## Checklist:

### DIFF
--- a/hooks/test
+++ b/hooks/test
@@ -3,8 +3,8 @@
 set -ex
 
 # Create a container with test and production postgres databases.
-docker pull postgres:12.0-alpine;
-docker run -d --env POSTGRES_HOST_AUTH_METHOD=trust --name pg postgres:12.0-alpine;
+docker pull postgres:12.8-alpine;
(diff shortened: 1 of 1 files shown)


mwbenowitz (approved):
This matches my understanding of how we run the application has changed since this file was created.

PR #1809, opened 2022-10-13
Fix Build API Documentation action

keithbauer:
## Description

Fix errors that cause the `Build API Documentation` action to fail.

## Motivation and Context

When `webpub_manifest_parser` was folded into `core/util` this caused the API documentation builds to fail.

## How Has This Been Tested?

In the webapp container in the `docs` folder running `make clean && make html` until build succeeded.

## Checklist:


mwbenowitz (approved):
👍

PR #1811, opened 2022-10-18
Database vacuum

thinkwebengineer:
## Description

A window between 01:00 and 03:00 local server time on Wednesdays has been created to allow for a weekly VACUUM FULL to to help clean up unused table rows and reclaim some disk space.  

## Motivation and Context

This should help improve query times and help prevent our databases from growing too out of hand.  While VACUUM FULL is running, there can be no other uses of the database, which effectively gives us a down time for 2 hours every Wednesday very early morning. [Simply-4170](https://jira.nypl.org/browse/SIMPLY-4170)

## How Has This Been Tested?

This has been tested locally by setting local date and time to Wednesday at 01:00 and allowing for the script to run.  There have been some instances where it does not fire again after moving the clock backwards on a currently running container but it does work for a fresh container every time.

## Checklist:


docker-compose.yml
```diff
@@ -60,7 +60,9 @@ services:
       - type: bind
         source: .
         target: /home/simplified/circulation
-        read_only: true
+        read_only: false
+      - "/etc/timezone:/etc/timezone:ro"
+      - "/etc/localtime:/etc/localtime:ro"
```
mwbenowitz: I'm assuming that this pertains to the timezone stuff, but could you clarify what this is doing?

core/scripts.py
```diff
@@ -2680,6 +2709,49 @@ def run(self, cmd_args=None):
...
+                INDEX_CLEANUP { AUTO | ON | OFF }
+                PROCESS_TOAST [ boolean ]
+                TRUNCATE [ boolean ]
+        """
+        today = datetime.now().strftime("%m/%d/%Y")
+        start = time.time()
+        # Go back up to engine-level.
+        connection = self._db.get_bind()
+        # Get table names
+        all_db_tables = connection.table_names()
```
mwbenowitz, line 2744: I'd be curious as to what this outputs. Is it grabbing only the `public` tables? Or is this vacuuming a lot of the interior, system defined ones as well?

core/scripts.py
```diff
@@ -2680,6 +2709,49 @@ def run(self, cmd_args=None):
...
+                FREEZE [ boolean ]
+                VERBOSE [ boolean ]
+                ANALYZE [ boolean ]
+                DISABLE_PAGE_SKIPPING [ boolean ]
+                SKIP_LOCKED [ boolean ]
+                INDEX_CLEANUP { AUTO | ON | OFF }
+                PROCESS_TOAST [ boolean ]
+                TRUNCATE [ boolean ]
+        """
+        today = datetime.now().strftime("%m/%d/%Y")
```
mwbenowitz, line 2738: Can we add a log line for the start of this process? Will help clarify when/if it's hanging for an extended period.

mwbenowitz (commented):
A few minor questions about this, but overall looks really good.

Is there any need to document what `ntp` is doing? Will that have an impact/side-effect on the other scripts?

Also, it seems like there is some auto-linting happening in this commit in the scripts.py file. I appreciate the effort to clean up the codebase, but it's making it difficult to find the functional changes in the file. Definitely welcome generally linting fixes but perhaps we should limit them to standalone PRs, lets discuss approach to that.

core/scripts.py
thinkwebengineer: ['delegatedpatronidentifiers', 'representations', 'admins', 'adminroles', 'libraries', 'genres', 'subjects', 'patrons', 'integrationclients', 'datasources', 'externalintegrations', 'externalintegrations_libraries', 'externalintegrationslinks', 'configurationsettings', 'collections', 'timestamps', 'identifiers', 'coveragerecords', 'classifications', 'credentials', 'annotations', 'measurements', 'equivalents', 'editions', 'collections_libraries', 'collections_identifiers', 'customlists', 'resources', 'rightsstatus', 'lanes', 'cachedmarcfiles', 'contributions', 'contributors', 'drmdeviceidentifiers', 'licensepooldeliveries', 'deliverymechanisms', 'works', 'collections_customlists', 'resourcetransformations', 'hyperlinks', 'lanes_genres', 'lanes_customlists', 'workcoveragerecords', 'cachedfeeds', 'licensepools', 'workgenres', 'customlistentries', 'circulationevents', 'complaints', 'holds', 'licenses', 'loans'] is the list of tables that returns for `all_db_tables`

mwbenowitz (approved):
Looks good! Want to pay close attention in QA to updated cron schedule

PR #1813, opened 2022-10-24
Update code and tests for clever v3 change

thinkwebengineer:
## Description

Update clever api calls and tests to v3.0 accommodate for the deprecation of clever v1.1 

## Motivation and Context

The Clever API is deprecating their v1.1 API we are updating to v3.0.  This required sending requests to versioned end points and updating the fields we are searching to their new titles.

## How Has This Been Tested?

Our test suite has been updated to send and reflect the changes to the Clever data model and extracting new multi_role_user_id and user_type.  The real test will come upon merging into QA and actually testing the Clever login functionality as this would require setting up Docker HTTPS to test locally.

## Checklist:

### DIFF
--- a/api/clever/__init__.py
+++ b/api/clever/__init__.py
@@ -87,8 +87,10 @@ class CleverAuthenticationAPI(OAuthAuthenticationProvider):
     LOGIN_BUTTON_IMAGE = "CleverLoginButton280.png"
 
     SETTINGS = [
-        {"key": ExternalIntegration.USERNAME, "label": lgt("Client ID"), "required": True},
-        {"key": ExternalIntegration.PASSWORD, "label": lgt("Client Secret"), "required": True},
+        {"key": ExternalIntegration.USERNAME,
+            "label": lgt("Client ID"), "required": True},
+        {"key": ExternalIntegration.PASSWORD, "label": lgt(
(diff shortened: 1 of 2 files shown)


api/clever/__init__.py
```diff
@@ -251,32 +255,36 @@ def remote_patron_lookup(self, token):
...
+                           '/me', bearer_headers)
         data = result.get('data', {}) or {}
 
         identifier = data.get('id', None)
 
         if not identifier:
             return INVALID_CREDENTIALS.detailed(lgt("A valid Clever login is required."))
 
-        if result.get('type') not in self.SUPPORTED_USER_TYPES:
+        if result.get('user_type') not in self.SUPPORTED_USER_TYPES:
```
keithbauer: This looks great! Small question around this field getting renamed. I was looking at the [Clever API Explorer](https://dev.clever.com/reference/me) for this endpoint and it looks like this field is still named `type`, just want to make sure I'm looking at the right thing.

thinkwebengineer: Yeah, thanks for the catch.  I believe I was looking at some of the changes to the user model that do not affect us.  Changes reverted.

keithbauer (approved):
LGTM!

PR #1815, opened 2022-10-25
Fix malformed Authorization headers 500 error

thinkwebengineer:
## Description

Added a try/except block to account for malformed headers, updated tests.

## Motivation and Context

This is in response to [500 errors](https://jira.nypl.org/browse/SIMPLY-4217) reported with a malformed Authorization Header. 

## How Has This Been Tested?

New tests were added to combat this specific case of malformed authorization headers.

## Checklist:

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -3,7 +3,9 @@ import importlib
 import json
 import logging
 import re
-import urllib.request, urllib.parse, urllib.error
+import urllib.request
+import urllib.parse
+import urllib.error
 from abc import ABCMeta
 
 import flask
@@ -122,7 +124,7 @@ class PatronData(object):
                  neighborhood=None,
                  cached_neighborhood=None,
                  complete=True,
-    ):
+                 ):
         """Store basic information about a patron.
 
         :param permanent_id: A unique and unchanging identifier for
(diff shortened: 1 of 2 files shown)


api/authenticator.py
```diff
@@ -908,7 +916,10 @@ def decode_bearer_token_from_header(self, header):
         """Extract auth provider name and access token from an Authenticate
         header value.
         """
-        simplified_token = header.split(' ')[1]
+        try:
+            simplified_token = header.split(' ')[1]
+        except IndexError:
+            simplified_token = 'undefined'
```
keithbauer, line 922: This looks good, small suggestion would be to set this to `None` since there wouldn't be a valid token here.

tests/test_controller.py
```diff
@@ -864,13 +874,23 @@ def test_authenticated_patron_from_request(self):
...
                 'api.base_controller.BaseCirculationManagerController.authorization_header',
-                  lambda x: None
+            lambda x: None
         ):
             with self.request_context_with_library("/"):
                 result = self.controller.authenticated_patron_from_request()
                 assert 401 == result.status_code
                 assert None == flask.request.patron
 
+        # Incorrect authorization header format -> 401 error
```
keithbauer: I think this should be a 400 error instead to match the assertion below.

keithbauer (commented):
This looks great! Left some minor comments.

tests/test_controller.py
thinkwebengineer: Absolutely, thanks for the catch!

keithbauer (approved):

api/authenticator.py
thinkwebengineer: No, if we set this to None there are a cascade of errors that fall out.  The system has been tested to accept "Authorization: Bearer undefined"  when it receives this it passes on 'undefined' and this is handled elsewhere.

PR #1819, opened 2022-10-25
Add patron's root_lane URL if one exists

keithbauer:
## Description

Adds a patron's root_lane URL if one exists and defaults to None if the patron does not have a root lane.

## Motivation and Context

[SIMPLY-4216 Include a patron's root lane as part of patron data](https://jira.nypl.org/browse/SIMPLY-4216)

## How Has This Been Tested?

Added a new test for the basic auth temp token and updated the test for authentication via OAuth.

## Checklist:

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -29,6 +29,7 @@ from .config import (
     CannotLoadConfiguration,
     IntegrationException,
 )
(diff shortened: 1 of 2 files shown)


thinkwebengineer (approved):
I think this looks good

PR #1823, opened 2022-10-31
SIMPLY-4216 Fix: Use the library from the patron's root_lane

keithbauer:
## Description

Use the library from the patron's root_lane instead of getting it from the authenticator. This is more reliable and cleaner than getting it from the authenticator since the library is already tied to the root_lane if one exists.

## Motivation and Context

[SIMPLY-4216](https://jira.nypl.org/browse/SIMPLY-4216)

## Checklist:


thinkwebengineer (approved):
Looks good to me

PR #1824, opened 2022-11-01
Loans head removal

thinkwebengineer:
## Description

Remove support of HEAD call to /loans endpoint based upon 2 year old comment.

## Motivation and Context

While working on documentation there was a 2 year old note to remove support of HEAD method to /loans endpoint. Front end clients should be making the call to /patrons for authentication verification. Linked to [this](https://jira.nypl.org/browse/SIMPLY-4224) ticket.

## Checklist:


api/controller.py
```diff
@@ -1349,13 +1366,6 @@ def sync(self):
         if isinstance(response, Response):
             return response
 
-        # TODO: SimplyE used to make a HEAD request to the bookshelf feed
-        # as a quick way of checking authentication. Does this still happen?
-        # It shouldn't -- the patron profile feed should be used instead.
-        # If it's not used, we can take this out.
-        if flask.request.method=='HEAD':
```
keithbauer, line 1356: Since this is being removed, should `HEAD` be removed from [the available methods](https://github.com/NYPL-Simplified/circulation/blob/develop/api/routes.py#L438) for this endpoint?

keithbauer: Nevermind, I see that you've already removed it now, but it should probably retain the `GET` method.

keithbauer (approved):
This looks good to me!

PR #1825, opened 2022-11-02
Clever v3.0

thinkwebengineer:
## Description

Remove links search and call specific API end points to receive the proper data from Clever v3.0.  
Adjust the data model to reflect what Clever is sending.

## Motivation and Context

This is to update our Clever OAuth for their upcoming deprecation of v1.1 to their latest v3.0

## How Has This Been Tested?

The test suite was adjusted to account for the new data model

## Checklist:

### DIFF
--- a/api/clever/__init__.py
+++ b/api/clever/__init__.py
@@ -102,8 +102,7 @@ class CleverAuthenticationAPI(OAuthAuthenticationProvider):
     TOKEN_DATA_SOURCE_NAME = 'Clever'
 
     EXTERNAL_AUTHENTICATE_URL = (
-        "https://clever.com/oauth/authorize"
-        "?response_type=code&client_id=%(client_id)s&redirect_uri=%(oauth_callback_url)s&state=%(state)s"
+        "https://clever.com/oauth/authorize?response_type=code&client_id=%(client_id)s&redirect_uri=%(oauth_callback_url)s&state=%(state)s"
     )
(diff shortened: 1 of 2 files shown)


api/clever/__init__.py
```diff
@@ -261,21 +260,20 @@ def remote_patron_lookup(self, token):
...
-        user_link = [link for link in links if link['rel']
-                     == 'canonical'][0]['uri']
         # The canonical link includes the API version, so we use the base URL.
-        user = self._get(self.CLEVER_API_BASE_URL + user_link, bearer_headers)
+        user = self._get(self.CLEVER_API_VERSIONED_URL +
+                         'users/%s' % identifier, bearer_headers)
 
         user_data = user['data']
-        school_id = user_data['school']
+        school_id = user_data[user_type]['school']
```
keithbauer: Is this `school_id` and the `student_grade` below nested inside a `roles` key? Looking at the [API Explorer](https://dev.clever.com/reference/getuser-1) for `/users/{id}` it looks like it might be?

thinkwebengineer: Yeah thanks for the catch... according to [this](https://dev.clever.com/docs/users#users-object-example) looks like userdata['roles'][user_type]['school'] is the proper nesting...

keithbauer (approved):
Looks good to me now! 🚀

PR #1826, opened 2022-11-02
Remove delete method from endpoint

thinkwebengineer:
## Description

Remove deprecated DELETE method to `/loans/<identifier_type>/<path:identifier>` end point.

## Motivation and Context

Found a method that has not been declared and removed it.

## How Has This Been Tested?

No testing required.  The method was never declared in the code.

## Checklist:


keithbauer (approved):
Good catch!

PR #1829, opened 2022-11-03
Add logging to identify where login error occurs

thinkwebengineer:
## Description

Add logging to detect bug
## Motivation and Context

Added logging to detect where the break down occurs for the new Clever calls

## How Has This Been Tested?

It will be tested on qa
## Checklist:


keithbauer (commented):
Might want to bump this up to `info` instead of `debug`, I think `info` is the lowest level that we have enabled by default unless we edit the site configuration settings.

keithbauer (approved):

PR #1831, opened 2022-11-03
Clever v3 take 2

thinkwebengineer:
## Description

Update logging info and headers to track where clever login fails.

## How Has This Been Tested?

Testing on QA

## Checklist:

### DIFF
--- a/api/clever/__init__.py
+++ b/api/clever/__init__.py
@@ -184,7 +184,6 @@ class CleverAuthenticationAPI(OAuthAuthenticationProvider):
         response = self._get_token(payload, headers)
         invalid = INVALID_CREDENTIALS.detailed(
             lgt("A valid Clever login is required."))
-        self.log.info('get_token', response['access_token'])
         if not response:
             return invalid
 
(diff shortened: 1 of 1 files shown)


api/clever/__init__.py
```diff
@@ -254,15 +253,18 @@ def remote_patron_lookup(self, token):
...
 
         """
-        bearer_headers = {'Authorization': 'Bearer %s' % token}
+        bearer_headers = {
+            "accept": "application/json",
+            'Authorization': 'Bearer %s' % token
+        }
         result = self._get(self.CLEVER_API_VERSIONED_URL +
                            '/me', bearer_headers)
         data = result.get('data', {}) or {}
```
keithbauer, line 262: Should we log `data` too to get more context?

api/clever/__init__.py
```diff
@@ -254,15 +253,18 @@ def remote_patron_lookup(self, token):
             with the data listed above.
 
         """
-        bearer_headers = {'Authorization': 'Bearer %s' % token}
+        bearer_headers = {
+            "accept": "application/json",
```
keithbauer, line 257: Just personal preference but I think this looks better capitalized as `Accept`, but it is case-insensitive so it'll work either way.

keithbauer (approved):

PR #1833, opened 2022-11-03
response.to_dict causes fatal error Clever

thinkwebengineer:
## Checklist:

### DIFF
--- a/api/clever/__init__.py
+++ b/api/clever/__init__.py
@@ -264,7 +264,7 @@ class CleverAuthenticationAPI(OAuthAuthenticationProvider):
         identifier = data.get('id', None)
 
         user_type = data.get('type', None)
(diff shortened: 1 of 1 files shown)


api/clever/__init__.py
```diff
@@ -264,7 +264,7 @@ def remote_patron_lookup(self, token):
         identifier = data.get('id', None)
 
         user_type = data.get('type', None)
-        self.log.info('identifier', result.to_dict())
+        self.log.info('identifier', result)
```
keithbauer, line 267: This errors out due to it being `None`, we should log `data` to see what we're really getting back from Clever.

keithbauer: Or log `result`, since data might be `{}` when we set it.

keithbauer (approved):

PR #1837, opened 2022-11-10
Werkzeug error

thinkwebengineer:
## Description

Add POST route for '/' endpoint

## Motivation and Context

A werkzeug 405 error is the most common error across all New Relic instances of the CM.  This specifically targets and handles that error. 

## How Has This Been Tested?

A test was added to POST to '/' and verify 405 response code.

## Checklist:


keithbauer (approved):
Could probably remove the `print()` statement from the test but otherwise LGTM!

api/routes.py
```diff
@@ -275,6 +276,12 @@ def decorator(f):
 def index():
     return app.manager.index_controller()
 
+
+@app.route('/', methods=['POST'])
+def return_post():
+    # Handle 405 error for POST method to '/'
+    return Response('Method not allowed', 405)
+
```
mwbenowitz: Is there a need for this explicit route? Or could we just allow `POST` on root and allow the handler you put in the `__call__` method to handle it?

From an abstract perspective this is great, but it clashes a bit with the pattern established in this module.

mwbenowitz (commented):
This 100% solves the problem, only have a comment/thought about ensuring this maintains the patterns we've used in this module.

keithbauer (approved):

PR #1838, opened 2022-11-16
Remove unused turtle import from update_lane_size script

keithbauer:
## Description

Removes an unused import which causes a `ModuleNotFoundError: No module named 'tkinter'` from the `update_lane_size` script.

## Motivation and Context

While investigating [SIMPLY-4229](https://jira.nypl.org/browse/SIMPLY-4229) it was discovered that `update_lane_size` does not run due to the error described above.

## How Has This Been Tested?

Tested locally, script now runs.

## Checklist:

### DIFF
--- a/bin/update_lane_size
+++ b/bin/update_lane_size
@@ -3,7 +3,6 @@
 
 import os
 import sys
-from turtle import update
 from newrelic import agent
 
 


thinkwebengineer (approved):
Looks good to me

PR #1840, opened 2022-11-23
Clever v3 take 2

thinkwebengineer:
## Description

Update Clever tests and schemes to v3.0

## Motivation and Context

To upgrade to v3.0 before deprecation of v1.1

## How Has This Been Tested?

Existing tests pass, but this is difficult to test locally. And will need to be confirmed on qa.

## Checklist:


api/authenticator.py
```diff
@@ -2594,7 +2595,7 @@ def oauth_authentication_callback_url(cls, library_short_name):
         provider to demonstrate that it knows which URL a patron was
         redirected to.
         """
-        return url_for('oauth_callback', library_short_name=library_short_name, _external=True)
+        return url_for('oauth_callback', library_short_name=library_short_name, _external=True, _scheme='https')
```
mwbenowitz, line 2598: I'm a little concerned that this could have side effects if partners have OAuth configured and _aren't_ using `https` but that wouldn't be great so I think it's okay?

api/clever/__init__.py
```diff
@@ -44,6 +44,9 @@
     json_data = f.read()
     TITLE_I_NCES_IDS = json.loads(json_data)
 
+# NCES ID not guaranteed, and returns empty string
```
mwbenowitz, line 47: Could we add a brief TODO just noting that we will implement this on a development-library specific basis in the future.

mwbenowitz (approved):
👍

api/authenticator.py
thinkwebengineer: I think clever requires https now... and I think we should as well.  And possibly do automatically on prod? Was a requirement to get this working locally.

PR #1841, opened 2022-11-29
Database vacuum

thinkwebengineer:
## Description

Add TZ variable to allow for setting local timezone with cron.  Current default is America/New_York.

## Motivation and Context

This Solves the issue of the database vacuum starting at 6pm on Wednesdays because of running on UTC time

## How Has This Been Tested?

This will be tested to see if it is running at the right time on the QA servers.

## Checklist:


mwbenowitz (approved):
TZ stuff looks great! Just a heads up though, your local certs got committed here, should take those out!

PR #1854, opened 2022-12-20
SIMPLY-4238 HTTPS images

keithbauer:
## Description

Force images from the Axis360 API to be `https` instead of `http`.

## Motivation and Context

[SIMPLY-4238](https://jira.nypl.org/browse/SIMPLY-4238) This forces images from the Axis360 API to https for the Open eBooks project.

## How Has This Been Tested?

Automated tests

## Checklist:

### DIFF
--- a/api/axis.py
+++ b/api/axis.py
@@ -1125,6 +1125,8 @@ class BibliographicParser(Axis360Parser):
             element, 'axis:imageUrl', ns
         )
         if thumbnail_url:
+            # Switch out http for https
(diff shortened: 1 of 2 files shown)


mwbenowitz (approved):
This looks like a simple fix that should work for the Axis API. We should merge this in and verify on QA.

Is there an associated script we should be running to remediate any existing `http` links in the database?

keithbauer:
It might take some time but `axis_reaper`, and I think `axis_monitor`, will update resources to https as it processes them.

PR #1856, opened 2022-12-27
Bytecode 3873

thinkwebengineer:
## Description

Add the virtualenv package.
Allow for .pyc files.
Add repository for download of 3.10-dev. 

## Motivation and Context

There were issues with python being compiled by the incorrect python version leading to corrupted .pyc files disrupting the cron jobs until the container is restarted.

## How Has This Been Tested?

This will be tested in QA to see if the .pyc files are still being currupted.

## Checklist:


mwbenowitz (approved):
This looks good. As mentioned in the comment on the PR the real test will be running this in QA and verifying that nothing is corrupted and everything is running under 3.10. But I see no reason why those problems would occur.

If successful this would put us on a version of Python that will eol in 2026, which would be great.

PR #1870, opened 2023-02-06
Fix pymarc build issue

thinkwebengineer:
## Motivation and Context

Builds failing due to an issue with pymarc 4.0.0.

## How Has This Been Tested?

Successful local build

## Checklist:


requirements-base.txt
```diff
@@ -72,7 +72,7 @@ pyfakefs==3.7
 Pygments==2.7.4
 PyJWT==2.4.0
 PyLD==1.0.5
-pymarc==4.0.0
+pymarc>=4.0.0
```
keithbauer: Could this be pinned to 4.2.2?

keithbauer (changes requested):
Since the circulation manager doesn't use pipenv could this Pipfile be removed?

Thanks for getting this updated!

requirements-base.txt
thinkwebengineer: Yup

PR #1874, opened 2023-02-22
Simply 4255 builtin type error

thinkwebengineer:
## Description

When there is a problem loading facets in the search function, that should return a ProblemDetail of the facet and not a lane.

## Motivation and Context

A bug was reported for [this](https://jira.nypl.org/browse/SIMPLY-4255) ticket.  The New relic logs gave a stack trace to this change.

## Checklist:


keithbauer (approved):
Looks good to me, not sure why tox decided to break now.

PR #1883, opened 2023-05-10
Add facet links to search

keithbauer:
## Description
Adds facet links to the search response which will allow for sorting by title and author.

## Motivation and Context
[OE-927 Add facet links to search](https://jira.nypl.org/browse/OE-927)

## How Has This Been Tested?
Modified `test_search_feed` test to also look for the ordering facet links. The `Facets.available_facets` class method needed to be monkeypatched to return the supported order facets since `config` is set to `None` along the way causing it to error out.

## Checklist:


mwbenowitz (approved):
👍

PR #1884, opened 2023-05-17
OE-929 patron is new

keithbauer:
## Description

Adds a new parameter, `is_new`,  for Clever and FirstBook (through the basic auth temp token) that denotes if this is the first time the database is seeing this patron.

## Motivation and Context
[OE-929](https://jira.nypl.org/browse/OE-929) Pass along a status indicator if a we're seeing a patron for the first time in our database.

## How Has This Been Tested?

Updated unit tests to check for `is_new`.

## Checklist:

### DIFF
--- a/api/authenticator.py
+++ b/api/authenticator.py
@@ -1455,7 +1455,13 @@ class AuthenticationProvider(OPDSAuthenticationFlow):
         :return: A Patron if one can be authenticated; a ProblemDetail
             if an error occurs; None if the credentials are missing or wrong.
         """
-        patron = self.authenticate(_db, header)
+        is_new = False
+        try:
+            patron, is_new = self.authenticate(_db, header)
+        except TypeError:
+            # This is a ProblemDetail
+            patron = self.authenticate(_db, header)
+
         if not isinstance(patron, Patron):
             return patron
         if PatronUtility.needs_external_sync(patron):
@@ -1466,7 +1472,7 @@ class AuthenticationProvider(OPDSAuthenticationFlow):
             # update. But we have a cached_neighborhood (which _is_ a
             # model field) to use in situations like this.
             patron.neighborhood = patron.cached_neighborhood
-        return patron
+        return patron, is_new
 
     def update_patron_metadata(self, patron):
         """Refresh our local record of this patron's account information.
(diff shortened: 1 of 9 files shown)


api/authenticator.py
```diff
@@ -2775,6 +2789,10 @@ def get_or_create_token(self, _db, patron):
     def basic_auth_temp_token(self, params, _db):
         """Generate and return a temporary token from HTTP Basic Auth credentials.
         """
+        short_name = self.authenticator.current_library_short_name
+        providers = self.authenticator.library_authenticators[short_name].providers
+        is_new = list(providers)[0].patron_is_new
```
mwbenowitz: How safe is it to assume that `patron_is_new` is always accurate/available from the 0-indexed provider?

api/authenticator.py
```diff
@@ -2023,6 +2032,7 @@ def _authenticate_from_credentials(self, _db, credentials):
             # Just make sure our local data is up to date with
             # whatever we just got from remote.
             self.apply_patrondata(patrondata, patron)
+            self.patron_is_new = False
```
mwbenowitz, line 2035: Setting this explicitly to false here and elsewhere might be safest, but do we _need_ to? It should be false by default, then we only have to change the line below where we can set it to `is_new`

mwbenowitz (approved):
Can't imagine there is a simpler way to add this, looks solid. And doesn't look susceptible to any side effects, seems very clean 👍

api/authenticator.py
keithbauer: Great point, better to err on the side of caution. I'll think of a better approach for this.

api/authenticator.py
keithbauer: When I was testing it needed this to toggle it back from True to False. When authenticating with a new patron the flag is set to True, but then on a subsequent authentication it was still set to True.

mwbenowitz: ah okay, so the class isn't (always?) initialized for each class. Then that makes sense!

api/authenticator.py
mwbenowitz: it's possible that there isn't a super easy way to do it. If there isn't it might be best to just handle the potential `IndexError` and set `is_new` to...false? We'd have to ask OEB what the desired fallback behavior would be

keithbauer: I think I've found a better way by filtering for `BasicAuthenticationProvider` then checking that for any patron_is_new

PR #1903, opened 2023-08-14
OE-917 [fix] Set the original facet to active

keithbauer:
## Description

Set the current search sorting facet to be active.

## Motivation and Context

While sorting facets are now included in the search feed there was no active facet set which caused the search results / client applications to experience unexpected behavior attempting to display the active facet. This now sets the active search sorting facet to be the one the patron has specified (or falls back to the default which must be defined somewhere). This feels a bit hacky to set the active facet on its `href`.

## How Has This Been Tested?

Manually in the CM admin and a basic test that checks for the `activefacet`.

## Checklist:


mwbenowitz (approved):
👍

PR #1905, opened 2023-08-14
Correctly set active search facet

keithbauer:
## Description

Correctly set the active search facet

## Motivation and Context

My previous hacky attempt didn't actually set the `@opds:activeFacet` in the OPDS feed. I thought the URL would pick up on it and do some background magic, but it didn't look like it did.

This now builds out a facet link by using `cls.facet_link` to ensure the `activeFacet` property is correctly set. But this generates a `.../feed/...` URL and we want a `.../search...` URL, so then we replace the href with what `annotator.search_url` builds before adding the facet to the OPDS feed.

## How Has This Been Tested?

More manual testing and verifying the search OPDS feed for the active facet is set through `  +@opds:activeFacet: "true"`

## Checklist:

### DIFF
--- a/core/opds.py
+++ b/core/opds.py
@@ -1029,13 +1029,15 @@ class AcquisitionFeed(OPDSFeed):
         for facet in all_order_facets:
             order = dict(parse_qsl(facet["href"])).get('order')
             if order in enabled_order_facets:
+                facets.order = order
+                if order == original_facet:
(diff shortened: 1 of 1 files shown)


mwbenowitz (approved):
This does look like a cleaner implementation. I'm not familiar enough with this code to know if this is a permanent fix, but it looks good to deploy and confirm!

PR #1920, opened 2024-03-22
NOREF Upgrade Ubuntu to 22.04

mwbenowitz:
## Description

Upgrade Ubuntu to LTS 22.04 and remove some dependencies that were necessary to install Python 3.10.

## Motivation and Context

Changes to legacy dependencies are causing builds to fail when trying to install Python 3.10 on Ubuntu 20.04.

## How Has This Been Tested?

Docker image has been built locally and confirmed to be stable/run as before

## Checklist:

- [] All new and existing tests passed.


keithbauer (approved):
LGTM! Was able to build and run this without any issues.

PR #1931, opened 2024-07-11
TGR-16 Node Docker Upgrade

keithbauer:
## Description

Bump node from 14.x to 20.x

## How Has This Been Tested?

Tested non-elasticsearch functionality locally since I'm having issues with that. Automated tests all pass so I'm reasonably confident that nothing broke. The admin interface looks good to me.

## Checklist:

### DIFF
--- a/api/admin/package.json
+++ b/api/admin/package.json
@@ -9,6 +9,6 @@
   "author": "NYPL",
   "license": "Apache-2.0",
   "dependencies": {
-    "simplified-circulation-web": "0.6.0-rc3"
+    "simplified-circulation-web": "0.6.0"
   }
 }
--- a/api/admin/package.json
+++ b/api/admin/package.json
@@ -9,6 +9,6 @@
   "author": "NYPL",
   "license": "Apache-2.0",
   "dependencies": {
-    "simplified-circulation-web": "0.6.0-rc"
+    "simplified-circulation-web": "0.6.0-rc3"
   }
 }
--- a/api/admin/package.json
+++ b/api/admin/package.json
@@ -8,10 +8,6 @@
   },
   "author": "NYPL",
   "license": "Apache-2.0",
-  "engines": {
-    "node": ">=8.0.0",
-    "npm": ">=5.0.0"
-  },
   "dependencies": {
     "simplified-circulation-web": "0.6.0-rc"
   }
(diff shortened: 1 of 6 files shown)


EdwinGuzman:
@keithbauer can you try updating the version of `simplified-circulation-web` in `api/admin/package.json` from `"0.5.16" to "0.6.0.-rc"? I published a new package version that should be compatible with node 20 and I will test it soon.

api/admin/package.json
```diff
@@ -8,11 +8,7 @@
   },
   "author": "NYPL",
   "license": "Apache-2.0",
-  "engines": {
-    "node": ">=8.0.0",
-    "npm": ">=5.0.0"
-  },
   "dependencies": {
-    "simplified-circulation-web": "0.5.16"
+    "simplified-circulation-web": "0.6.0-rc"
```
EdwinGuzman: @keithbauer can you try running this with version `0.6.0-rc3`? I think it'll be the last "release candidate" before I make the final stable version.

EdwinGuzman: It seems to work for me.

keithbauer: I just tried this out locally and it looks good to me on the pages that load without elasticsearch functionality.

api/admin/package.json
```diff
@@ -8,11 +8,7 @@
   },
   "author": "NYPL",
   "license": "Apache-2.0",
-  "engines": {
-    "node": ">=8.0.0",
-    "npm": ">=5.0.0"
-  },
   "dependencies": {
-    "simplified-circulation-web": "0.5.16"
+    "simplified-circulation-web": "0.6.0-rc3"
```
EdwinGuzman: @keithbauer one more update! Can you set this to `0.6.0`?
I'll leave it as that for now. It reduced issues flagged by dependabot from 53 to 30 but it's good enough for now...
I want to reassess the spreadsheet and make another release later on but this version and the update you have in this PR moves us along quite far already.

keithbauer: Sounds good, if I can get some of these python dependabot updates working I'll add them in a separate PR.

EdwinGuzman (approved):
This looks good. It doesn't seem like the changelog keeps tracks of changes but rather the releases do. Otherwise, I'd say to update the changelog in the repo, but now I remember that this app always worked on tagged releases, correct?

I think we can move forward with circulation-web `0.6.0`, find out how many issues are fixed by that update and the updates in this PR, and then make more changes if we need to in another release.

keithbauer:
Yeah, I track the changes in the tagged release notes.

